PR #10, opened 2017-09-07
Matrixordering

lfarv:
When looking at the matrix ordering question, I went through the `physics` module and found a few issues:

1. Function names: in matlab, all recent function names start with at, old ones don't. For python, I propose to remove the "at" initials if present and keep the matlab name. Ex.:
    `atlinopt` -> `at.linopt`
I also think that all public functions should be directly visible in the `at` package, since users will never remember if they belong to "physics" or "lattice" etc. module.

2. Output arguments: AT uses a lot the possibility in Matlab to check in a function the number of expected output arguments, and act accordingly. This does not exist in python, so there are several alternatives :
	- always consider the maximum number of arguments, and let the user put dummy variables if he is not interested in the result. This is what I tried in `find_orbit4`. Not very elegant, and may lead to unnecessary computations,
	- Add input keywords to let the user tell the function what he is interested in. Possibly more efficient, but not so elegant also.

	What is the most pythonic way ?

3. Matrix ordering:  we can respect the classical orientation of vectors and matrices:

	X=T.X0, where the transfer matrix T  is square and X, X0 are column vectors of coordinates

	Fortran-ordering is only needed in `atpass`. This condition is ensured by the intermediate functions `linepass` and `ringpass`. Outside, any ordering may be used, numpy takes care of it.  For performance reason, functions calling `linepass`/`ringpass` can build directly the input coordinate arrays directly in Fortran-order.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,9 @@ pyat/prefix
 pyat/integrator-src
 *.egg-info
 
+# dynamic integrator list
+atintegrators/passmethodlist.m
+
 # cmake
 CMakeCache.txt
 CMakeFiles
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ matrix:
       env: PYTHON=3
 
 install: |
+  python -c "import numpy; print(numpy.__version__)"
   if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
     if [[ $PYTHON == '3' ]]; then
       brew update
--- a/pyat/at.c
+++ b/pyat/at.c
@@ -206,7 +206,7 @@ static PyObject *at_atpass(PyObject *self, PyObject *args, PyObject *kwargs) {
         PyErr_SetString(PyExc_ValueError, "Failed to parse arguments to atpass");
         return NULL;
     }
-    if (PyArray_DIM(rin,PyArray_NDIM(rin)-1) != 6) {
+    if (PyArray_DIM(rin,0) != 6) {
         PyErr_SetString(PyExc_ValueError, "Numpy array is not 6D");
         return NULL;
     }
@@ -214,8 +214,8 @@ static PyObject *at_atpass(PyObject *self, PyObject *args, PyObject *kwargs) {
         PyErr_SetString(PyExc_ValueError, "rin is not a double array");
         return NULL;
     }
-    if ((PyArray_FLAGS(rin) & NPY_ARRAY_CARRAY_RO) != NPY_ARRAY_CARRAY_RO) {
-        PyErr_SetString(PyExc_ValueError, "rin is not C aligned");
+    if ((PyArray_FLAGS(rin) & NPY_ARRAY_FARRAY_RO) != NPY_ARRAY_FARRAY_RO) {
+        PyErr_SetString(PyExc_ValueError, "rin is not Fortran-aligned");
         return NULL;
     }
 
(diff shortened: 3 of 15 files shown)


lfarv:
It looks like the tests should be modified since `atpass` now accepts only 6xN Fortran-order input arrays. I do not know how these tests works...

willrogers:
Thanks for doing all this.  A few comments:

* I'm happy to update the tests.  It will help me understand the changes here
* For the most common functions, it makes sense to put them in the `at` namespace.  If the library gets large, it's not necessarily helpful to have all functions in the same namespace.  Numpy follows this approach
* I think you've copied in the comments from Matlab.  We'll need to make sure that they do still apply (e.g. variables in uppercase are not the same in Python, while they are in Matlab)
* In a few places you've added `# noinspection PyPep8Naming`.  Where possible I'd like to keep to PEP-8 naming (e.g. `keep_lattice` instead of `KeepLattice`, `xy_step` instead of `XYSTEP` etc.)  This will make the library nicer in the end I think

Your question 2 is the most difficult one and I will think about it.

willrogers:
I have updated the tests for:

* `atpass`
* `elements.py`
* `patpass.py`
* `lattice.py`

I have made a few small changes to the code as well.  I hope that's ok.

I deduce that the difference between `ringpass` and `linepass` is:

* `ringpass` may be called with multiple turns, but not refpts
* `linepass` may be called with refpts, but not multiple turns

Could they be combined into the same function?

lfarv:
You are right for the definition of `ringpass` and `linepass`: I just reproduced the Matlab behaviour. I agree they can be merged, it may even be more convenient in some cases. The function would then just be an interface to `atpass`.

Concerning the naming of variables and functions: going to PEP-8 convention forces to change the names of publicly visible items, function names and keywords. I have nothing against switching to PEP-8 naming, but the translation rules should be simple enough that one could guess the python name from the Matlab one.

Concerning test failures, some may be due to the shape of coordinate vectors: I considered single particle as 6-long vectors (shape (6,)) and set of particles as 6xN matrices (shape (6,N)). One could require single particles as 6x1 matrices, but I don't see it necessary.

willrogers:
There was a problem in some versions of numpy as below.

```
        if refpts is None:
            refpts = len(line)
        refs = uint32_refpts(refpts, len(line))
        r_in = numpy.asfortranarray(r_in).reshape((6, -1))
>       return atpass(line, r_in, 1, refs, int(KeepLattice))
E       ValueError: rin is not Fortran-aligned
```

I didn't realise that a `reshape()` could change the ordering, but it appears that it can.  I've changed it to ensure that the conversion to a fortran array is the last action before passing to atpass.

willrogers:
Might this cause us a problem?  I guess the only time we need to be sure of the ordering is when we pass the array to `atpass`.

```
>>> a = numpy.asfortranarray(numpy.zeros((6,2)))
>>> a.flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False
>>> a.reshape((3,4)).flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False
```

lfarv:
I think it is not a problem: as you said, the only place where order is important is the input of `atpass`. Calling it through `linepass`/`ringpass` takes care of it. Everywhere else, the user should not care about order.

willrogers:
I'm now happy to merge this.  The ordering is now switched and all the tests are passing.

We can make further changes (e.g. `ringpass` and `linepass`) in new pull requests.

You can merge or I will merge if I get an OK from you.

lfarv:
Go on, Will, it's ok for me !

PR #16, opened 2017-10-04
Pyat tweaks

willrogers:
@lfarv this branch merges `linepass()` and `ringpass()`.  Questions:

* I called the new function `lattice_pass()`. Would you like a different name?
* Should we reshape the array that is returned? 

It is currently 6xABC, where A is number of particles, B is number of refpts and C is number of turns, but we could make it 6xAxBxC, i.e. a 4D array. I think that should make it easier to use:

* `r_out[:,0,:,:]` is results for the first particle
* `r_out[:,:,1,:]` is all particles for all turns at the entrance to the second element

There are questions like whether you should 'squeeze' dimensions if they are of size 1.

### DIFF
--- a/pyat/at.c
+++ b/pyat/at.c
@@ -191,14 +191,14 @@ static PyObject *at_atpass(PyObject *self, PyObject *args, PyObject *kwargs) {
     double *drin, *drout;
     int num_turns;
     npy_uint32 num_particles, np6;
-    npy_uint32 numel;
+    npy_uint32 elem_index;
     npy_uint32 *refpts = NULL;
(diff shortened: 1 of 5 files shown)


lfarv:
@willrogers The name `lattice_pass` is fine for me. I agree with you that the 4D shape 6xAxBxC makes the use easier, especially if the array is not squeezed. I would let the squeezing done by the user, out of the function, if needed. But if you prefer doing it, I do not mind !

willrogers:
@lfarv happy to merge?

lfarv:
@willrogers ok, go on !

PR #20, opened 2017-10-12
correct phase advance (tune) computation in atplot splitted lattice with vertical bends

simoneliuzzo:
added option to atdivelem.m to keep rotations and translations at each slice. this option is then used by atplot.m


willrogers:
Hope it's not impolite to say but it's translations not traslations in English.

simoneliuzzo:
I fix it now! Thanks!

PR #22, opened 2017-10-17
Findtune fix

simoneliuzzo:
in this branch i removed the call to sr.bpmname (a local ESRF function) and added the option for NAFF


willrogers:
Hi, there was a problem with the Python build on Windows. If you merge master again you will probably get ticks instead of crosses.

PR #25, opened 2017-11-16
Atplot verticalbend

simoneliuzzo:
In a previous version of atplot the following problem appeared:

when a vertical bend is introduced using a pi/2 rotated dipole, the beta functions swap plane inside the vertical end, and the phase unwrapping is not correct. Eventually: the integer tune is augmented without need, the beta functions plot is strange (evident swap within the magnet).
nicaip_rotdip

To fix this we modified atbaseplot to exclude R1 and R2 from the fields that are kept only at the entrance and exit of the splitted magnets ('KeepAxis' option in atbaseplot). In this way the pi/2 rotation was applied and removed at each slice of the dipole in atplot, thus giving the correct integer tunes and proper beta functions plots.

nicaip_rotdip_keepaxis

<strong>This fix is however not correct when T1~=T2 and R1~=R2. This is used for example to displace girders or apply rotations about y or x (instead of s). </strong>

I consider this a too big loss in flexibility, so I propose to revert to the old atplot, keeping T1, T2, R1 and R2 only at the entance and exit of splitted elements (first and last slices only). 

To fix at least one of the two problems above (the swapping beta functions plot) I created a Vertical Bend PassMethod, that is simply a bend with fixed pi/2 rotation at entrance and exit.
 
nicaip_vertdip

The integer tunes are however are not well computed when this vertical bend is introduced.

Also atgeometry should be modified in this branch.


simoneliuzzo:
Following discussion with L.Farvacque, the KeepAxis flag is moved up to atplot. 
the result of `atplot(r,'labels',atgetcells(r,'Class','Quadrupole'),'KeepAxis');`
is correctly showing the integer tunes in a lattice with vertical bends. 

nicaip_vertdiporrotdip_keepaxisatplot

PR #27, opened 2017-12-04
Fix syntax errors

mfurseman:
I was attempting to build a compiled Matlab application against AT, however this failed as some files contained syntax errors. This provides the bare minimum necessary to resolve those errors.


carmignani:
Hello, the file splitelem.m should be completely deleted in my opinion.
@simoneliuzzo @lfarv

simoneliuzzo:
I agree for the file splitelem.m, it is not used, and it may not be used, as it is broken.

mfurseman:
Thank you for your comments. I've pushed a commit to remove splitelem.m.

carmignani:
We can merge the pull request, but first we should take the last change from the master!

willrogers:
Done.

PR #28, opened 2017-12-14
when the orbitin was passed, the dispersion was still computed on the…

carmignani:
… standard closed orbit, now the bug is fixed


simoneliuzzo:
Your modification seems correct to me. How could we miss something that big? probably we rarely use atlinopt with input orbit.

PR #39, opened 2018-03-12
Remove compiler warnings

lfarv:
Corrections to be safe if FringeBendEntrance/Exit is not in [1,2,3] : fallback to legacy treatment instead of crash, and to silence compiler warnings.


willrogers:
I don't really like commenting old code and leaving it there. The next person to change the code has to wonder why it's still there.

If you think it's relevant to keep it perhaps you could add a comment explaining why as well.

carmignani:
Those two lines are commented also in BndMPoleSymplectic4Pass.c. In the past, we had those boolean flags to decide if we had to add the fringe fields or not, then we decided to remove the flags but I think I forgot to do in the rad and quant version of the passmethod.
But I agree with @willrogers, it is not needed to keep the comment and we could delete the two lines.

lfarv:
I commented the lines for similarity with `BndMPoleSymplectic4Pass.c`, but I am happy to remove them !

PR #43, opened 2018-03-21
atx off-energy

carmignani:
the momentum compaction was computed on energy also when dpp was passed to atx

### DIFF
--- a/atmat/atphysics/ParameterSummaryFunctions/atx.m
+++ b/atmat/atphysics/ParameterSummaryFunctions/atx.m
@@ -110,7 +110,7 @@ end
         tuneper=lindata(end).mu/2/pi;			% tunes
         tunes=periods*tuneper;                  % tunes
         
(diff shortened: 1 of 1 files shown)


carmignani:
Sorry, I probably made a mistake, but I don't understand where. I first updated this branch with the last change of the master (only the elegant2AT converter), then I merged the branch without waiting for the tests, then I see now that the tests failed.
@willrogers, do you understand where was the problem? Unfortunately I already accepted the pull request.

willrogers:
Looks like it was a glitch somewhere - ignore it.

[It's a good idea to delete the branches once they're merged, otherwise you end up with dozens of branches on the repo that you don't need.]

carmignani:
Ok, thanks @willrogers! I was going to delete the branch, as I did for the converter one, but then I saw the red cross! In any case now the master passed all the controls, so it's fine and I delete the fixatx branch.

PR #47, opened 2018-03-23
Docs

willrogers:
Minor updates to the documentation. Added README for writing integrators, although it definitely needs to be improved.

Add `atpath` to `atmexall` so that it doesn't fail on missing paths.


lfarv:
Additionally the `'cd atmat'` command in the INSTALL document is useless and potentially dangerous: it could allow `atmexall` to run even if `atmat` is not in the path, which indicates a wrong installation. Step 4 should be instead `'add recursively atmat and atintegrators to the Matlab path'`
@willrogers : I cannot intervene in your branch to modify INSTALL myself (not in atcollab), so I let you do it !

willrogers:
Thanks for the comments. I am not sure I understand.

I understood that `atpath` was the command you run in order to 'load' AT - that is, it does the same thing as `'add recursively atmat and atintegrators to the Matlab path'`. Is that not true?

I don't mind what the correct procedure is, I just think we should get the documentation saying the right thing so that a new user can get started with no problems.

willrogers:
I pushed the docs branch to the atcollab at.

lfarv:
@willrogers : You are right, `atpath` does the right thing to put AT on the Matlab path, but this will be lost when you close Matlab. And you do not want to call `atpath` each time you start Matlab (since it's itself not in the path…). So the user should put permanently AT in his path, which is the standard thing to do with any Matlab package. There are different ways to do that: the easiest is to use the GUI ( "Set path" in the command bar), other users like to build explicitly their path in their `startup.m` file… I'll propose a sentence in the INSTALL file for you to check !

willrogers:
Could we merge these changes? I have a student who is beginning to work on pyAT and a Python middle layer application, and I would like to be able to improve the documentation.

None of these things are final, they can always be improved.

willrogers:
Close in favour of #78

PR #49, opened 2018-04-09
Atfittune total

carmignani:
The function atfittune can now also fit the total tune and not only the fractional part.

### DIFF
--- a/atmat/lattice/atfittune.m
+++ b/atmat/lattice/atfittune.m
@@ -6,10 +6,10 @@ function newring=atfittune(ring,varargin)
 %
 %RING:          Cell array
 %DPP:           Optional momentum deviation (default 0)
-%NEWTUNES:      Desired tune values (fractional part only)
+%NEWTUNES:      Desired tune values
 %QUADFAMILY1:   1st quadrupole family
 %QUADFAMILY2:   2nd quadrupole family
-%TOTALTUNE:     boolean flag. if true fit the total tune (default false)
+%TOTALTUNE:     if true the function fits the total tune (default false)
 %
 %QUADFAMILY may be:
 %   string: Family name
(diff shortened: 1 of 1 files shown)


atmat/lattice/atfittune.m
```diff
@@ -18,28 +19,53 @@
...
 else
     dpp=0;
-    [newtunes,famname1,famname2]=deal(varargin{:});
+    
+    if nargin==4
+        [newtunes,famname1,famname2]=deal(varargin{:});
+        totaltune=false;
+    else
+        [newtunes,famname1,famname2,totaltune]=deal(varargin{:});
+    end
```
simoneliuzzo: input parsing has become not so straight forward. Any better solution?

atmat/lattice/atfittune.m
```diff
@@ -18,28 +19,53 @@
...
+    if ~totaltune
+        % Compute initial tunes before fitting
+        [lindata, tunes] = atlinopt(ring,dpp); %#ok<ASGLU>
+        
+        % Take Derivative
+        [lindata, tunes1] = atlinopt(setqp(ring,idx1,kl1,delta),dpp); %#ok<ASGLU>
+        [lindata, tunes2] = atlinopt(setqp(ring,idx2,kl2,delta),dpp); %#ok<ASGLU>
+    else
+        % Compute initial tunes before fitting
+        lr=length(ring);
```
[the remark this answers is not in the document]
simoneliuzzo: the lr could be renamed lastpos and assigned to length(r) +1 to avoid the several +1 afterwards.

atmat/lattice/atfittune.m
```diff
@@ -1,14 +1,15 @@
 function newring=atfittune(ring,varargin)
 %ATFITTUNE fits linear tunes by scaling 2 quadrupole families
-% NEWRING = ATFITTUNE(RING,NEWTUNES,QUADFAMILY1,QUADFAMILY2)
+% NEWRING = ATFITTUNE(RING,NEWTUNES,QUADFAMILY1,QUADFAMILY2,TOTALTUNE)
```
simoneliuzzo: instead of TOTALTUNE we could use directly NEWTUNES:
* if [0 0]< NEWTUNES <[1 1] then the fast algorithm is used
* if NEWTUNES >= [1 1] then use the phase advance computation

atmat/lattice/atfittune.m
```diff
@@ -18,28 +19,53 @@
...
+        totaltune=false;
+    else
+        [newtunes,famname1,famname2,totaltune]=deal(varargin{:});
+    end
 end
  
 idx1=varelem(ring,famname1);
 idx2=varelem(ring,famname2);
-newtunes=newtunes-floor(newtunes);
+if ~totaltune
```
simoneliuzzo: see above comment

lfarv:
Hello ! Concerning the input syntax, I suggest checking the presence of a flag `'UseIntegerPart'` by using the function `getflag` on the 1st line of code. Argument parsing is easy, the flag may be anywhere in the argument list (after `ring`…) and by default the behavior remains unchanged.

carmignani:
Ok, so we agree that checking the 5th or 6th input is not very elegant.
There are two possible solutions:
- Fit the integer part always if newtunes is more than (1,1) (@simoneliuzzo );
- Check the presence of a flag 'UseIntegerPart' (@lfarv ).

lfarv:
3 arguments for my solution:

- non-ambiguous input for users,
- easy to implement,
- backward compatibility: if an integer part of the tune appears in some old code, it will still be ignored. If one wants to improve clarity, a warning could be issued in the case of an integer part specified but no `'UseIntegerPart'` flag.

carmignani:
I agree with the three arguments of @lfarv , so I add the flag.

carmignani:
Should we remove the if true of line 44 and the commented code from line 73?

lfarv:
Do you see any reason to go back to the very slow `fminsearch` ? If not, I think it's better to remove the unused parts !

Issue #52, opened 2018-05-23
mxGetPr is not recommended in matlab 2018a


carmignani:
Hello,
from matlab2018a the function mxGetPr is not recommended and mxGetDoubles should be used instead.
At some point probably mxGetPr won't work anymore, so we should do something before. 
mxGetPr is used in atelem.c

https://fr.mathworks.com/help/matlab/apiref/mxgetpr.html

lfarv:
Pull request #55 is merged into master and solves the issue

PR #55, opened 2018-07-16
Replace mxGetPr

lfarv:
As reported by @carmignani  in #52, Matlab R2018a introduces a new C API where:

1) The representation of complex variables is changed (interleaved real and imaginary parts instead of separate real and imaginary arrays (no problem for AT : no complex data accessible in integrators)
2) Data access should be done with specialized macros (`mxGetDoubles`, `mxGetInt32s`...) instead of `mxGetPr`

This new API is accessed when using the `-R2018a` flag in the `mex` command. This not the default yet but it will come in the future. The proposed modifications apply the new API when the -R2018a flag is used and keep the old one otherwise, as for the previous Matlab versions.

While doing this, all integrators have been converted to the Python-compatible `trackFunction` instead of `passFunction`


willrogers:
Looks fine to me, and the tests still pass.

carmignani:
It looks very good!

lfarv:
When converting the few last integrators to the `trackFunction`, I found some problems with data-ordering: see  #56. Until this is solved, I think that in pyat:

- the processing of axis rotations in Python looks wrong.
- Matrix66Pass and MatrixTijkPass are wrong

PR #58, opened 2018-07-26
atx gives also the damping partition numbers

carmignani:
I added the function DampJfromTau and I call that function from atx to compute the damping partition numbers using the damping times.


carmignani:
Automatic test failed. @willrogers, do you understand why? I changed only matlab code.

willrogers:
A Travis failure - I restarted the job and it passed this time.

lfarv:
Unless the function dampinJFromTau is useful somewhere else, I am not sure the computation is worth a dedicated function… It could be removed

carmignani:
I agree, it can be removed. I had that function in my local AT since a long time ago and I just added the function, but I can delete it now.

Issue #60, opened 2018-08-02
atfastring() with cavities of non-zero length


JonasKallestrup:
When atfastring() is used with a lattice containing cavities with non-zero length, then linear optics will change due to the movement of the cavity element to the start of the ring.

To fix this, atfastring() may check whether each cavity element has a length. If they do, then insert drift-spaces with correct length before/after and set cavity length to zero.

lfarv:
Fixed by merging pull request #61

PR #61, opened 2018-08-02
Update atfastring.m for non-zero length cavities

JonasKallestrup:
atfastring now modified to work with lattices where cavities have non-zero length. See issue: https://github.com/atcollab/at/issues/60


carmignani:
It looks good to me.

JonasKallestrup:
That is correct, @lfarv. I've moved the index shift to within the if-statement.

PR #64, opened 2018-08-23
Python data ordering

lfarv:
Two different points are addressed here:

**data-ordering**: it is now set (and checked) to Fortran-ordering for all element attributes.

**output arguments**: Matlab allow a function to check the number or requested output arguments and act consequently. Python has no such a feature. So I tried to set the following rules:

- in functions similar to find_xxx having an optional `refpts` argument: if `refpts` is `None` (the default), the function returns the result at the start of the 1st element. If `refpts` is given, the function return a tuple with the results at the beginning of the lattice, and the results at all referenced locations.
- all other output arguments potentially available in Matlab must be explicitly requested with a boolean input flag, and are returned in a tuple (see `find_m44`)

Here is a non-exhaustive list of changes:

1. In `track.py` and `at.c`: allow any iterable to describe a lattice (sequence or iterator). Also allow empty `refpts`, the coordinates are only available at the end of tracking in `rin`
2. In `elements.py`: new class `M66` 
3. In `lattice.c`: new functions to shift and tilt elements. New functions to select elements based on their class or attributes.
4. In `__init__.py`: new exception classes `AtError` and `AtWarning` to handle AT-specific errors.
5. In `physics.c`: all functions are modified for output arguments. all features available in Matlab are implemented. New functions: `find_sync_orbit`, `find_orbit6`, `find_m66`

### DIFF
--- a/atintegrators/IdTablePass.c
+++ b/atintegrators/IdTablePass.c
@@ -42,13 +42,8 @@ static double get_kick(double *r6, double *ktable)
     /*splin2(GLOBAL_y_map,GLOBAL_x_map,GLOBAL_xkick,GLOBAL_xkick2,GLOBAL_n,GLOBAL_nx,y,x,&f);*/
     
     /*biliniar interpolation*/
-#ifdef MATLAB_MEX_FILE
     /* Transpose coordinates because the kick-table is FORTRAN-ordered */
     linint(GLOBAL_y_map, GLOBAL_x_map, ktable, GLOBAL_ny_map, GLOBAL_nx_map, r6[2], r6[0], &f);
-#else
-    /* Assume the kick-table is C-ordered */
-    linint(GLOBAL_x_map, GLOBAL_y_map, ktable, GLOBAL_nx_map, GLOBAL_ny_map, r6[0], r6[2], &f);
-#endif
     return f;
  }
 
--- a/atintegrators/atelem.c
+++ b/atintegrators/atelem.c
@@ -190,8 +190,8 @@ static double *atGetDoubleArraySz(const PyObject *element, char *name, int *msz,
         PyErr_SetString(PyExc_RuntimeError, errmessage);
         return NULL;
     }
-    if ((PyArray_FLAGS(array) & NPY_ARRAY_CARRAY_RO) != NPY_ARRAY_CARRAY_RO) {
-        snprintf(errmessage, 60, "The attribute %s is not C-aligned.", name);
+    if ((PyArray_FLAGS(array) & NPY_ARRAY_FARRAY_RO) != NPY_ARRAY_FARRAY_RO) {
+        snprintf(errmessage, 60, "The attribute %s is not Fortran-aligned.", name);
         PyErr_SetString(PyExc_RuntimeError, errmessage);
         return NULL;
     }
--- a/atmat/atphysics/ParameterSummaryFunctions/atsummary.m
+++ b/atmat/atphysics/ParameterSummaryFunctions/atsummary.m
@@ -99,7 +99,10 @@ sum.etac = sum.gamma^(-2) - sum.compactionFactor;
 cavind = findcells(THERING,'HarmNumber');
 if ~isempty(cavind)
     freq = THERING{cavind(:,1)}.Frequency;
-    v_cav = THERING{cavind(:,1)}.Voltage;
+    v_cav = 0;
+    for i = 1:length(cavind)
+        v_cav = v_cav + THERING{cavind(:,i)}.Voltage;
+    end
 else
     % Default
     freq = 352.202e6;
--- a/atmat/atphysics/ParameterSummaryFunctions/atx.m
+++ b/atmat/atphysics/ParameterSummaryFunctions/atx.m
@@ -147,6 +147,8 @@ end
                 [chi,tns]=atdampingrates(m);
                 fs=abs(tns(3))/revperiod;
                 dampingtime=revperiod./chi;
+                alpha=1.0./dampingtime;
+                dampingJ=4.0*alpha/sum(alpha);
                 if any(radindex)
                     jmt=jmat(3);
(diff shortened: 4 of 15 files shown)


lfarv:
First look at errors in tests:

1. `refpts`: ordering is not necessary in most cases (except atlas). So I moved the tests in lattice_pass rather than `uint32_refpts`. I'll move it back, there is very little interest in asking for results in random order.
2. empty `refpts`: I kept the backward compatibility at the level of `lattice_pass`: the default is to give the output at the end of the lattice. Since `atpass` is not supposed to be called directly, this should not be a problem. But I added the possibility of specifying explicitly an empty list(`refpts=[]`). This is useful to avoid a lot of output if only the final coordinates are used (after possibly many turns). I propose to keep it like that.
3. length of orbit vectors: in Matlab, `findorbit4` returns 4x1 orbit, `findsyncorbit` returns 5x1 orbits, `findorbit6` returns 6x1 orbits. This is not logical since anyway orbits have to be padded to be used. I would like to standardize orbits to 6 coordinates.
3. orbit output in `find_m44`: I made a mistake by returning the orbit at the beginning of the lattice instead of at `refpts`, having in mind that once it's computed, it can be used somewhere else. If this is intended, one could compute the orbit explicitly and pass it to find_m44, it's as efficient. So I can revert the previous behavior. But I could also remove this output (and the corresponding input flag): it's one more particle to track in find_m44, it can be done outside ~as efficiently. As you like…

willrogers:
> refpts: ordering is not necessary in most cases (except atlas). So I moved the tests in lattice_pass rather than uint32_refpts. I'll move it back, there is very little interest in asking for results in random order.

I agree. Other points (I'm not finished, there's quite lot here):

* I think Octupole needs `REQUIRED_ATTRIBUTES = Multipole.REQUIRED_ATTRIBUTES`
* If you need `refpts_subset` as well as `refpts_iterator`, could you just use

```
def refpts_subset(ring, refpts):                                                                         
   return list(refpts_iterator(ring, refpts))
```

lfarv:
> I think Octupole needs REQUIRED_ATTRIBUTES = Multipole.REQUIRED_ATTRIBUTES

Ok, done !

> If you need refpts_subset as well as refpts_iterator, could you just use
```
def refpts_subset(ring, refpts):                                                                         
   return list(refpts_iterator(ring, refpts))

```
You are right, I do not think `refpts_subset` is necessary ! I'll remove it.

willrogers:
If I understand the changes correctly, passing `None` into `atpass` now means that the output is 6x0. The intention in this case is only to use the modified `rin`, not to use `rout`. Correct? If so, could `atpass` return None in this case?

lfarv:
@willrogers : You are right, the intention is to avoid useless output when only the final value is necessary and is taken from the modified `rin`
`atpass` (C-code) does not expect `None` as input. `refpts` is checked as being an integer array, possibly empty. I just removed the special treatment of the empty array, so that now an empty `refpts` means an empty `rout`. For consistency and C-code simplicity, I would prefer keeping the output shape as 6 x (n_particles x n_refs x n_turns), whatever the numbers are. The processing of `None`s is done in `lattice_pass` which is supposed to be the unique interface to python. There a `None` input (default) is converted to `len(ring)` and still means end of each turn, as before. But now an explicitly empty array means empty output. At that level, I could replace the output with `None` if desired. However this may make further processing more tricky: for instance

> `out_mat = numpy.squeeze(at.lattice_pass(ring, in_mat, refpts=refs, keep_lattice=keeplattice), axis=3)
> `

in find `m44` should be protected by a `None` test, while empty arrays with the right shape behave properly.

Another point: the previous behaviour of `get_s_pos` with `None` input resulted from a bug: indexing an array with `None` returns the whole array ! This was not intended nor documented in `get_s_pos`. There could be 2 meanings of `None` input: either all locations (`range(len(ring)+1)`) or the last one (`len(ring)`). I chose `len(ring)` because it's more commonly used (ring circumference), but I could switch the the other option.

willrogers:
> The processing of Nones is done in lattice_pass which is supposed to be the unique interface to python

I agree with this - the tests for `atpass` are just there to make sure that it is doing what we think it's doing. I will fix them to test for what we decide it should do. I will have another look when I am back at work tomorrow.

lfarv:
Finally, having atpass returning directly the right shape 6 x nparticles x nrefpts x nturns avoids any reshaping and makes data collecting in patpass much simpler (refpts is now handled). So I propose to keep this version.

willrogers:
>  Finally, having atpass returning directly the right shape 6 x nparticles x nrefpts x nturns avoids any reshaping and makes data collecting in patpass much simpler (refpts is now handled). So I propose to keep this version.

This is an improvement, thanks.

willrogers:
I pushed some commits to this branch. All the tests now pass apart from those in `physics.py` - I will look at those next.

> Another point: the previous behaviour of get_s_pos with None input resulted from a bug: indexing an array with None returns the whole array ! This was not intended nor documented in get_s_pos. There could be 2 meanings of None input: either all locations (range(len(ring)+1)) or the last one (len(ring)).

I chose the latter here as I thought it was more natural.

willrogers:
Now all the tests pass again. Final comments:

* `find_orbit4`, `find_sync_orbit` and `find_orbit6` have a lot of similar code. I wonder if there's a way of sharing some of it
* the other option for variable return options is to return `None` if a particular return value is not relevant. The caller then always knows how many values to expect
* I'm happy to standardise on 6D orbits if you think that is sensible

I'm happy to merge these changes if you are.

willrogers:
Oh, one more thing. Should we have modules `orbit` and `linear_optics` to match the directories in the Matlab version?

lfarv:
About return values: the question is when some additional computation is necessary to produce the desired values. It would be inefficient to do the work if the output is useless… Hence the input flag.

Concerning file organisation, one could turn `physics` into a python package with modules `orbit`, `linear_optics` and may more in the future. Otherwise it looks ready for merging ! Go on, @willrogers.

My next step is now the `atlinopt` conversion.

PR #68, opened 2018-09-24
Linopt

lfarv:
In this release:
- reorganisation into subpackages: `at.lattice` and `at.physics`
- in element creation functions, parameters are checked for their type and size
- the `linopt` function is now available

In addition:
- a bug in phase advances in the rarely used Matlab function `twissring` is corrected
- the Matlab test functions in `pyat/test_matlab` work again, but a test on the new `linopt` function would be useful.


willrogers:
Sorry, I have been very busy at work. This looks good, thank you. Some comments:

* does lattice need its own package when there's only one module? I don't particularly mind
* it's a bit confusing to me that a coupled ring uses `linopt` and an uncoupled ring uses `twissring`. I realise this is for consistency with Matlab, so if you prefer to keep it like that I don't mind
* I have pushed a commit that simplifies a test after your fixes to `get_twiss`
* I am pleased that you can run the Matlab comparison tests. They are fiddly to set up, but I think they are very useful to have. However, in order for them to run correctly, I need the additional lines:

```
(venv34) hgs15624@pc0072 ~/c/a/pyat (linopt)> git diff
diff --git a/pyat/test_matlab/utils.py b/pyat/test_matlab/utils.py
index c5b807b..205f310 100644
--- a/pyat/test_matlab/utils.py
+++ b/pyat/test_matlab/utils.py
@@ -18,5 +18,7 @@ def initialise_matlab():
     eng.addpath(os.path.join(ROOT_DIR, 'atmat/'))
     eng.addpath(os.path.join(ROOT_DIR, 'atmat/attrack/'))
     eng.addpath(os.path.join(ROOT_DIR, 'atmat/atphysics/'))
+    eng.addpath(os.path.join(ROOT_DIR, 'atmat/atphysics/LinearOptics'))
+    eng.addpath(os.path.join(ROOT_DIR, 'atmat/atphysics/Orbit'))
     eng.addpath(os.path.join(ROOT_DIR, 'atmat/lattice/'))
     return eng
```
I am not sure why you do not need these lines.

lfarv:
@willrogers : 

-  package names are set to mimic the Matlab directory names in the distribution. The `lattice` package is still almost empty but it will contain all the lattice-manipulation functions still to be written !
- both `linopt` and `twissring` (`get_twiss` in pyat) were in the original AT distribution. I do not know of any good reason to use `twissring` instead of `linopt` so we may safely get rid of it.
- Can you try the Matlab tests with the recursive path addition ?

willrogers:
Also, I timed execution of the `linopt` and `get_twiss` functions, and I found that `get_twiss` is about 10 times faster for the `dba.mat` file we are running our tests on. This sounds plausible as the calculation is simpler, but you might have another opinion.

The speed may be significant for the 'virtual accelerator' that we are building, so I'd like to keep `get_twiss`.

willrogers:
_Also_, should we remove the print statement:

```
Simulating 1 particles for 1 turns through 145 elements.
Simulating 8 particles for 1 turns through 145 elements.
Simulating 5 particles for 1 turns through 145 elements.
Simulating 5 particles for 1 turns through 145 elements.
Simulating 5 particles for 1 turns through 145 elements.
Simulating 1 particles for 1 turns through 145 elements.
Simulating 8 particles for 1 turns through 145 elements.
Simulating 5 particles for 1 turns through 145 elements.
Simulating 5 particles for 1 turns through 145 elements.
Simulating 5 particles for 1 turns through 145 elements.
Simulating 1 particles for 1 turns through 145 elements.
Simulating 8 particles for 1 turns through 145 elements.
```
?

It's now more noise than useful debugging.

lfarv:
@willrogers 
After working a little bit on `ohmienvelope` (equilibrium emittances), I think of rearranging the output of `linopt`: as for `find_orbit4`, `find_m44`, the idea would be to output first the results at the beginning of the lattice, and then optionally the results at refpts (if it is not None). It may help reducing the amount of computation to the minimum necessary.
If you think a variable number of output arguments causes problems, then one may opt for empty arrays (in all functions preferably). 
Let me know, before we can merge `linopt`

lfarv:
The reason for the chromaticity deviations is that in `get_twiss`, optics is computed for p=p0 and p=p0+dp, while in lint, it is for p0-0.5*dp and p0+0.5*dp, which is more correct. I changed `twiss_ring` to to the same as `linopt`. However the test now fail…
Here is also a proposal for `linopt` output. I think `get_twiss` should be similar. To be modified!

lfarv:
Cannot understand the test problem... Otherwise this version is ready for merging

willrogers:
Just a glitch in the Travis service.

lfarv:
2*pi increment is correct provided the phase advance is computed with `atan2` and not `atan`, as it is done in `atlinopt`. The modification in `twissring` was assuming this, which is wrong. #141 fixes that.

Issue #70, opened 2018-10-08
pyat: atpass error


ZeusMarti:
Hi all,

I just tried a simple code: 

```
import at
from at import load_mat
import numpy

m = load_mat.load('a25.mat')
rin = numpy.array((1e-6, 0, 0, 0, 0, 0))
print(rin)
at.atpass(m, rin, 1)
print(rin)
```

It runs well if I use the example "dba.mat", but with our lattice "a25.mat" I get the following error:

**Simulating 1 particles for 1 turns through 1469 elements.
Segmentation fault (core dumped)**

 It seems I can't attach matlab files, so if anyone is willing to help I can send the file.

Thanks,
Zeus

willrogers:
Hi,

Thanks for trying pyat, I'd be keen to help you. You could try renaming the file to `.txt` and see if it lets you attach the file by dragging it into this editor.

Cheers,
Will

willrogers:
Here's the example `dba.mat` renamed.

[dba.txt](https://github.com/atcollab/at/files/2456489/dba.txt)

ZeusMarti:
Ok, thanks @willrogers ...
[a25.txt](https://github.com/atcollab/at/files/2456500/a25.txt)

lfarv:
I cannot reproduce the bug, but the `dab.txt` lattice you sent is 145 elements long, and not 1469 as your 1st message says. Is it the right lattice ?

willrogers:
I added the `dba.txt` as an example. Zeus then added `a25.txt`and that's the one with the bug.

We've got a fix for this, should get a pull request in today.

willrogers:
The pull request #71 should fix the segfault. However, there is still difficulty loading the `.mat` file. That is because pyat requires the 'Class' field to be set on every element in the lattice. The Matlab version does some guessing to try to figure out the correct type.

@lfarv @carmignani what do you think is the correct behaviour? Do you think we should require the 'Class' field on every element?

T-Nicholls:
Hi @ZeusMarti,

The changes from #71 should resolve your original problem. However, as Will mentioned, your a25.mat file still will not load a valid ring, as some of your elements that do not have classes set have pass methods other than 'DriftPass' or 'IdentityPass'. To fix this set the 'Class' field to 'Corrector' on your correctors and to 'Quadrupole'* on your element with FamName of 'QS'.

[...]

PR #71, opened 2018-10-09
Corrector changes

T-Nicholls:
Added a Corrector element class, and a test for it. Also fixed the bug causing atpass to seg fault instead of raising a sensible error.


willrogers:
This should fix the segfault in #70.

PR #73, opened 2018-10-09
Equilibrium emittances

lfarv:
- New "tracking" package
- -C extension to compute the diffusion matrices (shared with Matlab)
- `radiation_on` and `radiation_off` functions to turn acceleration and radiation on/off on a lattice
- `ohmienvelope` function computing equilibrium emittances and damping rates with full 6X6 tracking


willrogers:
It's tricky for me to review all of this - there's a lot of new stuff. I might have to talk it through with an accelerator physicist. The tests show that we haven't changed any of the existing results.

It would be nice to get some more tests, so maybe I'll get my student to write some more.

pyat/at/lattice/lattice_info.py
```diff
@@ -0,0 +1,100 @@
...
+__all__ = ['get_energy', 'get_voltage', 'get_energy_loss']
+
+TWO_PI_ERROR = 1.E-4
+
+
+def get_energy(ring):
+    """
+    energy, periodicity = get_energy(ring) Get the ring energy
+    get_energy looks for the machine energy in:
+        1) the 1st 'RingParam' element
```
willrogers: I don't particularly like the RingParam element idea. It's not part of the lattice and would belong better in a lattice object. I'm happy to keep it for backwards compatibility with existing lattices though.

pyat/at/lattice/lattice_info.py
```diff
@@ -0,0 +1,100 @@
...
+    voltage, harm_number = get_voltage(ring) Get the accelerating voltage
+
+    PARAMETERS
+        ring  Ring structure
+
+    OUPUT
+        voltage    Total RF voltage
+        harmnumber Harmonic number
+    """
+    cavities = list(filter(checktype(elements.RFCavity), ring))
```
willrogers: Some people would say that 

```
cavities = [elem for elem in ring if isinstance(elem, RFCavity)]
```
is more pythonic.

pyat/test/test_atpass.py
```diff
@@ -5,11 +5,11 @@
 
 def test_incorrect_types_raises_value_error(rin):
```
willrogers: This should be

```
def test_incorrect_types_raises_TypeError(rin):
```
now.

willrogers:
Another comment, although not really related to the pull request:

The arguments to the element classes don't have a consistent convention: 

* Some are Python style: `length`, `poly_a`, `limits`
* Some are AT style: `PassMethod`, `MaxOrder`, `NumIntSteps`

I would prefer to change them all to the Python style unless there is difficulty with loading lattices.

pyat/at/lattice/lattice_info.py
lfarv: I agree with the idea of a lattice object, however in MATLAB the RingParam element is still necessary (before it was introduced, AT used global variables, even worse). So let’s keep it for compatibility.

lfarv:
About argument names: the problem is that the effective attribute names are looked for by the integrators, so they have to stay the same (‘Length’, ´PolynomA’ etc.). Changing the argument names implies a translation in the element factory function, while at the moment, except for positional arguments, each keyword argument is automatically stored as an attribute with the same name. In addition, when extracting or setting values in an existing lattice, the user has to know the real attribute name (‘FamName´, Length) in MATLAB style. If you think of an elegant way of sorting this out, let’s look at it !

pyat/at/lattice/lattice_info.py
lfarv: True… Some utility functions in utils.py are probably useless.

pyat/at/lattice/lattice_info.py
T-Nicholls: Excuse me if I'm missing something but, Is it not worth checking at least some of the other elements, rather than raising an error after only checking for RingParams and RFCavities? As an example, our dba ring doesn't have either but it does have Energy attached to all non-drift elements.

lfarv: @T-Nicholls 
short answer: I just copied the Matlab behaviour, but I can look for energy wherever it can be!

full answer: Energy is not an element attribute, so the good practice should be to set it in a single place, Cavity element, possibly inactive, or even better RingParam element. However the `Energy` attribute is necessary for all '*RadPass' passmethods. That's why`at.radiation_on` silently adds it to all radiating elements. Coming back to your example, a ring with radiation but no cavity would not work. And if it is not radiating (no *RadPass pass method in dipoles), the energy attribute is useless, though it is perfectly valid. So following @willrogers's idea, here is what I propose:

I will create a lattice object, such as all existing functions can be kept unchanged, working with any sequence of elements, including the new lattice object. The `get_energy` function disappears, replaced by the `energy` attribute of the lattice object. So functions requiring the energy (like `ohmi_envelope`)will need a lattice object as input rather than a simple sequence. At instantiation, the lattice object will look for energy wherever it is, RingParam, Cavity or anything else.

lfarv: @T-Nicholls 
short answer: I just copied the Matlab behaviour, but I can look for energy wherever it can be!

full answer: Energy is not an element attribute, so the good practice should be to set it in a single place, Cavity element, possibly inactive, or even better RingParam element. However the `Energy` attribute is necessary for all '*RadPass' passmethods. That's why`at.radiation_on` silently adds it to all radiating elements. Coming back to your example, a ring with radiation but no cavity would not work. And if it is not radiating (no *RadPass pass method in dipoles), the energy attribute is useless, though it is perfectly valid. So following @willrogers's idea, here is what I propose:

I will create a lattice object, such as all existing functions can be kept unchanged, working with any sequence of elements, including the new lattice object. The `get_energy` function disappears, replaced by the `energy` attribute of the lattice object. So functions requiring the energy (like `ohmi_envelope`)will need a lattice object as input rather than a simple sequence. At instantiation, the lattice object will look for energy wherever it is, RingParam, Cavity or anything else.

T-Nicholls: @lfarv, thank you for the clear explanation. I think that a lattice object is a good idea, and will be beneficial in other areas too.

lfarv:
Definition of  a `Lattice` object:
- derived from `list`
- accepts extended indexing as `Numpy.ndarray`
- `energy`, `voltage`, `energy_loss` are now properties of the `Lattice` object
- `radiation_on`, `radiation_off` are methods of the `Lattice` object
All existing function still work as before (I hope)

willrogers:
I'm generally happy to merge this, we can always make further tweaks in future. This work is very helpful to us at the moment!

PR #74, opened 2018-10-17
Class Guessing Implementation

T-Nicholls:
Lots of basic logic to hopefully better identify element Classes. This function will misidentify rather than not identify a class, though it is unlikely that this will have an effect on the essential functions or twiss data of the ring. The sanitisation function should also catch most PassMethod/Class incompatibilities that would otherwise raise an error during atpass.


pyat/at/load_mat.py
```diff
@@ -3,23 +3,146 @@
...
 
-FAMILY_MAPPING = {'AP': 'Aperture'}
+PASSMETHOD_MAPPING = {'CorrectorPass': 'Corrector', 'Matrix66Pass': 'M66',
+                      'CavityPass': 'RFCavity', 'QuadLinearPass': 'Quadrupole',
+                      'BendLinearPass': 'Bend', 'AperturePass': 'Aperture',
+                      'ThinMPolePass': 'ThinMultipole', 'DriftPass': 'Drift'}
 
 
-def load_element(element_array):
+def hasattrs(kwargs, *attributes):
```
willrogers: The implementation suggests this will return True if any of the attributes are present in kwargs, but the comment doesn't make that clear.

willrogers:
This is an excellent candidate for some unit tests - it will be easier to tweak the implementation once you have some tests.

T-Nicholls:
I think this is complete. Please ignore my comment about all elements having length, I realise now that they only do after element creation. My comment concerning the misidentifying skew quadrupoles is still valid though. I will remove both comments on my next commit.

pyat/test/conftest.py
```diff
@@ -3,9 +3,26 @@
...
 import pytest
+from at import load_mat
 
 
 @pytest.fixture
 def rin():
     rin = numpy.array(numpy.zeros((6, 1)), order='F')
     return rin
+
+@pytest.fixture(scope='session')
```
willrogers: Why do you need these fixtures? Can't you just use `load_mat.CLASSES` in the tests?

T-Nicholls: I tried it and it didn't work...I can't remember why though. I'll take another look.

T-Nicholls: Apologies, it doesn't work with pytest.mark.parametrize or passing it as an argument without fixtures. It does however work if you just define it at the top of the file and then don't pass it as an argument to the tests. I'll commit this change as it's simpler than using fixtures.

lfarv:
Is this ready for merging ? I would like to take benefit of it !

willrogers:
This looks generally OK. I'm still not sure why you redefine the constants at the top of `test_load_mat.py` - you could just use the constants in the code. Otherwise we can merge this.

T-Nicholls:
Corrected.
**Ready to merge**

PR #78, opened 2018-10-25
Miscellaneous changes

T-Nicholls:
A number of miscellaneous changes that I have compiled whilst using AT, plus the commits from @willrogers' pull request #47.


pyat/at/physics/orbit.py
```diff
@@ -76,7 +76,7 @@ def find_orbit4(ring, dp=0.0, refpts=None, guess=None, **kwargs):
     delta_matrix = numpy.zeros((6, 5), order='F')
     for i in range(4):
         delta_matrix[i, i] = step_size
-    id4 = numpy.asfortranarray(numpy.identity(4))
+    id4 = numpy.asfortranarray(numpy.eye(4))
```
willrogers: I'm not sure why you've changed `identity` to `eye` here.

T-Nicholls: Because it was causing a problem with our threading in atip and identity just imports and calls eye anyway. So I figured there was no downside to changing it.

willrogers: Except it was renamed to `identity` because it is a clearer name. I vote to reinstate it because it is easier to understand.

atmat/atmexall.m
```diff
@@ -5,6 +5,9 @@ function atmexall(varargin)
 % properly configured.
 % On Windows, Microsoft Visual C++ is required
 
+% Add AT directories to the Matlab path
+atpath
+
 PLATFORMOPTION = ['-D',computer,' ',sprintf('%s ',varargin{:})];
 LIBDL='';
 switch computer
```
lfarv: This results in a double definition of all the Matlab path ! It is useless if AT is correctly installed, and it is is not, I do not see how you call `atmexall`. More generally, I do not like modifications of the Matlab path without the user's intervention.

pyat/at/physics/orbit.py
lfarv: I have no problem with `eye` since it's the name for the similar Matlab function, but I let you, python experts decide.

willrogers:
I think you mentioned that Matlab path problem before. The misunderstanding comes because running `atpath` _is_ the way I use AT - perhaps you set the path by default at Matlab startup or something. But I don't mind, we will revert that commit.

lfarv:
Yes Will, you can run `atpath` to setup AT, but then you are responsible to call it. In this case, you know what you are doing, and then the path is lost at the end of the Matlab session, which is a protection against the risk I mentioned. But I note that the 'install' document still does not mention the different ways to correctly setup the Matlab path, either with the GUI, or with the `startup.m` file, or temporarily as you are doing. I also think we already discussed that, but the outcome has been lost…
Here is what I propose for point 4 in `INSTALL.md`

4. Insert recursively the directories _at_installation_/atmat and _at_installation_/atintegrators in the Matlab path. This can be done by
- using the GUI: open the  "Set Path" window, press "Add with subfolders", select _at_installation_/atmat. Repeat the operation for _at_installation_/atintegrators,
 - using the startup.m file: insert a line addpath(genpath(_at_installation_/atmat) and a similar one for atintegrators
 - temporarily modifying the path by running:
```
>> cd  at_installation/atmat
>> atpath

```

T-Nicholls:
New changes:
- Reverted `eye` to `identity`.
- Removed `atpath` from `atmexall.m`.
- Replaced point 3(`Start Matlab.`) with @lfarv's proposed new point in `INSTALL.md` and made small formatting adjustments.
- Updated the pyat readme to avoid strange behaviour relating to .pyc and .so files, and no longer mix `develop` and `install`, as discussed with @willrogers.

Are there any other changes that people would like added?

T-Nicholls:
The Corrector bug caused all correctors to have a length of 0, this caused some discrepancies in twiss data if a lattice had elements of class Correcor with a non-zero length.

lfarv:
Well done ! This corrector bug was also corrected in #76. Coming back to INSTALL.md, point 4 (cd at/atmat) is not necessary. Otherwise, it looks ok for me.

T-Nicholls:
@lfarv @willrogers 
INSTALL.md should now be correct, and following it works as intended for me.
I think it's now ready to merge.

PR #79, opened 2018-10-30
Kick angle in dipoles

carmignani:
I added the field kick angle in BndMPoleSymplectic4Pass and BndMPoleSymplectic4RadPass. So it is easier to  simulate the orbit correctors in the dipoles (as we will have in the esrf upgrade).


carmignani:
The change is like the one done in commit f998af800f201147b42f7a165ea36f309d8c71d9 by @yrtan for StrMPoleSymplectic4Pass.

carmignani:
Ok, I merge.

PR #80, opened 2018-11-02
Matlab tests

lfarv:
Following @T-Nicholls's idea, here is a new generic Matlab function to allow getting Matlab structures in Python. The `linopt` test is updated accordingly. An `ohmi_envelope` test is foreseen.


T-Nicholls:
This is great, I hadn't considered a generalised implementation. I will update my twiss data comparison test to use this syntax when I update my other tests and add RFCavity related tests inline with #76.

lfarv:
Merged into the linopt branch

PR #82, opened 2018-11-06
Fixed matlab bug in atlinopt

lfarv:
Work around a matlab different behaviour on different architectures concerning empty matrices


lfarv:
Problem fixed

Issue #86, opened 2018-11-12
MaxOrder is zero for magnets with zero strength


JonasKallestrup:
Creating an element that is turned off (PolynomB = 0 for quad, sext, etc.), the MaxOrder is automatically set to 0, since max([1 find(abs(elem.PolynomB)>0,1,'last') find(abs(elem.PolynomA)>0,1,'last')])-1 = 0.

If one wants to turn on the element, it is then not enough to make the PolynomB array non-zero; one must also change MaxOrder to the desired value manually, which seems like an unnecessary extra step, which can easily be forgotten. 

Instead, one could check the class of the element in atbaselem, and let MaxOrder = 1 for quads and 2 for sextupoles per default. I guess this may increase computation times somewhere (an element with MaxOrder = 1 is included in calculations, even though PolynomB(2) = 0), but if the number of turned-off elements is low, then it does not matter too much(?).

ZeusMarti:
Right, it happened to me once...

lfarv:
Solved by #93

PR #89, opened 2018-11-14
Suggested solution to MaxOrder issue

JonasKallestrup:
Suggested fix to #86 : 
If a magnetic element, e.g. a quadrupole, is created with zero PolynomB, then MaxOrder is set to 0, and must be changed whenever the element is turned on. 

Solution: Check for the class om atbaselem.

### DIFF
--- a/atmat/lattice/element_creation/atbaselem.m
+++ b/atmat/lattice/element_creation/atbaselem.m
@@ -22,7 +22,17 @@ if any(ab)
     if ~ab(1), elem.PolynomA=[]; end
     if ~ab(2), elem.PolynomB=[]; end
     if ~isfield(elem,'MaxOrder')
-        elem.MaxOrder=max([1 find(abs(elem.PolynomB)>0,1,'last') find(abs(elem.PolynomA)>0,1,'last')])-1;
+        MaxOrder1 = max([1 find(abs(elem.PolynomB)>0,1,'last') find(abs(elem.PolynomA)>0,1,'last')])-1;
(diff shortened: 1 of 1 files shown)


carmignani:
I don't like very much this solution, because you could run atbaselem by itself and it could happen that the field "class" is not there at all.
Then an element with class "Multipole", created with atmultipole, can have PolynomB and A up to the sextupole or even less, so I would definitely remove at least the case "Multipole" in your switch.
Maybe @lfarv can give a comment.

JonasKallestrup:
One could check that the field elem.Class exists. If not, then use the old solution (or even guess the class). 

Regarding multipoles: is it a problem, that MaxOrder = 3, even though that PolynomB(3) = 0 for the multipole; does it cause calculation errors somewhere? 
In principle, would it cause problems if all elements had MaxOrder = 3 - other than increase computational time somewhere in the code?

lfarv:
There is a more elegant solution, but I am away for another 2 weeks with very little network access. I’ll make a proposal when I’m back !

carmignani:
> Regarding multipoles: is it a problem, that MaxOrder = 3, even though that PolynomB(3) = 0 for the multipole; does it cause calculation errors somewhere?

It is a problem if PolynomB(3) does not exist, because you would have a crash of matlab. If it exist and it is 0, the tracking would be a little slower.

We could put this check of MaxOrder in atquadrupole and in atsextupole.

JonasKallestrup:
Solved in https://github.com/atcollab/at/pull/93

PR #90, opened 2018-11-20
Speed up linopt

T-Nicholls:
`linopt` takes significantly longer than `get_twiss`, even when uncoupled; this is because we always calculate `A` and `B`. Now, I think there is an argument to be made that uncoupled `linopt` should return the same as `get_twiss` i.e. don't calculate `A`, `B`, `C` and `gamma`, but regardless why are A & B so slow to calculate? It is because of the extensive use of multi_dot in their calculations, according to the first comment on this question (https://stackoverflow.com/questions/45852228/how-is-numpy-multi-dot-slower-than-numpy-dot#) it is because "multi dot has some setup-phase to optimize ordering". Now this means multi_dot is configured for large irregular matrices, which our 2 by 2 slices of `mstack` are not. Hence why linopt takes so long, because we use multi dot many times on small matrices. So I have replaced the use of multi_dot with the .dot syntax as it is significantly faster for our use case. I have seen a drastic time reduction after this change, with `linopt` taking between 1/4 and 1/5 of the time it used to take, and uncoupled only taking twice as long as `get_twiss` instead of 10 times longer. I am confident that by all measures this change reduces the duration of `linopt` but I have also had some fluctuations in my reference times that I cannot account for, and so I would be grateful if someone else could test this independently. That said here are my times:

Function                | old time | new time
get_twiss               |  0.0606  | N/A
linopt (coupled)     |  1.1385  | 0.2958
linopt (uncoupled) |  0.5366  | 0.1087

### DIFF
--- a/pyat/at/physics/linear.py
+++ b/pyat/at/physics/linear.py
@@ -2,7 +2,6 @@
 Coupled or non-coupled 4x4 linear motion
 """
 import numpy
-from numpy.linalg import multi_dot as md
 from math import sqrt, atan2, pi
 from ..physics import find_orbit4, find_m44, jmat
 from ..lattice import uint32_refpts, get_s_pos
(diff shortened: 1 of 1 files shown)


ZeusMarti:
Hi,

I tested it with our lattice, looks good!

OLD version:
get_twiss eapsed time 0.03600001335144043
linopt coupled elapsed time 0.7630000114440918
linopt elapsed time 0.38499999046325684

NEW version:
get_twiss elapsed time 0.030001163482666016
linopt coupled elapsed time 0.25001001358032227
linopt elapsed time 0.1800072193145752

T-Nicholls:
Thanks Zeus, that is great.

In addition to what I have mentioned above, I would like to note that as I just replaced multi_dot with .dot some of the lines have become more cluttered and harder to understand. So I would welcome anyone who understands the calculations taking place to improve the clarity of this section if they like.

T-Nicholls:
@lfarv I hope you enjoyed your time away.
I would like to draw your attention to this pull request and to #91, as starting from this Friday I am away for Christmas. I will not return until January 14th so there is no rush but please could you consider these changes so that I can continue work on AT when I get back.

PR #91, opened 2018-11-20
Extensive updates to the pyat tests

T-Nicholls:
Notes:
- I have not written any tests for lattice_object because I believe that there will soon be changes made to it.
- I have not written tests for a number of functions which I do not have the physics knowledge to understand; though many lines within these functions are "covered" despite them not having a specific test.
- I have left my test comparing the Python and Matlab outputs for `twiss`, but this could possibly be replaced with an uncoupled `linopt` test if there is a `coupled=False` equivalent in Matlab.
- In the sanitisation test I took out the use of 'Invalid' as a `PassMethod` as separately I would like to propose adding a `PassMethod` check in `sanitise_class`, checking that the pass method is valid for pyat. I'm unsure of how to do this though as a list of all "valid" pass methods would be quite unruly, perhaps a check against the names of the .so pass method files. At the very least I think that if the pass method string does not end in "Pass" an error should be raised.
- In the sanitisation test I used StrMPoleSymplectic4Pass as it was the first pass method not in `PASSMETHOD_MAPPING` that I could think of, better suggestions are welcome. I changed it from invalid because of the factors mentioned above, but I also changed it from `IdentityPass` as that was being caught by the very first "if" checking for elements with `IdentityPass` and a non-zero `Length`.
- I added `PolynomA` to a `hasattrs` check in class guessing to possibly catch `Multipole`s with no `PolynomB` but I'm not sure if any elements would lack it, possibly pure skew elements?
- Some of these tests are just for the sake of coverage and check against values from pyat running on my computer; though theoretically, that should be ok if they differ for anyone let me know. If you think they are overly redundant remove them; I thought I might as well leave them in, plus I like having a high coverage percentage.
- The degree of precision with which I am testing numpy arrays is entirely arbitrary so if there is a convention or greater consistency is required please change them. Edit: this caused some test failures on different platforms so a number of tests are now more consistent with their precision.
- With the addition of the dba lattice in the `test_matlab` directory should `test_physics` look there for its source? Furthermore, should we remove dba from `atmat/atdemos/atmatchExamples/ExampleATMATCH/dba.mat` or rename one of their versions for fear of having to files that should be the same but are out of sync?


T-Nicholls:
Another note, In python versions 3.3, 3.4, 3.5, 3.6 & 3.7 (maybe more, these are just the versions I have seen it in) when the tests are run three tracking error messages are printed to the console, but the tests still pass. This does not happen for python 2.7, I am not sure if this is something worth investigating or not as the tests do not fail. N.B. this does not just happen for me, it occurs on Travis CI and AppVeyor too.
```

============================= test session starts ==============================
platform linux -- Python 3.4.9, pytest-3.9.1, py-1.7.0, pluggy-0.8.0
rootdir: /home/jzs47954/git/at/pyat, inifile:
plugins: cov-2.6.0
collected 128 items                                                            

test/test_atpass.py .............                                        [ 10%]
test/test_basic_elements.py ...................                          [ 25%]
test/test_lattice.py ...................                                 [ 39%]
test/test_load_utils.py ........................................         [ 71%]
test/test_patpass.py ..                                                  [ 72%]
test/test_physics.py ............................                        [ 94%]
test/test_track.py .......                                               [100%]

========================== 128 passed in 1.30 seconds ==========================
Error in tracking element 0.
Error in tracking element 0.
Error in tracking element 0.

```

T-Nicholls:
The failed tests were caused by me measuring to the wrong degree of precision.
I am still unsure why Python 2.7 does not print the `Error in tracking element 0.` messages, but they are caused by three tests that check that errors are raised and so it is working as intended.

T-Nicholls:
1. `test_data_checks` is one of the tests that are primarily there for the sake of more complete coverage, so as mentioned above I do not mind if it is removed. For now, I will just remove the `float` and `int` assertions.
2. This is because I initially wrote the Matlab comparison test for `get_twiss` a while ago, you then wrote the test for `linopt` and introduced a more efficient way of transferring data between Python and Matlab. I updated my test to work with the new transfer method but did not bring it in-line with your `linopt` test style. This was mainly because I had the impression that `get_twiss` would be removed soon anyway, as after my speed improvements there seems to be no benefit in having both uncoupled `linopt` and `get_twiss`? If this is not the case then I am happy to look at those two tests and give them a more consistent structure.
3. Great, I think a consistent standard for `dp` in testing is beneficial, I will change all the tests to use (0.0, -0.01, 0.01).

T-Nicholls:
After discussion with Will, I am going to write a single test for `get_twiss` and `linopt` that tests both using `parametrize` to pass `twissring` or `atlinopt` and the relevant fields.
Since this will keep this pull request open for a little longer, I will also put the changes from #94 on here when appropriate; especially since I don't think there is that much to change.

lfarv:
If you keep this pull request open, there is something else I would mention : it would be nice to use `parametrize` to run the tests with different lattices. I have several test lattices and I am doing this by changing one line in the code, but having it included would be much better. Many results are depending on the presence of errors in the lattice (vertical emittance for instance), and the present one has none. If you agree, I can push another 2 lattices (`.mat` files) in the `test_matlab` directory (a developed lattice with errors, another one without 'Class' fields in the Matlab data to check the conversion). On the other hand I would prefer the outcome of #94 to be merged into `master` before: there should be no problem with that, and I could merge it with by branch, while this pull request would still concern only the `test*` files which I do not touch.

T-Nicholls:
I like the idea of testing on multiple lattices, I do not mind how/where you merge it.
I am happy to keep this pull request just for test related changes. I will put #94 on a separate pull request along with some changes I made to `load/utils.py`; to be merged before your multiple lattices changes, because they may have conflicts if you are testing the mapping. Hopefully not, as my changes are minor and should if anything class more elements correctly.

T-Nicholls:
N.B.
In the Matlab comparison tests `None` was to indicate that data0 (the first field of output) should be tested and `'-'`, or any string, indicated that all `refpts` should be used. 

Firstly I didn't like the `if` `else` for testing data0 on only one iteration of the test; so I removed it and now data0 is tested on each iteration. This meant `None` could be removed from `parametrize`, reducing the number of tests. Now whilst this change is subjective, I think it makes the actual test code cleaner; also, since the number of tests reduced, the overall runtime of the new Matlab comparison tests is roughly the same as that of the old.

I like the idea of an all `refpts` flag, both in the tests and in the main pyat code base, but I am unsure of using the 'any string' syntax. As `None` was no longer being used for the aforementioned reasons I have swapped to using it to mean all instead of `'-'`. I do not like the misnomer of using `None` to mean all, but in this case, as it is just tests I think it is fine. That said I am unsure if it is any more or less intuitive than before.

I have put an "all refpoints" flag on my list of things to consider how to nicely implement, as if done well I think it could be of benefit to pyat.

T-Nicholls:
Otherwise I think my changes should be rather uncontroversial. This also concludes all the test related changes I wish to make, and so I am now happy for this to be merged at your convenience.

T-Nicholls:
I would like to wait for #95 to be merged so that I can extend the test for `bool_repts` added here to cover the extra new behaviour added on #95.

lfarv:
I pushed 2 other lattices so that you can run the tests on 3 different lattices.

lfarv:
@T-Nicholls : I don't understand the rationale behind the changes in `load/utils.py`. In Matlab, official class names (those in the `CLASSES` variable) are case-sensitive and are always checked as such in the Matlab code. So checking the belonging to a set made sense. For other class names, anything can be tried… But I see 2 possibilities:

1. A strict control: either the class is an official one (`class_name` belongs to the `CLASS` set, where `Bend` is missing), or you fall back to the case where there is no class name, with possibly a warning like "Missing or invalid class name". The error "Class does not exist" disappears, one always tries to guess.

2. Anything may happen: then rather than looking in a row in 2 directories (`CLASS` and `CLASS_MAPPING`), why not grouping them in a single one, which can also be used later for checking `fam_name` ? Then a `try:/except KeyError:` would be more elegant than checking the belonging (by the way, `class_name in CLASS` is enough, you don't need to mention `keys()`).  Again, if checking `class_name` fails, one can try guessing from the parameters rather than failing.

A warning for invalid or missing class names is useful for detecting ill-formed Matlab files (which is not our goal here, but has its interest), but there should be an option to silence it because it may appear thousands of times in large files. Something like a `Quiet` boolean keyword argument defaulting to `False`.

In the files I uploaded, `errnoclass.mat` is the same as `err.mat`, but I removed all class names. I would be also interesting to run the tests with the file with strange classes that you used to tune your code (I don't have it).

T-Nicholls:
If I take a look at the ring from mml here at Diamond the `Class` attribute/field is fully capitalised e.g. `DRIFT`, now if we try to load it with our current class guessing it fails (`AttributeError: Class DRIFT does not exist.`). This was not a significant problem as I have a script in Matlab to correctly capitalise the elements and make other changes to the ring so it can be loaded into AT. I thought that doing this conversion in AT would be sensible, do we do things weirdly here at Diamond or am I missing something?

Regardless of case-sensitivity, I am more in favour of the 2nd possibility that you mentioned; as for the possibility of changing the current error to a warning and attempting to guess for elements with an invalid original class attribute, if you think it is more useful then I am happy with it.

I will take a look at your new lattices; as for lattices with strange classes, the only lattices I have tested it on directly are Zeus' lattice, the lattices in AT, and the different configurations of the Diamond lattice.

Here is the Diamond lattice, discussed above, before I have made any changes to it (renamed to .txt so it could be uploaded):
[vmx_original.txt](https://github.com/atcollab/at/files/2812454/vmx_original.txt)

T-Nicholls:
@lfarv, I've made a stab at what I think you were proposing; though you will likely want to propose changes I thought I'd make a first attempt.

lfarv:
@T-Nicholls,

1. `find_class_name` looks good,
2. I had a look to the Diamond lattice:
- in Matlab, tracking fails because of 2 unknown pass methods (ThinCorrectorPass and GWigSymplecticPass), which I replaced by CorrectorPass and Driftpass.  One consequence of non-standard class names is the fact that the elements are not drawn in the lattice plot (`atplot(RING)`).
- in Python, conversion fails because wigglers are converted into drifts (which is correct since the pass method is unknown), but then the GWigSymplecticPass is considered conflicting with a drift (which seems also correct). So the behaviour looks right…

lfarv:
Ok for me, but what about parametrizing the lattice file ?

T-Nicholls:
Ahh yes, I will take a look at implementing that tomorrow. I think a command line option for pytest would be the best way to do it, but I am open to input.

lfarv:
A command line option is a solution, but a `parametrize` has a few advantages: it runs all the tests automatically, and it allows to select the choice of lattices according to each test: some lattices may be more useful to test parsing, while others are more useful for optics… Some lattices can also be designed to trigger errors.

But this can be a second step !

T-Nicholls:
My main motivation behind letting the user specify what lattice(s) should be used was time, but I have just realised that the tests currently only take <2s. So, for now, I will make the tests always run on all lattices for now and we can choose specific lattices for specific tests at a later date.

T-Nicholls:
Disregard my previous comment, I have now taken a look at all the tests and the current behaviour is:

- In the normal Python tests we only use a loaded ring in `test_physics` and we already switch between `dba.mat` and `hmba.mat` for functions that do and do not require a cavity/radiation in the lattice.
- In `test_matlab` we currently just use `hmba.mat` for everything.

What I will do for now is use `parametrize` to run the Matlab comparison tests on both lattices where applicable. I will leave the Python tests as they are for now, but in the future I propose we add a variety of new tests both checking the loading from a number of `.mat` files and testing functions on multiple lattices.

willrogers:
My view: the Python tests should be as fast as possible. If they get slower it either disrupts development or discourages from running the tests.

The Matlab comparison tests are already quite slow, and I don't think it matters as much. These will be run less often.

T-Nicholls:
@lfarv please could we get your opinion on this pull request, are you happy for it to merge now?

lfarv:
Tested and running as expected ! A few comments, which don't prevent from merging as soon as you want:

- the time taken for the Matlab tests is dominated by the time for starting the Matlab process itself. In my setup, the time for actually running the tests is so short that adding some other tests is not a problem,
- is it possible to use the `scope='session'` argument in the `ml_dba` and similar fixtures? Or does the `lazy_fixture` function ensures that the files are loaded only once? I don't know what's happening at the moment, but this is a way to save time,
- running the Matlab comparison tests on `err.mat` will add more confidence in the results because of the errors introduced on all magnets, resulting non-negligible vertical emittance.

So go on with merging! I am very slow on my side on the lattice manipulations. My goal is to prepare plotting, which requires splitting the lattice in very short elements to give a smooth plot. Sorry for that (winter in the Alps requires quite some time on the slopes…)

willrogers:
@lfarv thanks, I agree with all the comments including skiing. Thanks for your help.

We have a potential request from the accelerator physicists here (perhaps you know of Johan Bengtsson) for an interactive gui that 

* allows you to change parameters on elements so that you can see plots of the beam parameters update
* allows you to optimise and match parameters between cells etc (I guess this is like the atmatch part of the Matlab code)

The examples he has are 'OPA' and 'Beta', but neither of these are stable or supported. If you are interested I could send you a copy of our draft specification (when it exists). It may be relevant to how you plan to implement plotting.

T-Nicholls:
@lfarv, you are correct about `scope='session'`, I will it in my next commit to AT.
My apologies I misunderstood, I thought that `err.mat` was a lattice designed to cause errors and so I did not include it. I will also look at adding it into the tests in the near future.
Thank you, and I hope you enjoy your time on the pistes; I am very jealous.

T-Nicholls:
So I have taken a look at `err.mat` and it raises an error during loading. It appears that elements 653, 1645, 2122, 2887, 3124, 3417, 3707, 3999; or all `Marker`s with a `FamName` starting "SBM-BM"; have a non-zero, and negative, `Length` whilst using `IdentityPass`. All 8 elements also have roughly the same value (-1.2e-5); however, when I changed these to 0 the lattice loaded correctly and the comparison tests passed. I am now re-confused as to the purpose of `err.mat`, so please could you clarify its intended use.

lfarv:
Ok @T-Nicholls, I see the problem. I did not notice it because the `IdentityPass` method ignores the length, and I tried it when the loading tests were not so stringent. I reported the bug to my ESRF friends: it comes from rounding errors in manipulating the lattice. I shall push a revised version soon, but your correction with length 0 is perfectly right. Now for the intended use:

Unlike the other lattices (`dba.mat` and `hmba.mat`), this lattice has alignment and rotation errors in the magnets (`T1`, `T2`, `R1` and `R2` fields). This results in a non-zero closed orbit, in distortions of the optics, in generation of vertical emittance, etc. The comparison with Matlab allows the validation of:
- closed orbit computation,
- optics distortion in tune, beta-functions, chromaticity,
- vertical emittance computation,

since none of these effects appear on the other lattices. So the success of the comparison tests you report is meaningful !

T-Nicholls:
@lfarv, if you are working on a revised version of `err.mat` might I suggest shortening the lattice. At 4176 elements it is significantly longer than our other test lattices; this, combined with the fact that linopt takes roughly 7.5 times longer than on an equivalent length lattice without errors, has resulted in the comparison tests taking several minutes to run. It is also worth noting that there are some small inconsistencies between the chromaticities and twiss/lin data between Matlab and Python, the other physics data comparison tests and the atpass comparison passed ok. Overall I think testing on a lattice with alignment and rotation errors is good, but as Will mentioned I think we should be conscious of the runtime of the tests.

PR #93, opened 2018-12-06
Set 'MaxOrder' for quadrupoles with zero strength

lfarv:
Proposal for solve #86


carmignani:
Ok, I think it is a good solution.
@JonasKallestrup

JonasKallestrup:
I agree with @carmignani - this resolves the problem well.

Issue #94, opened 2019-01-18
Refpts handling not working as intended?


T-Nicholls:
**TL; DR:**
How `refpts` are handled is inconsistent at best and erroneous at worst; furthermore, the docstrings only complicate things by contradicting each other. Please can someone outline the intended use syntax for `refpts`?

I have discovered an issue with the consistency of `refpts`, although refpoint related this is different to the issue discussed on #75.
The issue is primarily due to there being no max length check in all of our `refpts` handling, and the inconsistency/ambiguity of the docstrings of functions that use `refpts`.

**On the first point:**

Since, in `uint32_refpts`, we currently raise a ValueError if numerical `refpts` contain duplicates or are not in ascending order, therefore it would be intuitive to have a maximum length check.
Summary of behaviour:
```
============================================================================================
| refpts =                                | returns / raises                               |
============================================================================================
| numpy.array([0,...,2*len(ring)],        | Error raised later in function due to indexing |
|             dtype=uint32)               |  / calculation issues, or data of length       |
|                                         |  2*len(ring) is sucessfully returned.          |
--------------------------------------------------------------------------------------------
| numpy.ones(len(ring), dtype=bool)       | data at entrance to every element in the ring  |
--------------------------------------------------------------------------------------------
| numpy.ones(len(ring)+1, dtype=bool)     | above + data at the non-existant len(ring)+1th |
|                                         | element                                        |
--------------------------------------------------------------------------------------------
| numpy.ones(2*len(ring), dtype=bool)     | ValueError: refpts must be ascending           |
--------------------------------------------------------------------------------------------
| [0, 1, ..., len(ring)-2, len(ring)-1]   | data at entrance to every element in the ring  |
--------------------------------------------------------------------------------------------
| [0, 1, ..., len(ring)-1, len(ring)]     | above + data at the non-existant len(ring)+1th |
|                                         |  element                                       |
--------------------------------------------------------------------------------------------
| [0, 1, ..., 2*len(ring)-1, 2*len(ring)] | ValueError: refpts must be ascending           |
--------------------------------------------------------------------------------------------
| [2*len(ring)]                           | data at entrance to ring i.e. eqivelent to     |
|                                         |  putting refpts=0                              |
============================================================================================
```
The first isssue is that any array with `dtype=numpy.uint32` remains unparsed and so either causes issues in calculations or indexing later on in the function, or even worse it returns data at elements beyond the end of the lattice. The behaviour differs by function but is consistent inside each function.
[...]

PR #95, opened 2019-01-23
Update to refpts docstrings and handling

T-Nicholls:
The changes discussed on #94.
Minor notes:

- I may have put a little more checking logic into `uint32_refpts` and `bool_refpts` than you may want @lfarv; so let me know if you think any of it is overly complex or doesn't provide a meaningful benefit. I tried to make it as thorough as possible within the bounds we agreed, whilst keeping it simple.
- On that note, negative and wrap-around indexing remain available to the user; however, in their current forms, I am worried about one type of use case. Since when mixing indexing formats the `refpts`  must be provided in an unintuitive order. e.g. the intuitively ordered `[-1, 0, 2, 3, 5, 6]` would raise `ValueError('refpts should be given in ascending order')` and instead would have to be ordered like: `[0, 6, 2, 3, -1, 5]`.
- In some places, when I was adding the new refpts explanation, I also added a few other things to the docstrings; I did not fully complete the docstrings though. So, whilst better than before, they do not necessarily give a full explanation; this is because I intend to look at improving docstrings across pyat in the future.

### DIFF
--- a/pyat/at/lattice/utils.py
+++ b/pyat/at/lattice/utils.py
@@ -4,8 +4,13 @@ Helper functions for working with AT lattices.
 A lattice as understood by pyAT is any sequence of elements.  These functions
 are useful for working with these sequences.
 
-The refpts functions allow selecting a number of points in a lattice.
-Indexing runs from zero (the start of the first element) to n_elements (the end of the last element).
+The refpts allow functions to select points in the lattice, returned values are
+given at the entrance of each element specified in refpts.
+refpts can be: an integer in the range [-len(ring), len(ring)-1] selecting the
+element according to python indexing rules. As a special case, len(ring) is
+allowed and refers to the end of the last element, an ordered list of such
+integers without duplicates, a numpy array of booleans of maximum length
+len(ring)+1, where selected elements are True.
 """
 import numpy
 import itertools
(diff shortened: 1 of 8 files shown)


lfarv:
on bullet 1: a single remark on line 38 of 
`      refs = numpy.flatnonzero(refs)
`
This result fulfills automatically all the conditions tested later (non-negative, ascending, no duplicates). So 
`      return numpy.flatnonzero(refs)
`
Should be enough.
Otherwise, I'm happy with the modifications.

on bullet 2: you are right, but a user of negative indices should be aware of the meaning in python, and practically, the main usage is probably [-1] to reach the last element, which is simple enough. It looks good enough for me…

So go on, Tobyn

T-Nicholls:
Good point, done.
You are right that it is an infrequent use case and that most users will be aware of the correct usage, and so I am happy for this to be merged now.

T-Nicholls:
I do not know what is causing the build failure on Travis for Python 2 on OSX, but it is caused by some external change made recently; it's nothing to do with my changes, Travis builds of `master` now fail too.

T-Nicholls:
@willrogers has suggested a few improvements to the docstrings for clarity, as well as the addition of new tests to fully check the behaviour of `uint32_refpts`. So, I will make those changes and push them to this pull request today or tomorrow.

PR #96, opened 2019-02-10
Lattice and Element objects

lfarv:
Main changes in this pull request:

**Improvements of the `Lattice` object**

- `__add__` and `__mul__` methods
- More efficient slicing

**Improvements of the `Element` object**
Manipulations in preparation for plotting:

- Element.divide method
- Drift.insert method

**Improvements in physics function**

- The output of `linopt`, `ohmi_envelope` etc. is now a numpy ‘record array’ which allows to extract the different values as either:
`lindata[‘beta’]`, as before, or
`lindata.beta`, with attribute syntax, more convenient and similar to Matlab
- new function `get_mcf` to compute the momentum compaction factor

**Other changes**

- new `test_parameters` Matlab test for checking the circumference, energy, voltage, periodicity, harmonic number, energy loss/turn and momentum compaction factor,
- improved doc strings, corrected typos,
- line length limited to 80 in the modified files.


T-Nicholls:
These are some good changes; I am particularly happy about the switch to record arrays and mapping `K` to `PolynomB[0]`! I have also been working on a few updates to lattice object and so have a few additional changes I'd like to propose.

pyat/at/lattice/lattice_object.py
```diff
@@ -6,104 +6,154 @@
...
+    An AT lattice is a sequence of AT elements.
+    This object accepts extended indexing"""
 
-    _translate = dict(Energy='energy', Periodicity='periodicity', FamName='name')
+    _translate = {'Energy': 'energy', 'Periodicity': 'periodicity',
+                  'FamName': 'name'}
     _ignore = {'PassMethod', 'Length'}
 
-    def __init__(self, descr=None, **kwargs):
+    def __init__(self, descr=None, _scan=True, **kwargs):
```
T-Nicholls: Could we give `descr` a more informative name please, maybe `elements`? It would also be nice to list more of the arguments here i.e. `def __init__(self, elements=None, name='lattice', energy=None, periodicity=None, keep_all=False, **kwargs):`

pyat/at/lattice/lattice_object.py
```diff
@@ -170,7 +226,8 @@ def checkdipole(elem):
             n = mod_elem(self, checktype(elements.Quadrupole), quadrupole_func)
             print('{0} modified quadrupoles'.format(n))
```
T-Nicholls: Can we please either remove these print statements or have some way of suppressing them as they are being very frequently spammed in ATIP.

pyat/at/lattice/lattice_object.py
```diff
@@ -192,18 +249,21 @@ def repfunc(pass_method):
...
                     if not elem.PassMethod.endswith('RadPass'):
-                        elem.PassMethod = ''.join((elem.PassMethod[:-4], 'RadPass'))
+                        elem.PassMethod = ''.join(
+                            (elem.PassMethod[:-4], 'RadPass'))
                         elem.Energy = self.energy
             else:
                 def ff(elem):
                     elem.PassMethod = pass_method
                     elem.Energy = self.energy
             return ff
```
T-Nicholls: I think we should warn if the user switches the radiation on when it is already on, i.e.
```
if self._radiation:
    warn(AtWarning('Radiation already on, switching it on anyway.'))
```

pyat/at/lattice/lattice_object.py
```diff
@@ -225,40 +285,17 @@ def repfunc(pass_method):
             elif pass_method == 'auto':
                 def ff(elem):
                     if elem.PassMethod.endswith('RadPass'):
-                        elem.PassMethod = ''.join((elem.PassMethod[:-7], 'Pass'))
+                        elem.PassMethod = ''.join(
+                            (elem.PassMethod[:-7], 'Pass'))
             else:
                 def ff(elem):
                     elem.PassMethod = pass_method
             return ff
```
T-Nicholls: Likewise:
```
if not self._radiation:
    warn(AtWarning('Radiation already off, switching it off anyway.'))
```

pyat/at/lattice/lattice_object.py
```diff
@@ -274,19 +311,17 @@ def ohmi_envelope(self, *args, **kwargs):
...
 
+Lattice.get_s_pos = get_s_pos
+Lattice.get_mcf = get_mcf
+Lattice.find_orbit4 = find_orbit4
+Lattice.find_sync_orbit = find_sync_orbit
+Lattice.find_orbit6 = find_orbit6
+Lattice.find_m44 = find_m44
+Lattice.find_m66 = find_m66
+
 if sys.version_info < (3, 0):
```
T-Nicholls: I believe that if we move these to the bottom of `__init__` then we can do it the same way for all python versions:
```
def __init__():
    ...
    ...
    self.linopt.__func__.__doc__ += linopt.__doc__
    self.ohmi_envelope.__func__.__doc__ += ohmi_envelope.__doc__
```

pyat/at/load/utils.py
```diff
@@ -166,6 +166,6 @@ def error_message(message, *args):
         sanitise_class(index, class_name, elem_dict)
     cl = getattr(elements, class_name)
     # Remove mandatory attributes from the keyword arguments.
-    elem_args = [elem_dict.pop(attr) for attr in cl.REQUIRED_ATTRIBUTES]
-    element = cl(*elem_args, **elem_dict)
+    elem_args = (elem_dict.pop(attr, None) for attr in cl.REQUIRED_ATTRIBUTES)
+    element = cl(*(el for el in elem_args if el is not None), **elem_dict)
```
T-Nicholls: I think that `arg` would be a more appropriate name than `el` here.

pyat/at/physics/radiation.py
```diff
@@ -31,36 +31,39 @@ def ohmi_envelope(ring, refpts=None, orbit=None, keep_lattice=False):
...
-        R66                 (6, 6) equilibrium envelope matrix R
-        R44                 (4, 4) betatron emittance matrix (dpp = 0)
-        T66                 (6, 6) transfer matrix from the start of the ring
+        emit is a record array with fields:
+        r66                 (6, 6) equilibrium envelope matrix R
+        r44                 (4, 4) betatron emittance matrix (dpp = 0)
+        m66                 (6, 6) transfer matrix from the start of the ring
         orbit6              (6,) closed orbit
         emitXY              betatron emittance projected on xxp and yyp
         emitXYZ             6x6 emittance projected on xxp, yyp, ldp
```
T-Nicholls: Specify sizes for `emitXY` and `emitXYZ`, I think this is especially important for `emitXYZ` as otherwise the "6x6" may cause confusion.
```
emitXY               (2,) betatron emittance projected on xxp and yyp
emitXYZ             (3,) 6x6 emittance projected on xxp, yyp, ldp
```

pyat/at/lattice/elements.py
```diff
@@ -174,21 +357,42 @@ def __init__(self, family_name, length, BendingAngle, k=0.0, EntranceAngle=0.0,
...
+        FringeQuadEntrance  0: no fringe fiels effect (default)
+                            1: Lee-Whiting's thin lens limit formula
+                            2: elegant-like
+        FringeQuadExit
+        fringeIntM0     Integrals for FringeQuad method 2
+        fringeIntP0
+        KickAngle       Correction deviation angles (H, V)
         """
         poly_b = kwargs.pop('PolynomB', numpy.array([0, k]))
+        _ = kwargs.pop('K', None)
```
T-Nicholls: Same thing here?

T-Nicholls:
My ideas for lattice object only really differed from yours on the following:
- I was inheriting directly from object and storing the list of elements on a private attribute as this is the way we do it in Pytac. The only significant behavioural difference is that we return a list from `__getitem__` rather than having to create a new lattice object to return. I'm guessing that is why you added the _scan argument which I am opposed to since it could result in an incomplete lattice if passed by a user. When the ring should be scanned should be determined by the status of `energy` etc. perhaps something like: `if None not in [energy, periodicity, radiation]:`

Which brings me to things I forgot in the review:
- I propose a `radiation` argument so the radiation state of the lattice can be specified, also could we change `_radiation_on` to just `_radiation` so as to avoid confusion with the `radiation_on` method.
- If no `name` is passed is would it be reasonable to take the name off of the `RingParam`?

A few things Will mentioned in discussions on this topic:
- A `radiation_status` method on the lattice to access the private `_radiation` attribute.
- A method to check the lattice for any inconsistencies; differences between elements of energy, radiation status, periodicity or multiple `RingParam`s in the lattice. I am currently working on a draft of this.

I am sure I will have missed something and so I might try and merge your changes into my version of `lattice_object.py` to check for inconsistencies.

pyat/at/lattice/elements.py
```diff
@@ -61,10 +66,94 @@ def differ(v1, v2):
...
+
+        Return a list of elements equivalent to the original.
+
+        Example:
+
+        >>> Drift('dr', 0.5).divide([0.2, 0.6, 0.2])
+        [Drift('dr', 0.1), Drift('dr', 0.3), Drift('dr', 0.1)]
+        """
+        # by default, the element is indivisible
+        return [self]
```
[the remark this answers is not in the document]
lfarv: For plotting, I need to divide all elements in small pieces to get a smooth plot. Doing this, I'll iterate through all elements and call the divide method. I do not want to test if each element has a divide method. Further more, some long elements like IDs may be difficult to split. So it makes sense for an element to return itself if it is indivisible!

pyat/at/lattice/elements.py
```diff
@@ -91,81 +180,175 @@ def __init__(self, family_name, limits, **kwargs):
...
+                        with 2 objects
+                        The 1st object is the location of the elements to
+                        be inserted, given as a fraction of the Drift length.
+                        The 2nd object is an element to be inserted at that
+                        location. If None, self will be divided but no element
+                        will be inserted.
+
+        Return a list of elements.
+
+        Drifts with negative lengths may be generated if necessary.
```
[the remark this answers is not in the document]
lfarv: Negative drifts are very useful in modelling! For example, one way to adjust the magnetic length of a quadrupole with its strength consists in surrounding the quadrupole with drifts corresponding to its mechanical length, plus variable negative drifts adjusted together with the quadrupole length as a function of its strength. So I would keep negative drifts, however I agree that `frac`, the location, should be constrained between 0 and 1.

pyat/at/lattice/elements.py
```diff
@@ -91,81 +180,175 @@ def __init__(self, family_name, limits, **kwargs):
...
         """
         kwargs.setdefault('PassMethod', 'DriftPass')
-        super(Drift, self).__init__(family_name, Length=kwargs.pop('Length', length), **kwargs)
+        super(Drift, self).__init__(family_name,
+                                    Length=kwargs.pop('Length', length),
+                                    **kwargs)
+
+    def insert(self, insert_list):
+        """insert elements inside a drift
+
```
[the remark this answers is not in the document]
lfarv: OK, I'll try and be more explicit !

pyat/at/lattice/elements.py
```diff
@@ -91,81 +180,175 @@ def __init__(self, family_name, limits, **kwargs):
...
+        FringeQuadEntrance  0: no fringe fiels effect (default)
+                            1: Lee-Whiting's thin lens limit formula
+                            2: elegant-like
+        FringeQuadExit
+        fringeIntM0     Integrals for FringeQuad method 2
+        fringeIntP0
+        KickAngle       Correction deviation angles (H, V)
         """
         poly_b = kwargs.pop('PolynomB', numpy.array([0, k]))
+        _ = kwargs.pop('K', None)
```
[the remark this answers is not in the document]
lfarv: `K` was required in Matlab, for some `PassMethod`s, and possibly conflicting with `PolynomB[1]`. It should not appear any more as a Python attribute. I added that when I got `K`s left over after converting a Matlab lattice. I'll check if it is still necessary and if not. remove the line…

pyat/at/lattice/lattice_object.py
```diff
@@ -6,104 +6,154 @@
 from warnings import warn
 from math import pi
 from . import elements, checktype, AtWarning, AtError
-from ..physics import find_orbit4, find_orbit6, find_sync_orbit, find_m44, find_m66, linopt, ohmi_envelope
+from ..lattice import get_s_pos
+from ..physics import find_orbit4, find_orbit6, find_sync_orbit, find_m44, \
+    find_m66, linopt, ohmi_envelope, get_mcf
```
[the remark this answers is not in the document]
lfarv: I'll propose something, and let you refine it if necessary. What I would like is to refer imports with respect to their package (`at.lattice`, `at.physics`, `at.load`, `at.tracking`), but not to their module (`at.lattice.utils`), leaving the possibility to move them inside the package if needed. For me, only the packages should be public, their internal components should not be.

pyat/at/lattice/lattice_object.py
```diff
@@ -6,104 +6,154 @@
 from warnings import warn
 from math import pi
```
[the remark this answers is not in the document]
lfarv: I'll do that!

pyat/at/lattice/lattice_object.py
lfarv: You are right, `elements` looks more adequate.

pyat/at/lattice/lattice_object.py
lfarv: I can remove those prints!

pyat/at/lattice/lattice_object.py
```diff
@@ -6,104 +6,154 @@
...
+from ..physics import find_orbit4, find_orbit6, find_sync_orbit, find_m44, \
+    find_m66, linopt, ohmi_envelope, get_mcf
 
 __all__ = ['Lattice']
 
 TWO_PI_ERROR = 1.E-4
 
 
+def _scan_list(descr, keep_all=False, **kwargs):
+    """Extract the lattice attributes from an element list"""
```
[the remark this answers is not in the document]
lfarv: OK, I can look at that!

pyat/at/physics/radiation.py
lfarv: Good remark!

lfarv:
@T-Nicholls : thanks for your careful review. It's very useful for improving the code quality! Can you tell me how you can highlight and comment a few lines of code? I tried and never succeeded… I'll answer you different points step by step, so this concerns radiation:

I initially thought of using a `radiation` boolean property to set the radiation state of the lattice. But in fact turning radiation on may mean different things:
- turning longitudinal motion on: this means turning only cavities on. By the way, `radiation_on` is misleading in that case since there is no synchrotron radiation involved (and this is the only thing to do on hadron machines which do not radiate),
- turning only radiation without longitudinal motion (no cavity, the beam decelerates). There are several possibilities depending whether you want radiation in the quads (usually neglected), in other multipoles or not…,
- turning on everything together, which is the most common use for lepton machines and is the default behaviour at the moment.

This is why we need a method with several arguments and not a property. A `radiation` argument has the same problem: what exactly do you want to turn on?

So I will change the private attribute `_radiation_on` to `_radiation`, and create a boolean read-only property `radiation` to check the status, but no argument…
And if you have name less ambiguous than radiation to refer to all this, feel free to propose. But I would keep a single name and method for everything, for simplicity and consistency with Matlab.

lfarv:
Now, concerning the inheritance of the `Lattice` object:
I chose to inherit from `list` rather than embedding for 2 reasons;

- On the principle: for me a lattice does not CONTAIN a sequence of elements, it IS a sequence of elements with some additional metadata. It's how one usually answers the choice between embedding and inheriting.
- practically: to get the sequence interface, one needs to re-implement ALL the magic methods (`__len__`, `__contains__`, `__iter__`, `__getitem__`, `__setitem__`,  `__add__`,…), so that a lattice object can be indexed, iterated on, enumerated, zipped etc.. This is what the standard `collections.UserList` class is doing. It is mentioned that with new classes in python, inheriting directly from `list` is the preferred way. However it is the best choice if embedding is selected. 

Now why did I re-implement a few of those methods? There is a bad reason and a good one;

- The bad reason: I tried to return a `Lattice` object for the list manipulation instead of a pure `list` in an attempt to preserve the metadata. After thinking, it appears that only `energy` is worth keeping. For a subset of a lattice, `periodicity` and `name` make no sense. So returning a `list` and casting later to a `Lattice` is as simple. Let me give a simple example: I have a periodic lattice as a `Lattice` object `ring` describing one super-period and with `periodicity` set to 32. If now I want to break the periodicity for some reason, I need to expand the lattice into a single long super-period. With the present object, I would write
```
expanded_ring = ring*32 # this is a new lattice object
expanded_ring.periodicity = 1
expanded_ring.name = 'Expanded ring'
```
If I do not reimplement `__mul__`, `ring*32` would be a list and I would write:
```
expanded_ring = Lattice(ring*32, energy=ring.energy, periodicity=1, name='Expanded ring')
```
This may look even better. So I propose to go back to returning list for all manipulations of the sequence of elements.
- The good reason (concerning only `__getitem__`): I want to implement extended indexing as in `numpy` arrays. so that `ring[refpts]` is a subset of `ring`, for any valid `refpts`. This allows things like:
```
for quadrupole, strength in zip(ring[quads], quad_strengths):
   quadrupole.K = strength

```
`__getitem__` returning a pure `list` is OK for that.

To make this easier, I have a version of `Lattice` inheriting from a `numpy` array of objects instead of a `list`. Advantage: this behaviour is automatically inherited. Drawback: inheriting from a `numpy` array is tricky, and some useful methods of lists (`append`, `extend`, + and * operators…) need to be re-implemented.

As a conclusion, I see 3 possibilities, in preference order:

1. inheriting from `list` (with the modification I mentioned)
2. embedding the list by inheriting from `UserList` (inheriting directly from `object` is a waste of effort)
3. inheriting from `numpy.ndarray`

pyat/at/lattice/elements.py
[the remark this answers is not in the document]
T-Nicholls: Ok, that makes sense for that use.

pyat/at/lattice/elements.py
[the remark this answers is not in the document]
T-Nicholls: I had not considered that possibility.

pyat/at/lattice/elements.py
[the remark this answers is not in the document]
T-Nicholls: That is very strange, if an element has the required attribute it should be popped in `load.utils` by `elem_args = (elem_dict.pop(attr, None) for attr in cl.REQUIRED_ATTRIBUTES)`

pyat/at/lattice/lattice_object.py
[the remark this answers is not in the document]
T-Nicholls: I would be happy with that system.

T-Nicholls:
Thank you. In Files changed, if you click on the blue plus, that pops up when you hover over a line of code, it lets you comment on that line; it also includes a few lines above the selected line, but I am not sure how GitHub determines how many lines about to include.

I see the problem, I had been considering radiation in a much more all or nothing way, in fact I had gone so far as to enforce the radiation state during initialisation in my draft of lattice object, i.e. if one element has `RadPass` then call `radiation_on` to ensure they are all on. Also, since the radiation state of the lattice is not as discrete as I had imagined, I do not think my suggestion for warnings is as warranted, I think in this case the public boolean is sufficient. Overall I do agree that it is difficult to accurately represent the situation concisely, but I do not have any ideas better than yours.

On inheritance vs embedding, strictly speaking if you are inheriting then returning the same type as self is the right correct thing to do, and so as my main concern was about returning a lattice object containing a small number of elements, I chose embedding as to me a lattice object implies it is a complete `Lattice`. I think that the benefits gained by inheriting from `numpy.ndarray` are outweighed by having to re-implement the list methods; As such I am happy with inheriting from `list` and re-implementing `__getitem__` with extended indexing that returns a list. Side note, one possibility for preserving `energy` is to add it as an attribute to all elements during lattice creation, the Diamond lattice already has `Energy` on every element, this option would also open up the possibility of `ohmi_envelope` accepting rings other than lattice objects.

pyat/at/lattice/lattice_object.py
lfarv: No, it does not work in Python 3
```
>       Lattice.linopt.__func__.__doc__ += linopt.__doc__
E       AttributeError: 'function' object has no attribute '__func__'

```

lfarv: I don't really understand what's happening, but it works…

pyat/at/lattice/elements.py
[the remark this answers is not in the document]
lfarv: You are right, I don't get unexpected `K`'s anymore, it was probably remaining from a previous version!

pyat/at/lattice/lattice_object.py
T-Nicholls: Ok, that makes sense.

lfarv:
There is now a single scan in `Lattice.__init__` to collect energy, periodicity, name and radiation status. This is only done if one of  `[energy, periodicity, name]` is not provided as a keyword argument. Otherwise, a simpler scan is still performed to get the radiation status. My question is: should we really separate the 2 cases, or always perform the full scan which should not take much longer?

I recall that the "radiation status" (`Lattice.radiation`) means "There is at least one element changing the particle energy", and this should prevent from calling `linopt` (though many users want to do it anyway, since the results are almost correct).

I think that all other remarks have been answered, if not tell me.

lfarv:
I pushed corrected versions of `err.mat` and `errnoclass.mat`. I do notice a very long time when trying to include them in the tests (~1 mn on my machine for only the `ohmi_envelope` tests), which I do not understand since any interactive call to `ohmi_envelope` with the same arguments take less than 1s, both in Python and Matlab…

T-Nicholls:
I think the new separated scan behaviour is fine as it acts as if you specify `[energy, periodicity, name]` then the other `RingParam` attributes are not set on the lattice object, whereas if you don't specify all of them then the other `RingParam` attributes are set to the lattice object; I think this is a useful behaviour.

In `load/utils.py` is there ever a case where we can't pass `index` to `element_from_dict`? If there isn't we should remove `None` as and have no default, and then we could simplify `err_message` slightly.

I will look into the timings for `ohmi_envelope`, but if I remember correctly the rough estimates of timings I took previously did seem to justify the extra run time.

Otherwise, everything else looks good from my point of view.

T-Nicholls:
In `test_ohmi_envelope` there are two calls to Matlab, the first for the data at the specified `refpts`, and the second for the data at the entrance to the ring.
A summary of my timings, each the mean of 100 iterations:

*my base times for ohmi_envelope on err.mat:*

| refpts= | Python | Matlab |
| ------- | ------ | ------ |
| 0 | 0.75 | 2.76 |
| all | 1.42 | 3.67 |

*test_ohmi_envelope run on err.mat:*

| refpts= | Python call | 1st Matlab call | 2nd Matlab call |
| ------- | ----------- | --------------- | --------------- |
| 0 | 0.747 | 7.715 | 7.683 |
| [0, 1, 2, -1] | 0.715 | 7.686 | 7.692 |
| all | 1.462 | 8.729 | 7.822 |

*test_ohmi_envelope run on hmba.mat:*

| refpts= | Python call | 1st Matlab call | 2nd Matlab call |
| ------- | ----------- | --------------- | --------------- |
| 0 | 0.023 | 0.082 | 0.087 |
| [0, 1, 2, -1] | 0.025 | 0.087 | 0.091 |
| all | 0.047 | 0.109 | 0.093 |

So the expected timing difference, arising from a longer lattice and the introduction of error, can be seen by comparing the `hmba.mat` and `err.mat` python times. You are right however, there is clearly something else going on with the Matlab calls for `err.mat`. Though we see approximately the same increase in time taken between `refpts=0` and `refpts=all` in my base times and the test times, the test times for `err.mat` are consistently ~5 seconds longer than they should be.
The same phenomenon can be seen with linopt:

*hmba ring * 32 (a similar length lattice):*

| refpts= | Python call | Matlab call |
| ------- | ----------- | ----------- |
| 0 | 0.118 | 0.686 |
| [0, 1, 2, -1] | 0.122 | 0.693 |
| all | 0.698 | 0.936 |

*err linopt calls:*

| refpts= | Python call | Matlab call |
| ------- | ----------- | ----------- |
| 0 | 0.712 | 5.831 |
| [0, 1, 2, -1] | 0.716 | 5.776 |
| all | 1.428 | 6.005 |

Here too we see the ~5 second gain. I used hmba * 32 as the comparison to check that it wasn't the length of the lattice causing the problem. I'm really not sure what's causing this discrepancy, I guess it must be something specific to `err.mat`, but not its length, that causes a problem with `engine` or `pyproxy` as they are the only parts I haven't tested independently.

lfarv:
Here are my timing results. These are measured with Matlab to avoid the overhead due to inter-process communication (Python calls in Matlab are native library calls).

_err.mat lattice, linopt call:_

```
refpts      python        Matlab
[0,1,2,-1]   0.51          0.63
all          1.31          0.79

```

_err.mat lattice, single ohmi_envelope call:_

```
refpts      python       Matlab
[0,1,2,-1]   0.60         0.73
all          1.46         0.81

```
So it seems that for me the computation of the solution is slightly faster in python, while the transfer all along the ring is longer in python !

Coming to the behaviour in pytest, I can imagine two effects: the overhead due to access to the Matlab engine, which is a separate process. This should be constant for any call. And the data transfer/conversion from python numpy arrays to Matlab format. I have already seen that this is very slow. Of course the data size is much bigger for err.mat (mainly on returned data). I do not know which fraction is taken in the matlab engine itself and which in the conversion to numpy arrays in python. Is the `numpy.rollaxis(numpy.asarray(...` line included in your timing measurements?

Anyway this is worrying because the comparison using a "real life" example like `err.mat` is really important. For the moment, I can run the comparison tests under Matlab, which is less convenient but much faster. I am also concerned by the small deviations between both. I can understand the discrepancy for the '`data0`' results which are not exactly equivalent: in Matlab, beta etc. are transferred all along the ring to the end, adding numerical errors, while in python they are the direct solution. For chromaticity, even with the same `ddp`, there is still a small unexplained difference.

lfarv:
@T-Nicholls: about` load/utils`: `element_from_dict` can be used for converting a single Matlab structure (not as part of a lattice) to a python element. Called directly from Matlab for instance, I can create and test various python elements. `index` is used only for information purpose, it plays no role in the conversion process.

T-Nicholls:
No, I only timed the calls themselves; the statements I timed were:
```
py_emit0, py_beamdata, py_emit = physics.ohmi_envelope(py_lattice, refpts)
ml_emit = engine.pyproxy('atx', ml_lattice, 0.0, _ml_refs(refpts, nelems))
ml_emit0, ml_params = engine.pyproxy('atx', ml_lattice, 0.0, _ml_refs(0, nelems), nargout=2)

physics.linopt(py_lattice, 0.0, refpts, True)
engine.pyproxy('atlinopt', ml_lattice, 0.0, _ml_refs(refpts, nelems), nargout=3)
```
So from my point of view there is nothing in the test script that can explain the timings, but I'll look into other possibilities further.
I agree that if we can get this extra delay sorted out for `err.mat` then it would be good to have it as a real-life example to run the tests on. As for the numerical differences, I have not run any data comparison on `err.mat` for linopt yet so I'll take a look later.

Ok, that is a reasonable use for `element_from_dict` that I had not thought of.

T-Nicholls:
In `pyproxy` I mesured the time taken for everything, `linopt` call and data processing, and returned only the time taken not the data:

| refpts | Python call | Matlab call total time | Time taken in pyproxy |
| ------ | ----------- | ---------------------- | --------------------- |
| 0 | 0.735680317879 | 6.32676494122 | 1.0007701 |
| [0, 1, 2, -1] | 0.730126309395 | 6.12473249435 | 0.9645413 |
| all | 1.37959778309 | 6.19832570553 | 1.1267016 |

So the delay is not with the call or data processing inside `pyproxy` and it is not from returning the results through `engine`. This leads me to conclude that the issue must lie with passing the lattice through `engine`. To see if it was due to the long length of `err.mat` I decided to test the time taken for `linopt` calls on 2 * the Diamond lattice as `err.mat` is 1.95 times longer than `diad.mat`, 4176 vs 2144 elements:

| refpts | Python call | Matlab call total time |
| ------ | ----------- | ---------------------- |
| 0 | 1.14216058254 | 11.0220449924 |
| [0, 1, 2, -1] | 1.14704861641 | 11.0079421043 |
| all | 2.35984389782 | 11.4856152058 |

To my surprise it had double the delay, just to be sure I timed the `linopt` calls for the regular length Diamond lattice:

| refpts | Python call | Matlab call total time |
| ------ | ----------- | ---------------------- |
| 0 | 0.702683663368 | 5.72734401226 |
| [0, 1, 2, -1] | 0.701047325134 | 5.65509998798 |
| all | 1.27996981144 | 6.03293995857 |

So to me, this implies that some property of `err.mat` and `diad.mat` that has roughly the same value/effect for both is causing an issue within `engine`. For now I do not have any theories but in conclusion, it is not due to:

 - anything in the Python test script
 - anything in pyproxy
 - related to returning the data
 - related to lattice length
 - related to lattice memory usage (a prior theory of mine, whilst the full memory usage of a variable is hard to measure in Python `err.mat` has a file size of 1.1MB to `diad.mat`'s 181kB)

The deeper I look into this, the more confused I become as to what the direct cause could be!

willrogers:
I don't think I have a lot to add to this discussion.

* I agree we should understand the slow lattice loading time
* Once we decide what the different meanings of radiation are we should at least write them down somewhere
* I might have chosen to inherit Lattice from object and merely delegate the sequence operations to a list, but you present a good case so I'm happy with that

This generally looks good to me, thanks.

lfarv:
@T-Nicholls : The timing results are really puzzling… I have no new idea!

@willrogers : I added a definition of `radiation` to its doc string: _"If True, at least one element modifies the beam energy"_, which is what prevent some functions like `linopt`, `find_orbit4`, etc. to give valid results.

Otherwise, if you agree with the last modifications, I would like to merge at that point before going further on, to avoid deviating too much from master.

T-Nicholls:
@lfarv 
I am happy to merge this to avoid further delay, I will keep the timing issue in the back of my mind.
One note on changing `numpy.md` to `.dot` in `ohmi_envelope`, do not expect a speed increase as I previously tested it and found only a 2.8% time reduction compared to using `md`. The change is not a statistically significant decrease in the time taken; however, since it is not an increase and `.dot` is consistent with `linopt` I am happy with the change.

PR #98, opened 2019-03-07
Basic pyat examples.

T-Nicholls:
@willrogers received a request from a new user of pyat for some examples of the use of pyat to help them get started. If anyone has any other suggestions for things to include I would be happy to add them.


willrogers:
Shesh (is that the right name?) asked how to load a Matlab file from a `.m` file. I said that he had to load it in Matlab then save as a `.mat` file. Perhaps we should have instructions on how to do that.

Issue #100, opened 2019-03-11
Select elements with wildcards in PyAt


abdomit:
Hello,

First I'd like to thanks all the people working on PyAt. It is really great for Matlab-allergic people (like me). I am a PyAt user since a month, and it's working great.

But one thing I couldn't find in PyAt is a way to easily get elements from the lattice by their FamName, using a string supporting wildcards. This is something possible in Matlab-At, and it is really convenient (specially when the lattice you use is only made of Multipole elements).

For instance I would like to get all focusing quadrupole in my ring by just typing ring['QD*']. Maybe there is already a way to do it in PyAt, but I couldn't find it.

I did a quick implementation of this [here](https://github.com/abdomit/at/tree/string_indexing). It is just 5 lines long, doesn't support Python2 and don't address the \_\_setitem__ method. If there is no other way of doing this and you think it is worth it, I could do a pull request with it, trying to make it cleaner in the meantime.

Thanks,
Benoit

lfarv:
Hello @abdomit,

I thought of that, but I figured out that you can get the same thing either as an iterator with:

`qd_iter=(elem for elem in ring if fnmatch.fnmatch(elem.FamName, 'QD*'))`

for a single usage, or as a list with:

`qd_list=[elem for elem in ring if fnmatch.fnmatch(elem.FamName, 'QD*')]`

if you want to use the list several times… And as elements are indeed references, modifying the element in `qd_list` does modify it in `ring`. Modifying lattices using only simple commands give something like:

```
qd_strength = [elem.PolynomB[1] for elem in ring if fnmatch.fnmatch(elem.FamName, 'QD*')]
```
or in the other direction:
```
qd_iter=(elem for elem in ring if fnmatch.fnmatch(elem.FamName, 'QD*'))
for elem, strength in zip(qd_iter, qd_strength):
    elem.PolynomB[1] = strength
```
I have 2 comments on indexing directly the lattice:

1. It creates a new list, which you can avoid using iterators if you do not need to keep the list
2. It is restricted to a match on `FamName`, while you can imagine any other selection criteria

However, if there is a need for it, why not!

P.S. selecting focusing quads looking for 'QD*' might give surprising results!

abdomit:
Hello Laurent,

> I have 2 comments on indexing directly the lattice:
> 
>  1. It creates a new list, which you can avoid using iterators if you do not need to keep the list

Ok, but an iterator is not very user-friendy. You cannot address the nth element with

    qd_list[n]

[...]

PR #102, opened 2019-03-28
Add new get elements method to lattice to return elements by family or class.

T-Nicholls:
A first draft of my proposed solution to #100.
Perhaps having two separate `get_elements_by_family` and `get_elements_by_class` would be more elegant than my current implementation; what is everyone's opinion on this?
I would welcome any feedback or improvements that anyone might have.

### DIFF
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -2,12 +2,15 @@
 import sys
 import copy
 import numpy
+from fnmatch import fnmatch
 from scipy.constants import physical_constants as cst
 from warnings import warn
+from at.load import CLASS_MAPPING
 from at.lattice import elements, get_s_pos, checktype, uint32_refpts
 from at.lattice import AtWarning, AtError
 from at.physics import find_orbit4, find_orbit6, find_sync_orbit, find_m44
 from at.physics import find_m66, linopt, ohmi_envelope, get_mcf
+
 
 __all__ = ['Lattice']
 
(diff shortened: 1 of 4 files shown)


willrogers:
cc @lfarv @abdomit

abdomit:
Thanks @T-Nicholls for addressing #100!

I prefer simple names, so having just one get_elements function is fine. However I am not so comfortable with the idea that a str could be either a class name or a FamName. What if someone calls an element 'ap' (a 2 letter-acronym is easy to find), do a get_element('ap') and get all Aperture elements instead? Maybe one could say that giving a str to get_elements will just address FamName?

Also throwing an exception if there is no element matching the str seems a bit excessive. Returning an empty list is also fine, don't you think?

lfarv:
I agree with @abdomit on the "string" option for the key: it should refer to `FamName` only. The "Class" name is a pure Matlab feature and does not appear any more in the Python object. We should forget it.

I also think that we should get an empty list in case no element matches. For me it is not an error.

Otherwise:
- Shouldn't this function belong to `lattice.utils` and be available more generally to any sequence of elements, with a corresponding `Lattice` method, like it is done for `get_s_pos`?
- Checking versus the `str` type ignores the Python 2 `unicode` type. But Python 2 is at end of life, so this looks good enough for me.

Final note: this function ends up with either
`elems = [elem for elem in self if fnmatch(elem.FamName, key)]`
or
`elems = [elem for elem in self if isinstance(elem, type(key))]`
or
`elems = [elem for elem in self if isinstance(elem, key)]`

For me, using any of these single lines looks as simple and clear as a function call. But I have nothing against the function!

T-Nicholls:
Thank you both for the feedback, I agree with everything said so I'll make the changes.
However, I will use `numpy.issubdtype` for now since we might as well support unicode strings.

T-Nicholls:
@lfarv @abdomit 
When you are both happy with this I would like to put close this pull request and move these changes onto the end on #103 as the test is failing because it relies on a fixture from that branch and we would have a conflict on merging the second one of these branches anyway, so I think it is easier for me to move it manually.

abdomit:
@T-Nicholls: I would only suggest that you add an optional verbose mode to prevent being flooded by messages when calling get_elements with a str as parameters.

Otherwise it looks nice to me. Thanks again.

lfarv:
Hi @T-Nicholls. What happened to this branch? It looked ready for merging, but I do not see it any more…

T-Nicholls:
@lfarv 
As mentioned above I moved these changes onto #103 because the test I wrote for `get_elements` relied on changes made on that branch. There were also conflicts between the two, so I figured I might as well sort it out on my end and put it all on one pull request since as, for me,  #103 is nearly ready to merge. Sorry for any confusion.

PR #105, opened 2019-04-05
emittance fix

lfarv:
Fixed the Nan values for emittance of uncoupled machines


willrogers:
This has been merged in #99.

PR #106, opened 2019-04-08
Enable the automatic building and upload of wheels.

T-Nicholls:
The current behaviour is that wheels are built on every AppVeyor/Travis build, but only uploaded to the **test** PyPI if a tag is pushed. If we are happy with the wheels on test PyPI we can then upload them to the real PyPI, which should address #5.


willrogers:
Let's add Python 3.7.

willrogers:
@lfarv the idea behind this pull request is that whenever a tag is pushed to the repo, Travis and Appveyor will build Python wheels (for Linux, Mac and Windows) and send them to PyPI. This will make pyat much more user-friendly.

Do you have any thoughts? The best project names I can find still available on PyPI are `at-python` and `py-at`, although `python-at` has not been updated for ages so I might be able to persuade the owner to transfer that one.

lfarv:
Thanks for the explanation! Installing pyat with pip will make installation much easier. But I can't be of any help for that, so it's up to you. Concerning names, having both `at-python`and `python-at` for unrelated projects seems really confusing. In such a case, I would push for `py-at`. Getting `python-at` would of course be even better…

willrogers:
I thought about this a bit more and I have changed my mind. I think we should call it 'accelerator-toolbox' i.e. `pip install accelerator-toolbox`, then `import at` as before. Then it's not ambiguous at all.

lfarv:
@willrogers: I fully agree with this proposal!

willrogers:
@T-Nicholls can you register 'accelerator-toolbox' on PyPI and test PyPI please?

.appveyor.yml
```diff
@@ -1,18 +1,56 @@
...
-  - py.test test
+  - ps: |
+      if ($env:CIBW_BEFORE_BUILD -eq "pip install numpy") {
+        rm "build" -r -fo
+        python -m pip install --upgrade cibuildwheel
+        & cmd /c 'cibuildwheel 2>&1' --output-dir wheelhouse
+        python setup.py sdist --dist-dir=wheelhouse
+        if ($env:APPVEYOR_REPO_TAG -eq $true) {
+          python -m pip install twine
+          python -m twine upload --repository-url https://test.pypi.org/legacy/ wheelhouse/*
```
willrogers: This 'legacy' doesn't look right to me.

.travis.yml
```diff
@@ -21,26 +27,43 @@ matrix:
...
+  - python setup.py develop
 
+script: |
+  if [[ $CIBW_BEFORE_BUILD != '' ]]; then
+    rm -rf build
+    python -m pip install --upgrade cibuildwheel
+    cibuildwheel --output-dir wheelhouse
+    if [[ $TRAVIS_TAG ]]; then
+      python -m pip install twine
+      python -m twine upload --repository-url https://test.pypi.org/legacy/ wheelhouse/*.whl
```
willrogers: And this one.

T-Nicholls:
The official docs use 'legacy' with this comment `The legacy in the URL above refers to the legacy API for PyPI, which may change as the Warehouse project progresses.`, but I will try it on my testing branch without the legacy and see if it works. As a side note when we push directly to pypi not test pypi we can ditch the whole `--repositiory-url` all together.

T-Nicholls:
Without the 'legacy' we get:
```
python : HTTPError: 405 Client Error: Method Not Allowed for url: https://test.pypi.org/
At line:9 char:3
+   python -m twine upload --repository-url https://test.pypi.org/ whee ...
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (HTTPError: 405 .../test.pypi.org/:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
Command executed with exception: HTTPError: 405 Client Error: Method Not Allowed for url: https://test.pypi.org/
```

willrogers:
Well, it looks like it should be `https://test.pypi.org/pypi` similarly to `https://pypi.org/pypi`:

https://packaging.python.org/guides/migrating-to-pypi-org/

T-Nicholls:
I get the this error `HTTPError: 410 Client Error: This API has moved to https://test.pypi.org/legacy/. See TODO for more information. for url: https://test.pypi.org/pypi`
I will try one more combination but it looks like we will have to stick with 'legacy' for now.

willrogers:
It's not a problem, I was just a bit surprised that the 'legacy' one is still the right one to use.

T-Nicholls:
For me this is now ready to merge.
In the last two commits I:
- removed some superfluous lines from both the .yml files
- made wheels only build on tag to speed up normal build times
- fixed tests failing but build still passing on appveyor I also made sure it dosn't happen to other commands in future too
- found and fixed a bug in our pass method file extension checking in python <3.3 on Windows

T-Nicholls:
On Will's suggestion wheels will now only be built if the name of the tag begins with 'pyat', so our pyAT releases are distinct from our releases of AT a whole.

PR #107, opened 2019-04-10
Fix import cycles

lfarv:
As suggested by @T-Nicholls, `get_ring_properties` has been moved to `lattice.utils`, together with the helper function `_scanner`. This solves the cycle encountered in `radiation.py`. Not sure it solves everything…
I'm also not not fully satisfied with the function name _scanner, but I would keep it 'private'. All suggestions are welcome.

### DIFF
--- a/pyat/at/__init__.py
+++ b/pyat/at/__init__.py
@@ -2,6 +2,7 @@
 
 # Make all functions visible in the at namespace:
 from .lattice import *
+from .load import *
 from .tracking import *
 from .physics import *
-from .load import *
+from .plot import *
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -4,16 +4,14 @@ import copy
 import numpy
 import math
 from scipy.constants import physical_constants as cst
-from warnings import warn
 from at.lattice import elements, get_s_pos, get_elements, uint32_refpts
-from at.lattice import AtWarning, AtError
+from at.lattice import AtError
+from at.lattice.utils import _scanner
 from at.physics import find_orbit4, find_orbit6, find_sync_orbit, find_m44
 from at.physics import find_m66, linopt, ohmi_envelope, get_mcf
 from at.plot import plot_beta, plot_trajectory
 
-__all__ = ['Lattice', 'get_ring_properties']
-
-TWO_PI_ERROR = 1.E-4
+__all__ = ['Lattice']
 
 
 class Lattice(list):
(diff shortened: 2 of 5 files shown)


willrogers:
I think `_scanner` is a bit tricksy (a colloquialism there). If I read it correctly, it:

* returns a generator that will yield the elements:
** including RingParams if keep_all
** not including RingParams if not keep_all
* update the passed in attrs

I think this is to avoid iterating over the ring multiple times, but is the added complexity worth it? I seem to remember that @T-Nicholls raised the multiple iterations.

It would be simpler if `_scanner()` returned the derived attrs and we removed the ring params separately with a list comprehension. Then we can call `get_ring_params(elems)` from the lattice object and we don't need to import the private function.

T-Nicholls:
I only suggested doing it in one go due to personal belief, the performance benefit isn't noticeable for the size of rings we are using. So, if it is causing unnecessary code complexity, which it seems it is, I am in favour of ditching it for a simpler option.

lfarv:
@willrogers summarised perfectly the role of this iterator. I will try and explain the motivation for it. I fully agree that the time for any additional iteration over the elements is not a real concern here. But allowing any iterable (a sequence-like object OR an iterator) for initialisation, as it is the case for the underlying `list`, is a very interesting feature. For instance, the new `slice` method which creates a new `Lattice` does not require any intermediate copy of the elements: elements are sliced on the fly as the `list` constructor (`super(Lattice, self).__init__(elems`) is executed, `elems` being the "`slice_iterator`". This will also be the case for some future methods, for instance `reduce` doing the opposite of `slice`: removing unnecessary elements and packing successive identical ones. The problem is that this allows any scan of elements **after** the list is built, but **not before**, because it would consume the iterator. And in our case, looking for energy cannot be done after the list is built, since the `RingParam` element might have been removed. That's why it is done simultaneously…

If you see the necessity to separate these 2 functionalities, the `_scanner` function could be split into 2 iterators, the 1st one taking care of extracting the attributes, and being piped into a 2nd one removing the `RingParam` element. Only the 1st one needs to be `utils.py`. But it does not look so much clearer to me, and we loose a bit of efficiency by repeating the test of `RingParam` objects…

I am still not happy with exposing this iterator, because I don't think it can be of any use apart from **building** a Lattice object, or **simulating** a Lattice object, as it is the case for `ohmi_envelope`. I think we should push to use `Lattice` objects everywhere, even if some function may work with any sequence (as indicated in their docstring). And I welcome any suggestion for a better name than `_scanner`…

So I strongly push to keep the possibility of allowing iterators for initialising `Lattice` objects (without first making a list out of it!). This is a general trend for improving efficiency: in Python3, `zip`, `map`, `filter`, etc. now return iterators instead of `lists`, and this allows to use the features of the `itertools` module (`repeat` or `chain` to build rings from individual cells, for instance).

With this possibility, I would even reconsider the choice we made some time ago concerning `__getitem__`, that indexing a `Lattice` object returns a `list` and not a `Lattice`. With tools like `islice`, we could return a `Lattice` without any overhead nor intermediate copy. But that's another question…

willrogers:
OK, this required some thinking about.

I am happy with the requirement to accept iterators in the constructor to `Lattice`. I am also happy with 

* the proposal to use `Lattice` everywhere (in fact, I think the idea of just using a sequence of structures is now out of date, as no-one is likely to produce one without running code in pyat)
  * presumably we could just create a lattice on the fly in `ohmi_envelope()` though
* the proposal to return a lattice when indexing a lattice, this seems self-consistent to me.

Then scanner could return to a private function or static method.


----

Now we move on to opinions about software, don't worry too much if you disagree. Firstly the `_scanner()` function is quite long and doing more than one thing. It would be clearer if we could do a list comprehension after 

```
```

but I realise I don't even know how to do that from inside the list class. To revisit whether `Lattice` should inherit from `list`, there is the Liskov Substitution Principle: **is** the `Lattice` a `list`? I would argue no, it's a sequence. We also have [favour composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) - retain a reference to the list of elems instead - this would at least make my list comprehension above easier. The cost is delegating indexing methods to the list, but since we're doing extended indexing anyway that begins to look like a minor matter. 

In python 3 we could inherit from [collections.abc.Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence).

This was discussed at some length in #96, so if you retain your opinion we don't have to change anything.

lfarv:
I'll first comment on @willrogers's message, and then make a new proposal in the next message.
- I agree that the present `_scanner` has the problem of doing two unrelated things at the same time
- however, removing the `RingParam` elements after building the list more or less defeats the purpose of the iterator: you first build a temporary list before modifying it. I don't see how to do it with a list comprehension, but you could probably do it with a `for` loop and assignment statements. This is expensive.
- A reference to an embedded list attribute doesn't help on that: either you can build the list immediately (best solution), or you have to modify it later (with a list comprehension for a delegated list, I agree)
- So I still think that the only way to separate the two functionalities without introducing intermediate storage is to split the `_scanner` iterator in two: The first iterator runs through all elements, and only looks for energy, name, periodicity. It stays in `utils.py`. The second one runs through the first one and optionally eliminates the `RingParam` elements. It can be hidden in the `Lattice` class and is so simple that it can be written as a generator expression:
```
if keep_all:
    iter2 = iter1(...)
else:
    iter2 = (elem for elem in iter1(...) if not isinstance(elem, RingParam)
super(Lattice, self).__init__(iter2)
```

Concerning inheritance or composition, I confirm that for me a Lattice **is** a mutable sequence. So inheritance. Strictly speaking, you are right, it could inherit from [collections.abc.MutableSequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence). But this means implementing at least an adequate container plus 5  methods… Starting directly from a list simpler, and the list interface is well known. So I would stick to a `list` base class.

Now, another proposal…

lfarv:
Some time ago, we agreed that a `RingParam` is not a real lattice element. It’s a Matlab trick to store metadata. So I propose to eliminate it completely from Python.

- We move all the `_scanner` functionality to `load_mat`. But no need for an iterator there, we can rewrite it as as clean as we can, prepare the lattice step by step and eliminate `RingParam` elements. Then load_mat builds a `Lattice` object with the obtained metadata and returns it.
- The `Lattice` constructor sends its first argument (sequence-like object of iterator) directly to the `list` constructor, as it is. No more scan. It becomes very simple. And all `Lattice` manipulations are simpler and faster.
- We return a `Lattice` when indexing a `Lattice`
- The `get_ring_properties` function disappears
- All physics functions stay as they are: they can accept `Lattice` or sometimes sequence-like objects. Nothing changes for users, a `Lattice` can be accepted everywhere a sequence is. And you can still build a sequence of elements independently of any Matlab file, this still works.
- `ohmi_envelope` can accept a sequence only if `energy` is provided as input argument
- The `keep_all` keyword was introduced to make Matlab/python comparisons easier by keeping the same number of elements, so the same refpoints. It has to move from `Lattice.__init__` to `load_mat`. But I don’t think it concerns users. If `keep_all` is True, `load_mat` will replace the `RingParam` by a `Marker` with the same attributes (and no particular behaviour)

I think it makes everything simpler, pushes the use of `Lattice` objects and is easy to implement.

Loading lattices from Matlab files is a convenience for the time being. Maybe we should look for a native way of storing lattices (and no need for `RingParam` there)

T-Nicholls:
I was in favour of splitting the iterator between `Lattice` and `get_ring_properties` but now I think @lfarv's proposal addresses all the problems well and doesn't seem to have any drawbacks for me, so unless I've missed something I'd support pushing to use `Lattice` objects more.

willrogers:
I like @lfarv's latest proposal.

At the moment it is tricky to create a lattice unless you happen to have one in Matlab. In my opinion it is tricky to create a lattice in Matlab as well. A text format would be best, and I'm perfectly happy to adopt another code's format if it is not incompatible. @lfarv would know most about these I expect.

lfarv:
Ok, I’ll go that way and propose something’s by the end of the week.

lfarv:
The purpose of this pull request is now moved to #110

PR #108, opened 2019-04-15
Added option 'RelativeShift' to atsetshift

carmignani:
I added the optional flag 'RelativeShift' to atsetshift and to atshiftelem to allow to add a displacement on top of the existing T1 and T2.

### DIFF
--- a/atmat/lattice/atsetshift.m
+++ b/atmat/lattice/atsetshift.m
@@ -10,16 +10,20 @@ function ring=atsetshift(varargin)
 %	X  ->  X-DX
 %   Y  ->  Y-DY
 %
-% ATSETSHIFT(ELEMINDEX, DX, DY) Uses the global variable THERING
+% RING=ATSETSHIFT(RING, ELEMINDEX, DX, DY, 'RelativeShift')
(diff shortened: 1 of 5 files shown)


simoneliuzzo:
In my opinion this change should go along with the same change in atsettilt and other error setting related functions. 
Also the default should remain absolute error setting for backward compatibility.

carmignani:
Thanks @simoneliuzzo for your comment. I added the option 'RelativeTilt' to atsettilt, as you suggested. The default is for both tilt and shift an absolute error setting, not relative.
It would be good to add the same options to the python functions.

lfarv:
For tilts: if the initial R1 and R2 are not pure rotations, the multiplications may not be commutative. R1 multiplication is now correct.
I added tests on flags: a typo in a flags would have it been silently ignored, with wrong results.

lfarv:
Made the similar modification in python introducing the 'relative' keyword in
`shift_elem`, `tilt_elem`, `set_shift`, `set_tilt`

willrogers:
I haven't checked the physics but this looks sensible to me.

lfarv:
No real physics: just a convenient option for accumulating different kinds of errors!

PR #109, opened 2019-04-18
added method sbreak to Lattice object in pyat

carmignani:
I'm not expert in python, so tell me if the code is not very good!
The method does the same of the Matlab AT function atsbreak.

### DIFF
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -210,6 +210,31 @@ class Lattice(list):
         """Return a deep copy"""
         return copy.deepcopy(self)
 
+    def sbreak(self, s):
+        """
+        Create a new lattice with a marker with name 'sbreak' at positions s.
+        If s is a list of positions, many markers are added to the lattice.
+        """
+        if type(s)!=list:
+            s=[s]
+        for ii in list(range(len(s))):
+            if s[ii]<=0:
+                self=Lattice([elements.Marker('sbreak')]+self)
+            elif s[ii]>self.get_s_pos(len(self)):
+                self=Lattice(self+[elements.Marker('sbreak')])
+            else:
+                spos=self.get_s_pos(list(range(0,len(self))))
+                mask_before= spos<=s[ii]
+                before=self[mask_before]
+                after=self[spos>s[ii]]
+                elem=before[-1]
+                selem=spos[mask_before][-1]
+                frac=(s[ii]-selem)/elem.Length
+                elems=elem.divide([frac, 1-frac])
+                elems=[elems[0],elements.Marker('sbreak'),elems[1]]
+                self=Lattice(before[0:-1]+elems+after)
+        return self
+            
     def slice(self, size=None, slices=1):
         """Create a new lattice by slicing the range of interest into small
         elements


lfarv:
Hi @carmignani. I'll start by making two comments:

1. On the design: it would be more flexible to add a second argument: a list of elements, as long as s, giving the elements to be inserted, or a single element broadcast to the size of s. Its default value can be the Marker you mention,
2. On the implementation:
  - rather than comparing types, you can use '`isinstance(list)`' which compares with a type and its descendents. But there are other iterables types which could be used for `s`. So a better test is to check whether `s` can be iterated upon to not.
  - calling the `Lattice` constructor many times is expensive since it makes a new copy of the whole lattice each time. It is much more efficient to call `Lattice()` once, as in the slice method, with an argument which iterates through the original lattice and inserts the elements when appropriate.

I do not remember if there is an equivalent in Matlab. In any case, I would try and make similar things in both versions

T-Nicholls:
@lfarv it can be found [here](https://github.com/atcollab/at/blob/master/atmat/lattice/atsbreak.m)
@carmignani, a few points, based on your current code, they may be less relevant if you take Laurent's advice:
- it is worth noting that the base class methods `insert` and `append` are available because `Lattice` inherits from list. For example, your first two `Lattice` constructor calls could be replaced by `self.insert(0, elements.Marker('sbreak'))` and `self.append(elements.Marker('sbreak'))` respectively.
- about the edges of s values, we allow negative s positions and just treat them as 0 and we treat s positions larger than the length of the lattice as lattice length; if this is intended then a comment stating it would be nice, as my gut instinct would have been to throw an error.

lfarv:
Ideally for me, it could look like this:
```
def sbreak(self, s, break_elems=None):

    def sbreak_iterator(elems, insert_it):
        """Iterate over elements and produces the elements of the new lattice"""

        # get the 1st insertion
        try:
            insert_s, insert_el = insert_it.next()
        except StopIteration:
            insert_s = sys.float_info.max

        s_end = 0.0
        for el in elems:
            s_beg = s_end
            s_end += el.length
            if insert_s < s_end:
                # deliver the element with inserted markers
                frac = (insert_s-s_beg)/el.Length
                yield from el.insert(frac, insert_el)    # Python 3 only
                # get the next insertion
                try:
                    insert_s, insert_el = insert_it.next()
                except StopIteration:
                    insert_s = sys.float_info.max
            else:
                # deliver the element unchanged
                yield el

    # set default insertion
    if break_elems is None:
        break_elems = elements.Marker('sbreak')
    # broadcast s and break_elems to arrays of same size
    s, break_elems = numpy.broadcast_arrays(s, break_elems)
    # create an iterator over the elements to be inserted
    it = zip(s, break_elems)
    
    return Lattice(sbreak_iterator(self, it), **vars(self))
```
It may be wrong, I didn't try, but here is the idea: the elements of the new lattice are produced as they are needed.
But I found that `element.insert` is defined for `Drift`s only. One could imagine allowing insertion of zero-length elements in any `LongElement`

lfarv:
Here is a new proposal for the sbreak method:

- It does not need the `element.insert` method, it uses `element.divide`
- It handles multiple breaks inside the same element, multiple breaks at the same location, it silently skips the breaks outside the lattice,

```
def sbreak(self, s, break_elems=None):
    def sbreak_iterator(elems, insertions):
        """Iterate over elements and produces the elements of the new lattice"""

        def next_mk():
            """Extract the next element to insert"""
            try:
                return next(insertions)
            except StopIteration:
                return sys.float_info.max, None

        s_end = 0.0
        # get the 1st insertion
        smk, mk = next_mk()
        # skip all insertions at negative s
        while smk < s_end:
            smk, mk = next_mk()

        for elem in elems:
            s_beg = s_end
            s_end += elem.Length
            # loop over all insertions within the element
            while smk < s_end:
                frac = (s_end - smk) / elem.Length
                el0, elem = elem.divide([1.0 - frac, frac])
                if el0.Length > 0.0:  # skip 0-length elements
                    yield el0
                yield mk
                smk, mk = next_mk()
            yield elem

    # set default insertion
    if break_elems is None:
        break_elems = elements.Marker('sbreak')
    # broadcast s and break_elems to arrays of same size
    # and create an iterator over the elements to be inserted
    iter_mk = zip(*numpy.broadcast_arrays(s, break_elems))

    return Lattice(sbreak_iterator(self, iter_mk), **vars(self))
```

lfarv:
Replaced by #132

PR #110, opened 2019-04-20
Remove RingParam

lfarv:
According to the discussion in #107, I removed the `RingParam` element, moved all the `Lattice` coherence tests to `load_mat`, simplified the `Lattice` initialisation and return `Lattice` objects when indexing a `Lattice`.

### DIFF
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -57,13 +57,13 @@ class Element(object):
 
     def __str__(self):
         first3 = ['FamName', 'Length', 'PassMethod']
-        attrs = dict((k, getattr(self, k)) for k in self)
+        attrs = dict((k, v) for k, v in vars(self).items())
         keywords = ['\t{0} : {1!s}'.format(k, attrs.pop(k)) for k in first3]
         keywords += ['\t{0} : {1!s}'.format(k, v) for k, v in attrs.items()]
         return '\n'.join((self.__class__.__name__ + ':', '\n'.join(keywords)))
 
     def __repr__(self):
-        attrs = dict((k, getattr(self, k)) for k in self)
+        attrs = dict((k, v) for k, v in vars(self).items())
         arguments = [attrs.pop(k, getattr(self, k)) for k in
                      self.REQUIRED_ATTRIBUTES]
         defelem = self.__class__(*arguments)
@@ -462,3 +462,14 @@ class Corrector(LongElement):
         super(Corrector, self).__init__(family_name,
                                         Length=kwargs.pop('Length', length),
                                         KickAngle=kick_angle, **kwargs)
+
+
+class _RingParam(Element):
+    """Private class for Matlab RingParam element"""
+    REQUIRED_ATTRIBUTES = Element.REQUIRED_ATTRIBUTES + ['Energy']
+    _conversions = dict(Element._conversions, Energy=float, Periodicity=int)
+
+    def __init__(self, family_name, energy, **kwargs):
+        kwargs.setdefault('Periodicity', 1)
+        kwargs.setdefault('PassMethod', 'IdentityPass')
+        super(_RingParam, self).__init__(family_name, Energy=energy, **kwargs)
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -469,14 +469,3 @@ class Corrector(LongElement):
         super(Corrector, self).__init__(family_name,
                                         Length=kwargs.pop('Length', length),
                                         KickAngle=kick_angle, **kwargs)
-
-
-class _RingParam(Element):
-    """Private class for Matlab RingParam element"""
-    REQUIRED_ATTRIBUTES = Element.REQUIRED_ATTRIBUTES + ['Energy']
-    _conversions = dict(Element._conversions, Energy=float, Periodicity=int)
-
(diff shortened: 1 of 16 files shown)


lfarv:
The displacement of coherence tests has some side effects when creating a Lattice explicitly, without using a Matlab file:
- Energy is not extracted from the element list any more: there is no `RingParam` element to define it and the python element constructors don't expect `energy` attributes, except from for the `Cavity` element. So the search for energy is almost useless. `energy` has to be provided at Lattice creation!
- The default for `name` and `periodicity` are `""` and 1. No more search.

This is the counterpart for making a Lattice initialisation without any overhead. Lattice creation from a Matlab file works as before.

The problem with `ohmi_envelope` is more complicated than expected: we need a perfect equality between the energy provided to `find_mpole_raddiff_matrix` and the one in the attributes of each element with a `'*RadPass'` pass method. So giving `energy` as an input parameter of `ohmi_envelope` may lead to wrong results. There is no problem with a `Lattice` object since `radiation_on` forces the energy of each radiating element to the one of the lattice. However I will change the strategy for energy detection in `load_mat` to respect this requirement:

- `energy` will be extracted from `RingParam`, cavities and elements with *RadPass method. Any discrepancy will now be an error,
- `energy` from all other elements will be ignored and discarded.

With the same tests, we can think of restoring the get_ring_properties function if we really need it.

pyat/at/lattice/elements.py
```diff
@@ -473,3 +457,14 @@ def __init__(self, family_name, length, kick_angle, **kwargs):
         super(Corrector, self).__init__(family_name,
                                         Length=kwargs.pop('Length', length),
                                         KickAngle=kick_angle, **kwargs)
+
+
+class _RingParam(Element):
+    """Private class for Matlab RingParam element"""
```
willrogers: It might be helpful to explain why this class is private.

pyat/at/load/matfile.py
```diff
@@ -21,7 +25,103 @@ def _load_element(index, element_array, check=True, quiet=False):
     return element_from_dict(kwargs, index=index, check=check, quiet=quiet)
 
 
-def load_mat(filename, key=None, check=True, quiet=False):
+def _scanner(elems, **kwargs):
```
willrogers: Now this isn't a generator, it seems to me that it could be used whenever making a lattice from a list and so could be public?

pyat/at/physics/radiation.py
```diff
@@ -23,8 +22,7 @@
                   ('emitXYZ', numpy.float64, (3,))]
 
 
-def ohmi_envelope(ring, refpts=None, orbit=None, keep_lattice=False,
-                  energy=None):
+def ohmi_envelope(ring, refpts=None, orbit=None, keep_lattice=False):
```
willrogers: We should update the docstring to remove energy and say that a Lattice object is required.

pyat/test/test_lattice_object.py
```diff
@@ -41,62 +40,62 @@ def test_lattice_creation_from_lattice_inherits_attributes():
...
-def test_lattice_gets_attributes_from_elements():
-    d = elements.Dipole('d1', 1, BendingAngle=numpy.pi, Energy=3.e+6,
-                        PassMethod='BndMPoleSymplectic4RadPass')
-    l = Lattice([d])
-    assert len(l) == 1
-    assert l.name == ''
-    assert l.energy == 3.e+6
-    assert l.periodicity == 2
-    assert l._radiation is True
+#
```
willrogers: Commented tests: some should be deleted, others should run on the `_scanner()` function where these things are checked now?

pyat/at/lattice/elements.py
lfarv: It's now removed from elements and is a private class in` at.load.utils`.

pyat/test/test_lattice_object.py
lfarv: Commented tests are now deleted. For checking the internals of `load_mat`, there are two possibilities:

1. Ignoring the implementation and checking only the result as it is documented: for this, one would need a set of Matlab files with different configurations and checking the expected behaviour. This looks very heavy.
2. Checking the implementation by importing private functions: then the tests must be updated if the implementation changes. I'll try to restore some of the removed tests in that way

pyat/at/physics/radiation.py
lfarv: Done!

pyat/at/load/matfile.py
lfarv: Since the `RingParam` element does not exist any more, a large part of this scan is now useless. Do the remaining tests have any interest for a list created in another way than a Matlab file? If someone is building a lattice by hand, he should be able to specify `energy` and `periodicity` to the `Lattice` constructor. I recall that `energy` is not provided to any python element constructor (unless you put it in a keyword argument). Again, no more scan at `Lattice` creation removes the overhead. Since any subscripting now creates a new Lattice, this has some interest.
For the time being, `load_mat` may still change as I am looking at importing other formats, and I would like to reuse as much code as possible. Should a scan of `list`s for energy or periodicity still be useful, I would wait a little bit.

lfarv:
- The tests of extraction of lattice parameters from a list of elements are now moved to `test_load_utils`
- A new `save_mat` function is now available

Though the `at.load` package will still evolve, for me this is now ready for merging.

PR #113, opened 2019-04-29
Remove dependencies

lfarv:
Here is a solution to the dependency loop problem:
The `Lattice` object does not depend any more on the `at.physics` package.
To do this, all methods relying on the `at.physics` package are not defined any more in the `lattice_object` module. Instead, they are defined in the module where the underlying physical function is implemented. And similarly for the `at.plot` package. 
So the import order is now set to:
```
from .lattice import *
from .tracking import *
from .physics import *
from .plot import *
from .load import *
```
where no package depends on the following ones. This is shown on the following debug output obtained with a printout at the beginning and end of each of the sub-packages `__init__` module:

**Branch master:**
```
lattice start
  physics start
    tracking start
    tracking end
  physics end
  plot start
  plot end
lattice end
load start
load end
```
**Branch remove_dependencies:**
```
lattice start
lattice end
tracking start
tracking end
physics start
physics end
plot start
plot end
load start
load end
```
The `import` statements have also been rearranged inside the sub-packages so that there is also no loop within the sub-packages.

As side effects, `import at` is faster (0.701s instead of 0.797s in average on my machine) and the handling of the optional Matplotlib package is simpler.

I also set a decorator to check the `Lattice.radiation` status on `linopt`, `get_mcf`…

Finally I have 2 questions:

1.  I set `_ohmi_envelope` as a private function, so that it should be called only as a `Lattice` method. If you think it should stay accessible as a standard function, it’s easy set it public again,

3.  at.plot is still imported in `at/__init__.py`. Removing it would have 2 advantages
- Faster import for applications not using graphics (0.465s instead of 0.701s)
- The message about non-availability of Matplotlib appears when import `at.plot` is explicitly executed, rather than at the initial `import at`

And 1 drawback: the necessity to explicitly `import at.plot` to have access to the graphics commands
What is your preference ?

Sorry for the many changes in `lattice.orbit`: I try to cut the lines at 80 characters and review the doc strings when I touch a file…

### DIFF
--- a/pyat/at/__init__.py
+++ b/pyat/at/__init__.py
@@ -4,5 +4,4 @@
 from .lattice import *
 from .tracking import *
 from .physics import *
-from .plot import *
 from .load import *
--- a/pyat/at/__init__.py
+++ b/pyat/at/__init__.py
@@ -4,4 +4,5 @@
 from .lattice import *
 from .tracking import *
 from .physics import *
+from .plot import *
 from .load import *
--- a/pyat/at/lattice/__init__.py
+++ b/pyat/at/lattice/__init__.py
@@ -4,8 +4,7 @@ Helper functions for working with AT lattices.
 A lattice as understood by pyAT is any sequence of elements.  These functions
 are useful for working with these sequences.
 """
-print('lattice start')
-from .utils import *
+
 from .elements import *
+from .utils import *
 from .lattice_object import *
-print('lattice end')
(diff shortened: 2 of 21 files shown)


willrogers:
1. Since we have the slightly unusual separation of methods from their owning class and this allows using them as functions (and we do this with other functions), I think we may as well allow using `ohmi_envelope()` as a function.
2. I would definitely make `at.plot` an optional import, especially as it slows down the imports for everyone
3. We need to make it clear in the lattice module / class that methods are being attached from elsewhere, otherwise developers could potentially get quite confused. `help()` works well for users.

willrogers:
Fine by me, merge when you are ready.

PR #114, opened 2019-05-09
Check the "at.Element" attributes when set

lfarv:
The type and shape of `Element` attributes are checked at the creation of the element, but not when an attribute is set later. This allows things like:

`elem.FamName = [0, 0]`
or
`elem.PolynomB = ‘abcd’`

Here we move the checks to the `setattr` function, which secures completely  the `Element`.
In addition, we also prevent modifying `MaxOrder`, `PolynomA` or `PolynomB` in an inconsistent way.

### DIFF
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -28,7 +28,9 @@ class Element(object):
     """Base of pyat elements"""
 
     REQUIRED_ATTRIBUTES = ['FamName']
-    _conversions = dict(R1=_array66, R2=_array66, T1=lambda v: _array(v, (6,)),
+    _conversions = dict(FamName=str, PassMethod=str, Length=float,
+                        R1=_array66, R2=_array66,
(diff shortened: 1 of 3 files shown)


willrogers:
I will have a look at this.

On an unrelated note I am submitting a paper to ICALEPCS with the title 'pyAT, Pytac and pythonSoftIoc: a Pure Python Virtual Accelerator'. If you would like to be listed as an author I can easily add you as I have found your account in Jacow.

willrogers:
I think this is a definite improvement.

We should probably add a couple more tests, but since you are doing all this work that is useful for us we can do that some time, feel free to merge this.

lfarv:
@willrogers:

1. On ICALEPCS: you are the initiator of the paper, so I let you decide whether I should appear !
2. On tests: I agree that additional tests would be useful, I find the test sequence incredibly helpful while developing. I added one, please add more if necessary.

PR #115, opened 2019-05-13
Default MaxOrder determination

lfarv:
When exchanging lattices between Matlab and python, I noticed a difference in the way the default value for `MaxOrder` is computed:

**In python**: the default value for `MaxOrder` is 0 for `at.Dipole`, 1 for `at.Quadrupole`, 2 for `at.Sextupole`. Erroneously it is 0 for `at.Octupole`. While this looks coherent, in practice it may be confusing:
- for a focusing dipole (k != 0), the k value is correctly introduced in `PolynomB`, but will not be taken into account until `MaxOrder` is explicitly set to 1 (instead of 0)
- more generally, if additional multipoles are set in any ThinMultipole descendent (that is all magnets) by specifying `PolynomB=[…]` in the constructor, they will not be automatically taken into account.

**In Matlab**: The base value (0 for dipoles, 1 for quadrupoles,…) is incremented to take into account the highest non-zero coefficient of `PolynomA` or `PolynomB`. Of course, if `MaxOrder` is explicitly specified, it is honoured.

This difference does not appear when reading a .mat file, because all the attributes (including `MaxOrder`) are correctly converted. However, for defining or transfering lattice in text files (work in progress), a consistent behaviour is highly preferable. So I propose to switch to the Matlab convention. In all cases where `MaxOrder` is explicitly specified, it will not change anything. In other cases, the declared multipoles (including the dipole focusing strength) will be automatically activated.

I added a few tests to check this behaviour

### DIFF
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -13,7 +13,7 @@ from inspect import getmembers, isdatadescriptor
 def _array(value, shape=(-1,), dtype=numpy.float64):
     # Ensure proper ordering(F) and alignment(A) for "C" access in integrators
     return numpy.require(value, dtype=dtype, requirements=['F', 'A']).reshape(
-        shape)
+        shape, order='F')
 
 
 def _array66(value):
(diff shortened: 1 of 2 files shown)


willrogers:
Sorry, I've been on holiday. This looks fine to me.

PR #120, opened 2019-06-21
Misc changes

T-Nicholls:
Some miscellaneous changes I had left and a first draft of a Python radiation integrals function.
On `rad_int.py`:
- It is all very "rough and ready" and I do hope to work on improving it in the near future, but Will encouraged me to add the draft so there is some progress if I don't.
- The integrals differ from those calculated in Matlab:
    ```
    python = [8.6345849e-02, 8.6572883e-01, 1.2019837e-01, -7.5242001e-03, 1.7475844e-04]
    matlab = [8.7814944e-02, 8.6572883e-01, 1.2019837e-01, -7.6047592e-03, 1.7793325e-04]
    ```
- This is why it is not in `radiation.py`, where I assume it should go once completed?
- I've included `test_int.py` which I was using to compare the Python values with those from Matlab, but I guess would form the basis for a future comparison test.


lfarv:
The radiation integrals are computed here by integrating numerically inside the dipoles, splitting them in ~100 slices. Instead, one should use the analytical integrals as given [here](http://accelconf.web.cern.ch/AccelConf/p73/PDF/PAC1973_0900.PDF), which is more accurate.
I would like to have this branch merged as soon as possible for the numerous PEP8 conformity changes, and for `rad_int.py`:
- either leave it out of the merge, giving me a few days to validate the analytical integration,
- or merging as it is, knowing that the results are still preliminary (probably simpler)

willrogers:
OK, I've merged this branch but we're not depending on the radiation integrals yet so feel free to change the implementation.

PR #119, opened 2019-06-20
Wiggler and Exact Hamiltionian pass methods

T-Nicholls:
To provide a bit of background information, the Accelerator Physics group here at Diamond is still using an old modified version of AT1.3 and we would like to get them using AT2.0. After discussion with them on the topic, most of the additions to AT that they have requested are on this pull request.

This pull request consists of:
- @mashl's versions of `GWigSymlecticPass` and `GWigSymlecticRadPass` from #38 and #59 updated to work with pyat.
- @simoneliuzzo's pass method help from #33.
- @willrogers' `ExactHamiltionianPass` from #45, originally written by James Rowland.
- `BndStrMPoleSymplectic4Pass`, updated to work with pyat, and `idtable.m` sent to me by the Accelerator Physics group here at Diamond.
- The addition of a pyat `Wiggler` class for use with the new wiggler pass methods.

@ZeusMarti, if I remember correctly you were interested in / working on pass methods like those in this pull request, is that the case? If so, what progress did you make in this area?

Issue #17, opened 2017-10-11
no help for passmethods


simoneliuzzo:
it would be nice if we could type for ex. help DriftPass and have some message describing the PassMethod DriftPass.c . 
This can be doen creating appropriate DriftPass.m comment files in atintegrators.

simoneliuzzo:
I opened a branch for this development called pass_method_help

simoneliuzzo:
for those interested, take a look at the output of `>> help DriftPass` in matlab !

willrogers:
See atcollab/at#33

lnadolski:
@simoneliuzzo 
are you asking to get in the help file the mapping related to each passmethod?

PR #121, opened 2019-06-26
Wiggler changes

willrogers:
Continuation of #119.

I have got AP to do some testing and will update this branch with any necessary changes.


willrogers:
AP at Diamond is now using this branch and are happy. I think we need to merge this, it has been open for a long time.

@lfarv @swhite2401 @carmignani @lnadolski

willrogers:
@lfarv you mention documentation for pass methods, but I don't really see any documentation for any of them. Is #17 relevant?

lfarv:
Concerning radiation integrals:
`ringpara` successively computes radiation integrals for
- dipoles in `CalRadInt`
- wigglers in `RadIntegrals`

For clarity, I would prefer renaming these to `DipoleIntegrals` and `WigglerIntegrals`. However, considering the discussion in #168, there is much more to do (having `atsummary` and `ringpara` consistent, using analytical integral instead of numerical...). So I suggest to keep it as it is for now, and open a new pull request to answer #168.

Concerning the Matlab wiggler creation functions: element creation functions like `drift`, `quadrupole`, `wiggler`… are replaced by `atdrift`, `atquadrupole`, `atwiggler`… and are kept for compatibility reason. So one should put back the deprecation warning which has been removed from `wiggler` and have `wiggler` call `atwiggler`, as it is done for other functions. Also `atwiggler` has small errors and could be re-written with the new standard for processing default values and keyword arguments (see `atquadrupole` for instance), but these are details…

Concerning documentation, this branch introduces new PassMethods for which the behaviour is not obvious: for instance, I do not see what `BndStrMPoleSymplectic4Pass` does (straight or bent?), I do not see how to use `ExactHamiltonianPass`, and, for wigglers, what are the pros and cons of `GWigSymplecticPass` versus `IdTablePass`. I agree that other ('well-known") PassMethods are not documented, but adding small `BndStrMPoleSymplectic4Pass.m` to provide help would be useful. But again, that is not blocking…

So go on , and I'll the open a new branch for radiation integrals in Matlab.

willrogers:
I have addressed these comments as best I can and I will merge this once the tests have passed. If any problems have been introduced we will fix them as soon as possible.

PR #122, opened 2019-07-05
Compute analytically the radiation integrals

lfarv:
The 5 radiation integrals are computed analytically by `get_radiation_integrals` in `radiation.py`. 2 new functions  allow a comparison between results obtained from the radiation integrals (`radiation_parameters`) and from ohmi_envelope (`envelope_parameters`).
A slight difference with Matlab in i4 is under investigation


lfarv:
This is now OK for me. New features:
- a `Lattice.circumference` property,
- a `get_radiation_integrals` function / `Lattice` method to compute the radiation integrals,
- `radiation_parameters` and `envelope_parameters` functions and Lattice methods returning an object summarising some lattice properties, such that `print(ring.radiation_parameters())` gives a readable summary.

Function names can still be modified if there are better names!

lfarv:
Ok, @willrogers, go on merging.
You are right, comparison tests are very useful: I could fix a small bug in my formula for I4.
Here the comparison between numerical integration in Matlab (for I4 and I5) and analytical integrals in Python shows a very good agreement.

I see that the correction for NumIntSteps is still in a pending pull request, so I add it here.

I have also a question on the updates of the Pypi repository: when is it updated? Since pyat is still evolving quite fast, I would push for rather frequent updates…

willrogers:
In principle, if I push a tag `pyat-x.y.z`, wheels should be built and uploaded to [test PyPI](https://test.pypi.org/project/accelerator-toolbox/). If these work successfully I can then download them and upload them to the real PyPI.

I'm happy to try that now (0.0.2), then when we've merged the Diamond changes I can make a 0.0.3 release.

PR #124, opened 2019-07-09
New formats for loading/saving files

lfarv:
Load and save AT lattices in different formats:
- load and save binary .mat files
- load and save text .m files: format used by Matlab, each element is represented by the Matlab command which is used to create the element,
- load and save text .repr files: a new format, similar to .m, but each element is represented
by its python ‘repr’ string
- load from a Matlab cell array: can be used when run in python under Matlab to convert a Matlab lattice into a python lattice. 

In addition to the specific functions
`load_mat` / `save_mat`
`load_m` / `save_m`
`load_repr` / `save_repr`

there is a generic  Lattice method` ring.save(filepath)` which selects the format from the file extension and a corresponding static Lattice method `Lattice.load(filepath)`

Adding new formats should be easy.

This required a significant change in the Lattice constructor, to allow building lattices in a more flexible way. Here is an excerpt from the Lattice docstring:

```
Lattice(elems, **params)     Create a new lattice object

…
…

Custom iterators:
Instead of running through 'elems', the Lattice constructor can use one
or several custom iterators.

Lattice(*args, iterator=it, **params)

The iterator "it" is called as "it(params, *args)" and must return
a python iterator over AT elements for building the lattice. It must also
fill the "params" dictionary used to set the Lattice attributes.

        params is the dictionary of lattice parameters. It is initialised
               with the keywords of the lattice constructor. It can be
               modified by the custom iterator to add or remove parameters.
               Finally, all remaining parameters will be set as Lattice
               attributes.
        *args  all positional arguments of the Lattice constructor are sent
               to the custom iterator.

    An iterator can be:
        - a "generator" which yields elements from scratch.
          Examples: a list, or a file iterator,
        - a "filter" which runs through an input iterator, processes each
          element, possibly adds parameters to the params dictionary
          and yields the processed elements.

    Example of chaining iterators (taken from "load_mat"):

    Lattice(ringparam_filter, matfile_generator, filename,
            iterator=params_filter, **params)

    params_filter(params, ringparam_filter, *args)
        runs through ringparam_filter(params, *args), looks for energy and
        periodicity if not yet defined.

    ringparam_filter(params, matfile_generator, *args)
        runs through matfile_generator(params, *args), looks for RingParam
        elements, fills params with their information and discards them,

    matfile_generator(params, filename)
        opens filename and generates AT elements for each cell of the
        Matlab cell array representing the lattice,

```
Such iterators can be re-used, for instance `ringparam_filter` is used for both .m and .mat files.

As a side effect, there is also a new `copy` keyword in the method `radiation_on`. It defaults to `False`, but if set `True`, instead on modifying the lattice in-place, it returns a shallow copy of the lattice with radiation set on and leaves the lattice itself unchanged. This is convenient to keep both radiating and non-radiating versions of the lattice.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,7 +21,6 @@ install:
   - venv\Scripts\activate
   - cd pyat
   - pip install --only-binary=numpy,scipy -r requirements.txt
-  - python setup.py build
 
 before_build:
   - python setup.py develop
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,9 +40,10 @@ install: |
   fi
   cd pyat
   pip install --only-binary=numpy,scipy -r requirements.txt
-  python setup.py build
 
 before_script:
+    # Use develop as we will be running the tests from the directory
+    # containing the at package.
   - python setup.py develop
 
 script: |
--- a/pyat/README.rst
+++ b/pyat/README.rst
@@ -5,7 +5,7 @@ pyAT is a Python interface to the pass methods defined in Accelerator Toolbox,
 implemented by compiling the C code used in the AT 'integrators' plus a Python
 extension.
 
-It supports Python 2.7 and 3.3 to 3.6.
+It supports Python 2.7 and 3.4 to 3.7.
 
 
 Installation preparation (Windows)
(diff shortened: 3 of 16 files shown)


willrogers:
I've been on holiday, so I haven't looked at this yet. In principle it sounds good. We may want to load the Tracy file format (the owner of Tracy Johan Bengtsson is working at Diamond at the moment).

We have also been looking at implementing the `atmatch` function in pyat.

willrogers:
Proposal:

```
diff --git a/pyat/at/lattice/elements.py b/pyat/at/lattice/elements.py
index cc7aa9c..9815819 100644
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -68,10 +68,13 @@ class Element(object):
                      self.REQUIRED_ATTRIBUTES]
         defelem = self.__class__(*arguments)
         keywords = ['{0!r}'.format(arg) for arg in arguments]
-        keywords += ['{0}={1!r}'.format(k, v) for k, v in attrs.items()
+        keywords += ['{0}={1!r}'.format(k, v) for k, v in sorted(attrs.items())
                      if not numpy.array_equal(v, getattr(defelem, k, None))]
         return '{0}({1})'.format(self.__class__.__name__, ', '.join(keywords))
 
+    def __eq__(self, other):
+        return repr(self) == repr(other)
+
     def divide(self, frac):
```

If we sort the attributes in the `__repr__()` method we can use it to define `__eq__()`.

lfarv:
- `Elements` attributes sorted in `repr` as proposed by @willrogers 
- In preparation for a future `Wiggler` element, a dummy `Wiggler` (acting as a `Drift`) is added so that its processing in `radiation_on/off` can be implemented

T-Nicholls:
Would it be possible for the lattice to only be returned from `radiation_on/off` if `copy=True`, as currently calling, for example, `ring.radiation_on()` returns the lattice? I see no reason you would want the lattice returned if `copy=False`, and currently if I do not wish to have the lattice printed and fill up my terminal I must use something like `_ = ring.radiation_off()`.

willrogers:
I think this is ok to merge.

PR #125, opened 2019-07-11
Simplify setup

willrogers:
@lfarv @swhite2401 @T-Nicholls I believe these changes should improve `setup.py`.

In my testing this should work with both `python setup.py install` and `python setup.py develop` with no preceding build step, and it successfully builds the wheels on Travis as before.

### DIFF
--- a/pyat/setup.py
+++ b/pyat/setup.py
@@ -1,5 +1,7 @@
-import os
 import glob
+from io import open
+import os
+from os.path import abspath, basename, dirname, exists, join, splitext
 import sys
 import shutil
 try:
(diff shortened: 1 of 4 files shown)


pyat/setup.py
```diff
@@ -8,84 +10,92 @@
...
+# sufficient.
+at_source = abspath(join(here,'at.c'))
+integrator_src_orig = abspath(join(here, '..', 'atintegrators'))
+integrator_src = abspath(join(here, 'integrator-src'))
+diffmatrix_source = abspath(
+    join(here, '..', 'atmat', 'atphysics', 'Radiation')
+)
+
+if exists(integrator_src_orig):
+    diffmatrix_source = abspath(join(here, '../atmat/atphysics/Radiation'))
```
lfarv: How does this definition of diff_matrix_source differ from the one just above,

willrogers:
For me the failure case was:

* the files are copied in the build step
* however, when you run `install` the glob that determines which extensions to build (`pass_methods = glob.glob(join(integrator_src, '*Pass.c'))`) is run before the build step
* for a first clone, if you run `python setup.py install`, 
  * you don't find any pass methods when you first run
  * you run the build step: copy the integrators and build only `at.c` and `diffmatrix.c`
  * I think this was the problem seen by @swhite2401 
* if you now run `python setup.py install`
  * the pass methods are already there so you find them
  * you run the build step: copy the integrators again and build `at`, `diffmatrix` and all the pass methods

That's why running `python setup.py build` followed by `python setup.py install` always worked.

Unless you have a new clone or remove `integrator-src` and `build` you don't see any problems.

I understand your comment about `python setup.py --help` but this is an improvement I think.

T-Nicholls:
I take it by the readdition of twine upload to appveyor.yml that the wheels also build successfully on Appveryor? If so then this looks good and is probably a better solution than only copy during build anyway.

willrogers:
Yes, wheel building works on Travis and Appveyor.

lfarv:
OK, perfect for me!

PR #127, opened 2019-08-19
remove numpy options incompatible with numpy < 1.14

lfarv:
@T-Nicholls  noticed that the functions `at.save_m` and `at.save_repr` were using numpy print options incompatible with numpy < 1.14. This is solved by using compatible options.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,7 +7,8 @@
 image: Visual Studio 2015
 platform: x64
 environment:
-  CIBW_BEFORE_BUILD: "pip install numpy"
+  # Build against the oldest supported version of numpy.
+  CIBW_BEFORE_BUILD: "python -m pip install numpy==1.14.3"
   matrix:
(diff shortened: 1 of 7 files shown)


lfarv:
This should allow running `save_repr` with numpy < 1. 14. This also fixes a bug when saving elements with matrix attributes (`R1`, `R2` for instance), where the `repr` string was extending over several lines, which caused a crash when reading.

For me, it's ready, I'll be away for about 3 weeks.

willrogers:
There is another reason to support 1.14.3 and above - older versions don't work with Python 3.7. So I suggest requiring 1.14.3 and above. I have made this change in #130.

lfarv:
@willrogers: Although numpy >= 1.14.3 is now ensured since #130, these print options are simpler than before, and in addition this branch fixes a bug in the processing of `.repr` files (multiline elements). So for me it's ready for merging.

PR #135, opened 2019-10-18
Atenergy

simoneliuzzo:
atenergy does not fail if energy field is set in all elements identical (as set from atsetenergy)
plotB0curlyH is modified to use atenergy


atmat/atplot/plotfunctions/plotB0curlyh.m
```diff
@@ -1,7 +1,8 @@
 function plotdata=plotB0curlyh(lindata,ring,dpp,varargin)
 
 spl=299792458;
```
lfarv: Could be `PhysConstant.speed_of_light_in_vacuum.value`

PR #138, opened 2019-10-31
Fix behaviour of lattice_pass for non-F_CONTIGUOUS input

lfarv:
For multi-particule input and C-ordered input array,  the `lattice_pass` tracking was correct, but the value of `r_in` was not reflecting the coordinates at the end of the tracking, but still at the beginning.
When using multi-particle C-ordered input arrays (Python default), `lattice_pass` must make a copy at both input and output of the tracking. To mitigate this efficiency reduction, there are 3 possibilities:

1. Implementing transparently both copies (up to now, only the copy at input was done). This is the present solution in this pull request,
2. Implementing both copies with a warning on efficiency-loss when using non-F_CONTIGUOUS input,
3. enforcing the use of F_CONTIGUOUS input.

What is your preferred solution ?

### DIFF
--- a/pyat/at/tracking/track.py
+++ b/pyat/at/tracking/track.py
@@ -23,28 +23,30 @@ def lattice_pass(lattice, r_in, nturns=1, refpts=None, keep_lattice=False):
 
     PARAMETERS
         lattice:    iterable of AT elements
-        r_in:       6xN array: input coordinates of N particles
+        r_in:       (6, N) array: input coordinates of N particles.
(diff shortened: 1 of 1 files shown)


willrogers:
For me this is ok. I would add a warning once we find that there is a performance problem that is affecting someone, but we don't necessarily need one yet.

lfarv:
@willrogers : I do not expect performance problems except for repetitive calls with large input data. So we'll see!

PR #140, opened 2019-11-05
Improve type coercion in Element constructors

lfarv:
Attribute types are enforced even in some cases where attributes are not strictly required by the element class. This done by moving are conversions in the base Element type.


lfarv:
Attribute types are enforced even in some cases where attributes are not strictly required by the element class. This done by moving are conversions in the base `Element` type.

lfarv:
Merged in to master

PR #141, opened 2019-12-04
Fix phase advance computation in twissring.m

lfarv:
Correction of the bug in the phase advance computation in `twissring` reported by @lnadolski. We now use `atan2` instead of `atan` so that unwrapping with 2*pi is correct.


lnadolski:
The squeeze function is not properly set.

MX = atan2(squeeze(MS(1,2,:)), squeeze(MS(1,1,:)*bx-MS(1,2,:)*ax));
MY = atan2(squeeze(MS(3,4,:)), squeeze(MS(3,3,:)*by-MS(3,4,:)*ay));

PR #144, opened 2020-01-02
removed warning from new matlab versions

lfarv:
Correct the `getcellstruct` function where new Matlab versions issue a warning.


lfarv:
@willrogers : a new definition of physical constants in script (classical electron radius, electron mass) caused the tests to fail. I modified the target value for the energy loss test to match the new result, and relaxed the accuracy test so that old results still succeed

lfarv:
Tests are now successful, except for AppVeyor which complains for `RuntimeError: Python 3.5 or later is required`. I will anyway merge since only Matlab code is changed. @willrogers, can you look at the AppVeyor problem ?

PR #145, opened 2020-01-14
Appveyor: drop Python 3.4 and add Python 3.8.

willrogers:
Python 3.4 is at end of life so we don't need to worry too much about
supporting it any more. Support is no longer working correctly on
Appveyor.


willrogers:
@lfarv I don't think there's any need to keep testing Python 3.4, and I've added Python 3.8 and fixed a few warnings that caused test failures.

PR #150, opened 2020-03-24
added dp/p option

simoneliuzzo:
Option for lattice matching off energy. 
Also includes a modification of the input parsing for atmatch, that should not change the previous input logic (if there was any logic a part chronologically adding stuff...).
Adds also an additional optional input for optimset with parallel pool. This input will be more useful later when/if switching from optimset to optimoptions


simoneliuzzo:
apparently yes. atmatch has a dpp option. I did not compare the master with this 3 years old version.

PR #153, opened 2020-04-22
Possibility to set a default values in atgetfieldvalues

lfarv:
An option of `atgetfieldvalues(...,'Default',default_value)` allows to specify a default value for missing attributes of an element, rather that `NaN` or `[]` which were used, and are still used by default.


carmignani:
I also tested it. It works and it is useful!

PR #154, opened 2020-04-28
Load elegant tracy

willrogers:
This adds capability to parse Tracy's `.lat` files and Elegant's `.lte` files.

Since these are based only on existing lattice files, I can't make them complete, but they are a useful start.

Any comments are welcome, including any knowledge you have of the file formats.


willrogers:
Also, I haven't really taken the time to comment the functions or make the tests complete, so if there's anything that looks questionable I will update it.

willrogers:
@lfarv any comments?

PR #156, opened 2020-05-12
Added average optics function calculation

swhite2401:
Hello,
I added the function avlinopt to linear.py, it is based on the matlab sr.model function used to compute the average beta,D and mu at chosen ring elements.
I did a comparison and it looks ok to me, please let me know in case you find something odd.
Cheers,
Simon

### DIFF
--- a/pyat/at/physics/linear.py
+++ b/pyat/at/physics/linear.py
@@ -7,7 +7,7 @@ from at.lattice import Lattice, check_radiation, uint32_refpts, get_s_pos
 from at.tracking import lattice_pass
 from at.physics import find_orbit4, find_m44, jmat
 
-__all__ = ['get_twiss', 'linopt', 'get_mcf']
+__all__ = ['get_twiss', 'linopt', 'avlinopt', 'get_mcf']
 
 DDP = 1e-8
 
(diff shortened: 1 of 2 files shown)


lcarver:
Few things left undefined in the return.
I ran: 

```
ring = at.Lattice(ring)

res = at.avlinopt(ring)

```
and got an error:

`NameError: avespos is not defined `

swhite2401:
I have fixed the bug with avespos not defined.
On the request of @lcarver

lcarver:
Looks fine to me. Did a few tests of the different features, got what seemed to be sensible results out.

lfarv:
I added in the Matlab tests a test of this new function. The agreement is better that 1.0e-20 on-momentum, but off-momentum I had to relax the tolerance to 1.0E-10 for the dispersion.

swhite2401:
Hello Laurent,
Only the dispersion fails? That is odd...

swhite2401:
Thanks Laurent! Sure if you have improvements they are more than welcome!

I thought pyat did not allow negative length, did that change?

lfarv:
We had a discussion on negative lengths, they are useful and should work. But there may be problems somewhere, please report !

lfarv:
Proposal for vectorised code.

swhite2401:
Very nice! If you agree it is ok for me to merge.

Concerning negative lengths, it is just that I did not know you introduced it. So far, no problems on my side.

lfarv:
I found a problem with refpts. I’ll correct as soon as I can get a computer…

PR #157, opened 2020-05-15
Harmonic analysis

lcarver:
The aim of this pull request is to include some python based harmonic analysis (python version of SUSSIX/NAFF). It incorporates a module developed by Jaime Maria Coello De Portugal - Martinez Vazquez from CERN around 2016. This is a reduced version keeping only the key components.
Full code can be found at: https://github.com/pylhc/harpy

I added into the at/physics directory two files. one is the reduced harmonic_analysis module (called harmonic_analysis.py). The other is called analysis.py and contains a function called compute_tunes which can be called to easily make use of the harmonic analysis module. 

I was not sure of the best way of organising these kinds of functions, so if you feel it should be changed then it is a good idea to do it now before it gets out of control.

A simple test of this function can be done using the code below

```
import numpy as np
import at

amplitudes = [3., 0.5, 0.1]
tunes = [0.23, 0.16, 0.46]
turns = np.arange(100)

centroid = 0
for amp, tune in zip(amplitudes, tunes):
    centroid += amp * np.sin(2. * np.pi * turns * tune)

tunes, amps = at.physics.analysis.compute_tunes(centroid, num_harmonics=25) 

plt.stem(tunes, amps)
plt.show()

```


lcarver:
After a discussion with @swhite2401 outside of git, he suggested it is better to use this as an opportunity to write a get_tune function with several different available methods. This function will be put into linear.py

lcarver:
Added a get_tune function in linear.py (in commit c46478303333b74aac9f5b4b400e35f4273a8431 )
Now there is a get_tune function which is attached to the ring lattice object. I included some descriptions in the help for general usage, but for some help I have included some code below that can be used to start you off as a test. 
```
import numpy as np
import at

ring = at.Lattice(ring)

print('LINOPT')
print(ring.get_tune())

print('FFT')
print(ring.get_tune(method='fft', nturns=128, ampl=1e-6))
print(ring.get_tune(method='fft', nturns=256, ampl=1e-6))
print(ring.get_tune(method='fft', nturns=1024, ampl=1e-6))
print(ring.get_tune(method='fft', nturns=4096, ampl=1e-6))

print('HARMONIC')
output_x, output_y = ring.get_tune(method='harmonic', nturns=128, ampl=1e-6, nharmonics=20)
tunes_x, amp_x = output_x.T
tunes_y, amp_y = output_y.T

fig,ax = plt.subplots()
ax.stem(tunes_x,amp_x)
ax.set_xlim(0,1)
ax.set_title('Horizontal')
plt.show()

fig,ax = plt.subplots()
ax.stem(tunes_y,amp_y)
ax.set_xlim(0,1)
ax.set_title('Vertical')
plt.show()


```

swhite2401:
Hello Lee,
I have proposed an alternative implementation, please take a look and let me know.
We can always roll back.

swhite2401:
Ah and we should also allow for dp!=0

lcarver:
I think your method is better as it involves more of the features that were written in the original harmonic analysis scripts (for example specifying the hanning window). A comment relating to this is that it seems unnecessary to specify num_harmonics when anyway we are just returning the highest amplitude harmonic as the tune. Maybe additional generalisation can be put in place where the get_tune only asks for the highest harmonic.

But (as we discussed outside of git) some better organisation of these functions are needed. To achieve this I have merged the master into this branch and we will create a script called nonlinear.py. I already have some functions for computation of a detuning module that can go into this and @swhite2401 will develop some other functions for RDTs and some others.

lcarver:
As we are doing a lot of development now we will close this request without merging and make a new pull request when we are ready

swhite2401:
-Concerning num_harmonics I still see an interest for it in case the max of the spectrum is not the tune

-Also I would like to leave this request open for Laurent to review it. get_tune is of general interest and it would be very useful to merge it into the master such that it can be part of future developments like fittune or amplitude_detuning

lfarv:
@lcarver: please don't close the pull request. If you are afraid of a non-desired merge, you could try to set it as 'draft' (top right of the window, below 'reviewers'), but I don't know what it does exactly...

lcarver:
Hi Laurent, I wrote the message saying I would close without merge. But then Simon advised me not to so I deleted the message. But of course the emails are still sent. This branch will stay as it is for your review.
Sorry for the confusion.
Cheers,
Lee

lfarv:
@lcarver: Concerning physics, I let you go on with development, but I have already a comment about plotting: for several reasons, in particular
- `matplotlib` is optional to run payt,
- `matplotlib` import time is penalising when it is not necessary,

it has been decided that plotting must be explicitly requested by calling
`import at.plot
`
As a consequence, no module should import `matplotlib` unless it is in the `plot` sub-package.
A class such as `HarmonicPlotter`, purely dedicated to plotting,  can be moved to the `plot` package, while plotting methods of more general classes can be added to the class from within the `plot` package like it is done for instance for `Lattice.plot_beta`, at the end of ` pyat/at/plot/specific.py`.

Also, do not change the matplotlib backend: as indicated [here](https://matplotlib.org/tutorials/introductory/usage.html#backends), the backend is normally setup in the user's `matplotlibrc` file or `MPLBACKEND` environment variable. It can be programmatically changed with `matplotlib.use()`, but only before any figure is created! Since available backends depend on the platform and installation, don't try to force a non-available one...

lcarver:
Several things. 

First, I merged pyat_fittunechrom into this branch to bring it up to date with Simons latest devs. 

Then I PEP8-ified (PEPonised?) the linear.py and harmonic_analysis.py. 

I also deleted the Class HarmonicPlotter as there was really no need for it (it came as part of the original module). 

Ready for review.

lfarv:
@lcarver : It now works as expected. Still a few PEP8 warning (double quotes), the most important being that `_compute_coef_goertzel` should be a static method. Otherwise, it's OK for me.

lcarver:
Where are you seeing the PEP8 warning? We are running a function on the ESRF machines called PEP8 and it doesn't show this as a warning.

lcarver:
Few comments implementing Laurents request. Have also merged with master into this branch so this branch is ready to merge.

PR #158, opened 2020-05-18
Pyat fittunechrom

swhite2401:
Several changes here:

-this branch integrates avlinopt, this is a mistake

-utils.py: 
modifed get_elements to return refpts (optional)
added set_value_refpts, get_value_refpts to set and get attributes based on refpts and attribute name

-added a sub-package matching and module globalfit.py
Several functions all used in fit_tune_chrom that allows to fit the tune an chromaticity of the ring
this is based on the matlab implementation using 2 families
Minor improvement allows to ask for a tolerance and number of iterations: the function loops until one of these is reached


lcarver:
Hello,
Few comments on utils:

1. the indexes that are returned from get_elements should be a np array. This can then be used for masking and other things more conveniently.
2. would be useful to add in the help of get_elements that the output is elems, inds= ring.get_elements(...). 
3. do you think that get and set_value_refpts should be a part of the lattice object? 
4. I am having troubles with the set_value_refpts, see below code:

```
import numpy as np
import at

ring = at.Lattice(ring)

elems, inds = ring.get_elements('QF8*', return_refpts=True)
ks1 = at.lattice.utils.get_value_refpts(ring, inds, 'K')  
at.lattice.utils.set_value_refpts(ring, inds, 'K', ks1/2)  
ks2 = at.lattice.utils.get_value_refpts(ring, inds, 'K')  
```
ks1 and ks2 are the same. i think you either have to return the modified lattice or this is ok if being used on a lattice object. I'm not so sure, maybe I am using it wrong. I am interested in Laurents opinion.

lcarver:
For the fittunechrom:

I played around a bit and the tune and the chromaticity modification seems to work very well
The help needs to be expanded a lot for the inputs as it took a little digging to figure out specifically what the function wanted, anyway it was easily run with e.g.: 
at.matching.fit_tune_chrom(ring, 'tune', 'QF1*', 'QD2*', [0.1,0.25])
at.matching.fit_tune_chrom(ring, 'chrom', 'SD*', 'SF*', [0.1,0.25])

1. At the moment the elements are selected by specifying the key that it used in get_elements. Probably this should also be able to accept a list of refpts to achieve the same thing.
2. Would be useful to add a check that the return from get_elements is not an empty list (assert fam1i is not [] ) or something

swhite2401:
-Help has been improved as suggested
-set_values_refpts was modified to work also with properties
-the function does not take a list of refpts as input, this could be considered as a future development

swhite2401:
Hello Lee,
I have committed a proposal for your harmonic analysis code and integration in at.
It is quite different from your implementation. Please take a look and let me know.
We can always rool back.

lfarv:
On `get_elements`

1) DO NOT USE `index`: this is extremely slow: each element has to be compared with each elements of the lattice (until it is found). So it is `len(elems)` loops on the lattice, and on `average len(elems)*len(lattice)/2` tests for equality. It is much faster to repeat the search with 
`[id for id, elem in enumerate(ring) if check(elem)]`
(single loop)

2) Given that, could this feature be in a separate function so that we avoid flags ?
`get_indexes` returns a list of indexes
`get_elements` returns a list of elements (which can be as simple as ring[get_indexes(…)]

3) In fact, these functions consist in a single line: they have no added value compared to a line like
`[check(elem) for elem in ring]` (for a boolean mask), which is faster (no overhead for function call), easier to understand (you do not need to know what the function does), and more flexible (you can choose a boolean mask, an integer list, an iterator if convenient…). In particular, if you need both elements and indexes, you can get them simultaneously! 

To summarize: rather than a single function with a flag, I prefer two functions. I have nothing against adding a new function, but personally, I would use standard python syntax rather than the function call.

swhite2401:
I personally like single functions with flags rather than multiplying the number of functions that do almost the same thing.
However this is a question of taste and an easy fix: the last commit includes 2 functions as suggested.

Concerning the python syntax, on the principle I agree, however for someone like me whose python is not fantastic having a function can be useful and the fact that it exists does not prevent from using the python syntax.

lfarv:
Another suggestion:
```
def fit_tune(ring, key1, key2, newval, tol=1.0e-12, dp=0, niter=3)
    _fit_tune_chrom(ring, 1, _get_tune, key1, key2, newval, tol=tol, dp=dp, niter=niter)

def fit_chrom(ring, key1, key2, newval, tol=1.0e-12, dp=0, niter=3)
    _fit_tune_chrom(ring, 2, _get_chrom, key1, key2, newval, tol=tol, dp=dp, niter=niter)

def _fit_tune_chrom(ring, order, get_data, key1, key2, newval, tol=1.0e-12, dp=0, niter=3)
    almost unchanged...
```
The interface is simple and it eliminates the keyword `varname` and the function `_get_opt_data`

I also suggest that you put @willrogers  as reviewer, he will help in defining a consistent interface for the whole project.

And then details:
- in `_get_tune`, you can set `get_chrom` to `false` if you need only the tunes (3 times faster)
- I get a large number of warnings about PEP 8 conformance (spaces around operators, blank lines...). If you can open the code in PyCharm, it shows everything

willrogers:
We can put in some checks for some of the PEP-8 styles so that the tests fail.

I know it's a bit annoying to start with, but it ensures that we keep all the code consistent, which I think is really valuable for a large project.

I agree with Laurent, `fit_chrom` and `fit_tune` are different things, even if they do them in the same way. Switching what happens based on `varname` looks like Matlab code to me.

swhite2401:
-corrected PEP8 errors (also corrected in utils.py and linear.py)
-suggestion by Laurent implemented
-> can we merge this part?

Conerning Will's comments, PEP-8 style checks could be useful, I must admit that my editor was not showing them, I just asked IT group to install pyCharm on our systems.

Now on the coding philosophy, fit_tune and fit_chrom are mathematically identical, just the variables and observable are different. Having 2 functions with the same code inside is for me counter productive. It is also difficult to maintain in case a change is necessary in this code that is copied over. That being said, I was not happy with varname and the proposal by Laurent solves this issue.

swhite2401:
On a different topic, I think having a function to set the strength of a magnet (only the main multipole) without having to specify the order of PolynomB would be very useful.

This is similar to the attributes K and H for quadrupoles and sextupoles but with the same name for all magnet in order to be able to call it in a generic manner.

I will make a proposal in this sense if you don't mind.

lfarv:
Another question on the interface: the magnet selection is done through a pattern matching. What about a selection through refpoints ? `key1, key2` would be replaced by `refpts1, refpts2`.

**Advantage**: it's more general: any magnet combination can be represented, while it may not be possible with pattern matching (Ex: all QF1 except in the injection cell).

**Drawback**: In simple cases, calling needs 2 additional steps:
`fit_tune(ring, get_refpts(ring, key1), get_refpts(ring, key2), ...).
`

I leave the question open, my preference is for refpoints.

About the code, I still get warnings for double-quoted strings for docstrings.
Otherwise I'm happy with that version.

swhite2401:
Thanks Laurent.
I agree with you refpts is more flexible, can we add a check in the function
to search or not for refpts?
Something like:

`if not isinstance(key, (list, np.array)): get_refpts(ring,key)`

lfarv:
@swhite2401 : You are right, keys could be tested in `_fit_tune_chrom`, but also in `get_refpts` itself (returning the input unchanged in case of refpoints). But I would keep it simpler:

- fit_tune accepts refpoints, as most other AT functions, and does not rely anymore on `get_refpts`,
- it's up to the user to generate the refpoints, directly in python style, or using any convenience function, according to his preferences. Anyway it's a single statement!
- it avoids tricky checks for refpoints (they can be `list`s, `ndarray`s, or even simple integers)

I think it's cleaner that way, but you may disagree and additional tests do not change the interface…

swhite2401:
No you are right, I put it in and merge this part then

PR #162, opened 2020-07-02
get_chrom convenience functions

lcarver:
Added some functions to easily compute the chromaticity (linear or non-linear). Please see below for an example on how to use it. I wrote it in such a way that the user has full control over each of the tune computations so all functionalities of get_tune can be used and also included tune tracking for the more difficult cases.

```
import numpy as np
import at

ring = at.Lattice(ring)

print(ring.get_chrom())
print(ring.get_chrom(method='scan'))
print(ring.get_chrom(method='scan', dp_range=np.linspace(-1e-3,1e-3,11), tune_method='laskar', nturns=512, ampl=5e-6, fmin_x=0, fmax_x=0.5, fmin_y=0, fmax_y=0.5, remove_dc=True) )

chrom_x, chrom_y = ring.get_chrom_nonlinear(fit_order=3, dp_range=np.linspace(-1e-2,1e-2,21))
print(chrom_x)
print(chrom_y)


chrom_x_sus, chrom_y_sus = ring.get_chrom_nonlinear(fit_order=3, dp_range=np.linspace(-2e-3,2e-3,21), tune_method='laskar', nturns=1024, ampl=30e-6, fmin_x=0, fmax_x=0.5, fmin_y=0, fmax_y=0.5, track_tune=True, window=0.05, remove_dc=True)
print(chrom_x_sus)
print(chrom_y_sus)

```

When debugging I found a need for the remove_DC flag as I found the DC component of the beam spectrum in the laskar method was larger than the tune, in this case by setting the remove_dc flag to True the mean TbT position is removed from the bunch TbT data before being sent for spectral analysis. 

Some PEP8ifying still do but wanted to wait until it was approved.

### DIFF
--- a/pyat/at/physics/linear.py
+++ b/pyat/at/physics/linear.py
@@ -2,14 +2,14 @@
 Coupled or non-coupled 4x4 linear motion
 """
 import numpy
-from math import sqrt, atan2, pi
+from math import sqrt, atan2, pi, factorial
 from at.lattice import Lattice, check_radiation, uint32_refpts, get_s_pos, \
     bool_refpts
 from at.tracking import lattice_pass
 from at.physics import find_orbit4, find_m44, jmat
 from .harmonic_analysis import get_tunes_harmonic
 
-__all__ = ['get_twiss', 'linopt', 'avlinopt', 'get_mcf', 'get_tune']
+__all__ = ['get_twiss', 'linopt', 'avlinopt', 'get_mcf', 'get_tune','get_chrom', 'get_chrom_nonlinear']
 
 DDP = 1e-8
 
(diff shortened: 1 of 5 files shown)


willrogers:
Hi, I've noted this. I've got a lot on at work this week.

I would like to use this branch to try putting style checks on, so we automatically catch PEP-8 violations.

swhite2401:
Hi @willrogers , thanks for taking a look. 
I have made many changes today and I would qualify this as work in progress.
It is however a very good testbench for PEP8 as we have not cleaned up the code so far
so you can expect many violations.

lcarver:
This latest implementation by @swhite2401  is a significant improvement and is the one that should be reviewed. I committed some small changes after, adding the beta into the imaginary part of the gen_centroid and updated the comment in the chromaticity function.

willrogers:
I can't really check the physics, but I have two comments:

First, we are repeating calculations: if you follow `tunes_vs_amp()` then you call `find_orbit4()`, then `get_tune()`. But since `get_tune()` with those arguments just uses `linopt()` you could just call `linopt()` and get both the tunes and the closed orbit.

It's not so easy to spot these things but some of the computations are pretty expensive and it is worth avoiding duplication.

Second, you're not adding any tests. The tests for these sorts of things simply record the known values for a given lattice, often `dba.mat` or `hmba.mat`. These tests mean that if someone inadvertently changes the algorithm later then it will be noticed. Of course we need to choose an accuracy such that the tests don't fail just because they are being run on a different architecture or some such thing. See `test/test_physics.py` for examples.

What do you think?

I'll add the PEP-8 checks on a different branch.

swhite2401:
Hi @willrogers I agree with your comments and implemented the requested modifications

willrogers:
Thanks! This is fine to merge for me.

pyat/at/physics/linear.py
```diff
@@ -525,49 +527,86 @@ def get_tune(ring, method='linopt', **kwargs):
...
         tunes = np.array([Qx,Qy])
     """
 
-    def gen_centroid(ring, ampl, nturns):
-        p0 = numpy.zeros((6, 2))
-        p0[0, 0] = ampl
-        p0[2, 1] = ampl
+    def gen_centroid(ring, ampl, nturns, dp, remove_dc):
+        orbit, _ = find_orbit4(ring, dp)
+        ld, _, _, _ = linopt(ring, dp)
```
lfarv, line 538: To avoid a double computation of `find_orbit4`, you should use
```
orbit, _ = find_orbit4(ring, dp)
ld, _, _, _ = linopt(ring, dp, orbit=orbit)
```

swhite2401:
I have made the changes in the get_chrom branch, I will merge it right away in master

swhite2401:
Actually I have the following check that fails:

Travis CI - Pull Request Action required after 3m — Build Errored 

Any idea where this could come from @lfarv @willrogers ?

willrogers:
It looks like a network glitch. I re-ran the job and it has succeeded.

swhite2401:
Ok perfect I merge then

PR #163, opened 2020-07-07
Enable flake8 checks using Travis.

willrogers:
Some violations ignored, some fixed, others for discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,4 +60,5 @@ script: |
   else
     python -m pip install pytest-cov
     python -m pytest test --cov-report term-missing --cov=at
+    flake8 at
   fi
--- a/pyat/.flake8
+++ b/pyat/.flake8
@@ -1,0 +1,8 @@
+[flake8]
+
+max-line-length = 88
+# Allow importing * for package managment.
+ignore = 
+    F401
+    F403
+    W503  # no longer recommended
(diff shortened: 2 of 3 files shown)


willrogers:
@lfarv @swhite2401 @lcarver you can see style failures here:

https://travis-ci.com/github/atcollab/at/jobs/358416359

We can disable categories if we don't agree with them. I can fix them if we do.

lfarv:
Is it possible to disable a category in particular cases (using comments for instance)? If not, there are some categories that may be annoying. I spotted "ambiguous variable name" for instance: for well-known quantities, using their usual symbol makes sense. Or "module level import not at top of file", where in some cases, looking at import errors allows to adapt the behaviour to different configurations. Practically, I would keep only the tests related to the layout of the code (whitespace, blank lines, indents…), but not ones interfering with the coding. Most code editors make a similar job but are more easily configurable. Anyway, I'm open for discussion !

willrogers:
You can disable entire categories in the configuration file. You can also disable a category for a particular file.

You can disable for a specific line with a comment e.g. `# noqa: F402`. I deduce that you are using Pycharm because sometimes you have put `noinspection PyPep8Naming` (for example) but this comment only applies to Pycharm.

I assume that you can configure Pycharm to use these flake8 checks but I haven't double-checked.

willrogers:
Generally I am in favour of automated checks, but it seems it's not going to be that straightforward to agree on them, so I will close this PR.

Issue #164, opened 2020-07-10
closed orbit in pyat ohmi envelop


swhite2401:
Hello @willrogers @lfarv @lcarver ,

I was working on quatum diffusion and saw a potential issue in ohmi_envelop

I the present code there is this line:
`    bb = [find_mpole_raddiff_matrix(elem, orbit, energy)
          if elem.PassMethod.endswith('RadPass') else b0 for elem in ring]`

while I would have expected

`    bb = [find_mpole_raddiff_matrix(elem, orbit[i], energy)
          if elem.PassMethod.endswith('RadPass') else b0 for i,elem in enumerate(ring)]`

to properly take into account the closed orbit at elem, it seems to me that in the present implementation 
the initial orbit is taken for all elements.

Cheers,.
Simon

swhite2401:
In matlab at it is implemented as follows

`B(radindex)=cellfun(@findmpoleraddiffmatrix,...
    ring(radindex),orb(radindex),'UniformOutput',false);`

I suppose this confirms the problem, I will correct it in the branch quantdiff so
that the bugfix is implemented together with the addition of quantum diffusion

lfarv:
@swhite2401 : I agree, it's a bug. Thanks !

PR #166, opened 2020-07-14
Quantdiff

swhite2401:
Hello,
I have added function to compute diffusion matrices and quantum diffusion in radiation.py
This is useful multi-particle tracking to obtain equilibrium emittances


lcarver:
I ran the following code and got an error:

```
import at
import numpy as np
from at.physics.radiation import gen_quantdiff_elem

ring = at.load_mat('./test_matlab/hmba.mat',mat_key='RING')
ring = at.Lattice(ring)
ring.radiation_on(quadrupole_pass='auto')

qd = gen_quantdiff_elem(ring)
ring.append(qd)

rin = np.zeros((6,1))
at.track.lattice_pass(ring, rin, nturns=100)

```
this returns an error:

```
Error in tracking element 121.
RuntimeError: The attribute Lmatp is not Fortran-aligned.

```


The error can be fixed by including the following in gen_quantdiff_elem
`    lmat = numpy.asfortranarray(_lmat(dmat))`

swhite2401:
@lfarv @willrogers any objection in merging this branch?
It also fixes bug #164

pyat/at/physics/radiation.py
```diff
@@ -22,6 +24,57 @@
...
+        orbit, _ = find_orbit6(ring, keep_lattice=keep_lattice)
+        keep_lattice = True
+
+    orbs = numpy.squeeze(
+        lattice_pass(ring, orbit.copy(order='K'), refpts=allrefs,
+                     keep_lattice=keep_lattice), axis=(1, 3)).T
+    b0 = numpy.zeros((6, 6))
+    bb = [find_mpole_raddiff_matrix(elem, orbs[i], energy)
+          if elem.PassMethod.endswith('RadPass') else b0
+          for i, elem in enumerate(ring)]
```
lfarv, line 56: Why not using a single iterator for both `ring` and `orbs`? It's more efficient…
```suggestion
    bb = [find_mpole_raddiff_matrix(elem, elemorb, energy)
          if elem.PassMethod.endswith('RadPass') else b0
          for elem, elemorb in zip(ring,orbs)]
```

swhite2401: Good point thanks!

swhite2401:
For the double quote the pep8 command on rnice does not show it so I keep missing it, I ll try to think of putting it in properly in the future

Issue #171, opened 2020-08-11
get_twiss() usefull?


swhite2401:
Hello @lfarv , @willrogers ,

I took a look at linear.py and it seems to me that get_twiss() does not have any added value as it does exactly the same as linopt with coupled=False.
Is there any good reason to keep it?

If this function is required, it could at least be 

`def get_twiss(ring):
    return linopt(ring, coupled=False)`

to remove unnecessary duplicated code. What do you think?

T-Nicholls:
Hi @swhite2401,

They are almost identical however, unlike `linopt`, `get_twiss` does not calculate `A`, `B`, `C`, or `gamma` and, as a result, it is 1.5-2x faster than uncoupled `linopt`. Though the run time difference isn't as large as it used to be (#90), `get_twiss` may still useful for high-frequency recalculations of twiss data where `A`, `B`, `C`, and `gamma` are not needed.

willrogers:
Thanks @T-Nicholls, I remember now. As @swhite2401 didn't realise this (perfectly reasonably) then it would be worth adding a comment to the `get_twiss()` function stating this explicitly.

swhite2401:
Hi @T-Nicholls,

I just checked and get_twiss is faster by 30%. The 2 lines that make the difference are:

`        lindata['A'] = [ms.dot(A.dot(_jmt.dot(ms.T.dot(_jmt.T))))
                        for ms in msa]
        lindata['B'] = [ms.dot(B.dot(_jmt.dot(ms.T.dot(_jmt.T))))`

Duplicating a 100 lines of code is not efficient and maintenance gets more complicated. Also
AT users are used to work with linopt and I think get_twiss() will simply never be used.

Having a faster minimal twiss computation could however be useful in some cases like matching
(not implemented yet in pyAT) so I suggest that an option is added as input argument in linopt
to return or not A and B and get rid of get_twiss()

This will improve readability and future maintenance/devlopments.

swhite2401:
@willrogers @T-Nicholls 
I have started a branch to adapt linopt to transfer lines, following this discussion I also propose an implementation that matches the speed of get_twiss() for coupled = False by returning NaNs for A,B,C,g

This would allow to have a single function the compute twiss functions

You can take a look at the branch twissline

willrogers:
Yes, I think this is appropriate, as long as we are clear that the speeds are equivalent.

[...]

PR #172, opened 2020-08-13
adapted linopt for tranfer line

swhite2401:
Hello @lfarv 
I have implemented a potential adaptation of linopt for transfer lines.

-additional optional argument twiin to consider a transfer line

Also included in this branch:
-do not return unnecessary outputs for uncoupled lattice (speed-up)
-remove get_twiss()

Could you please take a look and let me know what you think?

(I haven't fixed the help and PEP8 yet)

Thanks!


pyat/at/physics/linear.py
```diff
@@ -305,17 +217,25 @@ def analyze(r44):
...
+        a0_b, b0_b, tune_b = _closure(B)
+        tune = numpy.array([tune_a, tune_b])
+
+    if get_chrom:      
+        if twiin is not None:
+            orbit[0:5] += numpy.append(disp0*0.5*ddp, 0.5 * ddp)
         d0_up, tune_up, _, l_up = linopt(ring, dp + 0.5 * ddp, uintrefs,
                                          keep_lattice=True,
-                                         coupled=coupled)
+                                         coupled=coupled, twiin=twiin)
```
lfarv, line 230: You are changing the initial orbit `orbit`, but it is not used. Probably it should be introduced in `twiin`?

swhite2401:
orbit is the pointer to twiin['closed_orbit'] so I think it is fine.
Am I missing something?

lfarv:
OK, you are right, I missed it…
But to avoid confusion between `orbit` in `twiin` and the `orbit` argument, I would suggest taking `orbit` out of twiin and using the existing argument for both cases:
- for a ring: if orbit is none, orbit=findorbit4
- for a transfer line, if orbit is None, orbit is zero

Then for `get_chrom`, no need to increment orbit and to multiply the tests, it could be:
```
        if twiin is not None:
            dorbit = numpy.append(disp0*0.5*ddp, 0.5 * ddp)
            kwup = dict(orbit=orbit+dorbit, twiin=twiin)
            kwdown = dict(orbit=orbit-dorbit, twiin=twiin)
        else:
            kwup = {}
            kwdown = {}

        d0_up, tune_up, _, l_up = linopt(ring, dp=dp + 0.5, refpts=uintrefs,
                                         keep_lattice=True,
                                         coupled=coupled, **kwup)
        d0_down, tune_down, _, l_down = linopt(ring, dp - 0.5, refpts=uintrefs,
                                               keep_lattice=True,
                                               coupled=coupled, **kwdown)

        chrom = (tune_up - tune_down) / ddp
        dispersion = (l_up['closed_orbit'] -
                      l_down['closed_orbit'])[:, :4] / ddp
        disp0 = (d0_up['closed_orbit'] - d0_down['closed_orbit'])[:4] / ddp
```
I also have a doubt: `orbit` and `dorbit` should have dimension (6,). The append statement adds a fifth value to a (4,) dispersion. Is there a zero missing?

Finally, the same variable `disp0` is used 1st as the initial dispersion in `twiin` and then for the computed value at the beginning of the structure. One could avoid the first definition by defining the orbit displacement `dorbit` directly in the argument checking section, instead of `disp0`:

```
        try:      
            dorbit = numpy.append(0.5*ddp*twiin['dispersion'], 0.5 * ddp)
        except KeyError: 
            print('Dispersion not found in twiin, setting to zero')
            dorbit = numpy.append(numpy.zeros((4,)), 0.5 * ddp)
```

willrogers:
Just a comment that you are free to ignore: there's a tendency to use very compact variable names and I find it makes it a bit more difficult to understand, partly because I am not as familiar with the physics.

e.g. is `twissin` or `twiss_in` not more understandable than `twiin`?

swhite2401:
Thanks Laurent, I have implemented your suggestions.
Concerning orbit the answer is yes, this is why only `orbit[1:5]` was incremented in the previous version.

We could possibly remove additional tests by requesting the type `lindata` for twiin but thhis removes some flexibility

swhite2401:
I agree `twiss_in` is better I'll change this now

lfarv:
To get the tests passed, you need to remove the test on `get_twiss` in `test/test_physics.py`. But there is also a problem in test_linopt_uncoupled, where "gamma" for an uncoupled line should be 1 all along the line and not nan. But it looks really useless for an uncoupled line, so probably removing the test is OK.

swhite2401:
Yes I was waiting for @willrogers  to give his green light to remove `get_twiss()` to start modifying `test_physics.py`

willrogers:
I should do this testing but I can't guarantee I can do it promptly enough not to hold you up. So I suggest you delete `get_twiss()` and I can always put it back manually.

swhite2401:
Ok then let me put back `get_twiss()` so that the checks pass and you have both functions to do your test.
Once you are satisfied we can decide to remote it or not.

willrogers:
See comment on #171.

swhite2401:
Ok I will remove get_twiss() as agreed and then this part is probably ready for merging

lfarv (approved):

willrogers:
OK for me.

willrogers:
I can resolve this conflict if you like.

swhite2401:
I was still working on that one, I restore it and do the merge when ready

PR #174, opened 2020-08-19
v0 of matching functions

swhite2401:
Hello @lcarver, @lfarv @simoneliuzzo @carmignani 
I have started working on matching functions for pyAT.

You will find attached a example matching job to adjust the phase advance of the ESRF standard cell
while keep other parameters constant.
The extension has to be set back to `.py` in order to run it.

[match_ebs_std.txt](https://github.com/atcollab/at/files/5096044/match_ebs_std.txt)

This is a -1 version comments are most welcome!


lfarv:
Hello @swhite2401, and happy new year!

2021: I finally looked at this "matching' pull request!

The structure looks to me really simple and elegant. So if I have some ideas to make the implementation simpler, I'll make only a few comments on the structure:

### Variables: ###
nothing to say about the two types of variables: they should cover most cases and it is possible to add new ones if necessary (coupled variables?). I have proposals for renaming the 2 classes;

- a "RingVariable" deals with the values of attributes of one or several elements, so I would propose "ElementVariable".
- a "FunVariable" acts globally on the ring (see the `vfun(val, ring, *args)` function), so for me this would be the "RingVariable" class.

### Constraints: ###
This is less clear for me, and your example uses a `get_obs` function which is very tricky. I would prefer building a sequence of simple constraints representing each a single line of your `get_obs` function… 

- A "RingConstraint" takes its values from `at.linopt`, and it is associated with a given `refpts` (a single refpoint, I guess?). A more explicit name could be "LinoptConstraint". It can support several "targets". Rather than introducing these targets in the constructor, they could be added with a `add_target(name, value, bounds, weight,...)` method. A sequence of such constraints could replace your big evaluation function, if we can organise them so that a single call to `at.linopt` is done. The idea being: one "LinoptConstraint" per refpoint.
- A "Constraint" applies more generally to the whole ring through a user-supplied function. I am not sure that it needs several "targets", but why not. Maybe I did not find good examples for that one…
- Finally, I did not understand the use of "MultiConstraints", which looks just like a sequence of constraints. We could feed match with a sequence of constraints like it is done for variables. But again, maybe I missed the purpose…

If you don't mind, I propose to work on optimising the implementation, but in the meantime, could you give some examples of building the various types of variables and constraints, to help checking that we cover most use cases?

swhite2401:
Hello @lfarv and happy new year!

I am very happy that you had time to look at this and that you like it. Please make proposals, I have very little time to look at these things these days.

To answer you:

**Variables:**

For me RingVariable was something already defined in AT, while FunVariable could be an arbitrary function, it could be a global or acting on a family but also on very few elements (for instance changing the DL gradient, or move an element and change its field with a fixed relation)
Changing to ElementVariable is a good proposal, the other could be confusing but I have no better proposal

**Constraints**

I started with what you propose but then this was my way how sending a single call to linopt (and yes a single reftps per constraint). Multiconstraint I think was there to combine different constraint types I believe, but honestly I don't remember and I agree it does not seem very useful anymore.

RingConstraint: I agree LinoptConstraint sound like a better name, if you find a way to have one constraint object per refts with a single call to linopt (or a single evaluate function) that would be great

Constraint: it was indeed ment to be much more general, it could apply to the full ring (tune) or single or multiple refpts, for instance match the w function, RDTs or anything not defined in linopt (although I started implementing the w-function in some branch. So Again RingConstraint may not be the best but I do not have a better proposal
Multiple target again was ment to minimize the calls to linopt or other function used to get the observable, the idea is that the user can define a single function to compute all observables that it used in evaluate.

I tried to optimize the speed an minimize function calls using these Constraint builder but this might not be the best, in case you have other proposals I would be very interested to see them.

I will come up with some of the examples I mentioned in this message to illustrate what I just said.

swhite2401:
@lfarv , to complement. MutliConstraint was necessary to concatenate several constraints, since as you mention each is constructed using multiple targets this function was necessary to merge several such objects in a single matching job with a single value to minize.

If we move to a structure where 1 Constraint -> 1 target then the merging will occur in the match function I suppose so MultiConstraint will not be required anymore, or will be hidden from the user

swhite2401:
@lfarv, here some examples of usage

`import at
import os
import numpy as np
import time
from matching import FunVariable, RingVariable, RingConstraint, Constraints


path = '../../at/pyat/test_matlab/'
filename = 'hmba.mat'
key = 'RING'
latticef = path+filename
ring = at.load.matfile.load_mat(latticef,key=key)
ring = at.Lattice(ring)


#Define some lattice elements that need to be moved, with some properties used by the matching

qf1 = at.get_refpts(ring,'QF1*')
qd2 = at.get_refpts(ring,'QD2*')
qd3 = at.get_refpts(ring,'QD3*')
qf4 = at.get_refpts(ring,'QF4*')
qd5 = at.get_refpts(ring,'QD5*')
qf6 = at.get_refpts(ring,'QF6*')
qf8 = at.get_refpts(ring,'QF8*')


refpts= [qf1, qd2, qd3, qf4, qd5, qf6, qf8]
names = ['qf1', 'qd2', 'qd3', 'qf4', 'qd5', 'qf6', 'qf8']
atname = 'PolynomB'
order = 1
bounds = [[0,5],[-5,0],[-5,0],[0,5],[-5,0],[0,5],[0,5]]


#Example of RingVariables, it is possible to change the attribute of one or several elements (tune family)
variables=[]
for i,r in enumerate(refpts):
    var = RingVariable(ring,r,atname,varname=names[i],order=order,bounds=bounds[i])
    variables.append(var)


#Example of FunVariable, the user is responsible to define the get/set function

#In this example sets PolynomB[1], can be done with the simpler RingVariable
def set_fun(val,ring,refpts):
    at.set_value_refpts(ring, refpts, 'PolynomB', val,order=1, increment=False)

def get_fun(ring,refpts):
    return at.get_value_refpts(ring, refpts, 'PolynomB',order=1)

#Or use an other function not handled by ring variable (real usage)
#Warning these functions do not exist!
def set_fun(val,ring,refpts):
    at.set_shift(ring, refpts,val)

def get_fun(ring,refpts):
    return at.getshift(ring, refpts)

variables=[]
for i,r in enumerate(refpts):
    var = FunVariable(ring,set_fun,get_fun,varname=names[i],args=(r,),bounds=bounds[i])
    variables.append(var)


#RingConstraints uses at.linopt and can constrain any of its output
#this function was not fully implemented and is missing some features like the weight
#it is not clear is linopt outputs as array are handled properly
#this part requires further developments, no example given


#Example of Constraints, the observable is defined by the user
def get_obs(ring,sf):
    refpts = sf
    l0,tune,chrom,ld=at.linopt(ring,dp=0,refpts=refpts,get_chrom=True,
                               keep_lattice=True, coupled=False)
    musf0 = np.squeeze(ld[ld.idx==sf[0]].mu)
    musf1 = np.squeeze(ld[ld.idx==sf[1]].mu)
    dmusf = np.squeeze(musf1 - musf0)/(2*np.pi)
    dmusf = dmusf-np.floor(dmusf)
    return dmusf

names = np.array(['dmuxsf','dmuysf'])
weights = np.array([1, 1])
sf = at.get_refpts(ring,'SF*')
sf = [sf[1],sf[3]]
init = get_obs(ring,sf)
   
constraints = []
for name,ini,w in zip(names,init,weights):
    constraints.append(Constraints.build(name,ini,weight=w))
constraint = Constraints(ring,get_obs,constraints,args=(sf,))


#Example on the use of MultiConstraints, here additional constraint on the initial beta
def get_betxy0(ring):
    l0,_,_,_=at.linopt(ring,dp=0)
    return l0.beta

bx = Constraints.build('bx',7.0)
by = Constraints.build('by',4.0)
constraints=[bx,by]
constraintb = Constraints(ring,get_betxy0,constraints)

constraintqb = MultiConstraints([constraint,constraintb])`

lfarv:
I committed a version with simpler `Variable` classes. The main differences are:

- renamed `FunVariable` to `Variable` (generic variable) and `RingVariable` to `ElementVariable`
- removed the `ring` argument: not needed
- the methods are renamed to `Variable.set` and `Variable.get`
- the order of arguments of the `vfun` (now called `setfun`) user-defined function is now `vfun(ring, val, *args)` instead of `vfun(val, ring, *args)` for analogy with `setattr` and `Variable.set`

I slightly modified your example (without fully understanding it) and it still runs. Here it is: [match_ebs_std.txt](https://github.com/atcollab/at/files/5788481/match_ebs_std.txt).

@swhite2401, tell me if we can go on!

lfarv:
I still think that we don't need the "MultipleConstraints": we can give to `match` a sequence of constraints: they can be either single-target and evaluate to a scalar, or multi-target and evaluate to an array. But in all cases we can concatenate all their evaluations to build the global evaluation function for `least_squares`.

Now for optimising the computations, I see two possibilities:

1. The simplest one: we define a single constraint per "computation": for instance a `LinoptContraints` (calls `at.linopt`), an `EnvelopeConstraints` (calls `at.ohmi_envelope`) and a generic `Constraints` (the existing one, where the user is responsible for building the evaluation). The LinoptContraints and EnvelopeConstraints have several targets, but in addition to the present ones, each target needs its `refpts` input. And we can send to `match` any sequence of those.

2. A more friendly one where we split `LinoptContraints` and `EnvelopeConstraints` with one constraint per refpts, still multi-target since we may ask for several variables at the same point. The call to `at.linopt` will be done outside the class object (in `match -> fun`). But then we need to modify the prototype of the evaluate method so that if gets the `lindata` at its own refpts.

swhite2401:
Hello @lfarv, this all sounds very good. I agree solution number 2 is the nicest and that the user's effort should be minimized.
Most cases will indeed be covered by linopt.

I can think of few necessary options:
-match transfer line
-turn coupling on/off to optimize speed
-in case atx is called do not call linopt but use the output of atx

Finally Getting rid of MultiConstraints would make it easier, but then match has to check the instance of the object to decide which evaluation method to use and gather all refpts in order to send a single call to atx or linopt per iteration.
I would keep the general Constraint as is, it is the responsability of the user to optimize it properly and can be consisdered an "expert" method

swhite2401:
Hello @lfarv, I can see you have progressed a lot!
Is it still under development, or can I run a few tests?

lfarv:
Hello @swhite2401: the interface should now be stable.
This answers most of you remarks: 
- all `at.linopt` parameters can be specified when initialising `LinoptConstraints`: `dp`, `coupled`, `twiss_in`, `orbit`…
- only useful refpoints are computed,
- off-momentum is computed only if the dispersion or chromaticity appears in the targets
- a single target may be associated with several refpoints
- it is possible to give a user-defined evaluation function (for example computing a phase advance between 2 points: see `help(LinoptConstraints)`)

I'll now work on:
- adapting your test the new interface,
- since `atx` combines results from `linopt` and `ohmi_envelope` which are independent (apart from orbit computation), I propose to split into two independent constraints: `LinoptConstraints` and `EnvelopeConstraints`,
- adding an `Orbit4Constraints`: though the orbit appears in linopt results, this will be much faster when dealing only with orbit parameters (steering a transfer line, computing bumps).

But now feel free to try it and let me know your findings!

swhite2401:
Great! Thanks for your contributions!

I will test after the restart most likely, however I already have a small comment, when mu appears in the constraint you ask to compute linopt for the full lattice.
This significantly slows down the matching and is not absolutely necessary as most of the time only the fractional part of the phase matters, maybe this could be left as an option 'UseInteger' for instance?

lfarv:
> when mu appears in the constraint you ask to compute linopt for the full lattice.
> This significantly slows down the matching and is not absolutely necessary as most of the time only the fractional part of the phase matters, maybe this could be left as an option 'UseInteger' for instance?

Yes! I had already noticed and commented…

lfarv:
The last version has small improvements (doc strings and outputs), and I have a few question for the next steps:
- I don't like the keyword "`order`", I would prefer "`index`". And there are limitations: what about indexing in a n-D array, like `m44` or defining slices? I'll look for a more powerful way.
- We could have match optionally varying a copy of the lattice instead of working in-place, like what is done for `Lattice.radiation_on`

Finally, here is the updated test file: [match_ebs_std.txt](https://github.com/atcollab/at/files/5802388/match_ebs_std.txt), looking significantly simpler.

lfarv:
Indexing works well: one can set "order" to a tuple for n-D arrays, and use slice objects and ndarrays for indexing

swhite2401:
Order was inherited from another piece of code I wrote, I agree index is much better and you solution is nice
Optionally working on a copy could indeed be useful in some cases

I have question: why did you separate Orbit4Constraints in separate module? You intend to write a separate match function?

lfarv:
@swhite2401 : The closed orbit is indeed part of linopt data, and is handled in `LinoptConstraints`.
But for problems which do not involve parameters other than orbit, `Orbit4Constraints` is much faster (`find_orbit4` is maybe 5x faster that `linopt`). So the only reason is the efficiency for steering problems or bump computation.

Keeping the idea of associating a "contraints" class to each time-consuming computation, so that all targets are evaluated in one shot, the next class would deal with the results of `ohmi_envelope`: local values(`r44`, `r66`, `emitXY`…) or global ones (6D-`tunes`, `mode_emittances`, `damping_times`). An `EnvelopeConstraint` is as easy to implement as `Orbit4Constraints` , except for 1 point:
- either we switch radiation on/off at the beginning of the `compute` method (so for each iteration),
- or we handle in parallel the two ring, and apply the variations simultaneously to both (it's probably faster), but then we must attribute in `match` the right ring to each container.

I am looking at both solutions, you ideas are welcome! If you think of more usual parameters which are not covered at the moment, tell me.

I'll rename "order" to "index", but otherwise I am happy with the current interface, so feel free to try it and give me your comments.

Concerning the copy/no copy option, I think that a signature like:
`newring = match(ring,...)` could be the best solution: a shallow copy at the beginning of `match` is cheap, and if you don't want to keep the previous ring, you just write `ring = match(ring,...`). No flag!

swhite2401:
Sure I understand the motivations. I was just wondering why you started a new module (file) instead of keeping everything in the same file.
It could simplify the imports to have everything in the same place.

Ok for the shallow copy, I have no strong opinion on that point.

I'll think about the radiation issue, I do not have a better idea than what you proposed for now with a preference for the 2nd option.

lfarv:
Sorry @swhite2401, I misunderstood your comment!
Concerning imports it was "decided" earlier that the internal organisation of packages was "private": meaning that within packages ("matching" in our case), modules could be split, grouped or reorganised at any time. All imports must refer to packages, and not inner modules (see how imports are declared in modules).
In this particular case, I wanted to check that the base classes are self-consistent so that writing new "*Constraints" classes was easy: importing the base class is enough. That's why I turned the "_access" utilities into static methods rather than private functions.
Now that's checked, and I can easily move back `Orbit4Constraints` into matching.

swhite2401:
Hello @lfarv, I have looked at your last iteration and it seems very nice, I think it is could be merged soon.
What do you think? Do you have anything else on your list?

lfarv:
For me, this is now ready for merging. The thresholds for the convergence of the tests are really platform-dependent and took some trials to be set, and finally only the obsolete "AppVeyor" test is failing, which is normal.

Reviewers, any comments?

swhite2401:
As I said this looks very good, I have done a few tests and everything worked as expected.
You can merge, thanks a lot for your help!

lfarv (approved):

PR #177, opened 2020-08-26
Atbeam

lcarver:
Added a file called particles.py into the tracking folder. This contains two functions, one to compute the sigma_matrix and one to generate particles according to this matrix (following atsigma and atbeam). 

Simple test below:
```

import numpy
import at

sigM  = at.sigma_matrix([1,1,1], twiss=True) #lattice can be provided here with twiss=False  
parts = at.beam(10000, sigM)

plt.hist(parts[0,:])
plt.show()
```

### DIFF
--- a/pyat/at/tracking/particles.py
+++ b/pyat/at/tracking/particles.py
@@ -106,17 +106,17 @@ def beam(np, sigma, orbit=None):
     v = numpy.random.normal(size=(sigma.shape[0], np))
 
     try:
-        lmat = numpy.linalg.cholesky(sigma).T
+        lmat = numpy.linalg.cholesky(sigma)
 #        needs the .T to be the same as matlab
 
     except numpy.linalg.LinAlgError:
         row_idx = numpy.array([0, 1, 4, 5])
-        a1 = numpy.linalg.cholesky(sigma[row_idx[:, None], row_idx]).T
+        a1 = numpy.linalg.cholesky(sigma[row_idx[:, None], row_idx])
         a = numpy.bmat([[a1, numpy.zeros((4, 2))], [numpy.zeros((2, 6))]])
         row_idx = numpy.array([0, 1, 4, 5, 2, 3])
         lmat = a[row_idx[:, None], row_idx]
 
-    particle_dist = numpy.squeeze(numpy.asarray(numpy.dot(lmat.T, v)))
+    particle_dist = numpy.squeeze(numpy.asarray(numpy.dot(lmat, v)))
 
     if orbit is not None:
         if (not isinstance(orbit, (numpy.ndarray, list)) or
--- a/pyat/at/tracking/particles.py
+++ b/pyat/at/tracking/particles.py
@@ -3,11 +3,12 @@ Functions relating to particle generation
 """
 import numpy
 from at.physics import ohmi_envelope
+from at.lattice.lattice_object import Lattice
 
 __all__ = ['beam', 'sigma_matrix']
 
 
-def sigma_matrix(argin, twiss=False):
+def sigma_matrix(argin):
     """
     Calculate the correlation matrix to be used for particle generation
 
@@ -34,7 +35,7 @@ def sigma_matrix(argin, twiss=False):
         sigma_matrix    correlation matrix (either 2x2, 4x4 or 6x6)
 
     """
-    if not twiss:
+    if isinstance(argin, Lattice):
         argin = argin.radiation_on(copy=True)
 
         emit0, beamdata, emit = ohmi_envelope(argin, refpts=[0])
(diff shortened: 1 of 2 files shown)


pyat/at/tracking/particles.py
```diff
@@ -0,0 +1,131 @@
...
+
+    OUTPUT
+        sigma_matrix    correlation matrix (either 2x2, 4x4 or 6x6)
+
+    """
+    if not twiss:
+        cflag = False
+        if not argin.radiation:
+            argin.radiation_on()
+            cflag = True
```
lfarv, line 41: Rather that modifying in-place and reverting later, you could use:
`argin=argin.radiation_on(copy=True)`
to work on a local copy. No need to turn radiation off later

pyat/at/tracking/particles.py
```diff
@@ -0,0 +1,131 @@
...
+
+        emit0, beamdata, emit = ohmi_envelope(argin, refpts=[0])
+        sig_matrix = emit.r66[0]
+
+        if cflag:
+            argin.radiation_off()
+
+
+    else:
+        if len(argin) == 2:
```
lfarv, line 51: If you rely upon the number of arguments to guess what inputs are, why not checking 
len(argin)==1
to guess that it is a Lattice object ?
Then, no need for the `twiss` flag

pyat/at/tracking/particles.py
```diff
@@ -0,0 +1,131 @@
...
+        elif len(argin) == 3:
+            [bx, ax, epsx] = argin
+            sig_matrix = epsx * numpy.array([
+                                    [bx, -ax],
+                                    [-ax, (1 + ax * ax)/bx]
+                                            ])
+
+        elif len(argin) == 6:
+            [bx, ax, epsx, by, ay, epsy] = argin
+            sig_matrix = numpy.bmat([
```
lfarv, line 67: Rather than `numpy.bmat` which return a matrix object, `numpy.block` returns a standard nd-array, like emit.r66 in the first case.

pyat/at/tracking/particles.py
```diff
@@ -0,0 +1,131 @@
...
+
+    OUTPUT
+        particle_dist   a matrix of shape (M, np) where M is shape of
+                        sigma matrix
+    """
+    v = numpy.random.normal(size=(sigma.shape[0], np))
+
+    try:
+        lmat = numpy.linalg.cholesky(sigma).T
+#        needs the .T to be the same as matlab
```
lfarv, line 110: Since later you use
`numpy.dot(lmat.T, v)`
you can avoid transposing twice!

pyat/at/tracking/particles.py
```diff
@@ -0,0 +1,131 @@
...
+
+    except numpy.linalg.LinAlgError:
+        row_idx = numpy.array([0, 1, 4, 5])
+        a1 = numpy.linalg.cholesky(sigma[row_idx[:, None], row_idx]).T
+        a = numpy.bmat([[a1, numpy.zeros((4, 2))], [numpy.zeros((2, 6))]])
+        row_idx = numpy.array([0, 1, 4, 5, 2, 3])
+        lmat = a[row_idx[:, None], row_idx]
+
+    particle_dist = numpy.squeeze(numpy.asarray(numpy.dot(lmat.T, v)))
```
lfarv, line 119: Same as above concerning `bmat`: if you use `block`, maybe you do not need the `asarray` call (to be checked)

lfarv:
@lcarver: at the top of the page, it says that you want to merge into `non-linear`. Should not it be into `master` ?

pyat/at/tracking/particles.py
lcarver: If argin is a lattice object then len(argin) = nelems not 1. So I resorted to a flag. Is there a robust way to check if an argument is a lattice object?

pyat/at/tracking/particles.py
lcarver: you're right. Fixed.

pyat/at/tracking/particles.py
lcarver: I think numpy.block was the function I was looking for. I had some problems with exactly this function that was returning a matrix object. numpy.block works in the same way but return a nd-array so it's perfect.

pyat/at/tracking/particles.py
lcarver: As above. Changing to numpy.block means I no longer need the asarray.

pyat/at/tracking/particles.py
lcarver: Agreed. Much better solution.

lcarver:
This branch was originally cloned from non-linear (which was a mistake by me). As we plan to merge non-linear into master, I thought it best to merge this branch into non-linear first then we merge them all together.

pyat/at/tracking/particles.py
lfarv: Ah ! I understood (wrongly) that you were expecting a variable number of arguments with the syntax `def sigma_matrix(*argin)`. Now that I see that you want a single argument which is either a Lattice object or a tuple, you could test `isinstance(argin, Lattice)`

lcarver: `isinstance(argin, at.Lattice)` returns True for a Lattice object. But it also returns positive for `isinstance(argin, list)`.

lfarv: Normal, since a Lattice is a list (but not the opposite). That's why you need to test for a Lattice, so that any other list of values like [energy_spread, bunch_length] returns False.

lcarver: ah yes, you are right. I was thinking of the reverse which is not needed. A list will return False for a lattice object check. Thanks a lot. Sorry for the confusion

lcarver:
All comments by @lfarv have been implemented.

Issue #168, opened 2020-07-24
Calculation of synchrotron radiation integrals not consistent


TeresiaOlsson:
The calculation of synchrotron radiation integrals doesn't seem to be done consistently in all functions resulting in different output depending on which function you use. Especially output like energy loss etc is no longer correct when including IDs using the GWigSymplecticRadPass function.

I suggest to implement a function that calculates the radiation integrals correctly and then call that function from all other functions that use the radiation integrals.

carmignani:
I think ringpara computes correctly the radiation integrals, at least without wigglers. But I agree, we can write a function just for the radiation integrals and then use it everywhere they have to be computed.
What functions did you use for the radiation integrals?

T-Nicholls:
Hi @TeresiaOlsson,
Have you taken a look at #121 yet, it might address some of the problems you mention? It incorporates some work I did last year to try and get the updated (and hopefully numerically correct) `GWigSymplecticPass` and `GWigSymplecticRadPass` methods working with AT2.0 and pyAT. The last I heard the changes had been passed to the Diamond AP group for testing, but I am unsure if they concluded whether the functions were accurate or not; @willrogers may have a better idea of their status.

TeresiaOlsson:
Hi,

I'm in the Diamond AP group and we are testing it now. So far it seems like the passmethods are working correctly, but other functions need updating to output correct numbers when including them in the lattice.

We tried atsummary, atx and ringpara and none seem to include the effect of the wigglers on the global parameters. The cause seems to be that the calculation of the radiation integrals doesn't include it. There are also other functions like atsetRFCavity, atenergy etc that needs updating to work correctly with wigglers.

Then we will take a look at it and propose changes.

lnadolski:
dear @TeresiaOlsson 

Can you send us an full example for helping us to debug ?

Laurent.

T-Nicholls:
I am glad to hear that it is being worked on and that the pass methods are correct.
I definitely agree that all the functions should be consistent and centralising the calculation of the radiation integrals would be sensible.
I am not surprised that `atsummary`, `atx`, and the other functions do not provide the correct numbers as they were not updated for wigglers. I am, however, concerned that `ringpara` doesn't function correctly, as it was updated to use `RadIntegrals.m` to calculate the added effects of the wigglers on the radiation integrals.

TeresiaOlsson:
[...]

PR #167, opened 2020-07-23
Bug correction in atcavityon and atcavityoff

carmignani:
There was a bug in the functions atcavityon and atcavityoff reported by TeresiaOlsson in #161.

PR #178, opened 2020-08-28
Centralize the computation of radiation integrals

lfarv:
Following the discussion in #168, this branches groups the computation of radiation integrals in Matlab in 2 new functions:
DipoleRadiation and WigglerRadiation. Now the results of:

- `ringpara` and `atsummary` in Matlab
- `radiation_parameters` in pyAT

all agree, at least without wigglers.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,7 +35,7 @@ build_script:
       if ($env:APPVEYOR_REPO_TAG_NAME -eq $null) {
         $env:APPVEYOR_REPO_TAG_NAME = "None"
       }
-      if ($env:APPVEYOR_REPO_TAG_NAME.StartsWith("pyat") -And ($env:PYTHON -eq 36)) {
+      if ($env:APPVEYOR_REPO_TAG_NAME.StartsWith("pyat") -And ($env:PYTHON -eq 37)) {
         rm "build" -r -fo
         python -m pip install --upgrade cibuildwheel
         & cmd /c 'cibuildwheel 2>&1' --output-dir wheelhouse
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -18,6 +18,10 @@ environment:
 
 install:
   - set "PATH=C:\Python%PYTHON%-x64;C:\Python%PYTHON%-x64\Scripts;%PATH%"
+  # This reverts a change made in setuptools v50.0.0, to continue to use
+  # distutils instead of setuptools for building extensions.
+  # Unfortunately I don't know what the correct fix is.
+  - set "SETUPTOOLS_USE_DISTUTILS=stdlib"
   - python -m virtualenv venv
   - venv\Scripts\activate
   - cd pyat
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,8 @@ before_script:
 
 script: |
   PYTHON_VERSION=$(python -V 2>&1 | cut -d ' ' -f 2 | cut -d '.' -f 1,2)
-  if [[ $TRAVIS_TAG == pyat* ]] && [[ $PYTHON_VERSION == 3.6 ]]; then
+  if [[ $TRAVIS_TAG == pyat* ]] && [[ $PYTHON_VERSION == 3.7 ]]; then
+    # Use cibuildwheel to build all wheels for the current platform.
     rm -rf build
     python -m pip install --upgrade cibuildwheel
     cibuildwheel --output-dir wheelhouse
--- a/atintegrators/DriftPass.m
+++ b/atintegrators/DriftPass.m
@@ -1,5 +1,5 @@
-% DriftPass.m Help file for DriftPass.c
-%  DriftPass.c - Integrator for Drift spaces
+function varargout=DriftPass(varargin)
+%  DriftPass - Integrator for Drift spaces
 % 
 %  The transformation performed is:
 % 
@@ -60,4 +60,8 @@
 %   >> ELEMENT=atdrift('DRIFT',L);
 %   >> outcoord=DriftPass(ELEMENT,[0 0 0 0 0 0]')
 %
-%see also: atdrift, ringpass, linepass, DriftPass.c  
+%see also: atdrift, ringpass, linepass, DriftPass.c
+
+error('AT:MissingMexFile','"%s.%s" is missing. Did you run "atmexall" ?',...
+    mfilename,mexext);
+end
--- a/atintegrators/GWigSymplecticPass.c
+++ b/atintegrators/GWigSymplecticPass.c
@@ -120,6 +120,8 @@ void GWigSymplecticPass(double *r, double Energy, double Ltot, double Lw,
     int c;
     double *r6;
     struct gwig Wig;
+    /* Energy is defined in the lattice in eV but GeV is used by the gwig code. */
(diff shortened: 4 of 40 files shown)


lfarv:
The contribution of wigglers done in `RadIntegrals.m` looks wrong to me for several reasons:

- it crashes for wigglers with more than 1 harmonic,
- The dispersion(eta, eta') is numerically integrated in the wiggler field with `ode15s`, but the initial value is set to (0, 0) at the entrance of the wiggler instead of the value of the ring dispersion,
- The differential equation for eta' uses B [T] instead of B/Bro [1/m]: apparently the normalisation with energy is missing,
- The computation of I5 is an obscure formula involving the wiggler parameters and the initial alpha and beta functions, but NOT the dispersion ???

So unless I miss something, the results are wrong.
For now, **the wiggler contribution is removed** !

lfarv:
Looking at `RadIntegrals.m` (which I would rename `WigglerRadiation.m`), it misses references but the principle looks correct, at least for planar wigglers. However the implementation has bugs.

1. Has there been any check of the contribution of wigglers to the radiation integrals?
2. To work on it, we need get a description of the magnetic field model corresponding to the input parameters of the wiggler. Anyway this is also needed for users, so that they can describe their devices in AT. @willrogers, do you have more information on that?

willrogers:
Maybe @TeresiaOlsson?

lfarv:
Sorry for the reviewers: following the merge of #167 into the master, and probably some wrong manipulation from me, the list of modified files of much longer than necessary. In fact, differences from the present master branch are in only 7 files:

- `atsummary.m`
- `ringpara.m`
- `atenergy.m`
- `wiggler.m`
- `DipoleRadiation.m` (new)
- `radiation.py`
- `ring_parameters.py`

For Github experts: is it possible to reset this list?

lfarv:
The field model corresponding to the `GWigSymplecticPass` is given in section VI of [this paper](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.68.046502): _"A general wiggler integrator for beam dynamic studies"_. I'll document the input parameters following that…

lfarv:
Given the field model, it appears that:

- `atwiggler` in Matlab misses the `Energy` parameter,
- the set By=[1, 1, 0, 1, 1, 0], Bx=[] represents a simple planar infinitely wide horizontal wiggler with a single field harmonic (the fundamental), which should cover the basic usage,
- these 2 parameters could bet set as default values and provided as keyword parameters, which simplifies the declaration of a wiggler in most cases,
- We are left with 4 mandatory  positional parameters: Length, Period_length, Bmax, Energy. And 4 keywords: n_step, n_meth, By, Bx.

Together with the documentation of the pass method, this is a topic for another pull request.

Concerning the computation of radiation integrals, I still think that the present computation is wrong and should be removed. This branch is then ready for merging, **without the contribution wigglers to the integrals**.

lfarv:
Problems with Appveyor builds… @willrogers, can you help?

willrogers:
Perhaps @mashl could give some opinion on `RadIntegrals.m`, it seems like it came from him originally?

I'm not sure what's happening in the Appveyor failures, I'll have a look this afternoon hopefully.

mashl:
dear @lfarv 
I checked the RadIntegrals.m for various types of insertion devices including linear and elliptical IDs. 
It has some problems with the cases in which the ID located in the straight sections with dispersion. in the last modification of this code I fixed these bugs and I'm can share the last version of RadIntegrals.m if you want.

lfarv:
Dear @mashl , nice to hear from you ! Indeed, the question of dispersion is one point I noted, since you start the integration from (eta, eta')=(0,0) instead of the machine dispersion. But I have 2 other questions for you:

1. In the differential equation, I see `dy(2) = -(By(x).*By(x)./100)*y(1)+By(x)/10;` which I interpret as
`eta''=-By^2*eta + By`. B is in Tesla, and I do not see the normalisation with energy, it should be B/(B.rho) in m^-1.
 
2. For a planar undulator, I see 
```
I5=(((Lw^4)/(4*(pi^4)*(rhw^5)))*((3/5/pi)+(3/16))*gammax*LT)-...
     (((9*(Lw^3))/(40*(pi^4)*(rhw^5)))*alphax*LT)+...
     ((Lw^2)*betax*LT/(15*(pi^3)*(rhw^5)));
 ```
Do you have explanations for this formula ? I am surprised by the fact that for the chromatic invariant, the dispersion does not appear anywhere! For the helical case, the formula for H is correct, but `betax`, `alphax`, `gammax` are the initial values, while they should be the current ones.

And of course we are interested by an updated version of your code,
Thanks for your help

willrogers:
I have pushed a commit that should fix the Appveyor build.

mashl:
@lfarv It is my pleasure to have a contribution on this modification. 
about your questions, the analytic formula for computing the dispersion is not working for the real lattice with errors, therefore, in the new version, I used the particle tracking method to compute the dispersion function through the IDs. For this propose, I made some changes in the pass method to get the trajectory of the off-momentum particles through the ID. 
having the exact form of eta and eta' helps me to compute all integrals explicitly and I don't use the mentioned formula in the new version of the code. 
For the I_5 integral, I saw the mentioned formula in an article. however, at the moment I can't remember the reference but I can find it if it's necessary. 
I made many changes in my AT functions, let me download the last version of the AT and add the new codes to it and share the compatible codes with you.

lfarv:
I think that the integration and validation of the radiation integrals in wigglers will take some time now. So I would like to merge this branch now and open a new pull request for wiggler radiation. This will immediately provide this:

- all output of radiation integrals come from a single computation,
- wrong results for wigglers are removed,
- single entry point for a future contribution of wigglers

Any last comment ?

TeresiaOlsson:
It sounds very promising to me. @lfarv, we have Elegant results for the Diamond-II CDR lattice which we have benchmarked against analytic formulas and Tracy.  Would that help with testing? The lattice has substantial dispersion in some of the straights and when testing it we found a bug in Tracy that was related to the integration of the dispersion so I think it's a good lattice for noticing that effect.

lfarv:
@TeresiaOlsson: that looks perfect for testing. As I said, I would like this branch to be merged, and also #172 which I need, to start from a clean state, and then open a new pull request specifically for wiggler radiation.

mashl:
dear @lfarv 
I prepared the modified version of RadIntegrals and test it for a couple of lattices, I think this version could fix the dispersion bug. I can send them here or start a new pull request

lfarv:
@mashl : I suggest that you wait for the merge of this branch and then open a new pull request. It will avoid mixing different things. There is now in `ringpara` a commented call to `WigglerRadiation(ring,lindata)` which should be the entry to your code. Thanks!

TeresiaOlsson:
I think this merge introduced errors in atsummary and ringpara that wasn't there before. For the Diamond-II lattice atsummary now gives the error:

"Unrecognized function or variable 'E0'.
Error in atsummary (line 34)
elseif isfield(E0,'LatticeFile')"

whereas ringpara gives the wrong momentum compaction. I tried running the example lattices in machine_data, but there was similar problems.

Anyone else that also has these problems?

lfarv:
@TeresiaOlsson: There was indeed a bug for lattice without RingParam element. This has been corrected in #187

PR #181, opened 2020-09-02
Wiggler creation methods

lfarv:
The `GWigSymplecticPass` integrator is now documented in the` GWigSymplecticPass.m` file (for Matlab).
It is not clear to be what is the equivalent for integrator documentation in Python…

Looking at the `GWigSymplecticPass` integrator, it appears that the wiggler creation functions (Python and Matlab) can be made simpler.

For simple planar horizontal wigglers, the mandatory parameters can be reduced to:
Name,  Length,  Period_length, Bmax, Energy

All other parameters can be set with keywords. As for other element creation functions, keyword names are set identical to the attribute names of the created element, so we have:

`PassMethod` (default: 'GWigSimplecticPass')
`Nstep` (default: 5)
`Nmeth` (default: 4)
`By` (default: [1, 1, 0, 1, 1, 0])
`Bx` (default: [])

The declaration of wigglers is then simpler (for instance creating an arbitrary horizontal wiggler only needs specifying explicitly `By`). But this new argument list is incompatible with the initial one since the energy parameter is now moved to the 5th location.  However since the wiggler has just been released, I think this is worth the effort.

### DIFF
--- a/atintegrators/DriftPass.m
+++ b/atintegrators/DriftPass.m
@@ -1,5 +1,5 @@
-% DriftPass.m Help file for DriftPass.c
-%  DriftPass.c - Integrator for Drift spaces
+function varargout=DriftPass(varargin)
+%  DriftPass - Integrator for Drift spaces
 % 
 %  The transformation performed is:
 % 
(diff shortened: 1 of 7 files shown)


willrogers:
Hi @lfarv,

I realise that you're clearing up after me a bit here, I appreciate it. These changes look good to me. If you merge `master` into this branch then the Appveyor build should pass.

Cheers,
Will

lfarv:
@willrogers : It took me some time to understand the physics of this new integrator, so I arrived a bit late to intervene in the development… I hope it"s still time to polish a little bit the interface !

lfarv:
@willrogers : I got nothing new from the `origin/master` branch ! so no change to ta AppVeyoir build. Is there something more to do?

willrogers:
Sorry, I fixed the appveyor build on the `radiation_integrals` branch but that one has not yet been merged.

willrogers:
I have added that commit to master, so you could try again.

willrogers:
ok

PR #184, opened 2020-09-04
Twissline

swhite2401:
@lfarv , I regenerated this one because I was still improving linopt and introducing some features that will be needed for the w-function computation by linopt


swhite2401:
@lfarv all tests passed I merge back in to the master

lfarv:
Ok, go on. I was waiting for this, so I got at least a first version which works…

PR #188, opened 2020-09-24
Simplified radiation integrals in wigglers

lfarv:
By neglecting the self-induced dispersion in computation of I4 and I5, radiation integrals in wigglers are computed faster and more accurately


lfarv:
@TeresiaOlsson: I ran your lattice with the various IDs, are is what I get. Can you check this?

[Diamond-II ATnew.xlsx](https://github.com/atcollab/at/files/5274136/Diamond-II.ATnew.xlsx)

lfarv:
I the computation is correct, I'll then go through:

- the update of `atsummary` to have it consistent with `ringpara`
- fixing the missing wiggler contribution in `atenergy` pointed out by @mashl 
- the update the python version accordingly

TeresiaOlsson:
Thank you @lfarv. I will check it. So I understand correctly, is the self-dispersion included in these numbers or not?

lfarv:
@TeresiaOlsson: the self-induced dispersion is neglected in I4 and I5, except as a lower limit for the I5 contribution. But in any case, it's completely negligible compared to other contributions (a few tens of microns…). It's included in I1 (the computation is obvious), though it is also negligible there.

TeresiaOlsson:
There is something odd with K04, but it was there before too so I have to check that. Maybe I have done something wrong in the lattice file for that one.

Otherwise, I think I1 looks really good now, but I'm a bit surprised that the self-dispersion is negligible for I5. Even for the strong wigglers I12 and I15 that seems to be the case when I calculate the effect on the emittance. I get the same emittance contribution using your AT results as I get from Elegant.

TeresiaOlsson:
@lfarv No, sorry. I did something wrong again. The self-dispersion isn't at all negligible for I12 and I15. I get the following emittance contributions:

I12: Elegant: -15.50097252 pm rad, AT: -8.861909069 pm rad
I15: Elegant: -14.26326934 pm rad, AT: -9.257037645 pm rad

lfarv:
@TeresiaOlsson: You are right, but as I told you, I take the effect of the self-induced dispersion as a lower limit for the I5 contribution (in case the lattice dispersion is 0, or less that ~100um). For the wiggler I12, I get for the I5 integral:
- lattice contribution alone:    **1.15E-12** (because of zero dispersion)
- wiggler contribution alone: **1.95E-07** (analytical computation)

So in this case, the final result is only due to the the self-induced dispersion. Now we have 1.88E-07 from Elegant and 1.95E-07 from AT. I cannot add both contributions in AT, but the results should be larger than 1.95E-07, so I do not understand that the Elegant value is lower. It's however a 3.5% difference which I think is acceptable, but cannot explain your difference in emittance. I2 and I3 are similar, so the wiggler description is the same. There must be something else, I'm looking…

For K04, I2 and I3 are different, so probably the wiggler I'm using by running your scripts is not the same that the one used in Elegant.

lfarv:
For I12, I find indeed -15pm in AT, as Elegant does !! I check now for I15

lfarv:
-14pm for I15, it looks correct…

TeresiaOlsson:
Hmm. Did you get this without doing any changes?

lfarv:
Absolutely no change !

TeresiaOlsson:
Interesting, but very nice! Are these latest changes in the wiggler_integrals branch? Then I can test it myself and check what I did. Sounds like something is wrong in my calculation of the emittance contribution from the integrals.

lfarv:
In fact, the reduction is due to the increase of I2 (strong wiggler, both codes agree), while I5 contributes on the other hand to a (small) increase. But It's not so sensitive to I5!

lfarv:
No change in the wiggler_integrals, please try it !

lfarv:
To go on with I12/I15: a wiggler in a zero dispersion straight is a "damping wiggler": its contribution to the excitation (emittance increase, I5) is small, unless the period is very large, while its radiated power (damping, I2) is large therefore reducing the emittance. I think I could completely remove the self-induced contribution without changing a lot the emittance (the reduction would be slightly larger). In a dispersive section, it's completely different.

TeresiaOlsson:
That makes sense. And in the dispersive midstraights the results look good because the lattice dispersion is so large compared to the self-dispersion. Do you think the current implementation could run into problems if we for example try to put I12 in the position of K02? So a strong wiggler in a section with large dispersion?

TeresiaOlsson:
I figured out my error so I also get it to agree now. A bit worse for I15, but still within what I think could be considered acceptable between codes.

I12: Elegant: -15.50097252 pm rad, AT: -15.25016664 pm rad
I15: Elegant: -14.26326934 pm rad, AT: -13.95016051 pm rad

So I would say it works :)

lfarv:
No problem, you can try I12 in K02 location ! As soon as the dispersion if larger than 1 mm, it is dominant by far. The most critical point is long periods and large field, since the self-induced dispersion scales with Bmax^2*Lw^2.

lnadolski:
> I figured out my error so I also get it to agree now. A bit worse for I15, but still within what I think could be considered acceptable between codes.
> 
> I12: Elegant: -15.50097252 pm rad, AT: -15.25016664 pm rad
> I15: Elegant: -14.26326934 pm rad, AT: -13.95016051 pm rad
> 
> So I would say it works :)

Hi Teresia, when for your comparison with ELEGANT, are you sure to slice enough your dipoles in your computation.

TeresiaOlsson:
@lnadolski I think so. Our Elegant lattice is more tested than the AT lattice so if there is a problem with the dipoles I would expect it to rather be in the AT lattice, but for the bare lattice we have good agreement between Elegant and AT.

TeresiaOlsson:
@lfarv Are you planning to do some more changes to this branch before merging it to `master`?

lfarv:
@TeresiaOlsson: Still remains the upgrade of `atenergy` and the python version, plus the possibility of analytical integration of I3 for the case of a single harmonic.

TeresiaOlsson:
@lfarv Then I won't tell my colleagues to start using it just yet. They might be eager.

lfarv:
For me the branch is now ready (except for python). However, we still miss tests for:
- wigglers with kx ~= 0,
- multi-harmonic wigglers,
- vertical wigglers,
- elliptical or helicoidal wigglers.

TeresiaOlsson:
@lfarv I asked my colleagues about it today, but they had no experience of using any of these in Elegant except for elliptical wigglers, but then only with kick maps which doesn't include the radiation effects. So I'm not sure if Elegant can be used for testing these.

lfarv:
I tested as much as I could the vertical, elliptical and helicoidal wigglers.
The python version is now on par with the Matlab version, so for me the branch is ready for merging.

mashl:
@lfarv , I want to thank you for your comprehensive effort on this issue.
currently, I work on adding the wiggler effect to _atx_, it works on the previous version but I faced some problems in updating the codes. I will inform you when I get a reliable result.

lfarv:
Last tuning before merging, if no objection !

lfarv:
The radiation integrals now give two independent ways of computing the damping times. Comparing the results give a clear indication that `GWigSymplecticRadPass` is wrong. This is documented in #190.

@mashl: I guess that you computation of the diffusion matrix is built on top of `GWigSymplecticRadPass`. It is probably affected by this bug.

lfarv:
Wigglers with a PassMethod of 'DriftPass' are ignored in integral computation so that it's easy to 'open' a wiggler simply by setting its PassMethod to 'Driftpass'

PR #191, opened 2020-10-11
Cavity with beam loading

carmignani:
Hello,
I wrote this pass method in 2016 and I updated it a few times. I put it in the repository because other people (@lcarver) are interested on it.
I use the matlab function [mxGetComplexDoubles](https://www.mathworks.com/help/matlab/apiref/mxgetcomplexdoubles.html), calling it with the function atGetComplexArray, to get a complex field (BeamVoltage). It works in matlab, but I have no idea of the similar function in python, so the pass method does not work in pyAT at the moment.
@willrogers or @lfarv maybe can help me to have a version working also in pyAT.


lfarv:
@carmignani : I implemented the access to complex attributes in python.
But I have a few questions on this integrator:
- Can you give a simple example to test the integrator ?
- To allow `atradoff` to turnoff the cavity, it would be better to rename the integrator to `RFBeamLoadingCavityPass` since `atradoff` looks for *CavityPass at the end of the passmethod name to turn the element into `IdentityPass` or `DriftPass`,
- It looks like the passmethod modifies the `VoltageBeam` attribute. This is forbidden in Matlab (a function cannot modify its arguments), and it will probably fail in python. Is this desired, and is there another way to achieve the result ?

carmignani:
Hello Laurent,
I will give a simple example for the integrator.
Ok for the change of name.
It is not really needed to change the VoltageBeam field of the element in the matlab/python memory, but it was convenient to have its value at the end of the simulation. It's a trick to be able to split the simulation and restart it from the exact VoltageBeam needed at that turn.
I did not do many tests with the mxGetComplexDoubles function, because when I did the real simulations I was using the old matlab functions mxGetPr and mxGetPi, but I think this trick works with matlab.
If you think it is better to remove this feature, we can decide to pass the value of VoltageBeam and not the pointer as it is done now.

lfarv:
After many trials, I discovered that though the `complex.h` header is available in Microsoft Visual C,  [the standard arithmetic operators are not defined on complex types](https://docs.microsoft.com/en-us/cpp/c-runtime-library/complex-math-support?view=msvc-160). Unless we rewrite the code to avoid complex numbers or switch to C++, we have to skip this integrator on Windows.

swhite2401:
I close this one, it is integrated in #516

PR #192, opened 2020-11-19
Rebuilt Matlab tests

lfarv:
Matlab tests are much faster by trying to connect to an existing shared Matlab session:
- if there is on the local machine a running Matlab session shared with the identifier "pytest", the connection is immediate and the tests take a few seconds,
- if there is no such Matlab session, pytest will start a new Matlab process as before.

The pre-existing Matlab session must be of the same version as the one started by the Matlab engine for python. `pytest` will not change the Matlab path of an existing session, so the Matlab user is responsible for setting the Matlab path to the one being tested before running the test.

An interactive Matlab session can be shared by typing at the Matlab prompt:

`>> matlab.engine.shareEngine('pytest');
`

All tests have been reorganised and augmented.


willrogers (approved):
This looks fine to me. I think these tests are very useful, it's just a shame that they are so difficult to run.

lfarv:
I agree with you, @willrogers, these tests are very useful for both versions. They are still tricky because of Matlab tuning, but at least now they are fast !

PR #198, opened 2020-11-28
Add octave build commands

matsievskiysv:
Add support for build with octave.
Some functions do not work with octave though. More work is needed. Maybe this could be pulled into separate branch?

### DIFF
--- a/atmat/atmatch/atmatch.m
+++ b/atmat/atmatch/atmatch.m
@@ -131,7 +131,7 @@ dpp = p.Results.dpp;
 usepar = p.Results.UseParallel;
 
 
-options=optimset(minimizer);
+options=optimset(func2str(minimizer)); % NOTE: func2str is needed for Octave compatability
 
 IniVals=atGetVariableValue(Ring,Variables);
 splitvar=@(varvec) reshape(mat2cell(varvec,cellfun(@length,IniVals),1),size(Variables));
--- a/atmat/attrack/linepass.m
+++ b/atmat/attrack/linepass.m
@@ -60,7 +60,7 @@ function [Rout,varargout] = linepass(line,Rin,refpts,varargin)
 %
 % Rfin=LINEPASS(...,PREFUNC)
 % Rfin=LINEPASS(...,PREFUNC,POSTFUNC)
-% Rfin=LINEPASS(...,function_handle.empty,POSTFUNC)
+% Rfin=LINEPASS(...,cell(),POSTFUNC)
 %    PREFUNC and POSTFUNC are function handles, PREFUNC is called
 %    immediately before tracking each element, POSTFUNC is called
 %    immediately after each element. Functions are called as:
@@ -90,7 +90,7 @@ nhist=getoption(struct(args{~funcargs}), 'nhist',1);
 
 newlattice = double(~keeplattice);
 
-[prefunc,postfunc]=parseargs({function_handle.empty,function_handle.empty},...
+[prefunc,postfunc]=parseargs({cell(),cell()},...
     args(funcargs));
 
 try
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,9 @@ Contents.m
 #Local documentation
 doc_html/
 athelp.mat
+
+# C LSP server cache
+.ccls-cache
+
+# octave files
+octave-workspace
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -44,3 +44,22 @@ Installation process:
     You can now use `athelp` to list all main functions.
 
 6. Update html doc - not yet documented.
+
+### Notes on Octave ###
+
+For the most part AT code works with [Octave](https://www.gnu.org/software/octave/).
+At least version 6 is required.
+
+In order to prepare octave environment you'll need to run `atoctave/bootstrap.m` function.
+It will install required Octave modules (Internet connection is needed), compile C sources and add
+paths.
+
+When in `at` folder, you may start an Octave CLI session with command:
+```bash
+> octave --eval 'cd atoctave;bootstrap;cd ..;' --persist
+```
+For GUI session, run command:
+```bash
+> octave --gui --eval 'cd atoctave;bootstrap;cd ..;' --persist
+```
+These commands will run `atoctave/bootstrap.m` for you.
--- a/atintegrators/atcommon.h
+++ b/atintegrators/atcommon.h
@@ -57,7 +57,9 @@
 #ifdef MATLAB_MEX_FILE
 /* Matlab only */
 #include <mex.h>
+#ifndef OCTAVE
 #include <matrix.h>
+#endif
 
 /* Get ready for R2018a C matrix API */
 #ifndef mxGetDoubles
--- a/atintegrators/elempass.h
+++ b/atintegrators/elempass.h
@@ -21,6 +21,8 @@
 #define ExportMode
 #elif defined(MACI64)
 #define ExportMode
+#elif defined(__GNUC__)
+#define ExportMode
 #else 
 /* Default - Windows */
 #define ExportMode __declspec(dllexport)
--- a/atintegrators/gwig.c
+++ b/atintegrators/gwig.c
@@ -23,12 +23,10 @@
  *  Accelerator Physics Group, Duke FEL Lab, www.fel.duke.edu  
  */
 
-#ifndef  GWIG
 #include "gwig.h"
 #include <stdlib.h>
 #include <math.h>
 #include <stdio.h>
-#endif
 
 void GWigGauge(struct gwig *pWig, double *X, int flag);
 void GWigPass_2nd(struct gwig *pWig, double *X);
--- a/atintegrators/gwig.h
+++ b/atintegrators/gwig.h
@@ -13,6 +13,9 @@
  *  Accelerator Physics Group, Duke FEL Lab, www.fel.duke.edu
 */
 
+#ifndef GWIG_H
+#define GWIG_H
+
 #define WHmax 200
 #define GWIG_EPS (1e-6)
 
@@ -88,3 +91,5 @@ struct gwigR{
   double Vkz[WHmax];
   double Vtz[WHmax];
 };
+
+#endif
--- a/atintegrators/mexpassmethod.m
+++ b/atintegrators/mexpassmethod.m
@@ -31,6 +31,10 @@ opt_parallel = false;
 
 PLATFORMOPTION=[varargin{:}];
 
+if isOctave
+  error('mexpassmethod does not work in Octave')
+endif
+
 %Additional platform-specific options for mex
 pdir=fileparts(mfilename('fullpath'));
 if ispc()
--- a/atmat/atdemos/IDModeling/trackWithU35KickMap.m
+++ b/atmat/atdemos/IDModeling/trackWithU35KickMap.m
@@ -2,7 +2,7 @@
 r=esrf;
 
 %create element from kick map and plot kick map
-U35elem=atidtable('U35',1,'U35.mat',6.04,'IDTablePass');
+U35elem=atidtable('U35',1,'U35.mat',6.04,'IdTablePass');
 xtable=U35elem.xtable;
 ytable=U35elem.ytable;
 xkick=U35elem.xkick;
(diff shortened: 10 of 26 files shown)


lfarv:
Hello @matsievskiysv,
I have a few comments on your proposals:
- Concerning the build commands: can you try to avoid makefiles ? These are not portable across architectures, and may need some tuning by the user. Using the `mkoctfile --mex`  or `mex` command within Octave should provide a universal way of building extensions, including on Windows. Is there any problem with that (I am not familiar with Octave)? Also, rather than modifying `atmexall` (already complicated enough), you can provide a dedicated build function for Octave.
- About missing Matlab functions (ex.: `endsWith`): one could add a top-level directory (`atoctave` for instance). It should be ignored by Matlab users (only `atmat` and `atintegrators` in the path), and should be put in 1st position by Octave users. There, you could put any function that simulates a missing Matlab function, or that replaces an `atmat` function for Octave users. Then you can keep the name `endsWith` (for instance) wherever it appears and provide your own version in `atoctave`.
- You can forget the files in the `atintegrators/user` directory: they are still there for historical reasons, by have not been upgraded for the python interface and are not compiled!

Finally, I agree with you idea of temporarily creating a dedicated branch for octave, until it can be merged into the main one.

matsievskiysv:
@lfarv
> Concerning the build commands: can you try to avoid makefiles ...
  
Replaced with file `bootstrapOctave.m`

> About missing Matlab functions ...
  
Done

> You can forget the files ...

Removed them from source list

> Finally, I agree with you idea of temporarily creating a dedicated branch for octave, until it can be merged into the main one.

As soon as you'll create an `octave` branch, I'll set it as this PR's merge target.

lfarv:
There is now a branch `octave_compatibility`. It adds an `atoctave` directory, where you should put all octave-specific .m files, so that they are ignored in Matlab.

matsievskiysv:
Ok. Demos seems to work except for the GUI stuff and ones that fail because of #200

matsievskiysv:
I don't know how the output of the demos should look like, so I attach graphical output:

---

**atfastring/testatfastring.m**

fastring-1
fastring-2

---

**atmatchExamples/bump/run_BumpFit.m**

bump

---

**atmatchExamples/ExampleATMATCH/runtotest_atmatch.m**

exampleATMATCH

---

**IDModeling/trackWithU35KickMap.m**

trackWithU35KickMap-1

trackWithU35KickMap-2

---

**opticsAndBeamSizes/linoptdemo.m**

*This one is obviously messed up*

linpotdemo

---

**opticsAndBeamSizes/ohmienvelopedemo.m**

ohmienvelopedemo-1

ohmienvelopedemo-2

---

**responseMatrix/findrespmdemo.m**

findrespmdemo-1

findrespmdemo-2

lfarv:
Hi @matsievskiysv, That's excellent ! There are still a few problems, but it's already very promising. On the demos, there is a problem with H/V coupling: you pointed the problem on atinopt, bit on ohmienvelope, on the right axis, the scale should be in the 1E-4 while you get 1E-18. Also, in findrespmdemo, the off-diagonal blocks are at zero, which is normal for the non-coupled ring, but not for the coupled ring. Finally, a detail, on some plots (bump fit, attach), you should get a synoptic representation of the ring magnets along the horizontal axis. I'll think about what could cause the tracking problems for could machines: it's like the tilted magnets would not couple horizontal position into the vertical plane…

lfarv:
Sorry to intervene again:

- For compilation of integrators: you should select  all ‘*Pass.c’ and ‘*Pass.cc’ files only. All other *.c or *.cc files are meant to be ‘#include’d in integrators, but not compiled independently.
- Can you avoid modifying all unused ‘atintegrators/user/*.c files ? The less modified files, the easier the review…
- I am worried by the requirement for the un-released Octave version 6. In my experience, system administrators are reluctant to continuously upgrade software packages to the last version, and users have often to leave with slightly older packages. This could block many potential users. I don't know how many functions are affected by the bug you mention, but at least the most critical ones could be re-written in the `atoctave` directory by extracting the problematic nested functions. For others, a partially functioning AT could be better than nothing.

lfarv:
For info: did you try `atdiag`? Here is the expected output:

test

matsievskiysv:
@lfarv  About `function_handle` for nested functions. It seems that I can wrap them into anonymous function:
```octave
[x,y,z] = dostuff(@handle, arg)		% replace this
[x,y,z] = dostuff(@(a) handle(a), arg)	% with this
```
Is this acceptable?

matsievskiysv:
By the way, Octave 6.1 [is out](https://www.gnu.org/software/octave/news/release/2020/11/26/octave-6.1.0-released.html)

matsievskiysv:
I get this for `atdiag`:
diag

Plot legend and annotations are missing.

matsievskiysv:
> For compilation of integrators: you should select all ‘*Pass.c’ and ‘*Pass.cc’ files only. All other *.c or *.cc files are meant to be ‘#include’d in integrators, but not compiled independently.

Edited

> Can you avoid modifying all unused ‘atintegrators/user/*.c files ? The less modified files, the easier the review…

Removed edits

> I am worried by the requirement for the un-released Octave version 6. In my experience, system administrators are reluctant to continuously upgrade software packages to the last version, and users have often to leave with slightly older packages. This could block many potential users. I don't know how many functions are affected by the bug you mention, but at least the most critical ones could be re-written in the atoctave directory by extracting the problematic nested functions. For others, a partially functioning AT could be better than nothing.

Unfortunately for now only 6.1 works. There's also some incompatibility in line continuations (`...`).

matsievskiysv:
> after small alterations of atplotsyn I get this

It also helped with some other plots

bump
atmatch1
atmatch2
atmatch3

atmat/attrack/linepass.m
```diff
@@ -90,7 +90,7 @@
 
 newlattice = double(~keeplattice);
 
-[prefunc,postfunc]=parseargs({function_handle.empty,function_handle.empty},...
+[prefunc,postfunc]=parseargs({cell(),cell()},...
```
lfarv, line 93: Works, but isn't there any way to create empty arrays on any type in Octave?

atmat/pubtools/lattice_tools/atreadbeta.m
```diff
@@ -160,7 +160,7 @@
 end
 
 
-evalin('base','global GLOBVAL');
+assignin('base','GLOBVAL',GLOBVAL);
```
lfarv, line 163: It is different from the original: here you create a local variable in the base workspace, identical to the global one (bad idea). Declaring the global variable in the base workspace makes it visible, but that's not so useful. The best is to suppress completely these statements (and the similar ones)

lfarv (commented):
I do not see incompatibilities with Matlab. One could merge this into the octave branch and see!

lfarv:
Now we are left with the problem of wrong results for coupled machine. I cannot get Octave 6.1 on MacOS, and I understood that you don't have access to Matlab. That makes checks difficult… The simpler check is just comparing numerical results for pure tracking over 1 turn of a coupled machine, then go to closed orbit, and then to transfer matrices. If you do not see an easier way, I could send you .mat files with Matlab results, can you read them? It may be slow since I cannot spend so much time on AT, but we can try.

matsievskiysv:
> I could send you .mat files with Matlab results, can you read them

I [should be able to](https://stackoverflow.com/questions/9915658/how-do-you-open-mat-files-in-octave), if `.mat` versions will match.

Maybe I'll be able to extract some information from https://github.com/atcollab/ATtraining examples as well.

matsievskiysv:
I've created a [docker](https://www.docker.com/) image with a [jupyter server](https://jupyter.org/) with [Octave](https://github.com/calysto/octave_kernel) and AT. Not all plots show, but it's sufficient for testing purposes.

@lfarv If you'll give me your email address, I could send you a link to my personal server so you could test AT there.

image

image

In order to use it:
1. Install [Docker](https://docs.docker.com/docker-for-mac/install/) (should be trivial)
1. Run command `docker run -d --name at -p 8888:8888 matsievskiysv/at`
1. Run `docker logs at --follow` and write down connection token (will appear on screen as `http://127.0.0.1:8888/?token=<token>`)
1. From web-browser go to `http://127.0.0.1:8888/` page and insert token there.
1. Create a notebook and execute commands. All AT should be available.

atmat/attrack/linepass.m
matsievskiysv: There's no `function_handle` class in Octave. In any case, anything empty will work: https://github.com/atcollab/at/blob/master/atmat/attrack/atpass.c#L403-L404

atmat/pubtools/lattice_tools/atreadbeta.m
matsievskiysv: Original statement made Octave crush. If these are not needed, I will gladly remove them.

willrogers:
This could be a really useful development, particularly allowing us to run Matlab tests and Python tests on pull requests and maybe compare the results. I'm very pleased someone is trying this, thanks!

I see Octave 6.1 was released on November 26th, but anyway since we have no Octave support at all at the moment I wouldn't be worried that it requires a new version.

atmat/pubtools/lattice_tools/atreadbeta.m
lfarv: Go on, remove all the similar lines that you pointed out

lfarv:
@matsievskiysv:

- I'll try the 'docker' way as soon as I'm back home (next monday probably). How shall I get your image?. If that fails, I'll try the access to your server,
- I propose that you continue the development of the octave port directly on the `octave_compatibility` branch. It will remove one intermediate step.

matsievskiysv:
> How shall I get your image?

`docker run -d --name at -p 8888:8888 matsievskiysv/at` command will download [it](https://hub.docker.com/r/matsievskiysv/at) automatically.

> I propose that you continue the development of the octave port directly on the octave_compatibility branch. It will remove one intermediate step.

It's fine by me

lfarv:
Here is a set of 2 files for checking `atlinop` with octave:

1. Remove the .txt extension which I added to get them accepted as attachments and put them anywhere in the octave path
2. execute checklinopt with either;

- `[results,expected]=checklinopt('coupled');`
- `[results,expected]=checklinopt('uncoupled');`

`results` and `expected` are structures containing the output of `atlinopt` in `lindata`, `tunes`, `chrom` fields.
'coupled' and 'uncoupled' run the same ring (spear2 example) without and with H/V coupling, introduced with quadrupole rotations.

[checklinopt.m.txt](https://github.com/atcollab/at/files/5637601/checklinopt.m.txt)
[checkring.mat.txt](https://github.com/atcollab/at/files/5637602/checkring.mat.txt)

matsievskiysv:
@lfarv Results look identical

matsievskiysv:
Got jupyter plots working. They do not look as good as Qt plots unfortunately.

image

atmat/atmatch/atmatch.m
```diff
@@ -131,7 +131,7 @@
 usepar = p.Results.UseParallel;
 
 
-options=optimset(minimizer);
+options=optimset(func2str(minimizer));
```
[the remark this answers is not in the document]
matsievskiysv: Done

atmat/pubtools/lattice_tools/atreadbeta.m
matsievskiysv: It broke some code. Particularly, `machine_data`. I'll need some time to fix it

lfarv:
@matsievskiysv: you should now be allowed to merge you fork into this branch. Let me know

matsievskiysv:
Yes, I can merge now. But I want to fix `evalin` issues first

matsievskiysv:
Don't know exactly why, but response matrix for coupled lattice now looks like this:

image

lfarv:
For `evalin` problems: some demos are using a obsolete way of keeping the 'active' ring structure in the global variable "`THERING`". Though this should still work, it is now recommended to set the ring structure as 1st argument of most functions.

For response matrix: it now looks good!

For `atlinopt`: fine if numerical results are good. I found out that the `linopt` demo take a non coupled ring (spear2), and introduces coupling by adding randomly generated tilts. This is not a good idea since the demo will give different results each time it is run!  But this does not explain why the coupling looked zero in you plot. Did you try again since the response matrix looks good?

matsievskiysv:
`linoptdemo`:
image
`ohmienvelopedemo`:
image
image

matsievskiysv:
Jupyter inline graphics looks a bit sad, but works.

image

PR #199, opened 2020-11-30
atlinopt improvements

lfarv:
This pull request adds 2 things:

#### 1. Consistent differentiation steps and across all functions:
there are 4 constants (at the moment):

> XYStep = 1.0E-8 :	Transverse step (for findorbitx, findmxx,...)
> DPStep = 1.0E-6 :	Momentum step (for dispersion and chromaticity)
> OrbConvergence = 1.0E-12 :	Convergence of findorbitx
> OrbMaxIter = 20 :	Max number of iterations for findorbitx

These constants can be modified either globally, or at the function level through keyword arguments. In Matlab, the default value is now different from the historical one of 6.05545e-6, possibly slightly modifying some results.

- For Matlab: the default values are defined in `initoptions.m`, and the constants can be obtained and modified with the `getoption` and `setoption` functions
- For python: the constants are handled in the` at.physics.diff_constants` module and accessed using the `at.physics.DConstant` class object.

#### 2. `atlinopt` can be used for transfer lines similarly to the equivalent python function


lfarv:
There is a very strange behaviour with the tests: all tests succeed on Windows, and on Linux for versions 2.7 to 3.5. But they fail on linux 3.6 to 3.8, and on macOS. But I cannot reproduce this at home: all tests succeed, on both linux and macOS. Since I cannot reproduce, I do not know where to look. The differences appear on horizontal chromaticity and longitudinal damping time ! And the values reported by the failing tests are indeed different slightly from the values computed here.

willrogers:
These sorts of things are tricky.

I suppose that we expect numerical differences, so either these differences are permissible and we allow a wider range, or there's actually a bug somewhere.

swhite2401:
Sorry I haven't had time to take a look yet, I think I can get to it next week

lfarv:
For sure we can expect numerical differences for different machines or libraries. But I cannot understand different results for the same machine. It seems that fairly large discrepancies (on chromaticity for instance) can all be related to very small deviations on the closed orbit computation (which is the 1st thing done everywhere). When introducing the same closed orbit for transfer matrix computation, I get results identical to the machine precision everywhere, instead of 1e-7 usually. I'm investigating why the closed orbit computation may differ on GitHub and locally (NumPy version ?), and I'll purpose modified tests allowing a reasonable deviation on closed orbit, but much more strict when using a fixed closed orbit. More in a few days !

lfarv:
@willrogers : I don not understand why Travis checks disappeared!

willrogers:
It appears that Travis has stopped offering free builds.

GitHub offers a similar service that is free (for now) but it would take a bit of work to move across.

lfarv:
Ready for merging !

The default values for differentiation steps are now:
- XYStep: 1.0E-8
- DPStep: 1.0E-6

Some more work is still necessary to find the best values, but everything is now implemented to make this easier.
Because of the new values, results are slightly different, which explains the changes in the test functions. An update of the `test_physics.py` module would be a nice improvement.

For @willrogers: when the tests fail apparently most of them stop with the mention "This test was cancelled" without any indication of the reason for cancelling, while only a few go to the end so that the test results can be visualised (see the las failure at commit bb821bf).

swhite2401:
@lfarv sorry I haven't gotten to this yet, next should be more quiet, feel free to merge I don't want to block anything

willrogers:
I approve that you are getting rid of the magic numbers with the mysterious origin.

I think that any one of the tests failing means that any that haven't completed yet are cancelled. I suppose this means that we use less of the computers that Github are providing us for free. There may be a way to change this if you think it's important.

lfarv:
### Optimisation of XYStep
To look at the influence of XYStep, we take a reference point in a lattice, and then apply a circular shift of lattice elements while keeping the same reference point. When we change the starting point of the lattice, global parameters like tunes, chromaticities, emittances should be constant, as well as local parameters at the chosen reference point (closed orbit, dispersion, beta functions...). In practice, the standard deviation of those values will be taken as a figure of merit to optimise XYStep. We shall plot the standard deviation over 20 starting points of the hmba lattice for Matlab results, for python results, and for the difference between Matlab and python results.

#### Test 1 : 4D, dp/p = 0
dp000

- The standard deviation of Matlab results and python results are almost equal,
- The standard deviation of the difference is much smaller than the intrinsic deviation of each version,
- The lower XYStep, the better.

#### Test 2 : 4D, dp/p = 0.01
dp010

- The same behaviour appears in vertical,
- In horizontal, there is an optimum around 3.0E-08,
- The difference between Matlab and python increases for smaller XYStep, though it is still far below the intrinsic fluctuation

#### Test 3 : 6D, tunes and damping times
6d

- Here the optimum is around 1.0E-08

lfarv:
The same test performed on various lattices shows an optimum for XYStep around **3.0E-08** to **1.0E-07**.
Example for the esrf lattice from `at/machine_data`:
#### 4D
esrf4d
#### 6D
esrf6d

lfarv:
### Optimisation of DPStep
The same procedure is applied when varying DPStep. The variables depending on DPStep are the chromaticities, the dispersion, the longitudinal tune and damping time

hmba3

- Here the standard deviation decreases when DPStep increases.
- The deviation on horizontal chromaticity is quite large

The same behaviour is shown on all lattices, though the orde of magnitude of horizontal chromaticity varies a lot.

The choice of DPStep is therefore a compromise between the stability of the result (versus starting point of the lattice) and accuracy, since increasing DPStep means averaging around the desired dp.

lfarv:
According to the previous results, the default values have been set to:
XYStep = 3.0E-08
DPStep = 3.0E-06
With these values, the fluctuation of parameters as a function of dp/p are shown on the next plots

hmbadp1

- vertically, the fluctuation of tune and beta function are very small, with no strong dependence on dp/p, and the difference between Matlab and python is even smaller (1.0E-14)
- horizontally, the discrepancy between Matlab and python is larger (1.0E-09), except for dp/p =0 !!!, but still smaller than the intrinsic fluctuation.

hmbadp2

- The fluctuation of horizontal chromaticity is quite large (1.0E-05, but can reach 1.0E-03 on some lattices), except for dp/p = 0

lfarv:
This is my final setup, and based on these results I'll try to review the tests in another pull request to stick more closely to the expected values.

willrogers:
Can you please leave a comment in the code explaining where the numbers came from this time? Otherwise in 5 years time someone might be puzzled again. I would refer back to this PR.

PR #209, opened 2020-12-12
Octave compatibilty

matsievskiysv:
Add support for [Octave](https://www.gnu.org/software/octave/); add [Jupyter](https://jupyter.org/) server files.


matsievskiysv:
@lfarv Fixed aforementioned incompatibility issues

lfarv:
@matsievskiysv: a last problem with `atplot`:

> \>\> atplot(mring)
> Conversion to logical from matlab.graphics.chart.primitive.Line is not possible.
> 
> Error in atplot (line 92)
> if (comment && curve.left)
> 
It should be:
`if comment && ~isempty(curve.left)`

Then Matlab plots correctly!

matsievskiysv:
Edited this line.
Does the legend look good in MatLab? In previous commits I've removed font size specifications to make jupyter plots look better.

FYI, submitted a [bug report](https://savannah.gnu.org/bugs/?59660) to Octave regarding messed up plot legends in jupyter.

lfarv:
The plot looks good enough in Matlab. Did you look at Octave native plots? From what I have seen, notebook plots are far from publication quality, and probably many users will prefer the native GUI for real work.

I started running your new demo notebooks with a Matlab kernel in Jupyter, and on the `octave_compatibilty` branch. Up to now it looks good, and again the notebook plots are rather bad. But jupyter gives an option for native plots (`%plot native `magic). Here is an example with inline plot,  I'll let you know the outcome for the demo notebooks.

Screenshot

lfarv:
Good news: all Jupyter demos are successful with Matlab. Incidentally, here is the difference in performance on `atfastring` (pure tracking) between native Matlab and dockerized Octave on my machine:

**Matlab**:

Screenshot Matlab

**Dockerized Octave**

Screenshot Octave

The puzzling point is that in Matlab, "fastring" in ~100 times faster than "normal" tracking, while in Octave the difference is less that a factor 4… I hope to get some time a native Octave running here to get more meaningful results.

Also I started reporting in #210 a list of precautions to keep in mind to avoid compatibility problems with Octave

matsievskiysv:
> Did you look at Octave native plots?

By default Octave uses QT graphic backend. It produces [nice looking plots](https://github.com/atcollab/at/pull/198#issuecomment-736627346).

> But jupyter gives an option for native plots

This would work only if running jupyter locally. Docker containers cannot interact with system GUI.

matsievskiysv:
I've got a a bit better result.

image

AFAIK, MatLab uses JIT compilation. There's an experimental support for JIT in Octave. I'll try to build an Octave version with JIT support and rerun this test.

lfarv:
I finally got a native Octave running, and I found a problem with `bootstrap`:

For the integrators, we need the search for all `'*Pass.c'` files, because any user may add his own integrators and have them compiled.

But for the others directories, we need a specific `mex` command for each: for instance to have `nafflib` compiled, we need `nafflib.c`, `modnaff.c,` `complexe.c` compiled **together** (in that order) into a single mex-file, so the command should be
`mex nafflib.c modnaff.c complexe.c`
This also applies to some extent to the other directories since the options may differ: only `atpass.c` and `findmpoleraddiffmatrix.c` need access to the integrators headers ('`-I../atintegrators`').
Also in `atphysics/NonLinearDynamics`, `RDTelegantAT.cpp` is ignored by `bootstrap`.

I don't think `-DAT_MODE` is necessary, it's defined internally in `ExactHamiltonianPass.cc`

matsievskiysv:
@lfarv edited bootstrap script, pulled latest changes from master and updated docker image. One demo broke after recent changes #213.

atintegrators/mexpassmethod.m
```diff
@@ -31,6 +31,10 @@ function mexpassmethod(PASSMETHODS, varargin)
 
 PLATFORMOPTION=[varargin{:}];
 
+if isOctave
+  error('mexpassmethod does not work in Octave')
+endif
+
```
lfarv, line 37: Should be `end` instead of `endif`

matsievskiysv: Fixed

matsievskiysv:
[Octave 6.1.1](https://packages.debian.org/bullseye/octave) is now properly packaged in Debian. With it I get the following function timings:
```
+ %Now, track with full ring and with fast ring and time the computation.
+ tic
+ z1=ringpass(ring,z0,500);
+ toc
Elapsed time is 0.493984 seconds.
+
+ tic
+ z1fast=ringpass(fastring,z0,500);
+ toc
Elapsed time is 0.00204992 seconds.
```
I'll update the docker image soon with the new Octave version.

matsievskiysv:
@lfarv I think that it's ready for master

lfarv (approved):
Ok for me!

lfarv:
Hi @matsievskiysv.
No problem on Matlab with this branch, so I'll be happy to merge.

Concerning JupyterHUB, I am stuck:

When I start JupyterHUB, I get the login page.
According the your instructions, I choose "Signup!".
Then I put 'admin' as username, set a password and press "Create User". The answer is:
"Your information has been sent to the admin"
Then I go to "https://localhost/hub/authorize", and I get a "403: Forbidden" error!

Tried again after `make purge; make; docker-compose up`, the result is always the same…

Anyway, there may still be some work on that (Jupiter lab?), so I'll be happy to merge as it is, since AT itself is OK on Matlab, Octave and python, and let you go on with Jupiter and Docker.
What do you think of that ?

matsievskiysv:
@lfarv after creating admin your you must click login button and enter the credentials. `admin` is a special user and does not need to be authorized before use (because there's nobody to authorize him).

lfarv:
Thanks, I got it running, and tested on my home network: perfect!

PR #211, opened 2020-12-14
Back compatibility old at element creation

ZeusMarti:
A 'class' field has been added for back compatibility in elements:

1. corrector
2. id_table_global
3. marker
4. multipole


lfarv:
Hello @ZeusMarti. Welcome improvement!
I suggest that for `marker.m`, `corrector.m` and` multipole.m `you follow the method used in `quadrupole.m` (for instance): the old constructor calls the new one. This avoids code duplication, and any modification is automatically available in the old-style constructor.

atintegrators/wiggler.m
```diff
@@ -20,7 +20,7 @@
 %  See also atdrift, atquadrupole, atsextupole, atsbend, atskewquad,
 %          atmultipole, atthinmultipole, atmarker, atcorrector
 
-ElemData = atwiggler(fname, Ltot, Lw, Bmax, Nstep, Nmeth, By, Bx, method);
+ElemData = atwiggler(fname, Ltot, Lw, Bmax, getenergy, method,'Nstep',Nstep, 'Nmeth',Nmeth, 'By',By, 'Bx', Bx);
```
lfarv, line 24: I cannot find `getenergy`. Am I missing a function?

ZeusMarti: Sorry, I thought it was an AT function, which is the standard way to get the energy?

lfarv: Good question !
For a given lattice, you can get the energy with `atenergy(ring)`. It looks in order for:

- a `RingParam` element (new-style lattices)
- a `Cavity` element
- the global variable `GLOBVAL.E0`
- an `Energy` field in any element of the ring

Without a ring, you are left with only the global GLOBVAL.E0, which is not guaranteed to exist…
I am not very pleased with the present wiggler definition, it would have been better to describe it with a Bmax/(B.ro) normalised field, but that's how it is!
`atradon` will force correctly the energy in all radiating elements, including wigglers, but you may want to track a wiggler without radiation. That's why `atwiggler` needs explicitly the energy (though if defaults to zero if not provided).

So no obvious solution…

ZeusMarti: Ok, I added the possibility to pass the energy as an extra argument in wiggler.m, if it is not there it will look at GLOBVAL.0, it it is not there it will get a default value 0. What do you think about it?

lfarv: I don't see anything better !

lfarv:
So is it ready for merging now?

ZeusMarti:
This is my first "pull request", tell me if I have to do something else. If it is ok to everyone please merge it.

lfarv:
Nothing more to do ! I merge now

Issue #215, opened 2021-01-11
Bug in DipoleRadiation.m


lfarv:
Hello @lfarv,
I think there is a bug in DipoleRadiation.m. The function looks for the field K of the dipoles, but K is not a mandatory field.
Now ringpara fails with the ESRF lattice, because the K value is not found in the dipoles.
The old ringpara was looking at both K and polynomB(2), if I remember well, and then it was giving a warning in case the two were different.

_Originally posted by @carmignani in https://github.com/atcollab/at/issues/168#issuecomment-757996700_

lfarv:
Fixed in #216

PR #216, opened 2021-01-11
Fixed bug in Matlab version for missing 'K' in dipole

lfarv:
Fix  bug reported by @carmignani (#215)


carmignani:
Hello Laurent,
thanks for the quick fix! I checked the new ringpara and it works well.

PR #219, opened 2021-01-21
Fixes in at.lattice.utils.py

lfarv:
- This makes `bool_refpts` accept integer scalars, empty list (`[]`) and `None`.
- A few functions are simplified,
- a new function `checkname` is added: it generates a filter function selecting elements with `FamName` matching a given pattern. This complements the existing function `checktype` and `checkattr`.

### DIFF
--- a/pyat/at/lattice/utils.py
+++ b/pyat/at/lattice/utils.py
@@ -59,11 +59,9 @@ def uint32_refpts(refpts, n_elements):
     Return a uint32 numpy array with contents as the indices of the selected
     elements.  This is used for indexing a lattice using explicit indices.
     """
-    refs = numpy.asarray(refpts).reshape(-1)
+    refs = numpy.ravel(refpts)
     if (refpts is None) or (refs.size == 0):
         return numpy.array([], dtype=numpy.uint32)
-    elif refs.size > n_elements+1:
-        raise ValueError('too many reftps given')
     elif numpy.issubdtype(refs.dtype, numpy.bool_):
         return numpy.flatnonzero(refs).astype(numpy.uint32)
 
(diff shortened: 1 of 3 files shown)


lfarv:
@swhite2401: I'm again questioning the keyword  'order' in `get_value_refpts` and `set_value_refpts`: 'index' should be better. I think you are the author of these functions, and there are only a few usages (in` globalfit.py`). So it may be the last opportunity to change, what do you think?

Otherwise, it's ready!

willrogers:
It seems to me that it would be useful to say 'alll refpts'. At the moment you have to use `range(len(ring) + 1)` or something like that, which seems awkward.

Any thoughts? Is there an easier way?

lfarv:
@willrogers : I agree, that would be very convenient!
One way would be to create a class with a singleton instance (similar to `None`), which could be used in function calls like:
`at.linopt(ring, refpts=at.AllRefpts)`
and tested in utility functions as:
`if refpts is at.AllRefpts:`

However, I would like to fix here a bug which prevents writing` bool_refpts(0, len(ring))` to refer to the 1st element (for instance), and keep this improvement for a next step.

willrogers:
We should add tests for these cases. There are tests that we can add to.

swhite2401:
> @swhite2401: I'm again questioning the keyword 'order' in `get_value_refpts` and `set_value_refpts`: 'index' should be better. I think you are the author of these functions, and there are only a few usages (in` globalfit.py`). So it may be the last opportunity to change, what do you think?
> 
> Otherwise, it's ready!

Hello @lfarv, I agree to change it. However, we have a certain number of ESRF operation tools calling these functions it will take a bit of testing, can we put this on a separate issue?

willrogers:
I won't block merging, it's just my general reminder about tests. I pushed a few more.

lfarv:
refpts_count has a `n_elements` input argument, in prevision of a future "all refpts" special value.

`'order'` is replaced by `'index'` in `get_value_refpts`/`set_value_refpts`. However the obsolete `'order'` keyword is still temporarily accepted.

Anything else?

willrogers (approved):
Fine by me.

swhite2401:
Ok for me too

swhite2401:
@lfarv , I have cleaned up all the dependencies of operation tools to the variable order (I think...).
You may remove the code added for backward compatibility

Issue #218, opened 2021-01-20
Make pyat-0.1.0 release using Github Actions


willrogers:
Previous releases were made using Travis CI, but we are now using Github Actions for CI.

Useful links:

* [cibuildwheel docs](https://cibuildwheel.readthedocs.io/en/stable/)
* [Example Gtihub Action for deploying to PyPI](https://github.com/joerick/cibuildwheel/blob/master/examples/github-deploy.yml)
* [Accelerator Toolbox on PyPI](https://pypi.org/project/accelerator-toolbox/)

It is a tricky process, but is mostly working now. Currently failing:

* pypy on MacOS. This is present on the 0.0.4 release.

willrogers:
@lfarv at the bottom of `at.c` I see:

```
#if PY_MAJOR_VERSION < 3
int main(int argc, char *argv[]) {
    /* Pass argv[0] to the Python interpreter */
    Py_SetProgramName(argv[0]);

    /* Initialize the Python interpreter.  Required. */
    Py_Initialize();

    /* Add a static module */
    initatpass();
    return 0;
}
#endif
```

As far as I can tell `Py_Initialize()` and `Py_SetProgramName()` are not required, because they are specific to embedding a Python interpreter; removing them makes it possible to build wheels for PyPy, not that that is particularly important.

Do you remember more? It is a while since I looked at this code. Do we need the main function at all?

willrogers:
If I remove that `main()` function all the tests pass on Windows, Mac and Linux.

lfarv:
Perfect! As far as I remember, that block of code came from a tutorial on writing extensions, but I agree that there is no need for a `main()` here. If everything works without it, go on!

willrogers:
Is there anything particular to add before I make a 0.1.0 pyat release?

@lfarv @swhite2401

lfarv:
@willrogers, @swhite2401 : I would just merge #219 (bug fix), otherwise it looks like a good time to push a 0.1.0 release. Any objection ?

willrogers:
Released here:

https://pypi.org/project/accelerator-toolbox/

PR #220, opened 2021-01-21
Add Github Action for building pyat wheels.

willrogers:
Each time a tag of the form pyat-* is pushed, all wheels
will be built and uploaded to Github as 'artifacts'. These
wheels may be downloaded, tested and uploaded to PyPI. In
future we could automatically upload wheels directly to PyPI.

Other changes:

* refactor Python testing workflow
* add tests on Python 3.9 and fix associated bug
* remove unnecessary function from at.c
* improve pyat README files
* update pyat version to 0.1.0 in preparation for release


willrogers:
See #218

lfarv (approved):
Ok for me!

PR #221, opened 2021-01-31
Parallelise integrators with openMP

lfarv:
openMP directives were introduced in some integrators some time ago by Xiabiao Huang, but were not activated.
They can now be optionally activated.

## Behaviour

In each integrator, the loop over the different particles is parallelised. This occurs if the number of particles is greater than 10 (by default, configurable). Below this number, nothing is changed, so functions like findorbit*x*, findm*xx*, linopt… are not accelerated. on the other hand, when tracking 1000 particles, I could see an acceleration by a factor 2 to 3 on a desktop machine.
At the moment, apart from the nb. particles threshold, everything (like the number of threads) is automatically handled by openMP. Additional functions may be added in the future to get more control.

## Requirements

An openMP-aware compiler is needed. This is what I observed:
#### Linux:
gcc is compatible, nothing to do.
#### MacOS:
For Matlab, we must use the openMP library provided by Matlab: nothing to do,
For python, an openMP library needs to be installed, preferably with a package manager:
`$ sudo port install libomp` or
`$ brew install libomp`
#### Windows:
Nothing to do with Visual Studio Community 2017, to be tested with other compilers.

### Matlab
Rebuild the integrators with:
```
>> atmexall -openmp
```
The threshold on the number of particles can be adjusted (to 5 in this example) by using:
```
>> atmexall -openmp -DOMP_PARTICLE_THRESHOLD=5
```
Notice: I experienced crashes when compiling with a Matlab version and running with an older version.

### Python
Set the environment variable `OPENMP` to a non-null value when building pyat. Example for `sh`/`bash`/`zsh`:
```
OPENMP=1 python setup.py build/install/develop
```
You may need to remove previously compiled integrators with:
```
rm -rf build
find at -name "*.pyc" -exec rm '{}' \;
find at -name "*.so" -exec rm '{}' \;
```
The threshold on the number of particles can be adjusted (to 5 in this example) by using:
```
OPENMP=1 OMP_PARTICLE_THRESHOLD=5 python setup.py build/install/develop
```
### Octave
Not yet implemented, in progress…


lfarv:
I still have to add documentation…

matsievskiysv:
Pushed edits for Octave build script.

Tested it using [`testTracking`](https://github.com/atcollab/at/blob/master/atmat/atdemos/tracking/testTracking.m) example with `N=30`:
```octave
tic;testTracking;toc
```
Got 240 and 295 seconds with and without OpenMP respectively.
@lfarv could you maybe suggest a better test case?

matsievskiysv (approved):

swhite2401:
hello all, 
I did not know this was implemented, could you point to an integrator that has it for me to see the code?

As a general comment on this, openMP is nice but relies on shared memory similar to matlab parfor or pyhton multithreading, as soon as you will want to use several nodes (i.e. in a cluster architecture) you will need something that can deal with distributed memory and openMP cannot. For this, an MPI implementation is needed. 
I also strongly discourage hydrid implementation combining the two. I know this is a much more significant development, but I would go for MPI right away and not spend too much time on this.
The main application for such development is multi-particle tracking and in my opinion it only makes sense if it can be used on a cluster.

swhite2401:
I took a look and it is in fact very simple, one question though: wouldn't it be simpler to handle this at the atpass.c level?
One could think of splitting the particle array into N array and run the parallelized loop on these.
This way the integrators are left unchanged, and for pyAT there is nothing to do as there the patpass function that implements the same at the python level using multithreading.

Only problem I can see is for the impedance elements that requires a knowledge of all particles, but again here I believe MPI is needed.

lfarv:
@swhite2401 : As you can see, the implementation is very simple, and none of the integrators is modified in this branch! The only new thing here is the selection of the adequate compile flags for different platforms. An implementation at the `atpass.c` level might be marginally more efficient but requires more work and indeed raises the problem of impedance-like elements.

`patpass` in python has the same problem (impedances), and as far as I could see is slightly less efficient than openMP.

 `parfor` in Matlab is apparently quite different as it does not use shared memory, but spawns several processes (so called "workers"), on a single machine or on several nodes, and circulates the data between them through .mat files. It's more like to MPI, and has a lot more overhead.

I agree that openMP should not be mixed with another parallelisation solution, that's why it's optional and defaults to "non-activated". But the openMP option offers almost for free a solution for any desktop machine, with very little overhead. 

@matsievskiysv : I'll add a test function to the distribution

swhite2401:
I fully agree, this is very nice and free as you, I was just suggesting further developments.
As far as I can see patpass is not usefull anymore, should it be removed to avoid conflicts?

Ok for me to merge, btw

lfarv:
I added a test function `trackMultipleParticles` in `demos/tracking`. It tracks 500 particles over 500 tunes (by default) and prints the elapsed time (without and with longitudinal motion). As an example, here are my results (MacOS, quad-core processor):

**Matlab without openMP**
\>> trackMultipleParticles()
Elapsed time is 9.9 seconds.
Elapsed time is 24.7 seconds.

**Matlab with openMP**
\>> trackMultipleParticles()
Elapsed time is 3.2 seconds.
Elapsed time is 7.1 seconds.

**_The gain is approximately a factor 3 !_**

**Octave (native) without openMP**
\>> trackMultipleParticles()
Elapsed time is 9.8 seconds.
Elapsed time is 24.6 seconds.

**Octave (native) with openMP**
\>> trackMultipleParticles()
Elapsed time is 6.2 seconds`                   `<--surprisingly long
Elapsed time is 9.8 seconds.

lfarv:
Correction: Octave results are equivalent to Matlab results. The problem was that in Octave, clear mex does not ensure that updated files will be loaded. The right results were obtained after reloading Octave.

**Octave (native) with openMP**
\>> trackMultipleParticles()
500 particles, 500 turns, without longitudinal motion: 3.2 s
500 particles, 500 turns,    with longitudinal motion: 7.1 s

willrogers:
Fine for me, and feel free to remove `patpass`.

matsievskiysv:
@lfarv enabled OpenMP in Jupyter image

lfarv:
The choice of the threshold for parallelisation is not obvious: here is a plot showing the ratio between tracking with and without openMP, for 1 to 10 particles (quad-core desktop machine):

test_opnmp

With a threshold 0 (blue curve, always parallelise), the overhead from spawning the threads increases the elapsed time for n_particles < 5, up to a factor 2 ! In this case, setting OMP_THRESHOLD=4 (red curve) suppresses the problem, and gives some improvement starting with 5 particles (such as in the computation of closed orbit or response matrices).

However, this depends on the machine: here is what I got on another one (Linux server):

test_opnmp2

Here, attempting to parallelise less than 10 particles may give terrible results. However the default threshold of 10 looks safe enough.

To make this check easy, I added a function `test_openmp(threshold)` in` atmat/atdemos/tracking`;
the 1st plot is obtained with

`>> test_openmp(4)
`

and the second with

`>> test_openmp()`

swhite2401:
Very interesting! I am not sure how openMP works, do you you initialize a new environment every time you enter a passmethod or the same is re-used for the whole tracking?

lfarv:
@swhite2401 : no idea… For your information, the 1st results comes from my Mac, the second from "grappa". In absolute "grappa" is also much slower, which is disappointing!

matsievskiysv:
Some considerations:
* Maybe different compilers have different [schedulers](http://jakascorner.com/blog/2016/06/omp-for-scheduling.html) enables by default? I worth investigating which one works best in this case.
* It's probably better changing `defatlt(shared)` to `default(none)`. Intuitively, reducing a number of shared variables should increase performance.
* Particle treshhold should be expressed in particle/thread units.

swhite2401:
@matsievskiysv thanks for you insight, I was actually documenting myself, and I agree we should limit the number of threads.
From what I have understood so far (correct me if I am wrong) openMP will spawn by default a number of threads = number of CPUs.
The grappa machine has 64 CPUs, so with low particle threshold there are many empty threads that will consume ressources, it could explain some of the issues with speed-up that Laurent observed.
I suggest to add a command that sets the max number of threads<=particle threshold (to be optimized of course)
I think this can be done by simply adding the  num_threads(nthreads) to the existing openMP commands

matsievskiysv:
It will be more convenient to use [envvars](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.cbcux01/ruomprun.htm). This way recompilation or code changes will not be needed.

matsievskiysv:
BTW, a large number of cores operating on the same memory region would decrease performance due to [cache coherence issues](https://en.wikipedia.org/wiki/Cache_coherence). It's better to split memory into big and continuous chunks. On the first glance [`static`](http://jakascorner.com/blog/2016/06/omp-for-scheduling.html#static) scheduler should work better in this case.

lfarv:
Looking at how we use of AT, I see 2 main cases:
1. single particle optics: closed orbit, transfer matrices and all their derivatives. This uses 4 to 6 particles (less than 10 in any case). There parallelisation does not bring a lot, OMP_PARTICLE_THRESHOLD should guarantee that at least we do not loose. The only question is to see whether we can accelerate a little bit the computation of closed orbits/transfer matrices or not. Obviously, this depends on the machine.
2. Study of populations: frequency maps, impedances, cavity beam loading… There we need a lot of particles, and in this case the more CPUs, the better. There is no reason to limit the number of threads.

In both cases, it does not look like a control of the number of threads is necessary, except if OpenMP creates useless threads, which seems stupid. If this control is necessary, apparently this can be done explicitly using a dedicated command:

>> omp_set_num_threads [3.2.1] [3.2.1]
Affects the number of threads used for subsequent parallel regions not specifying a num_threads clause, by setting the value of the first element of the nthreads-var ICV of the current task to num_threads.
**void omp_set_num_threads(int num_threads);** 

which seems better than hard-coded value or environment variables.

lfarv:
Concerning the most interesting case of many particles, here is a compilation of my results.

Results are obtained with
`>[t1, t2]=trackMultipleParticles();
`
which tracks 500 particles over 500 turns in the hmba test lattice. I take the average of two runs.
1st line: t1:  no longitudinal motion
2nd line: t2: with longitudinal motion (`atradon`)

|  | quad-core desktop | dual-core laptop | "grappa", 64 CPUs 
|- | - | - | - 
| without OpenMP no rad | 9.9 s | 9.8 s | 12.5 s 
| without OpenMP rad | 24.7 s | 26.3 s | 21.9 s
| |  |  |  
| with OpenMP no rad | 3.2 s | 6.5 s | 3.6 s 
| with OpenMP rad | 7.1 s | 14.0 s | **1.6 s** ??
| |  |  |  
| ratio no rad  | 3.2 | 1.5 | 3.5
| ratio rad  | 3.5 | 1.5 | 13.7
| theoretical ration: n CPUs  | 4 | 2 | 64

Remarks:
- for single-thread, the 3 machines are very similar,
- for multi-thread, the gain on the two Macs is close to the number of cores, which makes sense. For grappa, it seems obvious that the number of threads used is far from the total number of CPUs,
- grappa with openMP is the only example where the computation with radiation on is faster (twice faster) than without !!! To be investigated

matsievskiysv:
> If this control is necessary, apparently this can be done explicitly using a dedicated command, which seems better than hard-coded value or environment variables.

Manual OpenMP configuration would be tedious and error-prone.
More elaborate then `OMP_PARTICLE_THRESHOLD` condition may solve this issue. I propose adding user-controlled option `MIN_PARTICLES_PER_THREAD` (or something like that). Setting `MIN_PARTICLES_PER_THREAD` to, lets say, 10 would automatically disable parallel execution for the single particle optics case and distribute load more evenly for population studies (not to spawn 64 threads for 64 particle simulation).
It should not be hard to implement using `if` and `num_threads` arguments.

matsievskiysv:
I've implemented `MIN_PARTICLES_PER_THREAD` option in <https://github.com/matsievskiysv/at/tree/with_OpenMP>. It's controlled by envvar.
Benchmark:
|`MIN_PARTICLE_PER_THREAD`|Used CPUs|Without longitudinal motion|Without longitudinal motion|
|:--:|:--:|:--:|:--:|
|300|1|9.96|22.37|
|200|2|5.49|11.52|
|150|3|4.00|7.96|
|100|4|3.39|6.74|

swhite2401:
I am not in favor of using environment variables, I very much prefer @lfarv 's suggestion of using:

`omp_set_num_threads(int num_threads);`
where 
`num_threads = num_particle/num_particle_per_thread`
This by the can be implemented in an upper layer (before the code enters any openMP for loop) without any modification to the existing passmethods which I think should be avoided too, especially `include omp.h` should not be required

I think it is important to be able to set the number of threads. 
For instance  grappa is a multi-user machine and although it seems that the default does not use all CPUs it is potential problem issue for other systems where openMP could decide to fill the machine.

matsievskiysv:
I don't understand what do you have against envvars. For example, you may configure OpenMP from Matlab/Octave directly using `putenv('OMP_NUM_THREADS', LIMIT)`. `omp_set_num_threads` function would require header.

swhite2401:
envvars are setable and visible from anywhere by anyone, other instances/programs (that you don't know about) may change them. This might be ok when for your personal usage on your own desktop, but when you share accounts and environments with
others this is an evident source of problems and conflicts.

Headers can be avoided using:
`#pragma omp parallel for num_threads(nthreads)`

matsievskiysv:
It's not trivial to [set envars for another process](https://stackoverflow.com/questions/205064/is-there-a-way-to-change-the-environment-variables-of-another-process-in-unix) even when run as a same user. Doing so for a different user would be a security bug in the OS.

lfarv:
Hello @swhite2401 and @matsievskiysv. Sorry to arrive late: sunny day, no computer !
Looking at what happened today:
- Why would we set the minimum number of particles per thread to more than one ? With a quad-core machine, I already see a ~20% improvement with 6 particles and 4 cores used (interesting for `findm66`!). The only problem could be threads with 0 particle, I hope openMP is clever enough not to start threads which do nothing (but I'm not sure).
- So the main reason to control the number of threads is Simon's one: to avoid clogging a multi-user system for one user.
- Then I would favour minimal modifications to the integrators. Instead of recomputing the openMP parameters in each element, on each turn, these parameters, whatever they are, like test condition or number of threads, etc., could be computed once when entering `atpass.c`, and transmitted to the integrators through their last argument` struct parameters *`. We just need to extend this structure. That's not a trivial modification: we need to carry the information from the interface function `struct elem *trackFunction()` to the "compute" function (`StrMPoleSymplectic4RadPass()` for instance) and handle the case of direct call not going through `atpass`.

To summarise, getting the optimum from OpenMP may still take  a while. On the other hand, as it is now, this branch is safe (off by default, and protected from undesirable behaviour if turned on, thanks to `ORM_PARTICLE_THRESHOLD`), it gives some improvement on single-user machines, and also after some tests on compute nodes. So I propose to:
1. merge it as it is,
2. start to think of the best strategy to give more control.

Last remark: I'm also not fond of environment variables. I've always seen difficulties to have unaware users run with an adequate environment. It's really not intuitive. Except if, as suggested by @matsievskiysv, environment variables are both set and used within the process. But then it's really a loss of efficiency compared to argument passing, `setenv`/`getenv` is expensive.

swhite2401:
Hello @lfarv lucky you! Looks like you are going to have another nice day

I agree with most of what you say, and  we can merge the existing simple version.

Few additional comments or wishes for potential upgrades:
- changing integrators should be done only if necessary
- AT should not require openMP to compile:specific functions and` include` should be hidden by `#if defined(_OPENMP)` and openMP should be off by default
- I fully agree with setting parameters in `atpass.c`, however I do not see the need to pass anything to the integrators. Using `omp_set_num_threads` or setting the envvar `OMP_NUM_THREADS`, if this is the preferred option, should affect all subsequent parallel blocks if I understood well
- a test to make sure that the number of threads never exceeds the number of particles tracked could be useful, for this we may have to disable dynamic thread adjustments

lfarv:
Hello @swhite2401.

About bullet 2:
This is obviously necessary.

About your bullet 3:
If the additional controls are limited to setting the number of threads, you are right, calling `omp_set_num_threads()` from `atpass` does not need anything else in the integrators: it will affect all parallel blocks which don't specify `num_threads`.
This if my preferred choice: no need to change the integrators.
If we want more, you can just compute a condition for multi-threading, the number of threads, and possibly more (schedule, chunk_size), in atpass. And give these to the integrators to be used if the #pragma directives. No headers anywhere.

About bullet 4:
If `atpass` computes the number of threads based on the number of particles, this should be enough, isn't it?

swhite2401:
I am not in favor of automatically computing the number of threads, there are many reasons why the user may want to chose the number of threads without any constraints, some of which we have discussed already.

I would personally have number of threads as free parameter, with a max value =number of particles to avoid empty threads, and if not specified use the OMP default (what is done in the present implementation).

matsievskiysv:
> to avoid empty threads

Is there any indication that empty threads are spawned? I suspect that it doesn't do unnecessary things.

> user may want to chose the number of threads without any constraints

this is a default behaviour

---

So, IMHO this PR satisfies basic requirements and may be merged. Discussion may be continued in a separate issue with consideration for openMPI implementation.

swhite2401:
By default it probably doesn't, but if you ask for them maybe it does...to be checked
For example, the user sets `num_thread > num_particles`, in this case you could end up spawning useless threads because you have more threads than jobs.
In a similar fashion I don't think you should be allowed to create more threads than CPUs.
You think openMP is smart enough to override your commands?

matsievskiysv:
It depends on implementation. Here's a pseudocode from the [standard](https://www.openmp.org/spec-html/5.1/openmpsu40.html#x60-600002.6.1):
```
else if (dyn-var = true) and (ThreadsRequested ≤ThreadsAvailable)

then 1 ≤ number of threads ≤ThreadsRequested;

else if (dyn-var = true) and (ThreadsRequested > ThreadsAvailable)

then 1 ≤ number of threads ≤ThreadsAvailable;

else if (dyn-var = false) and (ThreadsRequested ≤ThreadsAvailable)

then number of threads = ThreadsRequested;

else if (dyn-var = false) and (ThreadsRequested > ThreadsAvailable)

then behavior is implementation defined; 
```
If dynamic thread adjustment is on, standard requires implementation to spawn no more then `ThreadsRequested` (has a value of `OMP_NUM_THREADS`). The specific number is implementation defined.

lfarv:
Basic OpenMP merged, the discussion may be continued in #222

PR #224, opened 2021-02-11
Fix some integrators for OpenMP

lfarv:
Though the integrators have included for long the OpenMP directives, some were wrong, possibly due to evolution of the code since they were introduced. As a consequence, some gave wrong results when parallelised. The correction consists in moving some variables inside the parallel block so that they are considered "private".


lfarv:
Only the integrators identified as failing were modified.
For safety, the OpenMP default scope was changed from `default(shared)` to `default(none)` which requires an explicit specification of all loop variables. In the future, this should be done for all integrators.

lfarv:
It's fixing a bug potentially giving wrong results with OpenMP, so I merge now

PR #226, opened 2021-02-21
OpenMP thread control

lfarv:
This implements the control of the number of threads used by OpenMP in AT.
It is based on the standard control of default values in AT:

- **Matlab**: `setoption('omp_num_threads', value)` / `getoption('omp_num_threads')`
- **Python**: `at.DConstant.omp_num_threads`

The `omp_num_threads` constant defines the maximum number of threads used in `atpass`, with a special value `0` meaning: leave the default behaviour of OpenMP unchanged (no interaction with the OpenMP library). If it is non-zero, `atpass` will take the minimum of (number_of_particles, number_of processors, `omp_num_threads`) to perform the tracking.

This default value is initialised at AT startup with the environment variable `OMP_NUM_THREADS` if it is defined, or at zero. It may be changed at any time, affecting the following invocations of `atpass`.
In addition, `ringpass`, `linepass` (Matlab) now have a new keyword argument `omp_num_threads` overloading the default value for a specific call.

In summary:

- The default behaviour is to leave the OpenMP behaviour untouched unless AT is started with the environment variable `OMP_NUM_THREADS` defined,

- using `setoption('omp_num_threads', n)` (Matlab) or `at.DConstant.omp_num_threads=n` (Python) will affect all the following trackings,

- The behaviour can be modified for a single tracking with a new keyword argument in `linepass` and `ringpass`.


pyat/at/physics/diff_constants.py
```diff
@@ -9,6 +9,7 @@ class _Dst(object):
     DPStep = 3.e-6           # Momentum step for dispersion and chromaticity
     OrbConvergence = 1.e-12  # Convergence criterion for orbit
     OrbMaxIter = 20          # Max. number of iterations for orbit
+    omp_num_threads = int(os.environ.get('ORM_NUM_THREADS',str(0)))
```
matsievskiysv, line 12: better change `str(0)` to `'0'`

matsievskiysv:
Why is OpenMP not enabled by default? It seems that it is [supported](https://en.wikipedia.org/wiki/OpenMP#Implementations) by the most wildly used compilers.

pyat/at/physics/diff_constants.py
lfarv: Of course!

lfarv:
@matsievskiysv: OpenMP is indeed supported by most compilers, but I would wait for more experience with it. For instance here are results from simply running `pytest`:

Matlab is using an OpenMP library provided by the Intel compiler. To avoid library conflicts, `atmexall` links AT with the Matlab library `libiomp5`. There is no such constraint for pyat unless if we need to run pyat inside Matlab, but if Matlab is installed, we can also use its libraries. So I tested both.

Most tests performed by `pytest` use less that 10 particles and therefore are not parallelised. Here is a summary of the elapsed time printed by `$ pytest test` in default configuration, OMP_PARTICLE_THRESHOLD=10, OpenMP default behaviour:
|   | iMac (4 procs) | Linux (raki, 16 procs) | Linux (grappa 128 procs) |
| - | - | - | - |
| OpenMP disabled | 3.3 s | 4.1 s | 3.1 s |
| Standard library (libgomp) | 3.3 s | 15.5 s | 8.7 s |
| Matlab library (libiomp5) | 3.6 s | 6.5 s | 5.4 s |

These tests are representative of a typical use of AT. Though for a large number of particles OpenMP is obviously beneficial, it may introduce some annoying overhead for many computations. So I would be careful before putting OpenMP in standard, leave users test it in their conditions and hopefully give some feedback.

matsievskiysv:
I guess, integrator implementations are not very OpenMP friendly. Most of the OpenMP implementations use [thread pools](https://stackoverflow.com/a/12617270/5745120) in order to mitigate thread creation overhead. Integrators are compiled into shared libraries and it's not obvious how these pools behave in this case.

swhite2401:
Hello @lfarv, sorry I was away last week. This looks good to me and I agree openMP should not be the default.
Is there a need for benchmarking on our side? I must admit that I personally don't have much time for this right now...

As far as I am concerned, if you are satisfied with the implementation it is ok to merge.

PR #229, opened 2021-02-26
[MPI] add AT parameter sweep script

matsievskiysv:
There's a common pattern to spawn multiple instances of AT on computational clusters for parameter sweep. I've heard that people usually do this using script interfacing with [slurm](https://slurm.schedmd.com/overview.html) itself. An easier solution is to use MPI. I think it would make sense to maintain a generic sweep script in this repo.

This PR adds rather trivial script `mpi_sweep.py` that allows distributing processing jobs across computational cluster. 
`mpi_sweep_example.py` is a simple example of its use. It runs simulations and dumps data into `output.csv` file.

Similar function may be created for Octave using [mpi plugin](https://octave.sourceforge.io/mpi/), I'll try to write it a bit later.
Matlab probably has something similar.

### DIFF
--- a/utils/mpi_sweep/README.md
+++ b/utils/mpi_sweep/README.md
@@ -1,0 +1,6 @@
+`mpi_sweep.py` script allows parameter sweeping on computational clusters via MPI.
+
+You may run example script locally `mpi_sweep_example.py`:
+```bash
+mpirun --host localhost:$(nproc) -n 3 ./mpi_sweep_example.py
+```
--- a/utils/mpi_sweep/mpi_sweep.py
+++ b/utils/mpi_sweep/mpi_sweep.py
@@ -1,0 +1,28 @@
+from mpi4py import MPI
+import numpy
+
+def mpi_sweep(generator, worker, collector):
+    """
+    Evaluate function on multiple MPI nodes and collect the result.
+    generator: function returning list of worker arguments
+    worker: worker function
+    collector: function collecting calculation results
+    """
+    comm = MPI.COMM_WORLD
+    size = comm.Get_size()
+    rank = comm.Get_rank()
+    name = MPI.Get_processor_name()
+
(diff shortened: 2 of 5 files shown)


matsievskiysv:
Added Octave script. For some reason [mpi plugin](https://octave.sourceforge.io/mpi/) does not provide `scatter`/`gather` functions, so messages had to be collected by hand.

lfarv:
I did not find any mention of MPI in Matlab documentation. However Matlab has an similar functionality in the "Parallel Computing Toolbox". Using instructions like "`parfor`" or "`parfeval`", it is possible to distribute the processing on a computational cluster.
I will try to reproduce your example using this, but it may take some time. So when your python and Octave examples are ready, feel free to merge them without waiting for Matlab.
Otherwise, anybody having experience with Matlab and MPI is welcome. One limitation of the "Parallel Computing Toolbox" is that it allows to use all the processor of a single node, but to connect to several nodes it requires installing another software called "Matlab Parallel Server"…

TeresiaOlsson:
I'm not sure MPI is the best choice if you just want to scan parameters. Usually you need MPI if you want to run many particles or bunches in a simulation where they aren't independent of each other (like instability simulations). In case you want to launch a lot of jobs that are independent, but with different parameters it can be better to submit the jobs as a job array to the cluster to not get stuck in the cluster queue or have to wait for communication between nodes to get the results.

matsievskiysv:
@TeresiaOlsson this script may be run locally on one node, utilizing all the threads. After that, collecting data is a simple matter of concatenating .csv output.
This script allows testing your parameter sweeps locally without slurm before actually running them on the cluster.

TeresiaOlsson:
Ah, okay. Then I understand the reason why you might want to use MPI.

lfarv (approved):

PR #231, opened 2021-02-27
Unified atmexall

lfarv:
`atmexall` is unified for Matlab and Octave to allow a similar control of all options: OpenMP, threshold on particle number, optimisation, …
`atmexall` now also  compiles the integrators (`mexpassmethod` is removed)


matsievskiysv (approved):
Don't know about Matlab, but Octave part seems fine.

swhite2401 (approved):
Ok for me

lfarv:
AT successfully compiles with Octave on Windows 10 !

PR #234, opened 2021-03-09
OpenMP on macOS

lfarv:
The OpenMP option on MacOS requires that `libomp` has been previously installed. This branch adds an error with information on how to do it if this has not been done.

### DIFF
--- a/atmat/atmexall.m
+++ b/atmat/atmexall.m
@@ -215,9 +215,9 @@ warning(oldwarns.state,oldwarns.identifier);
     end
 
     function include=select_omp()
-        if exist('/usr/local/include/omp.h', 'file')
+        if exist('/usr/local/include/omp.h', 'file')        % Homebrew
             include='-I/usr/local/include';
-        elseif exist('/opt/local/include/omp.h', 'file')
+        elseif exist('/opt/local/include/omp.h', 'file')    % MacPorts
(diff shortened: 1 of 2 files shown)


lfarv:
@willrogers: I had to put a dirty check on the python version to allow this on python2.7. If we can drop this version, would you agree to remove the tests with it?

willrogers:
I'm not sure I understand the question.

lfarv:
I think that after the 0.1.0 version we could drop the compatibility with python 2.7. However the tests still prevent that, so that I had to put in the code ugly tests on the python major version. I was just asking about the possibility to remove python 2.7 from the sequence of tests.

lfarv:
Also I had yesterday an unexpected failure of Windows tests (independently of python version). In that case, is there a way to restart the test sequence without pushing anything new?

willrogers:
Once it's agreed we will remove the Python 2.7 tests of course. I just commented on the Discussion about this.

> 
> 
> Also I had yesterday an unexpected failure of Windows tests (independently of python version). In that case, is there a way to restart the test sequence without pushing anything new?

Yes, if you click on the little red cross then select Details there is a button 'Re-run jobs'.

lfarv:
So I understand that we continue supporting python 2.7 for some time…

PR #238, opened 2021-03-22
Optics radiation

swhite2401:
This branch implements a 6D treatment of optics calculation: (much slower!)
-twiss function are obtained by extracting the symplectic part of m66
-a dispersion-like function is computed using the local energy deviation

Few other changes:
-set quadrupole_pass='auto' by default in radiation_on()
-function to adjust the cavities phase (similar to atsetcavity but deals only with TimeLag)
-faster symplectify function avoiding the use of linalg.inv
-corrected some leftover from matching branch: off-momentum optics computed only on request (get_w kwarg)
-function to compute U0 using findorbit6 -> get_energy_loss was considering only bends and wigglers
-function to scale dipole/quadrupole/sextupole with local energy: this was requested by the FCC users

I have not implemented anything in matlab, waiting for your feedback! Below an example usage:
For some reason tapering fails for the HMBA lattice...I suspect something with the DQs

`import at
import os
import numpy
import time
from at.physics import linopt, linopt_rad
from at.lattice import DConstant

#path = '../../at/pyat/test_matlab/'
#filename = 'hmba.mat'
#key = 'RING'

path = './'
filename = 'fcch_norad.mat'
key = 'ring'

#path = '../lattice/'
#filename = 'betamodel.mat'
#key = 'betamodel'

xy_step = 1.0e-9
dp_step = 1.0e-6

latticef = path+filename
ring = at.load.matfile.load_mat(latticef,key=key)
ring = at.Lattice(ring)

get_chrom = True

ring.radiation_off()
t0 = time.time()
l0,q,qp,l = linopt(ring,dp=0,refpts=range(len(ring)),get_chrom=get_chrom,
                   coupled=False, XYStep=xy_step, DPStep=dp_step)
print(time.time()-t0)

ring.radiation_on(quadrupole_pass='auto')
ring.set_cavity_phase()
ring.tapering(niter = 2, quadrupole=True, sextupole=True, XYStep=xy_step, DPStep=dp_step)
t0 = time.time()
l0rt,qrt,qprt,lrt = linopt_rad(ring,refpts=range(len(ring)),get_chrom=get_chrom,
                               coupled=False, XYStep=xy_step, DPStep=dp_step)
print(time.time()-t0)

print(q,qrt)
print(qp,qprt)

spos = ring.get_s_pos(range(len(ring)))

plt.subplot(231)
plt.plot(spos,(lrt.beta[:,0]-l.beta[:,0])/l.beta[:,0])
plt.subplot(232)
plt.plot(spos,(lrt.beta[:,1]-l.beta[:,1])/l.beta[:,1])
plt.subplot(233)
plt.plot(spos,lrt.dispersion[:,0]-l.dispersion[:,0])
plt.subplot(234)
plt.plot(spos,lrt.closed_orbit[:,0])
plt.plot(spos,lrt.closed_orbit[:,2])
plt.subplot(235)
plt.plot(spos,lrt.closed_orbit[:,4])
plt.subplot(236)
plt.plot(spos,lrt.closed_orbit[:,5])
plt.show()`


lfarv:
1. I do not like the change in `radiation_on`: modifying the default behaviour can generate problems with existing code which can be very hard to track. The tests are there to check that all modifications are transparent. If you modify the tests accordingly, you completely break the backward compatibility. If you need quadrupole radiation, just ask for it. This supports my idea the radiation must be turned on/off explicitly by the user, and not "silently" by AT when needed.
 
2. I am **strongly** against the `set_cavity_phase` function (it’s been an old discussion with Nicola!) : `TimeLag` should always be zero for main cavities, and should be used only for harmonic cavities. The AT reference system is such that:
* In transverse, the origin is the theoretical axis of magnetic elements,
* In longitudinal, the origin is the zero-crossing of the RF field.
Modifying the time origin is as if in transverse you would change the origin so that the closed orbit is zero: doing this, you hide the closed orbit information (or synchronous phase), with absolutely no benefit.
Leave the reference system where it should be, and let the beam centred on its closed orbit (transverse and longitudinal). Some may claim that even if the function is there, users are not obliged to use it, it’s better if it is not there. Keep it private if you want it.

3. `symplectify`: only one version should be left. You could take the faster one (through I doubt that for so small matrices it’s a big deal) and make it work for the 3 cases.

4. `get_u0`: nice addition. It will also work for impedance losses. A question about the function names: how to make the difference between `get_u0` and `get_energy_loss` easy to remember?

5. the main point: `linopt_rad`. It’s rather simple and gives results. But is there any theoretical support giving the physical meaning of the pseudo-betas ? The "symplectified"  matrix used for the processing represents a "virtual" lattice quite different from the real one! Without this support, it looks a little bit like "bricolage". Anyway, we need some notice in the doc string to make things more clear. 

6. I think there is a much better alternative using the A-matrix. The help of `at.amat` tells that this matrix represents a change of referential which turns the 1-turn matrix (2D, 4D or 6D) into a set of 1, 2 or 3 rotations (possibly damped). This is exactly what alpha,beta are doing in 2D with a transformation matrix:
```
[[ 1/sqrt(beta),            0]
 [alpha/sqrt(beta), sqrt(beta)]]
```
This is the 2D A-matrix. `at.amat` generalises this to 4D and 6D.
Now if you compute A’ * A, you get a symmetrical matrix with on the diagonal
`[beta1, gamma1, beta2, gamma2, beta3, gamma3]`, `-alpha1` at position [0, 1] or [1, 0] and so on.  So you get numbers similar to what you are looking for, but with a clear definition.
`amat` is ready, we just need to propagate the result to the desired refpoints

swhite2401:
Thanks for the comments! This is obviously a prototype and needs improvements.

1-This was a bit arbitrary I admit, but really accounting only for dipole gives wrong results in many cases, I would tend to think that the default should give the correct results. But it is true that I am breaking backward compatibility in this case...it can be reverted

2-I had the same opinion until I started tracking the FCC lattice: it is unstable with radiations ON if you do not tune the cavity phase properly. I therefore insist that this function is needed, we may add a warning with your comments.

3-I agree, difference is ~30% from my estimates, I will propose a single function

4-The naming is a problem in fact...we could potentially integrate get_u0 in get_energy_loss? But this means flags again...

5&6-In fact I could not find any reference beside the Ohmi paper on emittance but it clearly does not describe this numerical approach, more benchmarking is obviously needed and will be done with ptc_twiss that also has a 6D treatment (waiting on the EPFL colleagues for this part). Using the amat is a very good alternative, I will look into it, nevertheless I think it works only in the low coupling approximation right?

TeresiaOlsson:
I think a function to set the cavity phase is a good idea. If you don't set a phase on the main cavity it is in my experience much more complicated to figure out how to set the phase of the harmonic cavity correctly to get the RF bucket you want. It has annoyed me that I have to use atsetcavity to do it, when I just want to set the TimeLag.

swhite2401:
Maybe I was a bit unclear: set_cavity_phase() (I agree the name is misleading) is just a convenience function to set the origin of the synchrotron motion based on energy loss per turn, in other words hiding the synchronous phase as @lfarv said.
It allows to avoid 6D orbit convergence errors without any further actions such as defining an initial guess and to generate beam distribution centered around 0 for tracking without having to face transient oscillations for example.
There are other ways to handle this, for instance we may think of improving find_orbit6 and atbeam to automatically take into account U0.

For harmonic cavities, you do not need to use atsetcavity(), you can just directly set TimeLag like any other attribute.

lfarv:
@swhite2401. For point 2.: I cannot understand the problem: changing the time origin cannot stabilise an unstable machine ! The problem must lie in find_orbit6, where the initial guess of 0 may be unstable. Introducing the synchronous phase in the initial guess (with get_u0?) of find_orbit6 seems a cleaner solution, which costs only one `lattice_pass`.
Same for @TeresiaOlsson: why a shift of origin would make things simpler ?
Anyway, I am just trying to encourage good practice, but if you want the function…

For points 5&6: I don't think `amat` is limited to low coupling: It's a pure mathematical analysis of a 6x6 matrix. What may happen is that you cannot easily associate one motion to one plane. But it has to be checked. Otherwise, Simon, do you know what happens in `ptc_twiss`?

Sorry @swhite2401, you new comment appeared just now, it seems that we agree !

lfarv:
Also for the generation of particle populations: you generate it around the transverse closed orbit, right? So why longitudinal should be zero?

TeresiaOlsson:
@lfarv Because many formulas for harmonic cavities (and main cavities with beam loading) are derived using a convention with a phase offset relative to the synchronous phase (or main cavity phase). So in case you don't set the phase of the main cavity you have to shift the phase of your harmonic cavity correctly to account for that so the relative phase between main and harmonic cavity becomes correct. So it doesn't really have to do with AT, but rather which conventions other people have used and what makes it easier to translate that into AT.

@swhite2401 Sure, for harmonic cavities I set the TimeLag directly. I only use atsetcavity for the main cavity to avoid having to calculate what it should be myself from the radiation loss and remember the sign convention. By setting it I avoid having to shift my input beam or track more turns for the beam to find the longitudinal equilibrium. It also makes it easier to compare output or reconstruct RF buckets found by other codes such as mbtrack and Elegant because they both use a non-zero phase for the main cavity. If I set the same phase in AT as they use I avoid having to recalculate all results to the same phase reference afterwards.

swhite2401:
No reasons besides what Teresia mentions

lfarv:
@swhite2401 (and others) : a few minor comments on your example file:
In the imports section, you do not need to specify at sub-packages (though you may, if you want):
```
from at import linopt, linopt_rad
from at import DConstant
```
`load_mat` already returns a `Lattice` object, you do not need an additional copy:
```
ring = at.load_mat(latticef, key=key)  # is already a Lattice object
```
And even simpler: the generic `at.load_lattice` selects the loader as a function of the file extension:
```
```
would load `.mat`  as Matlab binaries, ` .m` files as Matlab text, `.lte` files as Elegant, `.lat` files as Tracy input and `.repr` files as AT python-native files. `help(at.load_lattice)` displays the up-to-date list of available formats. There may be compatibility problems, please report!

lfarv:
@swhite2401 and @TeresiaOlsson : the question of cavity phase is somewhat decoupled from the present topic of 6D optics. I propose that the `set_cavity_phase` function is moved to another dedicated pull request. I would then add a similar function in Matlab so that @TeresiaOlsson you would not have to use `atsetcavity`. With a warning in these functions to discourage their use.

lfarv:
In `get_U0`, the cavity PassMethod should be `IdentityPass` if the cavity length is 0 and `DriftPass` otherwise. A possiblity would be to require radiation **ON** to call the function (the user is responsible for the options), and then call
`radiation_off(dipole_pass=None, quadrupole_pass=None,...)` to turn off only the cavities. `radiation_off` will take care of cavity lengths.
Or use `DriftPass` which can also handle length=0. Less efficiently, but who cares here…

swhite2401:
I agree, let's keep things separate. I propose to have separate branches for get_U0() and set_cavity_phase() for which it seems we agree on and keep the 6D optics open on here because it needs quite some iterations still

swhite2401:
@lfarv , right! I did not think about this, I like you proposal of using radiation_off better

lfarv:
Hello @swhite2401: I feel that there are too many different and unrelated things in this pull request. May I suggest to:
- remove `pyFuncPass.c` (part of another pull request),
- revert `elements.py`, `lattice_object.py` and `test_physics.py` to their initial state (different topic),
- cleanup the "simplectify" things in `amat.c`,

so that we keep only the important parts in `amat.py` and `linear.py`?
Concerning `radiation.py`, `get_energy_loss` and `set_cavity_phase` are treated in #242, tapering stays here, it will have to be merged later.

This could help reviewing a lot…

swhite2401:
Hello @lfarv, I fully agree but I needed to put all these developments together such that they are available for our CERN colleagues to do some benchmarking with ptc/madx/bmad

I have already put the `PyFuncPass.c` and `elements.py` on a other branch and as you mentioned, `set_cavity()` and `get_U0` are treated elsewhere.
I also had in mind to create a separate branch for the tapering.

Can we keep this branch alive for benchmarking purpose, and update it with the master as the separate developments are merged?

In this case I would propose to create a new branch to introduce the amat development and maybe linopt_rad with the celan symplectify, what do you think?

lfarv:
@swhite2401: It's clear like this, this branch is for tests and we keep it as Draft. Feel free to open other branches for features intended to be merged.

swhite2401:
All the features from the branches are integrated in the master, I delete it

PR #240, opened 2021-03-24
find_m66 with radiation off

lfarv:
`find_m66` should work with or without radiation. The reference orbit is now chosen depending on the presence of a RF cavity: if there is at least one cavity, orbit6 is used. Otherwise, orbit4 is used.

### DIFF
--- a/atmat/atphysics/LinearOptics/findm66.m
+++ b/atmat/atphysics/LinearOptics/findm66.m
@@ -47,7 +47,12 @@ if length(varargs) >= 2	% FINDM66(RING,REFPTS,ORBITIN)
 end
 
 if  isempty(R0)
-    R0 = findorbit6(LATTICE,'XYStep',XYStep,'DPStep',DPStep);
+    cavities = atgetcells(LATTICE, 'PassMethod', @(elem,pass) endsWith(pass, 'CavityPass'));
+    if any(cavities)
+        R0 = findorbit6(LATTICE,'XYStep',XYStep,'DPStep',DPStep);
+    else
+        [~, R0] = findorbit4(LATTICE,0.0,'XYStep',XYStep);
+    end
 end
 
 if length(varargs) >= 1	% FINDM66(RING,REFPTS)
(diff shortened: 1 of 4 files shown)


lfarv:
The same treatment is applied to `atnuampl`

swhite2401:
This looks good and will be very useful to simplify the implementation of the 6D treatment of linopt.
In fact, I did think about a similar improvement but at the level of find_orbit6, maybe this could be considered as well?

swhite2401:
What if the radiations are ON without cavities?

bnash:
> This looks good and will be very useful to simplify the implementation of the 6D treatment of linopt.
> In fact, I did think about a similar improvement but at the level of find_orbit6, maybe this could be considered as well?

Hi @swhite2401, for a 6-D linopt, you may be interested in these functions to compute the "etaoids" and "betaoids" I wrote based on [E. Forest's 1998 paper](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.58.2481).

https://github.com/atcollab/at/blob/master/atmat/atphysics/LinearOptics/find_betaoids.m
https://github.com/atcollab/at/blob/master/atmat/atphysics/LinearOptics/find_etaoids.m

Meanwhile, I've been lurking and following the conversation at times. Happy to see how well all the development of pyat is going!

lfarv:
@swhite2401: I confirm the test on cavities: that's the only condition to get a 6D closed orbit (with or without radiation). Otherwise, if there is no radiation, there is a 4D closed orbit, and with radiation, no possible closed orbit and you get an error. However, you still have the possibility to explicitly provide the initial orbit and get m66 around a non-closed trajectory.

By the way, all this is in anticipation of a more strict check of radiation in Matlab

lfarv:
@bnash : Hi Boaz, nice to hear from you. What you point out is exactly what I want to push. Thanks for the reference !

swhite2401:
Yes in fact I had the trajectory in mind, which I find useful, isn't this cavity check discarding this option?
@bnash in fact go to hear from you and to see you are still active on the field! Indeed what you propose looks like the proper approach, thanks a lot!

lfarv:
I moved the test back to radiation check rather tan cavity check. The result is the same: failure if there is radiation but no cavity, but the error message is better: "singular matrix" in find_orbit6, rather than 'Radiation must be OFF" in find_orbit4.

PR #242, opened 2021-03-25
Set cavity phase

lfarv:
New features:

- The existing functions `get_energy_loss` (Python) and `atgetU0` (Matlab) have now an option **'method'** with 2 possible values:
  **'integral'**: Default value, it' s the previous computation based on the I2 integral. It works with or without radiation, but only takes into account dipoles and wigglers,
  **'tracking'**: new method using tracking over one turn without cavities. It works only with radiation ON,  but takes into account all radiating elements.
- New functions `set_cavity_phase` (Python) and `atSetCavityPhase` (Matlab). They modify the time reference by setting the `TimeLag` attribute of the cavities. Though not recommended, this may be convenient in some cases. As they use the energy loss per turn, they have the **'method'** option described above, and also a **'refpts'** option, giving the location of the cavities. This allows to ignore harmonic cavities if present. The default behaviour is to affect all cavities.
As the use of these functions is strongly discouraged, they emit a warning.
In Matlab, `atsetcavity` is modified to call both atgetU0 and atSetCavityPhase if necessary.


lfarv:
Modified the description of the pull request, ready for merging.

PR #243, opened 2021-03-29
Pypassmethods

swhite2401:
Hello @lfarv @willrogers , I have written this code to be able to call python passmethods.
For some reasons the windows python 2.7 build keeps failed when trying to compile PyFuncPass.c

I did build this branch from optics_radiations by mistake, the relevant changes here are in elements.py, and PyFuncPass.c
I would need your help with this compilation errors, on my side (linux) everything runs and compiles fine.
I can provide some example usage if needed.

Btw, are we still supporting python2?

Thanks in advance for your help!


willrogers:
We have a pull request open to remove support for Python 2.7. When that is merged this won't be a problem.

swhite2401:
Ah right very good!
If you have time, you opinion on PyFuncPass.c would be very helpful, I am really not very familiar with this Numpy C-API

willrogers:
We should wait until the Python 2 branch is merged first, and clean up this pull request so the changes are clear.

@lfarv you said you had some things that we can remove before merging #236.

swhite2401:
Ok yes I agree, meanwhile I have cleaned up the branch to keep only the changes related to python passmethods

lfarv:
@willrogers: I am also waiting for #236 to be merged! There are many small tricks for python 2.7 which are no more necessary, but it's not urgent to remove them. But I may take benefit of this for the pending pull requests. So please go on.

swhite2401:
master was merged to remove python2 tests, all have passed.
The compilation of at.c was giving waring "unused variable" for maxthread, it is now declared only when OpenMP is activated
I think this one is ready for review

lfarv:
This looks more complicated than the Matlab version: you need
- a specific element class: the Python pass method cannot be used for other classes
- 3 names to define the passmethod: "PyFuncPass", module name, function name.

Is not it possible to use the same strategy as in Matlab ?

At the beginning of at.c, we scan the lattice to load the pass methods. At the moment, we look for a compiled shared library (something like `"PassMethod.cpython-37m-darwin.so"`), load it and put its pointer in a table. If this fails, one could look for a python module (`"PassMethod.py"`), import it and put it in another table. The entry point in C integrators is fixed (`"trackFunction"`), one could decide that the function in the python module is also fixed (`"trackFunction"` also?).

The advantages are:
- the search for the Python function is done once at the beginning of `at.c`, and not each time the integrator is called (and only for a new lattice),
- The passmethod name fully identifies the integrator, whether C or Python. Of course we can change the search order so that Python integrators have precedence over C ones,
- the Python integrator may be applied to any element (no special class).

I did not look in detail if it can work, but I do not see any big problem.

swhite2401:
Thanks for your comments, I did not look at the matlab way in fact.
You do have a point and we can explore that solution, I think this solution allows to do everything you mention, it is just a bit more complicated.
It made it like this not to interfere with `at.c`,  but in fact you are right nothing prevents us to change it.

My personal feeling is that C should always have the priority, I will look into what you propose and see what is possible.
My only question is how in practice the C array will be handle by the python `trackfunction`, presently all conversions are handled into `PyFuncPass.c` which is convenient because the python `trackfunction` can be called either with `atpass`  or a standalone python code for testing without any modifications.

I will think about it and keep you updated.

lfarv:
Concerning the conversions, which are common for any Python passmethod, this part become part of `at.c`, when calling a Python passmethod:
```
PyObject *args, *kwargs;
kwargs = Buildkwargs(ElemData);
args = Buildargs(r_in, num_particles);
PyObject_Call(function, args, kwargs);
```
with `function` taken from a linked list (`integrator_list`) built at the beginning of `at.c` and associating each element of the ring with its `function`. It has to be modified to deal with both C and python pointers.
The integrator call itself (lines 316-318 of `at.c`) must be different for C and Python integrators: see lines 441-449 of `atpass.c` for the Matlab version.

Concerning the search order, you could imagine testing quickly a Python passmethod overloading the C one before modifying the C… Anyway, usually we should use different names.

swhite2401:
Ok thanks! I will come back with a new proposal

swhite2401:
@lfarv you will find a new proposal where everything is integrated in at.c, for now I kept the "old" version too.
Let me know what you think, I think I am properly de-referencing all objects but these things can be tricky sometimes, would you have a good way to check for memory leaks?

Other thing, in the process I notice that the C code related to the diffusion matrix is still contained in `atmat/...` would it make sense to have a common unique place for all C integrators? Also `findmpoleraddiffmatrix.c` has copies of the drift kick implementation present in `driftkick.c` , it seems that simple a `include` could avoid duplicating this part of the code.

lfarv:
@swhite2401 : Looks good, but could you put a sample python integrator in the `atintegrators` directory ? Something like a `pyDriftPass.py` similar to `mDriftPass.m` for Matlab.

Concerning `findmpoleraddiffmatrix.c`, it's not in integrator though it contains one. So it does not need the special processing applied to all the C-files of `atintegrators`. And you are right, it deserves some cleaning up and even more. It contains a copy of an old version on Str/BndMPoleSymplectic4Pass, augmented with diffusion stuff. The 1st point would be to update it with the last version of BndMPoleSymplectic4Pass (with the new options for edges), while limiting the code duplication. Then it could be nice to be able to add new elements like wigglers. Turn diffusion in a modular style, like it is for integrators. But that's another story…

swhite2401:
Ok, I will do. We may keep the clean-up for a later development.

swhite2401:
@lfarv , @willrogers , I have added an example + test as requested (IdentityPass) in `at/pyat/test.`
I think this one is ready to be merged, any final comments?

willrogers:
The Git history is really complex now, so I made #253 to clean it up.

swhite2401:
I delete this one that is now covered in #253

swhite2401:
Covered in #253

Issue #235, opened 2021-03-15
Bug in ringpara for momentum compaction factor


carmignani:
Hello,

ringpara does not compute correctly the momentum compaction factor for the ESRF storage ring lattice. It could be due to the presence of the dipole-quadrupoles, because the computation is correct for the old ESRF lattice, but I'm not sure.
ringpara computes the momentum compaction factor with this formula:
`alphac = I1/circ;`
An easy fix would be to use the function mcf, unless we want to keep using only formulas inside ringpara.

The momentum compaction of ringpara is used in the function atBunchLength.

lfarv:
@carmignani: `alphac` is indeed `I1 /circ`. So we can suspect a bug in the computation of `I1`.
I cannot reproduce the problem with the `'hmba'` test lattice, which has dipole-quadrupoles.
So can you make sure you are using the up-to-date version of AT and post the lattice which causes the problem? Thanks!

carmignani:
Hello @lfarv,
I don't have the hmba test lattice, but I did some more tests and it looks like the problem is not at the dipole-quadrupoles, but at the RF cavity pass methods.
I tried with a few lattices I had and in all cases the momentum compaction factor computed with ringpara is different than the one of the mcf function if the cavity pass method is not IdentityPass. With atradoff the two results are the same.

swhite2401:
Hell @lfarv @carmignani , I believe this relates to a more generic problem in AT: optics calculations are not valid for 6D motion. Same goes for mcf that uses findorbit4.
I am working on this problem and will have a solution to propose most likely by the end of the week. Meanwhile I suggest (as advised by @lfarv in the past) to turn OFF radiations and cavities for this kind of calculations.

lfarv:
@swhite2401, @carmignani : 
There is no "generic problem in AT" ! The momentum compaction factor is a **lattice** ("geometric") property, which CANNOT be computed in 6D ! The dispersion is the trajectory of a particle with CONSTANT off-momentum, how can you compute it if you encounter a cavity or radiation? What can be done in other codes is to **ignore** cavities when computing beta functions and dispersion. AT makes the things clear: you have to do it explicitly, which is for me the best approach.
So don't loose time in trying to find a solution to a non-existing problem: I'll immediately add the test which will prevent from producing wrong results. You have been warned !

@carmignani: the hmba test lattice is part of the distribution, with a few other ones, in the `at/pyat/test_matlab` folder.

swhite2401:
[...]

PR #244, opened 2021-03-31
Prepare the radiation check in Matlab

lfarv:
To avoid problems when calling linear optics functions with radiation while they must be called without (see #235), a test will be implemented in the future to prevent that, as it is done for Python. However it appears that a number of functions within AT are violating this rule (even `atdiag.m`)
In anticipation for these test, this pull request correct a number of these functions, mainly demos. There may be still be other ones…

At this point:
- **no check is implemented yet**,
- no behaviour is changed,
- It consists mainly in applying `atradoff` to the test lattices when necessary. Only `testTrackBBR.m` has been more deeply reorganised.


swhite2401:
Hello @lfarv, I support this development. On my side I see no problem changing demos, however, although fully justified, this change is quite severe and will impact a lot of past developments.

I would therefore propose to not deploy the check_radiation() before we have and alternative solution for computing optics in 6D and potentially start with a deprecation warning for a few months to give time to users to adapt their scripts and have a smooth transition.

I let @simoneliuzzo and @carmignani  comment on this because they will (hypothetically) suffer much more than me.

lfarv:
@swhite2401 : I understand your concern, and I agree with your strategy for the deployment of `check_radiation()`. However we are not yet at that point: for the moment, I started correcting some internal problems of AT (the ones I found). This is not an exhaustive list, and I fear some will remain. But it is necessary before applying the check.

As I mentioned in the description, no check is applied anywhere yet, and there is a bunch of other corrections to do.

PR #245, opened 2021-04-01
Improvement of the initial orbit guess in find_orbit6

lfarv:
The initial orbit guess on `findorbit6` (Matlab) and `find_orbit6` (Python) can be easily improved by using the energy loss per turn to derive the time lag. By default, this energy loss is computing with tracking to ensure that the right radiating elements are taken into account.
This improvement should eliminate one of the bad reasons to modify the TimeLag attribute of the main cavities.

Additionally this revealed a naughty Matlab crash occurring when tracking complex trajectories. This is now properly rejected.


swhite2401:
Very nice development! For the lattice I have tested the orbit search converges without issues.
I will add the tapering option that is ready for review to your new module once it is merged.

swhite2401 (approved):

PR #246, opened 2021-04-07
Upgrade of the Lattice object

lfarv:
#### New method and decorator
In preparation for a "copy" keyword in all functions functions modifying the lattice, the `Lattice` object has a new method:
`Lattice.replace(refpts)`
which returns a shallow copy of the lattice where the selected elements have been "deep copied" so that their modification will not affect the original lattice. Non-selected elements are common to both lattices.
Associated with that, a new decorator `make_copy(copy)` can be used to decorate any function like `func(ring,refpts,*args)` modifying the lattice so that it handles the copy keyword. See `set_value_refpts()` for an example.
As of today, the "copy" keyword is implemented on `Lattice.radiation_on()/off()` and `set_value_refpts()`. A series functions controlling the RF cavities is in preparation.

#### Upgraded methods
 A set of new or existing methods using `refpts` now accepts a 3rd way of specifying the points of interest:
1. integer or sequence of integer (sequence of indices)
2. sequence of booleans (logical mask)
3. filter function:  func(elem) returning True for selected elements. Such functions can be generated by the existing checktype, checkattr and check name functions.

`Lattice.select(refpts)`:  returns an iterator over the selected elements
_Example: `qp=ring.select(checktype(Quadrupole)` returns an iterator over all quadrupoles_
`Lattice.refcount(refpts)`: returns the number of selected elements
`Lattice.boolrefpts(refpts)`: conversion to a list of indices
`Lattice.uint32refpts(refpts)`: conversion to a logical mask
`Lattice.replace(refpts`): returns a new lattice with selected elements copied
'Lattice.get_s_pos(refpts): returns the location of selected elements
_Example: `s=ring.get_s_pos(checkname('MK*'))` returns the location of elements with name starting with MK_

This is also also valid for `get_value_refpts` and `set_value_refpts` `refpts_iterator`

### DIFF
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -185,12 +185,19 @@ class Lattice(list):
         return uint32_refs(refpts, len(self))
 
     def bool_refpts(self, refpts):
-        """Return a boolean numpy array of length n_elements + 1 where True elements
-        are selected.
+        """Return a boolean numpy array of length n_elements + 1 where
+        True elements are selected.
         """
         if callable(refpts):
             refpts = [refpts(el) for el in self]
         return bool_refs(refpts, len(self))
+
+    def rotate(self, n):
+        """Return a new lattice rotated left by n elements"""
+        if len(self) == 0:
+            return self.copy()
+        n = n % len(self)      # works for n<0
+        return self[n:] + self[:n]
 
     def update(self, *args, **kwargs):
         """Update the element attributes with the given arguments
@@ -335,9 +342,9 @@ class Lattice(list):
             None if copy == False
 
         KEYWORDS
-            copy=True           Return a shallow copy of the lattice. Only the
-                                modified attributes are replaced. Otherwise
-                                modify the lattice in-place.
+            copy=True   If True, return a shallow copy of the lattice. Only the
+                        modified elements are copied.
+                        If False, the modification is done in-place
         """
 
         def lattice_modify():
@@ -412,9 +419,12 @@ class Lattice(list):
             dipole_pass='auto'          PassMethod set on dipoles
             quadrupole_pass=None        PassMethod set on quadrupoles
             wiggler_pass='auto'         PassMethod set on wigglers
-            copy=False                  Return a shallow copy of the lattice and
-                                        replace only the modified attributes
-                                        Otherwise modify the lattice in-place.
+            copy=False  If False, the modification is done in-place,
+                        If True, return a shallow copy of the lattice. Only the
+                        radiating elements are copied with PassMethod modified.
+                        CAUTION: a shallow copy means that all non-radiating
+                        elements are shared with the original lattice.
+                        Any further modification will affect in both lattices.
 
             For PassMethod names, the convention is:
                 None            no change
@@ -464,9 +474,12 @@ class Lattice(list):
             dipole_pass='auto'          PassMethod set on dipoles
             quadrupole_pass=None        PassMethod set on quadrupoles
             wiggler_pass='auto'         PassMethod set on wigglers
-            copy=False                  Return a shallow copy of the lattice and
-                                        replace only the modified attributes
-                                        Otherwise modify the lattice in-place.
+            copy=False  If False, the modification is done in-place,
+                        If True, return a shallow copy of the lattice. Only the
+                        radiating elements are copied with PassMethod modified.
+                        CAUTION: a shallow copy means that all non-radiating
+                        elements are shared with the original lattice.
+                        Any further modification will affect in both lattices.
 
             For PassMethod names, the convention is:
                 None            no change
@@ -670,4 +683,4 @@ Lattice.bool_refpts = _bool_refs
 Lattice.get_elements = get_elements
 Lattice.get_s_pos = get_s_pos
 Lattice.select = refpts_iterator
-Lattice.count = refpts_len
+Lattice.refcount = refpts_len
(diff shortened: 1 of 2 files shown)


swhite2401:
I find this proposal very useful, for non python experts the difference between copy and deepcopy is not obvious and the use of ring.copy() has been a source of problems where a user though he was modifying only the copy while both objects were affected. These errors can be very difficult to track.
I was tricked a couple times by this in fact.

swhite2401:
I tempted to propose that `copy=True` always return a `deepcopy` actually. This would remove any ambiguity:
-if False: no copy made, you are modifying the original lattice
-if True: deepcopied, the original lattice is intact

A shallow copy of the Lattice object is not so useful in the end since most time object in the `deeper` level are modified

lfarv:
Added the "callable refpts" option in get_s_pos (description edited)

matsievskiysv (approved):

lfarv:
It's true that the notion of "deepness" of a copy may seem awkward. But it's inherent to the language, and not specific to lattices. For any object, b=a is not a copy of a, it's a new name for the same object. At some point, any python user will be confronted to that.

So, to support the choice of a shallow copy:

- it's much more efficient in terms of processor and memory
- it's adapted to the functionality provided by the proposed functions: when turning radiation on/off, you want the same lattice, there is no reason to modify anything :else in the copy. Similarly, when changing the RF frequency in linopt_rad to look at the dispersion: you just want to tune the cavity frequency up and down, at the smallest cost.
- The functionality provided by the `make_copy` decorator is the lightest possible, but rather difficult to implement for a user. On the other hand, creating a deep copy of a lattice is obvious and available in a single and simple command:
`ring_b = ring_a.deepcopy()`
So the idea is: if you just rely on the provided functions to modify a lattice, they are safe, you can use them. If you want to manually tune parameters and modify the attributes of elements, you create a deep copy and work on it.

In the future, one could provide more control by allowing more values to the '`copy`' keyword, rather than using a boolean. But it could also look more obscure…

swhite2401:
I agree this is inherent to python, and utility functions like `set_value_refpts` should not be allowed to generate deepcopies because this is clearly inefficient.

However one may think that a copy option in an upper level function, such as `radiation_on()` returns an independent object, I would suggest to put a warning in the help: "copy=True will generate a shallow copy, object not modified by this function will share references with the original ring. In order to create an independent new ring please use ring.deepcopy()' or similar to guide the user.

Also, as an example of failed scenario, calling `radiation_on(copy=True)` and then adjusting the tune which is standard practice, will result in different quadrupole gradients in both the parent and child objects, is that correct?

For your proposal to be entirely safe we would have to enforce using at functions with `copy=True` in all other AT functions changing attributes right? I don't think it is the case for now

lfarv:
@swhite2401:

1. OK, I'll be more explicit in the doc string concerning shallow copies!

2. About the radiation ON: I would rather call it a successful scenario: radiation on and off represent the same lattice, it makes no sense to change the tune in one and not in the other ! Unfortunately, at the moment this success is partial, since if you enable quadrupoles, they will be copied and therefore independent. I'll make a shallow copy of the involved elements (at the moment it's deep) so that only the `PassMethod` is independent. Then the lattice with and without radiation will be fully synchronised.

3. It's indeed planned to have all high-level functions modifying the lattice to offer the `copy` keyword. But the default value must be the same (`False` at the moment) for all functions. When the Lattice object is modified, we can go on.

lfarv:
After checking, `radiation_on(copy=True)` does make a shallow copy of the radiating elements, so the two rings are synchronised, provided that the change of tune is done with copy=False (`copy` keyword not yet implemented).

swhite2401:
Concerning 2. I am not sure I completely get your point, synchrotron radiations will drive the beam on a different closed orbit and shift the tune, it is not possible to have the same tune in both cases if the two lattices are synchronized.
On the contrary, I would think that for comparison purposes one would then to want the same tune for both case and by definition different quadrupole settings.

Anyway, for me it is fine to merge as long as the help clearly describes the expected behavior.

`fit_tune()` and `fit_chrom()` functions and matching in general will need to be modified with the `copy` keyword too in order to be coherent with other functions

lfarv:
Difficult question: you want to compare two lattices. Different lattices since magnets are different. The "electron" tune of one is equal to the "optical" tune of the other one. I do not know if the comparison makes sense, but anyway they are different and so deserve a deep copy. However, strictly speaking, the optical functions of the "retuned" lattice are not the original ones, they are given by turning off radiation on the modified lattice. Hence the motivation for synchronising with a shallow copy.

To summarise, the goal of the `copy` option is to get efficiently a **temporary** modification of a lattice, without further modifications for which the scope is questionable. In particular, this applies to RF frequency changes, or cavity `TimeLag` changes, which do not modify the basic lattice. For **persistent** modifications, I would advise either modifying in-place, or on a deep copy. And this confirms for me the choice of the `False` default option.

Looking now at the tapering you are implementing: the optical function of the tapered lattice will be distorted: they depend only on the magnetic lattice, not on the particle. They may not be relevant in such a case, but that's how they are… So I agree with the need for another formalism in such cases. And for these reasons, I do not think the copy keyword is useful for tapering: if the user needs to keep track of the original lattice, he is obliged to make a deep copy.

lfarv:
- Improved documentation for the `copy` keyword,
- new `Lattice.rotate` method,
- the new `Lattice.count` method is renamed as `Lattice.refcount` to avoid overloading the list.count method.

swhite2401:
Ok , I get your point and I agree with what you propose.
In fact I have completely removed the copy option in the tapering because I thought it did not make much sense.
Similarly, we may keep the `fit_tune()` and `fit_chrom()` function with the copy flag -> in-place modification only.

However, the `matching.match()` return a shallow copy with deepcopied modified elements by default: matching might be considered persistent no?

lfarv:
So I merge.

PR #247, opened 2021-04-08
included tapering function in radiation.py

swhite2401:
This function allows to rescale dipole/quadrupol/sextupole fields to cancel the effects of synchrotron radiations on closed orbit and linear optics.

### DIFF
--- a/pyat/at/physics/radiation.py
+++ b/pyat/at/physics/radiation.py
@@ -5,7 +5,7 @@ from math import sin, cos, tan, sqrt, sinh, cosh, pi
 import numpy
 from scipy.linalg import inv, det, solve_sylvester
 from at.lattice import Lattice, check_radiation, uint32_refpts
-from at.lattice import Element, Dipole, Wiggler, DConstant
+from at.lattice import Element, Dipole, Wiggler, DConstant, Multipole
 from at.lattice import get_refpts, get_cells, get_value_refpts
 from at.lattice import uint32_refpts, set_value_refpts
 from at.tracking import lattice_pass
@@ -411,9 +411,7 @@ def tapering(ring, multipoles=True, niter=1, **kwargs):
                          index=0)
 
     if multipoles:
-        mults = numpy.array([i for i, el in enumerate(ring) if
-                             hasattr(el, 'PolynomB') and
-                             hasattr(el, 'PolynomA')])
+        mults = get_refpts(ring, Multipole)
         k0 = get_value_refpts(ring, dipoles, 'PolynomB', index=0)
         _, o6 = find_orbit6(ring, refpts=range(len(ring)+1),
                             XYStep=xy_step, DPStep=dp_step)
--- a/pyat/at/physics/radiation.py
+++ b/pyat/at/physics/radiation.py
@@ -376,8 +376,7 @@ def gen_quantdiff_elem(ring, orbit=None):
 
 
 @check_radiation(True)
-def tapering(ring, multipoles=True, niter=1,
-             **kwargs):
+def tapering(ring, multipoles=True, niter=1, **kwargs):
     """
     Scales magnet strength with local energy to cancel the closed orbit
     and optics errors due to synchrotron radiations. PolynomB is used for
--- a/pyat/at/physics/radiation.py
+++ b/pyat/at/physics/radiation.py
@@ -399,7 +399,7 @@ def tapering(ring, multipoles=True, niter=1,
     dp_step = kwargs.pop('DPStep', DConstant.DPStep)
     dipoles = get_refpts(ring, Dipole)
     b0 = get_value_refpts(ring, dipoles, 'BendingAngle')
-    k0 = get_value_refpts(ring, dipoles,'PolynomB',index=0)
+    k0 = get_value_refpts(ring, dipoles, 'PolynomB', index=0)
     ld = get_value_refpts(ring, dipoles, 'Length')
 
     for i in range(niter):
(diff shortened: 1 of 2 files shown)


pyat/at/physics/radiation.py
```diff
@@ -372,5 +374,54 @@ def gen_quantdiff_elem(ring, orbit=None):
...
+    xy_step = kwargs.pop('XYStep', DConstant.XYStep)
+    dp_step = kwargs.pop('DPStep', DConstant.DPStep)
+    dipoles = get_refpts(ring, Dipole)
+    b0 = get_value_refpts(ring, dipoles, 'BendingAngle')
+    ld = get_value_refpts(ring, dipoles, 'Length')
+
+    for i in range(niter):
+        o0, _ = find_orbit6(ring, XYStep=xy_step, DPStep=dp_step)
+        o6 = numpy.squeeze(lattice_pass(ring, o0, refpts=range(len(ring))))
+        dpps = (o6[4, dipoles] + o6[4, dipoles+1]) / 2
```
lfarv, line 405: `o6[4, dipoles+1]` will fail if the lattice ends with a dipole. You should use `range(len(ring)+1)` to compute the orbit.

lfarv (changes requested):
naive question: would not it be better to have the loop on niter enclose all dipoles+quadrupoles+sextupoles ?

pyat/at/physics/radiation.py
swhite2401: You are right, I have included the proposed changes to dipoles/qudrupoles and sextupoles

Concerning the loop, I had the same questions but it turns out that only one iteration is needed for quadrupoles and sextupoles. This is explained by the fact that when the closed orbit is minimized by the dipole loop, multipoles do not contribute significantly to energy loss anymore. In order the speed-up the calculation I have moved them out of the loop.

swhite2401:
Changes applied

swhite2401:
I think this one is ready to merge, ok for you?

lfarv (commented):
Last question: unless you want to build special magnets, shouldn't you scale the whole `PolynomB` ?

swhite2401:
Good question, potentially yes, although corrector coils are generally present for combined function magnets to allow for tuning of the dipole components only.
Also at some point I wanted to consider the possibility of having circuits: scale groups of magnets.
I propose we consider this one as the ideal case, to be improved in the future, is that ok?
I will improve the docstring accordingly

lfarv (approved):
Ok, but for combined-function magnets it should be done. Before corrections are turned on… So OK to start like this!

swhite2401:
How would you handle this actually? PolynomB is generally used to model the correction coils: so on one side you add and on the other your remove?

lfarv:
I would start from a perfect machine, apply tapering once for ever, with a full scaling of Polynoms A and B, and then add errors and corrections. But I do not see any way to ensure that the user follows such a sequence…

swhite2401:
Yes very tricky...but you are right at least DQs should be treated as quadrupoles, I will improve this before the merge

lfarv:
I am in favour of full scaling: many kinds of combined-function magnets exist.  Correctors if present would be scaled as a side effect, but they can be retuned. Anyway, I am doubtful about tapering in the presence of errors and corrections. Does it make sense? The doc string might warn about that.

swhite2401:
Yes thinking about it I agree, I am implementing it, I will say clearly in the docstring that this method should not be applied on a lattice with errors

swhite2401:
Modifications done to scale all PolynomB and PolynomA componenents. Docstring warning added in case this method is used on a lattice with errors.

Ran into a deprecation warning when using `get_value_refpts` to get array attributes, fix implemented.
 in `utils.py`.

lfarv:
No : the modification in `get_value_refpts` breaks many things: for scalar attributes, the resulting array cannot be used any more in computations. It's an array of objects instead of and array of floats.
Don't use the function:
[scale*elem.PolynomB for elem in ring.select(refpts)]
gives you immediately a list of scaled polynomials. Simpler and more understandable.

swhite2401:
Ok this can be done, but still `get_value_refpts()` will fails for attributes arrays so we should do something about it no?

swhite2401:
I checked the doc:
https://numpy.org/neps/nep-0034-infer-dtype-is-object.html

This change was accepted so it will soon throw an error, however returning `numpy.array(list,dtype=object)` is the standard behavior in case `list` is a ragged nested sequence (with elements of different lengths)

What I propose is to leave things as they are for now, or ignore the deprecation warning.

Once the deprecation turns into an error we may have:
```
try:
    numpy.array(listt)
except:
    numpy.array(list,dtype=object)
```

lfarv:
A simpler solution (to be checked !):

```
    xy_step = kwargs.pop('XYStep', DConstant.XYStep)
    dp_step = kwargs.pop('DPStep', DConstant.DPStep)
    dips = get_cells(ring, Dipole)
    b0 = get_value_refpts(ring, dips, 'BendingAngle')
    ld = get_value_refpts(ring, dips, 'Length')
    pol0 = [(dip.polynomB, dip.polynomA) for dip in ring.select(dips)]

    for i in range(niter):
        _, o6 = find_orbit6(ring, refpts=range(len(ring)+1), XYStep=xy_step,
                            DPStep=dp_step)
        dpps = (o6[dips, 4] + o6[numpy.roll(dips, 1), 4]) / 2
        for dip, scale, r0, (polb0, pola0) in zip(ring.select(dips), 1.0+dpps, b0*ld*dpps, pol0):
            polb = scale * polb0
            polb[0] = r0
            dip.PolynomB = polb
            dip.PolynomA = scale * pola0

    if multipoles:
        mults = get_cells(ring, checktype((Quadrupole, Sextupole)))
        _, o6 = find_orbit6(ring, refpts=range(len(ring)+1), XYStep=xy_step,
                            DPStep=dp_step)
        dpps = (o6[mults, 4] + o6[numpy.roll(mults, 1), 4]) / 2
        for mult, scale in zip(ring.select(mults), 1.0+dpps):
            mult.PolynomB = scale * mult.PolynomB
            mult.PolynomA = scale * mult.PolynomA

```

lfarv:
`get/set_value_refpts` are perfect for scalars (either scalar attributes or array elements). For other types, the `ndarray` does not bring a lot compared to a standard sequence, so they are not really useful. I would keep them as they are.

swhite2401:
Thanks for the suggestions, I understand the logic but I like to keep the dipole `PolynomB[0]` separated because it is sufficient to cancel the horizontal orbit which is the minimal case.

I did not know you had enforced that PolynomB AND PolynomA should be defined for multipoles, I have simplified the second part accordingly.

Concerning `get/set_value_refpts`, I agree they can remain like this, however once the error is thrown we should remember to catch it and print out a meaningful error message, the python stack trace can be a little bit cryptic...

swhite2401:
Well this is what is done now:
-1st part iterate on PolynomB[0] for bends only -> minimal tapering
-2nd part scale all multipole of all magnets including Dipoles, but restore PolynomB[0] afterwards

find_orbit6+lattice_pass was removed to keep only find_orbit6 as you proposed (I had this historically because at some point I was iterating on a transfer line but this does not apply anymore)
Now searching multipoles based on instance of Multipole instead of 'PolynomB'

lfarv:
If you are happy with this one, OK for me !

PR #249, opened 2021-04-13
Cavity access

lfarv:
#### Global control:

**In Python**, a new method `Lattice.set_cavity()` sets the frequency, voltage, harmonic number and time lag, individually or simultaneously, on all cavities or optionally on a subset of cavities:
```
def set_cavity(ring, Voltage=None, Frequency=None, HarmNumber=None,
               TimeLag=None, refpts=None, copy=False)
```
A special value for the frequency: `'nominal'` (string) sets the frequency to the nominal value.

**In Matlab**, the existing function `atsetcavity()` is upgraded in the following way:
- the existing behaviour is kept unchanged, for backward compatibility:
```
NEWRING = ATSETCAVITY(RING,RFV,RADFLAG,HARM_NUMBER)
```
- a new mode is created, similar to the python syntax, which gives a choice of parameters:
```
%NEWRING=ATSETCAVITY(RING,...,'Frequency',FREQUENCY,...)
%   Set the cavity frequency [Hz].
%
%NEWRING=ATSETCAVITY(RING,...,'Voltage',VOLTAGE,...)
%   Set the total voltage (all cells) [V]
```

#### Single parameter control:

- New methods for the Lattice object:
`Lattice.get_rf_frequency(refpts=None)`, `Lattice.set_rf_frequency(frequency, refpts=None, copy=False)`, ...
give access to single parameters with the optional choice of a subset of cavities and the possibility of returning a new copy of the lattice,

- New Lattice properties:
`Lattice.frequency`, `Lattice.voltage`,...
give an easier syntax for the simple case of a single RF system. These properties raise exceptions if all cavities do not have the same frequency or harmonic number

### DIFF
--- a/pyat/at/physics/orbit.py
+++ b/pyat/at/physics/orbit.py
@@ -8,7 +8,7 @@ import scipy.constants as constants
 from at.lattice import AtWarning, AtError, check_radiation, DConstant
 from at.lattice import Lattice, get_s_pos, elements, uint32_refpts
 from at.tracking import lattice_pass
-from at.physics import get_energy_loss
+from at.physics import get_energy_loss, ELossMethod
 import warnings
 
 __all__ = ['find_orbit4', 'find_sync_orbit', 'find_orbit6']
@@ -283,7 +283,10 @@ def find_orbit6(ring, refpts=None, cavpts=None, guess=None, **kwargs):
         keep_lattice    Assume no lattice change since the previous tracking.
                         Default: False
         guess           Initial value for the closed orbit. It may help
-                        convergence. Default: (0, 0, 0, 0)
+                        convergence. The default is computed from the energy
+                        loss of the ring
+        method          Method for energy loss computation (see get_energy_loss)
+                        default: ELossMethod.TRACKING
         convergence     Convergence criterion. Default: 1.e-12
         max_iterations  Maximum number of iterations. Default: 20
         XYStep          Step size. Default: DConstant.XYStep
@@ -296,7 +299,7 @@ def find_orbit6(ring, refpts=None, cavpts=None, guess=None, **kwargs):
     max_iterations = kwargs.pop('max_iterations', DConstant.OrbMaxIter)
     xy_step = kwargs.pop('XYStep', DConstant.XYStep)
     dp_step = kwargs.pop('DPStep', DConstant.DPStep)
-    method = kwargs.pop('method', 'tracking')
+    method = kwargs.pop('method', ELossMethod.TRACKING)
 
     # Get revolution period
     l0 = get_s_pos(ring, len(ring))
--- a/pyat/at/physics/orbit.py
+++ b/pyat/at/physics/orbit.py
@@ -315,6 +315,7 @@ def find_orbit6(ring, refpts=None, cavpts=None, guess=None, **kwargs):
         try:
             rfv = ring.get_rf_voltage(cavpts=cavpts)
         except AtError:
+            # Cannot compute synchronous phase: keep zeros(6,)
             pass
         else:
             u0 = get_energy_loss(ring, method=method)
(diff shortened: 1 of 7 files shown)


matsievskiysv (approved):

matsievskiysv:
Note not for this PR in particular but for a whole package.
IMHO it would be better to use [enums](https://docs.python.org/3.5/library/enum.html) instead of the string arguments. It would make package more autocomplete friendly and allow [static type checking](http://mypy-lang.org/).

lfarv:
@matsievskiysv: very good suggestion.
To make the best use of static type checking, one would like a different enum class for each function using such arguments, with a potential problem to find simple and non-conflicting names for all these classes. Example: a `Method` class for `get_energy_loss` (with INTEGRAL and TRACKING members). You could imagine many other functions needing their own methods. Then you have to quality the method classes, leading to long class names…
Or define a global `Method` class for the whole AT package, listing them all, but weakening the type checking…
Any advice?

matsievskiysv:
Probably one could use the second approach in combination with [tagged unions](https://mypy.readthedocs.io/en/stable/literal_types.html#tagged-unions). But I don't have much experience with [Literal type annotations](https://mypy.readthedocs.io/en/stable/literal_types.html).

willrogers:
I don't think you need different enums for different functions as long as the items in the enum are compatible.

Would the items in `Method` be different for different functions? If there were a global enum what members would it need?

lfarv:
@willrogers, @matsievskiysv:
Going on with the `Method` enum: `get_energy_loss` looks for 'TRACKING' or 'INTEGRAL', while `get_tune` and `get_chrom` would use 'LINOPT', 'FFT' or 'LASKAR' (I would prefer 'NAFF' for the last one). The methods for these functions have nothing in common, but do represent the same kind of choice. There may be many other values. And I find "ELossMethod" and "TuneChromMethod" rather heavy…

A really different enum would be useful for ref. points: I would suggest Obs.ALL and Obs.END to refer to "All elements" and "End of lattice". @willrogers, you already mentioned that, I was looking for the right way to implement it, this one looks nice.

swhite2401 (approved):

swhite2401:
Before you merge, could you update the initial guess for `find_orbit6` to be zero if len(frf)>1
Thanks!

willrogers:
I don't think that one big enum makes sense. I do agree that enum names can be unwieldy. 

One option is implicitly to use the module name as a qualifier i.e. `Method` is in the `energy_loss` module so you can distinguish it if necessary. This potentially conflicts with importing everything into the `at` package (but I am not all that keen on that).

lfarv:
For cavity selection, "cavpts" is used instead of "refpts". This avoids a conflict for functions like find_orbit6 needing to select both observation points (refpts) and cavities (cavpts).

lfarv:
Created 2 enums to replace string keywords:
`ELossMethod` for energy loss computation
```
ELossMethod.INTEGRAL
ELossMethod.TRACKING
```
`Frf` for set_rf_frequency
```
Frf.NOMINAL
```

willrogers:
Should there be another `Frf` enum member rather than `None` for whatever `None` means in that function i.e. don't modify? You could use the value `None` for backward compatibilty. Also it might be neater to use a non-numeric value for `NOMINAL`.

pyat/at/physics/orbit.py
```diff
@@ -308,12 +311,16 @@ def find_orbit6(ring, refpts=None, guess=None, **kwargs):
...
-        rfv = ring.periodicity * sum(elem.Voltage for elem in cavities)
-        u0 = get_energy_loss(ring, method=method)
-        if u0 > rfv:
-            raise AtError('Missing RF voltage: unstable ring')
         ref_in = numpy.zeros((6,), order='F')
-        ref_in[5] = -constants.c / (2 * pi * f_rf) * asin(u0 / rfv)
+        try:
+            rfv = ring.get_rf_voltage(cavpts=cavpts)
+        except AtError:
+            pass
```
matsievskiysv, line 318: Maybe add a comment here? Otherwise this looks like a not implemented branch.

lfarv: Done.

Issue #251, opened 2021-04-15
6D matrix normalization, amat and phase advance


bnash:
@swhite2401 [has raised some questions](https://github.com/atcollab/at/issues/248#issuecomment-820170170) about how to find phase advance for the general 6D optics, and whether the [A matrix](https://github.com/atcollab/at/blob/master/atmat/atphysics/LinearOptics/amat.m) is defined correctly.

We can use this issue to discuss all of this.

bnash:
[Here](https://github.com/atcollab/at/blob/master/pyat/at/physics/amat.py) is Laurent's Python version of this function.

bnash:
See also [this 1998 reference](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.58.2481) by E. Forest, "Dispersive lattice functions in a six-dimensional pseudo-harmonic-oscillator"

swhite2401:
Hi @bnash,
Very good idea, so to summarize (from the tests I have done):
-getting optics function using betoids/etoids functions works very nicely
-getting the tune based on a one-turn map also work very well
-problem comes from: getting the phase advance from point 1-> 2 I get pi "jumps"

This the little code I used for testing (M->transfer map,M0->one tuen map, for from `find_m66`)
```
def get_mus(M,M0):
    A = amat(M0)
    M2 = numpy.dot(M0,inv(M))
    A2 = amat(M2)
    A2I = inv(A2)
    R12 = numpy.squeeze(multi_dot([A2I,M,A]))
    mux = numpy.arctan2(R12[0,1]-R12[1,0],R12[0,0]+R12[1,1])
    muy = numpy.arctan2(R12[2,3]-R12[3,2],R12[2,2]+R12[3,3])
    #print(M,'\n\n',M2,'\n\n',M0,'\n\n')
    #print(A,'\n\n',A2,'\n\n',AI,'\n\n',A2I,'\n\n')
    #print(A2I-AI,'\n\n',A2-A,'\n\n',M2-M0,'\n\n')
    return numpy.array([mux,muy])
```
You can see it is relatively simple, maybe I am missing something??

Interesting fact, the pi jump depends on:
-an arbitratry factor applied on M66 before computing the eigenvectors:  _,vv=eig(m66*fact)
-the method employed for inversion: inv,pinv,solve

Which makes me think we are facing numerical issue related to rounding errors.

Any advice?

Thanks!

bnash:
[...]

PR #252, opened 2021-04-19
Analysis of fully coupled 6D optics

lfarv:
We introduce here an alternative processing of fully coupled motion based on:
- Étienne Forest, Phys. Rev. E 58, 2481 – Published 1 August 1998
- Andrzej Wolski, Phys. Rev. ST Accel. Beams 9, 024001 – Published 3 February 2006


lfarv:
Still missing:

- Analysis of the initial point, to be similar with the `linopt` output,
- Tunes and damping times,
- Transfer line option.

lfarv:
The main changes are:
- `amat.py`, `amat.m`: added the Eigen values to the output, and faster computation of the tunes from the Eigen values,
- `linear6.py`, `atlinopt6.m`: these are the new functions for 6D analysis.

The function and module names are proposals, you may have better ideas. This version is operational, but the transfer line options and 4x4 processing (radiation OFF) are still not there.

lfarv:
Now ready for testing. Remaining questions:
- the name of the function. Better ideas?
- the name of the field containing the local optics parameters ("R" for now): a set of 3 6x6 matrices (one for each independent motion, x, z, s when there is no coupling). Named "Bk" (k=1,2,3) in Wolski's paper, or sometimes "R". 
- the conditions for transfer line mode: as it is now, technically it runs with either a 2x4x4 input, and then runs in 4D mode, or a 3x6x6 input. However, 6D results in a lot of NaNs. I am not sure of the interest of 6D mode in a transfer line (no longitudinal motion). So I am about to forbid 6D for transfer lies. Any advice ?

swhite2401:
Hello @lfarv,
Thanks for the implementation this is starting to look very good, however I have some remarks/comments.

**Implementation:**

-It would be very convenient to mimic the ouput of linopt: lindata0,q,q',lindata, this way we could integrate this method in other functions easily, especially for matching it is potentially very interesting as it seems to be faster than linopt
Damping times are not so interesting here but could be integrated in lindata
- a flag get-chrom that outputs chromaticty and chromatic functions, similar to what we have in linopt would be very useful

Unrelated bugs: doing my tests I found:
-find_orbit6 fails if TimeLag is non-zero, this has to be fixed
-linopt computes the chromatic functions but does not return them, I think this comes from the matching branch. This is expensive (more calls to linopt) and should probably be optional

**Optics**

The 4D calculations works very nicely and results agree for all the methods I have tested.

I have found some inconsistency in the presence of radiations. In this comparison I compare the 4D optics, and the lattice including radiations and tapering (to be as close as possible to 4D).
linopt6 is your implementation, and betaoids is the prototype I wrote based on the discussion with @bnash  in #251. You can see here that while the dispersion agrees very well, the beta functions are clearly off in linopt6 as they seems to have a slope related to energy loss...

FCC lattice (strong energy loss)
fcc_lat
FCC comparison betoids vs linopt6:
fcc_comp
HMBA lattice(low energy loss)
hmba

lfarv:
Hello @swhite2401, thanks for the extensive tests. I need some time to understand the deviations between the two methods, but I have answers for the "bugs" you report:
- **find_orbit6**: it was recently modified to include the energy loss in initial guess for the search, so that it starts from the synchronous phase. With `TimeLag`, you break the synchronous phase, so the starting point is less accurate. It would be possible to include the `TimeLag` effect in the initial guess, but this would work only for a single RF system. For a complex system with different frequencies and phases, as we already discussed, you need to find the solutions of sum(V_i * sin(2*pi*F_i + PHI_i)) = U0, numerically I guess, and take one of several possible solutions… At the moment, you may use the "`guess`" keyword to  set the starting point correctly (try with ct=0). Normally, when you design the RF system, you should know at which time you roughly get the desired voltage. And we may think of a way to solve the equation.
- **linopt**: you are right, there is some unused computation, that you introduced on 04/09/2020 ("reorganise linopt and introduce chromfunc"). I propose to open another pull request to control this computation with a flag and output the results.
- **output consistency**: One has to separate local data (called `lindata` for `linopt` and `elemdata` for `linopt6`) from global data (tunes, chromaticities, damping times,…). That's why the record array `ringdata` of `linopt6` is more flexible than the two parameters `nu, xsi` of `linopt`: you can but as many attributes as you like. Unfortunately it's too late to change `linopt` now… I do not know how to do.

I'll now look at your results !

swhite2401:
- **find_orbit6**: for simplicity I would suggest to set the guess to `numpy.zeros(6)` in case a non zero `Timelag` is found
- **linopt**: ok I will create a branch with a fix
- **output inconsistency**: I know this is not the cleanest but we may introduce an upper layer `get_optics()` that allows to select the method and reformats the output of linopt using record arrays

lfarv:
@swhite2401: Since we are doing the same thing, I would like to look at your implementation to see where we diverge.

Trying here, I do not get exactly you plots, I think your HMBA lattice is not the one from the repository; But I confirm that I get a drift of beta values (down), roughly by the same amount as on your plots. So the beta functions do not close! Problem!

This does not  appear when I run a lattice with cavity and no radiation, so it is indeed linked with radiation.
I continue investigating!

swhite2401:
Ah right, I am using my local ESRF lattice copy.
I will send you the lattice and my test script by mail, I used the conventions from Forrest + betaoids rather than the one form Wolski but in principle this should not mater..., a beta function drifting down could mean that the damping matrix is not properly canceled out when getting the twiss parameters

lfarv:
Corrected the beta drift. It's indeed a difference between Forest and Wolski. Equation 56 in Wolski's paper is wrong with radiation.  @swhite2401, can you check (no urgency, of course!).
I also added the A-matrix on the `elemdata` output

swhite2401:
Excellent news! I'll take a look and let you know!

swhite2401:
Did you look at consistency with linopt for a (strongly) coupled lattice?

lfarv:
Not yet, but I found a strange behaviour with your lattice. In my examples, linopt6 in 4D is in excellent agreement with linopt. It's not such with your lattice. So new questions…

swhite2401:
Where do you see problem with the 4D version? On my side it looks fine, see plot below for HMBA
(I removed dispersion since linopt6 - 4D does not return it - for now!)

new

lfarv:
Good enough for sure… So it's on my side. Does linopt6 agree now with your implementation ?

swhite2401:
Yes very nicely:

new

lfarv:
`linopt6` has now the same fields as `linopt` (when applicable), included the chromatic amplitude functions introduced in #254

lfarv:
`linopt6` is now on par with `linopt`. The provided informations are identical, the results in 4D are strictly equal, and the speed is equivalent. Here is a short evaluation I did locally, for one particular example. Times on Octave are surprisingly high, it may be due to my setup, no guarantee… Otherwise:
- when comparable, `linopt6` and `linopt` are equivalent,
- in my case, python is on average slightly faster than Matlab.

 times in ms | Matlab | python | Octave
----- | ------ | ------- |  -------
4D, linopt, no refpoint | 6.7 | 2.5 | 35
4D, **linopt6**, no refpoint | 4.5 | 3.5 | 90
| | | 
4D, linopt, 123 refpoints `get_chrom=True` | 23 | 35 | 96
4D, **linopt6**, 123 refpoints `get_chrom=True` | 27 | 15 | 166
| | | 
6D, **linopt6**, 123 refpoints | 19 | 30 | 118
| | | 
6D, **linopt6**, `get_w=True` | 80 | 60 | 530

For me, this is now ready for merging, except for one point: the function name ! Shall we keep `linopt6` ?

swhite2401:
For me the name is fine and we can merge
However this is a major improvement and there are 2 things that I think we should consider:
-create a new tag
-Re-organize the code to better integrate this new development

I do not like that output are different for linopt and linopt6 this is not nice and should be avoided for functions providing the "same" ouput. A similar problem can be found for lattice_pass and find_orbit6 this can be very confusing and annoying to the user.

What I propose is:
-keep linopt6 as it is
-create a new module called linopt4 similar to linopt6 that returns the SAME output structure
-in linear.py: keep a function linopt for backward compatibility that calls linopt4 and just changes its output to the present linopt structure. Introduce a new function get_optics? get_twiss? that has a additional enum as input OpticsMethod to switch form Edwards and Teng to Normal Mode

This would 1) make the code structure more understandable 2) allow to easily integrate either methods in existing functions using linopt such as matching, get_tune etc...

lfarv:
@swhite2401 : these are very good suggestions. 
I would even go further: the `coupled` keyword which was introduced for performance reasons is not available in `linopt6`. We could remove it by defining ` linopt2` (uncoupled motion, equivalent to `linopt(…, coupled=False)`) and `linopt4`, equivalent to `linopt(…, coupled=True)`.

The additional global function is not strictly necessary: I find quite readable to use the optics function name as a keyword. For instance a "`matching`" function could be defined as
```
def matching(ring, ..., method=linopt6, ...):
    ...
    lindata = method(ring, ...)
    ...
```
However, if you find a `get_optics` function useful, we can provide it.

This will require some reorganisation on the `linear.py` module. A number of files have already been modified in this PR for `linopt6`. To avoid putting too many things in it, I propose to finalise it now, and to open a new one for the reorganisation. The new tag can wait for that one!

Let me know…

swhite2401:
Perfect for me, I agree it is better to keep things separated.

For less experienced users I think a `get_optics` function with docstring briefly describing the 3 methods would be useful.

The enum proposal was just to remain consistent with your recent additions, but I must admit that I don't like it, for me string arguments are more readable, so what you propose is perfect.

swhite2401 (approved):

PR #253, opened 2021-04-19
Add the capability to use passmethods written in Python.

willrogers:
Added example and test.

@swhite2401 @lfarv the same as #243 but with all commits squashed to one. 

I would rename to `py_identity_pass.py` to match Python convention but no big deal.


lfarv:
Renaming ? I think the passmethod names must end with 'Pass'. `atradon/off` rely on this. And up to now, they use mixed case and no underscores…
Also, I would prefer a pyDriftPass: identity does not show anything on the result, so the test is not really convincing.

willrogers:
Yes, fair enough.

swhite2401:
Ok, I will write another example for a drift

lfarv:
`pyIdentityPass` should be in integrators:
`RuntimeError: PassMethod pyIdentityPass: library, module or trackFunction not found`

swhite2401:
atintegrators is not in your python path

lfarv:
So where should the python integrators go?

swhite2401:
Well, this is a good question, I was working under the assumption that they could go anywhere in the pythonpath

lfarv:
Then they have to be mixed with other python code… 
C integrators, including the ones added by users must be in `atintegrators`. Any way to force `atintegrators` it in the python path?

swhite2401:
Mmm we may add `__init__.py` in `atintegrators` then `pyat/integrator-sr` would be added automatically to the path right?

lfarv:
There is a `__init__.py` in `at/pyat/at/atintegrators`. Is this the place to put the python integrators? Is the empty __init__ file enough?

swhite2401:
Ok I did some test, you can add `__init__.py` in atintegrator but:
-setup-py needs to be modified to copy over *.py
-integrator-src name needs to be changed to be compatible with module/package names

Doing this you may import integrator_src.pyDriftPass (for example) directly

But not very nice, a folder pyintegrators with an empty __init__.py would be nicer in my opinion, because these functions are anyway not accessible from the matlab implementation, but I let you decide

swhite2401:
Ah wait you are right, there is already pyat/at/integrators, we just need to copy over *py from at/atintegrators and it should work

lfarv:
I tried to copy `*Pass.py `in `at/pyat/integrators`. For me it still does not work…

lfarv:
Adding a `__init__.py` file in a directory creates a package. But unfortunately it does not put the directory in the path. Looking at `sys.path`, only the top-level ("pyat") is in the path…
Anyway, I copied the files into my "site-packages" directory, and checked that it works,. But we need a rule for where to put the python integrators.

swhite2401:
Hello @lfarv @willrogers  , please see new implementation (see lines 144->169 in at.c):
-at.c will first look in at.integrators for python passmethods
-if it fails the search is extended to the whole pythonpath

I believe this sets a clear rule for python integrators:
-the c integrators always come first in case they have the same name
-official ones distributed with at are privileged over user's devs and located in at.integrators
-users developments -> anywhere in the python path, ignored if the name is already used

Would this be ok for you?

lfarv:
Perfect! Now it works beautifully!

A personal remark: I would have put the precedence order exactly opposite to yours:
- A user could overload a distributed integrator with his own version. As it is now, he is obliged to take a different name and change the lattice,
- python integrators, which should be limited to prototyping, could overload the stable C version.

For non-conflicting names, the order is irrelevant, so mine has no drawback…
@willrogers, any advice?

swhite2401:
Ok I can invert, I have no strong feelings about it, so the order would be:
1-pythonpath
2-at.integrators
3-C passmothod

Is that right?

willrogers:
I agree with @lfarv - if you look at linux configuration files there is a similar hierarchy e.g. git configuration:

* system default
* user config
* project config

It is most flexible that way.

swhite2401:
@lfarv, @willrogers  As agreed, the call/search order was modified:
1-search for python pass methods in the python path
2-search for python pass methods in at.integrators
3-search for C pass methods in at.integrators

Test functions pyIdentityPass.py and pyDriftPass.py were added in at.integrators and a tests in test_integrators.py with calls to these functions compare the python and C behavior in atpass

Anything else you would like to add?

lfarv (approved):
Ok for me!

PR #254, opened 2021-04-27
Chromatic functions in linopt

swhite2401:
This branch fixes a bug left after the matching development where chromatic function were computed but never used.
These are expensive computation and should be done on request only.

Chromatic function are defined in Brian W. Montague Report LEP Note 165, CERN, 1979

A flag get_w=False is added in linopt, the default behavior is back to the standard one.

Computing times for the ESRF-HMBA lattice:
default: 0.1440258026123047
get_chrom=True: 0.24343562126159668
get_w=True: 0.4185199737548828

Without this fix all case are 0.42

We are unfortunately accumulating flags and tests in linopt...any suggestions?

### DIFF
--- a/pyat/at/physics/linear.py
+++ b/pyat/at/physics/linear.py
@@ -258,7 +258,7 @@ def linopt(ring, dp=0.0, refpts=None, get_chrom=False, orbit=None,
             print('Mu not found in twiss_in, setting to zero')
             tune = numpy.zeros((2,))
 
-    if get_w:
+    if get_chrom:
         # noinspection PyUnboundLocalVariable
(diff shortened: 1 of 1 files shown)


lfarv (approved):

lfarv:
@swhite2401 : I hope your computation of W is correct: I copied it into linopt6 (#252)

swhite2401:
I hope it is as well, I used this:
https://mad.web.cern.ch/mad/webguide/manual.html#Ch1.S7.SS4

lfarv:
This explains why it differs from the original paper. Thanks !

PR #255, opened 2021-04-28
Better handling of multiple rf system

swhite2401:
**energy_loss.set_cavity_phase()**

In case multiple RF systems are found a least_square minimization is used to find the `TimeLag` from `U0` and the sum of the sine waves. A single TimeLag is applied to all RF systems.
This method is called only is multiple RF frequencies are found in `cavpts`

**orbit.find_orbit6()**

Instead of using the first cavity the determine the harmonic number and rf frequency, the following method is used:
1-look for a unique rf frequency using `cavity_access()`
2-If 1 fails, i.e. multiple rf frequencies are found, the cavity with the smallest frequency is used as reference
The initial guess was sometimes failing for non-zero `TimeLag` as it was not taken into account. 
Presently the only function setting the TimeLag is set_cavity_phase() which sets the initial point to zero. If a non-zero `TimeLag` is found  `numpy.zeros(6)` is used as initial guess, 
To be noted, in case the `TimeLag` is not set and multiple RF systems are present, the initial guess is forced to `numpy.zeros(6)` and it is possible that `find_orbit6()` does not converge.

### DIFF
--- a/pyat/at/physics/energy_loss.py
+++ b/pyat/at/physics/energy_loss.py
@@ -3,8 +3,10 @@ from enum import Enum
 from warnings import warn
 from math import sqrt, pi
 import numpy
-from at.lattice import Lattice, Dipole, Wiggler
+from scipy.optimize import least_squares
+from at.lattice import Lattice, Dipole, Wiggler, RFCavity
 from at.lattice import check_radiation, set_cavity, AtError
+from at.lattice import checktype, get_refpts
 from at.tracking import lattice_pass
 from at.physics import clight, Cgamma, e_mass
 
@@ -104,16 +106,29 @@ def set_cavity_phase(ring, method=ELossMethod.INTEGRAL,
     if cavpts is None and refpts is not None:
         warn(FutureWarning('You should use "cavpts" instead of "refpts"'))
         cavpts = refpts
-    rfv = ring.get_rf_voltage(cavpts=cavpts)
-    freq = ring.get_rf_frequency(cavpts=cavpts)
+    u0 = get_energy_loss(ring, method=method)
+    try:
+        rfv = ring.get_rf_voltage(cavpts=cavpts)
+        freq = ring.get_rf_frequency(cavpts=cavpts)
+    except AtError:
+        if cavpts is None:
+            cavpts = get_refpts(ring, RFCavity)
+        freq = numpy.array([cav.Frequency for cav in ring.select(cavpts)])
+        rfv = numpy.array([cav.Voltage for cav in ring.select(cavpts)])
+        if u0 > numpy.sum(rfv):
+            raise AtError('Not enough RF voltage: unstable ring')
+        ctmax = 1/numpy.amin(freq)*clight
+        eq = lambda x: numpy.sum(rfv*numpy.sin(2*pi*freq*x/clight))-u0
+        tl = least_squares(eq,0,bounds=(-ctmax,ctmax)).x[0]
+        set_cavity(ring, TimeLag=tl, cavpts=cavpts, copy=copy)
+    else:
+        if u0 > rfv:
+            raise AtError('Not enough RF voltage: unstable ring')
+        timelag = clight / (2*pi*freq) * numpy.arcsin(u0/rfv)
+        set_cavity(ring, TimeLag=timelag, cavpts=cavpts, copy=copy)
     print("\nThis function modifies the time reference\n"
           "This should be avoided, you have been warned!\n",
           file=sys.stderr)
-    u0 = get_energy_loss(ring, method=method)
-    if u0 > rfv:
-        raise AtError('Not enough RF voltage: unstable ring')
-    timelag = clight / (2*pi*freq) * numpy.arcsin(u0/rfv)
-    set_cavity(ring, TimeLag=timelag, cavpts=cavpts, copy=copy)
 
 
 Lattice.get_energy_loss = get_energy_loss
--- a/pyat/at/physics/energy_loss.py
+++ b/pyat/at/physics/energy_loss.py
@@ -119,7 +119,7 @@ def set_cavity_phase(ring, method=ELossMethod.INTEGRAL,
             raise AtError('Not enough RF voltage: unstable ring')
         ctmax = 1/numpy.amin(freq)*clight
         eq = lambda x: numpy.sum(rfv*numpy.sin(2*pi*freq*x/clight))-u0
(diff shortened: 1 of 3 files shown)


lfarv:
#### set_cavity_phase
No ! You are putting all cavities in phase ! The 1st RF goals I see (harmonic bunch-lengthening, fs modulation, bunch splitting/grouping…) all require an adequate phasing between the cavities. Putting them in phase ruins the goal ! What you have to do is:
- include the initial `TimeLag` in the equation (`sin(2*pi*freq*(x-timelag)/clight)`),
- **decrement all** `TimeLag`s with the solution.

This way, you shift the time origin to bring the synchronous phase to zero, which is the goal of `set_cavity_phase`, but keep the desired voltage shape. If for some particular purpose you need cavities in phase, they will stay in phase after running `set_cavity_phase`, so no problem.
There is still a "detail" to solve: the least square fit has 50% chance to find an unstable fixed point. You have to choose a solution with the right voltage slope.

#### find_orbit6
It is much simpler: now that you can solve the equation `sum( Vi(t_sync)) = U0`, you just set the initial guess to `t_sync` and you are done (provide you take a stable solution). No test, no choice. Just for efficiency, in the case of a single system, you can use the analytical solution (the present one) rather than the least squares.

However, you should keep the accessibility of the `guess` keyword, it's the only way to select the desired bucket, since in a complex system, you may have several buckets with different properties.

swhite2401:
Thanks for the feedback, in fact your are taking a step further, I just had in mind to facilitate `find_orbit6` convergence.
I am a bit busy these days but will implement your proposals as soon as I can!

swhite2401:
Corrected version:

**set_cavity_phase**
-least_square used to find the zero crossing of the derivative
-> use the side with positive slope to set the boundaries for the least_square minimization to find the synchronous phase
-The initial TimeLag is handled as advised for complex RF systems

**find_orbit6**
-Simplified version according to previous comments

I had to relax a bit the tolerance on test_ohmi_envelop for the comparison of emittances computed
with matlab, probably due to the different solution found by find_orbit6

I am not fully satisfied with this double call to least_square....any advices?

I hope I got all the signs rights...from the test I have done it seems to work fine but a double check would be good.

pyat/at/physics/energy_loss.py
```diff
@@ -81,6 +84,59 @@ def tracking(ring):
...
+    except AtError:
+        freq = numpy.array([cav.Frequency for cav in ring.select(cavpts)])
+        rfv = numpy.array([cav.Voltage for cav in ring.select(cavpts)])
+        tl0 = numpy.array([cav.TimeLag for cav in ring.select(cavpts)])
+        if u0 > numpy.sum(rfv):
+            raise AtError('Not enough RF voltage: unstable ring')
+        ctmax = 1/numpy.amin(freq)*clight/2
+        tt0 = tl0[numpy.argmin(freq)]
+        eq = lambda x: numpy.sum(rfv*numpy.sin(2*pi*freq*(x+tl0)/clight))-u0
+        deq = lambda x: numpy.sum(rfv*numpy.cos(2*pi*freq*(x+tl0)/clight))
```
lfarv, line 122: `CavityPass` uses a minus sign: `sin(2*pi*freq*(ct - timelag)/clight))`. Should not you use the same convention?

pyat/at/physics/energy_loss.py
```diff
@@ -104,16 +160,13 @@ def set_cavity_phase(ring, method=ELossMethod.INTEGRAL,
     if cavpts is None and refpts is not None:
         warn(FutureWarning('You should use "cavpts" instead of "refpts"'))
         cavpts = refpts
-    rfv = ring.get_rf_voltage(cavpts=cavpts)
-    freq = ring.get_rf_frequency(cavpts=cavpts)
+    elif cavpts is None:
+        cavpts = get_refpts(ring, RFCavity)
```
lfarv, line 164: Redundant test: caveats if properly handled in both `get_timelag_fromU0` and `set_value_refpts`

pyat/at/physics/energy_loss.py
```diff
@@ -81,6 +84,59 @@ def tracking(ring):
...
+            ts = least_squares(eq, (zero_diff-tt0)/2,
+                                 bounds=(-tt0, zero_diff)).x[0]
+        else:
+            ts = least_squares(eq, (ctmax-tt0+zero_diff)/2,
+                                 bounds=(zero_diff, ctmax-tt0)).x[0]
+        timelag = ts+tl0
+    else:
+        if u0 > rfv:
+            raise AtError('Not enough RF voltage: unstable ring')
+        timelag = clight/(2*pi*freq)*numpy.arcsin(u0/rfv) * \
```
lfarv, line 135: `cavpts` may be a boolean mask or a test function. So `len(cavpts)` is not always the number of refpoints. You can use `refpts_len(ring,refpts)` or `ring.refcount(refpts)` to get the number. Or better, keep it as a scalar, it should be broadcast to the right size when necessary.

pyat/at/physics/energy_loss.py
```diff
@@ -81,6 +84,59 @@ def tracking(ring):
...
+                  numpy.ones(len(cavpts))
+    return timelag, numpy.squeeze(numpy.unique(timelag-tl0))
```
lfarv, line 137: You could remove the expensive call to "`unique`" by defining `ts = timelag-tl0` in the "simple case, and returning it

lfarv:
Now I need some time to look at the solution of the equation itself, more on Monday evening…

pyat/at/physics/energy_loss.py
swhite2401: set_value_refpts does not handle refpts=None, I am not I understand what you mean

pyat/at/physics/energy_loss.py
swhite2401: The sign is consistent with this: `timelag = clight/(2*pi*freq)*numpy.arcsin(u0/rfv)`

pyat/at/physics/energy_loss.py
swhite2401: Done

pyat/at/physics/energy_loss.py
swhite2401: Changed to get_cells
I don't like that the function can return different "types" depending on the case so I prefer to leave it as is and return always a scalar

swhite2401:
Ok to merge?

lfarv (approved):
Ok for now, we'll see how it works.

PR #256, opened 2021-05-01
Argument checking in Matlab

lfarv:
The utility functions `getflag`, `getoption`, `getargs` are updated to:
- add new possibilities for argument checking,
- provide a better documentation,
- be compatible with new features introduced in Matlab 2020b and 2021a. AT itself stays compatible with Matlab >= 2016b, but users may use these new features in their code.


willrogers:
Fine by me.

PR #257, opened 2021-05-05
Linear optics options

lfarv:
This pull request is the continuation of #252. The `linopt6` function introduced there provides a new way of analysing the coupled motion. We try here to simplify the choice of the optics method with a set of 3 functions with compatible input and output:

- `linopt2` studies non-coupled motion (equivalent to `linopt(..., coupled=False, ...)`,
- `linopt4` uses the Sagan/Rubin analysis of 4D coupled motion (equivalent to `linopt(..., coupled=True, ...)`,
- `linopt6` uses the normal mode analysis for 4D or 6D motion.

`linopt2` and `linopt4` are derived from linopt, but provide the same interface as linopt6. This allows the choice of the method in higher level functions.

`linopt` is kept as it is, but is preferably replaced by `linopt2` or `linopt4`.


lfarv:
Ready for review. The 3 analysis method are available in both Python and Matlab
- `linopt6` is the most flexible choice: it work with or without longitudinal motion,
- `linopt2` is the fastest when it can be applied

For H/V coupled motion, the alpha and beta values are slightly different between `linopt4` and `linopt6` because of the different analysis. The independent modes are of course the same in the 2 methods, but they are expressed along different axes:
- `linopt4` returns the motion as two independent 2D motions in rotated axes: the beta values correspond the each of these motions, so along these rotated axis,
- `linopt6` returns the motion as independent 6D motions in the standard AT axes. beta values are really horizontal and vertical values, for the dominant mode.

In none of these 2 outputs the beam sizes are simply equal to `sqrt(betax*epsx)`…

But the results from linopt6 are maybe easier to use: the envelope matrix sigma can be easily built from the `R` output:
`sigma = eps0*R[0] + eps1*R[1] + eps2*R[2]`
with (eps0, eps1, eps2) the emittance of each mode
The beam sizes are then  `sigmax = sqrt(sigma[0,0])` and so on…

swhite2401:
Hello @lfarv, this looks nice.
Just one thing: in `get_tune()` I don't think `remove_dc` should be subtracting the closed orbit but really the mean of the oscillation like we implemented in the previous version: for oscillating beam sextupoles introduce static dipole kick that you are not accounting for here. Also I prefer that this method is indepedant of an optics calculation or orbit search.

swhite2401:
What is the issue with unstable lattices?

lfarv:
Handling of unstable machines in `linopt6`:
1. All the processing is based on the A-matrix,based itself on the Eigen values/vectors of the 1-turn transfer matrix. There, an unstable motion in one plane appears as a pair of real, inverse Eigen values instead of a pair of complex conjugate values. This gives a "0.0" tune, and the corresponding Eigen vectors are infinite or NaN. Up to that point, the "stable" planes are correct,
2. Then, when attributing each value/vector to the closest x/z/s plane, the scan fails so that the "valid" and "0" tunes end up randomly in x, z or s… I could not find any way to solve this, but it's not hopeless,
3. Even if the previous problems were solved, in the case of non-symplectic motion (radiation damping), deriving the R-matrices (alpha and betas), needs inverting A, so that even the stable planes are lost ! Only the phase advance could be preserved for stable planes. No solution there…

So the only solution at the moment is to raise an exception as soon as a single unstable plane is found.

`linopt2` and `linopt4` are corrected so that they give NaNs for unstable planes and correct results for others, rather than raising `ValueError`.

lfarv:
@swhite2401: `get_tune` is corrected to remove the effective average position, however it still depends on a preliminary optics calculation (as before) for 2 reasons:
1. the closed orbit is used to throw the particle with the desired amplitude,
2. the A-matrix is used to normalise the motion (though it is may be not absolutely necessary)

It now works with or without radiation, but apparently the detection of the longitudinal tune is problematic (amplitude problem?).
In addition, the 'fft' method gives very bad results. Should we keep it ? Or warn about it?

swhite2401:
I think what you propose is fine, having only 1 stable plane does not make much sense.

Concerning the tune it is fine like this as well, you are right we did have this dependency already. 

ftt is indeed less precise, probably more turns are needed to get "good" results, it is ok to remove it in `get_tune`.

Concerning Qs, I faced the same issue recently (using `get_tunes_harmonic`) I could solved it with:
-removing the dc part (maybe not necessary)
-setting fmin=eps (and fmax=1-eps?)
The problem for me was a strong(er) line at exactly 0, not sure if you had the same issue.

lfarv:
Last step: `get_mode_matrices` is modified to be similar to what is done in `linopt6`.

The R-matrices represent the beam modes invariant over 1 turn and are deduced from the A-matrix. For instance in 1D, the single R-matrix is:

[&beta;,&nbsp;-&alpha;],
[-&alpha;,&nbsp;&gamma;], and similarly for 4D and 6D motion

The formula used up to now was valid for symplectic transfer (no radiation):

R<sub>k</sub> = A * I<sub>k</sub> * A<sup>T</sup>, k = 1, 2, 3&nbsp;(equation 56 in ref [2], Wolsky)
with  I<sub>1</sub> = diag([1, 1, 0, 0, 0, 0]),  I<sub>2</sub> = diag([0, 0, 1, 1, 0, 0]) and  I<sub>3</sub> = diag([0, 0, 0, 0, 1, 1])

For no-symplectic transfer (with radiation), there is no closed solution, stricky speaking, because of damping. However, looking statistically at the equilibrium beam, the beam envelope does close over one turn. This leads a slightly different definition of the R-matrices (and consequently derived "&beta;" and "&eta;" function), in terms of statistical averages. The corresponding formula is:

R<sub>k</sub> = A * S * I<sub>k</sub> * A<sup>-1</sup> * S<sup>-1</sup>, k = 1, 2, 3 and S the anti-symmetrical matrix, (~equation 32 in ref [1], Forest)

This is now what is set in `linopt6` and `get_mode_matrices`. For symplectic transfer (for instance cavities ON but no damping), both results are identical. For non-symplectic transfer, the results are very slightly different, however it required a modification of the `test_physics.py` test module.

Unless you report a bug, this pull request is ready for merging. Given the significant additions to AT, this may deserve an increment of the version number. How can it be done @willrogers ?

lfarv:
### Summary of linear optics options
There are now have 3 options for linear optics computation: `linopt2`, `linopt4` and `linopt6`

#### Without radiation damping
Without longitudinal motion (4D transfer) or **with longitudinal motion** (cavities ON, 6D transfer) we get:
- without H/V coupling: `linopt2`, `linopt4` and `linopt6` give exactly the same results (tested at the 10<sup>-12</sup> level),
- with H/V coupling: `linopt4` and `linopt6` still give exactly the same results. However, since the definition of "&beta;s"  is different, the comparison is not obvious. As there cannot be a single definition "&beta;s", the easiest to understand results are given by the R-matrices, which are not displayed by `linopt4`. They can be easily derived and this could be the topic for a later development.

The choice is then a matter of speed

#### With radiation damping
- only `linopt6` is available. In this case, there is not "optical" closed solution, but only a "statistical" solution, given by the R-matrices. The difference with the results without radiation is typically at the 10<sup>-3</sup>, 10<sup>-4</sup> level (hmba lattice), due to closed orbit effects in sextupoles.

swhite2401:
All good for me to merge.

However before we change version we could profit a lot in other parts of the code from this development, mainly in:
-matching
-linear
-nonlinear
-fastring
-other in matlab/python?

I suggest we implement all these simplifications / improvement before incrementing the version number.

swhite2401 (approved):

PR #259, opened 2021-05-26
Improved closed orbit functions

lfarv:
The closed orbit functions are improved in both Matlab and python in the following way:
-  a new function `find_orbit(ring, refpts)` for python and `findorbit(ring, refpts)` for Matlab groups all other functions: if radiation is ON, it calls `find_orbit6` to get the 6D closed orbit. If radiation id OFF, it computes the on-momentum orbit with `find_orbit4`, but an off-momentum value may be specified with the keywords `dp` for off-momentum or `dct` for path lengthening. It will then call either `find_orbit4` or `find_sync_orbit` as necessary.
- All functions, including the new one, now accept the `orbit` keyword argument: if provided, the search for the closed orbit at origin will be skipped, and the function will propagate the provided orbit to the selected refpoints.

 These improvements will make the future `linopt*` functions much cleaner, but can be convenient elsewhere.

### DIFF
--- a/atmat/atphysics/Orbit/findorbit.m
+++ b/atmat/atphysics/Orbit/findorbit.m
@@ -36,7 +36,7 @@ if isempty(orbitin)
         if isfinite(dp) || isfinite(ct)
             warning('AT:linopt','In 6D, "dp" and "dct" are ignored');
         end
-        orbitin=x_orbit6(ring,varargs{:});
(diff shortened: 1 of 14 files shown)


swhite2401 (approved):
Ok for me

willrogers:
What are the planned `linopt*()` functions going to be?

PR #260, opened 2021-06-03
deprecated order argument

swhite2401:
This branches implements a few fixes on the matching modules

- globalfit: replace deprecated order argument with index
- matching: introduce optional argument UseInteger, if false (default) only the fractional part of the phase is matched. This is to avoid computing the optics at all refpts to match the integer part of the phase which represent an enormous loss in speed
- the fit results display the absolute variation to avoid nans in case the initial value is zero


lfarv:
Sorry, I'm away… No objection, go on.

PR #261, opened 2021-06-04
allow for long field names and arbitrary type in ndarray

swhite2401:
Fixes related to problems in saving ESRF lattices to mat files:
-allow for arbitrary type in ndarray
-allow for long field names in scipy savemat


lfarv (approved):
Looks good!

PR #262, opened 2021-06-09
Losses

swhite2401:
This branches implements the possibility for `atpass` to output loss maps:
-a flag `losses` is introduced for backward compatibility, if `False` outputs particle distribution, if `True` an additional dictionary with loss maps data is provided. The default is `False`
-consistency with matlab AT enforced: default aperture limit is 1m, when a particle hits the aperture its first coordinate is set to `nan`, the others to zero. This is now the case for `losses=True or False`
-The history buffer present in matlab is not yet implemented

### DIFF
--- a/pyat/at.c
+++ b/pyat/at.c
@@ -71,6 +71,7 @@ static struct LibraryListElement {
     struct LibraryListElement *Next;
 } *LibraryList = NULL;
 
+
 static PyObject *print_error(int elem_number, PyObject *rout)
 {
     printf("Error in tracking element %d.\n", elem_number);
@@ -104,23 +105,17 @@ static struct LibraryListElement* SearchLibraryList(struct LibraryListElement *h
 
 
 static void checkiflost(double *drin, npy_uint32 np, int num_elem, int num_turn, 
-        double *xnturn, double *xnelem, double *xcoord, double *xlostcoord, 
-        bool *xlost, double *histbuf, npy_uint32 ihist, npy_uint32 lhist)
+        int *xnturn, int *xnelem, bool *xlost, double *xlostcoord)
 {
-    int n, c;
+    unsigned int n, c;
     for (c=0; c<np; c++) {/* Loop over particles */
         if (!xlost[c]) {  /* No change if already marked */
            double *r6 = drin+c*6;
-           for (n=0; n<6; n++) {	/* I remove the check on the sixth coordinate N.C. */
-                if (!isfinite(r6[n]) || (fabs(r6[n])>LIMIT_AMPLITUDE)) {
-                    int h, k=ihist;
+           for (n=0; n<6; n++) {
+                if (!isfinite(r6[n]) || ((fabs(r6[n])>LIMIT_AMPLITUDE)&&n<5)) {
                     xlost[c] = 1;
                     xnturn[c] = num_turn;
                     xnelem[c] = num_elem;
-                    for (h=0; h<lhist; h++) {
-                        if (++k >= lhist) k=0;
-                        memcpy(xcoord+6*(np*h+c),histbuf+6*(np*k+c),6*sizeof(double));
-                    }
                     memcpy(xlostcoord+6*c,r6,6*sizeof(double));
                     r6[0] = NAN;
                     r6[1] = 0;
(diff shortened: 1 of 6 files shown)


swhite2401:
@lfarv , @willrogers : instead of making of copy of ring in all threads of patpass I create a global variable accessible to all threads. The speed-up is a factor 10 for the ESRF ring on Linux machines.
Tests are failing on macos and windows -> global variable not found, I have no idea why this fails. For now I have implemented platform dependent behavior:
-linux: use global variable
-macos, windows: pass ring as argument to all processes

Do you have any idea idea on this issue?

pyat/at.c
```diff
@@ -97,6 +103,56 @@ static struct LibraryListElement* SearchLibraryList(struct LibraryListElement *h
...
+
+static void checkiflost(double *drin, npy_uint32 np, int num_elem, int num_turn, 
+        int *xnturn, int *xnelem, bool *xlost, double *xlostcoord)
+{
+    unsigned int n, c;
+    for (c=0; c<np; c++) {/* Loop over particles */
+        if (!xlost[c]) {  /* No change if already marked */
+           double *r6 = drin+c*6;
+           for (n=0; n<6; n++) {
+                if (!isfinite(r6[n]) || ((fabs(r6[n])>LIMIT_AMPLITUDE)&&n<5)) {
```
lfarv, line 115: Why not:
```
           for (n=0; n<5; n++) {
                 if (!isfinite(r6[n]) || fabs(r6[n])>LIMIT_AMPLITUDE) {
```

pyat/at.c
```diff
@@ -97,6 +103,56 @@ static struct LibraryListElement* SearchLibraryList(struct LibraryListElement *h
...
+
+
+static void setlost(double *drin, npy_uint32 np)
+{
+    unsigned int n, c;
+    for (c=0; c<np; c++) {/* Loop over particles */
+        double *r6 = drin+c*6;
+        if (isfinite(r6[0])) {  /* No change if already marked */
+           for (n=0; n<6; n++) {
+                if (!isfinite(r6[n]) || ((fabs(r6[n])>LIMIT_AMPLITUDE)&&n<5)) {
```
lfarv, line 141: Same thing here:
```
           for (n=0; n<5; n++) {
                 if (!isfinite(r6[n]) || fabs(r6[n])>LIMIT_AMPLITUDE) {
```

lfarv:
I think there are two many different things in this pull request:
- Losses processing is fine,
- Modifications on `set_cavity_phase`, if relevant, should be put in another PR. I am against removing completely the warning. What I suggest is to turn the message into a real Python warning (it's presently a simple print), so that suppressing it is simple enough: `warnings.simplefilter("ignore", SetPhaseWarning)`, but needs an explicit action from the user. Since the closed orbit problem is solved, there is to more any justification to use it, and naive users should be discouraged. Let's open a new PR for discussing that. I know, @swhite2401, you tried to push that silently, but I'm back home !   ;-)
- Same for `patpass`: it was about to be removed since I've always seen it less efficient than openMP. But if you have counter-examples, we can look at them in another PR, and try to understand the "global variable" problem.

So I propose to limit this PR to the losses processing (`at.c`, `track.py` and `test_basic_elements.py`), so that it can be merged rapidly, and move the other modifications to two other pull requests.

swhite2401:
Hello @lfarv ,
Good to have you back! I will split this as requested I convert this one to draft then.

Few answers:

- Losses: ok, I will answer your comments in the PR

- `set_cavity_phase`, I do use this function a lot for tracking, I find it very useful and I know for a fact that many others do! so I disagree with you. There are reasons to use it and as long as you know what you are doing it is fine. There is not reason to pollute users that have been using it for years without problems to discourage new users from using it. New users are supposed to read the doc and the message there is pretty clear. Also, we are not responsible for the usage they make of the code. I will make a dedicated PR on this

`-patpass`, I would like to keep this one, the speed up is very good on linux and it is more more convenient that OpenMP since you do not need to rebuild AT

lfarv:
Concernant `lattice_pass`, I see two possibilities:

1. the present one, with an input flag and a number of output arguments depending on the flag,
2. adding a new function (`lattice_pass_losses`?), so that each Python function has a fixed signature. Both will call `atpass` with the adequate flag. 

I have no strong preference, what's your advice, @willrogers and @swhite2401 ?

swhite2401:
Presently #1 is implemented but #2 is definitely a good alternative, let me create this new PR to move the discussion there (this afternoon probably)

swhite2401:
Developments related to patpass and changes on the default AT behavior / warnings have been remove

swhite2401:
@lfarv, I cannot see the comments on at.c anymore: in any case I need to check the last coordinate is finite because passmethods set it to infinity when the particle hits the aperture.
I suppose this is to keep track/save of the transverse coordinates in the output in at.c.
To be honest I kept exactly the same structure as the matlab implementation without thinking too much, in case you have better solutions it is probably time to consider them before the merge.

lfarv (approved):
OK for me

lfarv:
@swhite2401 : You are right, that's why I removed my first comments

PR #263, opened 2021-06-13
add method =0 to disable dipole fringe field

lnadolski:
I propose for clarity to add a condition to disable dipole fringe field, with the condition method == 0.


lfarv (approved):
Good idea!

lfarv:
For me, it can be merged.

PR #264, opened 2021-06-14
Patpass improvements

swhite2401:
This branch reduces the overhead in patpass as follows
 Instead of making of copy of ring in all threads of patpass I create a global variable accessible to all threads. 

On grappa (64 cores) running 500 particles for 500 turns on the hmba test lattice with radiations off, reuse=False, average of single calls:
-Single process:7.5s
-OpenMP: 3.2s
-patpass: 0.44s

comp

Some odd observations: 

`timeit.repeat()` with multiple calls:
patpass:[0.4279227163642645, 0.4299731105566025, 0.44496859423816204, 0.45715007930994034, 0.46737251803278923]
 atpass +OMP:[3.062729252502322, 1.318743273615837, 1.9151567667722702, 1.4526983611285686, 0.7289203442633152]
atpass: [7.544847888872027, 7.541526518762112, 7.5362925082445145, 7.540230557322502, 7.538320714607835]

It looks like atpass with openMP is much slower on the first execution and after that fluctuates a lots, is this expected?

Tests are failing on macos and windows -> global variable not found, I have no idea why this fails. For now I have implemented platform dependent behavior:
-linux: use global variable
-macos, windows: pass ring as argument to all processes

Until this issue is understood I leave this PR as draft

### DIFF
--- a/atmat/atphysics/Orbit/findorbit.m
+++ b/atmat/atphysics/Orbit/findorbit.m
@@ -1,14 +1,59 @@
-function orbit  = findorbit(RING,D, varargin)
-%FINDORBIT Alias to the orbit search functions: (findorbit4, findsyncorbit,
-%          findorbit4 ...) 
-%  depending on the number of parameters, their types and argument options
-%  It can also return additional parameters.
-%  
-%  NOTES
-%  1. Temporary version of 7/18/00 - defaults to FINDORBIT4
+function [orbs,orbitin]  = findorbit(ring,varargin)
+%FINDORBIT find the closed orbit
+%
+%Depending on the lattice, FINDORBIT will:
+% - use findorbit6 if radiation is ON,
+% - use findsyncorbit if radiation is OFF and ct is specified,
+% - use findorbit4 otherwise
+%
+%[ORBIT,FIXEDPOINT]=FINDORBIT(RING,REFPTS)
+%
+% ORBIT:        6xNrefs, closed orbit at the selected reference points
+% FIXEDPOINT:   6x1 closed orbit at the entrance of the lattice
+%
+%[...]=FINDORBIT(RING,...,'dp',DP) Specify the momentum deviation when
+%   radiation is OFF (default: 0)
+%
+%[...]=FINDORBIT(RING,...,'dct',DCT) Specify the path lengthening when
+%   radiation is OFF (default: 0)
+%
+%[...]=FINDORBIT(RING,...,'orbit',ORBIT) Specify the orbit at the entrance
+%   of the ring, if known. FINDORBIT will then transfer it to the reference points
+%
+%[...]=FINDORBIT(RING,...,'guess',GUESS) The search will start with GUESS as
+%   initial conditions
+%
+%For other keywords, see the underlying functions.
 %
 % See also FINDORBIT4, FINDSYNCORBIT, FINDORBIT6.
 
-orbit = findorbit4(RING, D, varargin{:});
+[orbitin,varargs]=getoption(varargin,'orbit',[]);
+[refpts,varargs]=getargs(varargs,[],'check',@(arg) isnumeric(arg) || islogical(arg));
+if isempty(orbitin)
+    if check_radiation(ring)    % Radiation ON: 6D orbit
+        dp=getoption(varargs,'dp',NaN);
+        ct=getoption(varargs,'dct',NaN);
+        if isfinite(dp) || isfinite(ct)
+            warning('AT:linopt','In 6D, "dp" and "dct" are ignored');
+        end
+        orbitin=xorbit_6(ring,varargs{:});
+    else                        % Radiation OFF: 4D orbit
+        [~,varargs]=getoption(varargs,'DPStep');   % Take it out because findorbit4 does not accept it
+        [dp,varargs]=getoption(varargs,'dp',0);
+        orbitin=xorbit_dp(ring,dp,varargs{:});
+    end
+    args={'KeepLattice'};
+else
+    args={};
+end
 
-return
+if islogical(refpts)
+    refpts=find(refpts);
+end
+if isempty(refpts)
+    % return only the fixed point at the entrance of RING{1}
+    orbs=orbitin;
+else
+    orbs=linepass(ring,orbitin,refpts,args{:});
+end
+end
--- a/atmat/atphysics/Orbit/findorbit4.m
+++ b/atmat/atphysics/Orbit/findorbit4.m
@@ -1,4 +1,4 @@
-function [orb4,fixedpoint] = findorbit4(RING,dP,varargin)
+function [orb4,orbitin] = findorbit4(ring,dp,varargin)
 %FINDORBIT4 finds closed orbit in the 4-d transverse phase
 % space by numerically solving  for a fixed point of the one turn
 % map M calculated with LINEPASS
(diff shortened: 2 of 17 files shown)


lfarv:
Concerning `patpass`, maybe it does not work on Windows and MacOS because it's not supposed to work !

Basically, the availability of the resources of the parent process in the child processes depends on the "start" method of the child. Only the "fork" method allows this, since in that case, initially the child is a clone of the parent: see [here](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods).

And the "fork" method is unavailable on Windows and discouraged on MacOS !

So your approach is probably the only one, though I would prefer to separate completely the "Unix" and "Others" implementations to avoid various tests.

However, looking down in the [documentation](https://docs.python.org/3.7/library/multiprocessing.html#programming-guidelines), there are 2 paragraphs:
"**Explicitly pass resources to child processes**"
and
"**Global variables**"
which are not clear to me but deserve some attention. Is moving the global variable to the module level a solution?

lfarv:
I am surprised by your OpenMP results, since on grappa I've seen reductions of the processing time by factors between 10 and 15, while you get hardly a factor 3. For 64 cores, it's indeed disappointing. It's true that what happens inside OpenMP in "auto" mode is a kind of mystery. For instance I'm not surprised by the slower 1st run.  I'll have a look.

swhite2401:
Thanks @lfarv for the doc, in fact it is a bit complex.
The test I have done without the global variable showed no or very little improvement because of the huge overhead...I would simply suggest to disable this function for "other", we could think of writing a decorator such checkradiation()
What do you think?

swhite2401:
Yes OpenMP is a bit herratic...you can see the last case in my example shows a factor 10 improvement but I was indeed surprised of the difference, for completeness: no other tracking function is called before the "timed" ones

lfarv:
Apparently, multiprocessing results vary a lot with the computer. Here (iMac 2017, 4 cores), for the same example as @swhite2401 (hmba test lattice, radiation OFF, 500 particles, 500 turns), I get on average:

single process: 10.9 s
OpenMP: 3.1 s
patpass: 3.4 s  (without global variable)

1. All results are very stable (0.1 s fluctuation),
2. OpenMP is here slightly faster than patpass,
3. The ratio for both multiprocessing runs is close to the expected factor 4, even without the global variable in `patpass`.

So I think we should indeed keep both possibilities so the the user can select the fastest one for his case…

swhite2401:
Ok this makes sense in fact: the large the number of processes the larger the overhead of passing the ring argument.
I open this PR for review then!

Any comments on the implementation?

lfarv:
Given the strange results obtained up to now, I did systematic measurements again on both Linux (grappa) and MacOS.
There are some differences with what was observed previously.

1. on grappa, the behaviour of OpenMP with default parameters looked strange. However it can be improved by specifying in `lattice_pass` the desired number of threads. I varied rapidly between 32 and 128 (number of CPUs) and found an optimum around 56. Why? And why the default behaviour is so bad ?
2. Unlike what was observed by @swhite2401, the influence of setting `ring` as a global variable seems negligible. which makes sense to me since the impact of a simple copy of data compared to a long computation should not be tremendous. Simon, could you repeat your tests? I do not understand.

**Results on grappa (128 CPUs):**
| grappa| radiation OFF | radiation ON |
| - | - | - |
| single process | 11.6 | 25.6 |
| `patpass` without global | 0.83 | 1.00 |
| `patpass` with global | 0.64 | 0.80 |
| OpenMP | 4.90 | 5.75 |
| OpenMP, 56 threads | 1.00 | 1.30 |

**Results on MacOS (4 CPUs)**
| MacOS| radiation OFF | radiation ON |
| - | - | - |
| single process | 11.0 | 32.5 |
| `patpass` without global |3.3 | 9.0 |
| `patpass` with global | 3.4 | 9.1 |
| OpenMP | 3.1 | 9.1 |

Concerning the use of a global variable, it must be restricted to Linux. I also tested the "fork" start method on MacOS though it is discouraged (I don't know why) with no benefit. On Linux the benefit is small, and might create problems if in a next python version the new "forkserver" start method becomes the default.

Concerning the implementation, I did my tests with a much simplified version which I can propose.

Also, the `patpass` function should be a replacement of `lattice_pass` rather than `atpass` which is not intended to be called directly. It is maybe too late to change the function name, but at least the arguments should be the same: for instance the `reuse` keyword should be replaced by `KeepLattice`. And it might be ignored, assuming always False, rather than be rejected if True.

swhite2401:
Hello @lfarv, below updated results for `npart=numpy.zeros((6,500), nturns=500` averaged over 5 runs for different cases.

Here are my observations:

1. I can see a clear benefit of the global variable
2. The speed-up depends on i) the size of the ring ii) the number of copies (or processes)

I do not understand why you results are so different but it seems that tests done on a 4 CPU machine are not that relevant.
Are you sure you recompiled properly with OpenMP to test patpass? The 2 should not be activated together because you will overload your processors (I am asking because I did the mistake ;))

As an alternative we may look at the multiprocessing Manager that allows to share variables between processes:
https://docs.python.org/3/library/multiprocessing.html
I am not sure we can use it with the lattice object as it seems to apply only to lists, but since Lattice inherits from list there could be a way (I haven't looked at it...)?

A simpler implementation is of course more than welcome!


**Results on grappa (128 CPUs) for test hmba.mat:**
| grappa| radiation OFF | radiation ON |
| - | - | - |
| single process | 7.95 | 18.3 |
| `patpass` without global | 0.84 | 1.05 |
| `patpass` with global | 0.45 | 0.54 |

**Results on grappa (4 CPUs) for test hmba.mat:**
| grappa| radiation OFF | radiation ON |
| - | - | - |
| single process | 7.92 | 18.3 |
| `patpass` without global | 2.32 | 4.94 |
| `patpass` with global | 2.32 | 4.92 |

**Results on grappa (128 CPUs) for betamodel.mat (ESRF lattice, single run):**
| grappa| radiation OFF | radiation ON |
| - | - | - |
| single process | 360.09 | 674.2 |
| `patpass` without global | 38.26 | 41.34 |
| `patpass` with global | 4.61 | 7.93 |

lfarv:
@swhite2401: did you track with particles at zero ??? That's unfair, all numerical routines have  special treatment for zero argument. For hmba without radiation, that explains the difference. My results are for horizontal motion only. For radiation ON, this does not apply and results are more compatible.

Yes, for these tests I recompiled without OpenMP. Normally, for a single particle, OpenMP should be inactive (the threshold is at 10 particles), but I confirm that it degrades the performance of `patpass`. Another mystery…

Concerning shared variables, a list is good enough: `lattice_pass` works with any sequence. Anyway [multiprocessing.shared_memory](https://docs.python.org/3.8/library/multiprocessing.shared_memory.html#module-multiprocessing.shared_memory) allows to share anything, the example is given for a numpy array. It looks like a cleaner solution but requires more work to split the job and gather the results…

One could also look at multithreading: the "Global Interpreter Lock" prevents form running two python interpreters in parallel, but what needs to be parallelised is the C-code of `atpass` Maybe it can work?

swhite2401:
Well I did not think it would make a big difference since in AT the rin[0]=NaN has special treatment not zero.
In fact I do not see much different with `npart=numpy.zeros((6,500)+1.0e-6`, see table below (only for one case).
The benefits with global are the same (about factor 2 for this case).
Could you share your implementation? This is the only difference I can imagine that would explain our different results.

I think we have to keep a solution that does not imply to rebuild AT. This is what makes `multiprocessing` very convenient and we should keep it this way. In fact if this proves to work for all OS I would be in favor of getting rid of OpenMP to simplify the C code. 
The only real interest for me to parallelize a the C level would be to introduce a full MPI implementation.

Let's see if the `shared_memory` option allows it to run on windows and macos.

Question: is there a similar implementation in matlab using parfor?

**Results on grappa (128 CPUs) for test hmba.mat:**
| grappa| radiation OFF | radiation ON |
| - | - | - |
| single process | 7.93 | 18.26 |
| `patpass` without global | 1.02 | 1.10 |
| `patpass` with global | 0.45 | 0.57 |

lfarv:
@swhite2401 : I do not think of removing OpenMP because:
- except on grappa it gives better performance (it may be a problem with multi-user machines ? I get very fluctuating numbers)
- it accelerates everything (closed orbit, transfer matrices…) transparently, as soon as the number of particles is more than the threshold. From my experience, I use 4 instead of the default 10, which looks like a good optimum for desktop machines,
- it allows to track through elements involving collective effects (ImpedanceTable, BeamLoadingCavityPass) where `patpass` fails (by the way, one should put a test on passmethods when entering `patpass`),
- the "global variable" trick is undocumented, as far as I could see, and may fail in the future: it' a specificity of the fork start method. Also it may create problems for multithread programs (if a user wants to include AT in a multithreaded python program). In that respect, python multiprocessing looks more fragile.

Concerning recompilation, you can turn it the other way: I am using it permanently, except for timing mesurements, so I have to recompile to use `patpass` !

If someone wants to spend time on it, OpenMP and MPI can work together (see [here](http://www.intertwine-project.eu/sites/default/files/images/INTERTWinE_Best_Practice_Guide_MPI%2BOpenMP_1.1.pdf)), but this really needs work…

Here is my implementation: it should not change anything, I just simplified and matched the signature with lattice_pass:
[patpass.py.txt](https://github.com/atcollab/at/files/6680897/patpass.py.txt)

swhite2401:
Ok you are right, I have not observed any improvement on my side with OpenMP for standard AT use and the behavior is erratic which is something I don't like at all  but this could be due to different configurations as you mentioned.

The test can be added indeed also I need to check how it behaves with loss maps activated.

Once this is done, I propose to merge `patpass` as is after some clean-up. This is a clear improvement and simple parallelization method that I find very useful for some applications.

Concerning potential future issue: `shared_memory` is only available from python3.8 and is certainly a fall back solution in case this trick fails in future python releases, I have tried the `manager.list()` available in earlier versions, but it returns a `ListProxy` object and converting back to pure python `list` for `atpass` has large overhead.

lfarv:
Ok for merging as is: missing the global variable has no impact on desktop machines, so we can keep it like that. Did you try my version ? As soon as you are happy with your checks, let's go on, we'll see later if it can be improved.

In the end, the fastest way to run multi-particle tracking must be evaluated for each platform, users can choose…

swhite2401:
Ok I think this is close to the final version. I have removed unused argument and added the possibility to disable the global variable. The output is consistent with `atpass` with and without losses and `r_in` is also modified in the same way.
In case` ImpedanceTablePass` is found (`BeamloadingCavityPass` not distributed?) or `npart=1 atpass` is called.

lfarv:
I get a crash for `refpts=[].` This is commonly used in `lattice_pass` to get the results at the end of the tracking (in `r_in`) without outputting the values at the end of each turn. Even if this behaviour cannot be reproduced, it should not crash, but return an empty array of the adequate shape.

Also, the results in `r_in` are different of those of `lattice_pass`: you are returning the last refpoint, while `lattice_pass` returns the end of the ring, even if it's not a refpoint. To reproduce that, one should add  "end of the ring" to the refpoints,  take the last value to fill `r_in` (as you are doing), and remove it from the `r_out` if necessary… Quite tricky. It's the consequence of a bad design of `lattice_pass` (my fault…) and I have no elegant solution, apart from stating that getting results in `r_in` is not allowed in `patpass`. It would simplify `format_results`.

By the way, `format_result` illustrates the inconvenience of a variable output signature, rather than defining separate functions, as we discussed [here](https://github.com/atcollab/at/pull/262#issuecomment-860523294). It can probably still be changed, with also duplicating `patpass` (`patpass_losses`).

I also would like to accept all the arguments of `lattice_pass` so that `patpass` behaves as an exact replacement. These are `keep_lattice` and `omp_num_threads`, which can be ignored, even silently, since they do not affect the results. A `**kwargs` argument is enough, unless we want to signal stupid keywords. And similarly for `pool_size` in lattice_pass (could it be a sinonym of  `omp_num_threads` ?)

swhite2401:
Thanks! I can propose this...it is not very elegant but fixes all the above issues I believe.

I agree the variable output function is painful but multiplying functions is also not very nice, we want to keep the code compact. Ideally we would like to have atpass output losses all the time (empty when losses=False) or even the  perform the full computation as the overhead is minor (7.92s w.o, 7.95s with).
Providing the very few users of pyAT, now is probably the time to improve the design of central functions like this one, we could start by replacing all references to lattice_pass in AT by the "new" function with losses and add a deprecation warning (similarly we can remove all reference to linopt and replace them by your upcoming function).

What do you think? My feeling is that if design issues are identified they should be fixed...and released with the next AT version

lfarv:
OK for this version (you are right, it's not elegant…).

Concerning the "cleaning" you propose: `atpass` is not intended to be called directly: I checked, apart from tests, if appears in pyat only in `lattice_pass` and `patpass` (6 occurrences in 2 functions). So there could be a new C function identical to `atpass`, but always returning losses, possibly empty. And for compatibility, it's easy to provide a wrapper in C totally ignoring losses, both input and output.
That would clean at least the lower level, in a 'compact' way.
Concerning lattice_pass, it's used in many places (~30 occurrences), and everywhere the losses results is not relevant. I think that's also to case for user code. So having an additional unused output is there really inelegant, that' s why I am more in favour of another function for losses studies. Or keep the keyword if you prefer…

For now, I propose to merge `patpass` as it is, and start cleaning the code step by step.

PS: BeamLoadingCavityPass (#191) is pending: I expected @carmignani to provide a test/example function (it's use is not obvious), but it should be added to the distribution. Lengthening the list of "forbidden" passmethods has to be easy enough.

lfarv (approved):
OK

swhite2401:
I have added `_collective` list in `lattice.elements` that can be easily modified when other forbidden passmethods come up
I will then merge this one as is, pending the improvements discussed above

willrogers:
This is another one where it might be worth tidying up the commits.

swhite2401:
@willrogers sure I can squash all commits when merging or you have a better way to do it?

willrogers:
No, I think that's fine.

PR #265, opened 2021-06-14
Default behavior

swhite2401:
This branch is a placeholder for discussion, for now 2 changes are propose that as a user I would find very beneficial:

-remove warning messages in energy_loss, in my opinion a warning in the `docstring` is sufficient

-set default `radiation_on()` with `quadrupole_pass=auto`: this has been the source of many mistakes and wasted time at ESRF, I do not see any benefit in leaving this disabled by default. With modern computer, the overhead is not so relevant anymore and for standard "on-axis" usage results will be unchanged and will remain correct for DA computations, off-momentum or strong radiation lattices etc... 

This is a draft, and changes were applied only to pyAT as examples. I believe it would be good to have an open discussion on these topics that seem to be recurrent requests/issues.

Please comment and propose other modifications in case you have any.

### DIFF
--- a/atmat/atphysics/LongitudinalDynamics/atSetCavityPhase.m
+++ b/atmat/atphysics/LongitudinalDynamics/atSetCavityPhase.m
@@ -10,6 +10,9 @@ function newring=atSetCavityPhase(ring,varargin)
 %   CAVPTS is the location of RF cavities. This allows to ignore harmonic
 %   cavities.
 %
+%WARNING: This function modifies the time reference,
+%this should be avoided
+%
 %NEWRING=SETCAVITYPHASE(...,'method',METHOD)
 %   Choose the method for computing the energy loss per turn
 %
@@ -27,9 +30,6 @@ freq=unique(atgetfieldvalues(ring(cavities),'Frequency'));
 if length(freq) > 1
     error('AT:NoFrequency','RF frequency not equal for all cavities');
 end
-warning('AT:CavityTimeLag',...
-    ['\nThis function modifies the time reference\n',...
-       'This should be avoided, you have been warned\n']);
 U0=atgetU0(ring,'periods',1,varargs);
 if U0 > rfv
     error('AT:NoEnoughRFVoltage','Not enough RF voltage, unstable ring.');
--- a/atmat/atphysics/LongitudinalDynamics/atsetcavity.m
+++ b/atmat/atphysics/LongitudinalDynamics/atsetcavity.m
@@ -1,6 +1,9 @@
 function ring = atsetcavity(ring,varargin)
 %ATSECAVITY Set the cavity parameters
 %
+%WARNING: This function modifies the time reference,
+%this should be avoided
+%
 %ATSETCAVITY may be used in two modes:
 %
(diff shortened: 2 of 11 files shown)


simoneliuzzo (approved):

lcarver:
I am 100% supportive of radiation_on using quadrupole_pass='auto' as default. 

The warning message removal is fine for me.

lcarver (approved):

lfarv:
Nothing against having `quadrupole_pass=auto` as default if we accept to break the backwards compatibility: the same code may now give different results.

Just a question for my understanding: why would you put radiation in quads, which represent less than 1.e-3 of total losses for hmba (a lattice with "low" dipoles losses) and a reasonable closed orbit ? And 0 for all the time consuming matching studies performed with zero closed orbit ? I would think that radiation in quads is just the final touch when all the works is done…

lnadolski:
Dear All
We should keep radiation in quad. This is an important feature depending on the accelerator.
The solution proposed should be generic and not specific to a given accelerator.
Laurent.

swhite2401:
@lfarv, just to answer your question: many machines are now looking at options with misaligned quadrupoles to introduce anti-bends, off energy operation etc... so radiation in quads is needed. I fully agree that the default should provide correct results for generic lattices, which is not the case presently. The overhead for the ESRF lattice is 20% which is acceptable for most applications.
Matching is the only case where this can be detrimental but it is done with radiations off generally, so no issues there.

lfarv:
@swhite2401 : I see your point, and again I have nothing against changing the default behaviour: the penalty of 20% is acceptable. But do these cases represent the majority of AT uses ?
@lnadolski : the question is about the default behaviour: should we keep the present behaviour where quadrupole radiation must be explicitly requested, or switch to @swhite2401's proposal to have it enabled unless explicitly disabled?

@swhite2401 and @lnadolski, I let you come to an agreement, in any case overwriting the default behaviour is easy, just a keyword argument to set…

lfarv:
Since we are discussing default values, there is another one which is for me more important:

the default passmethod for quadrupoles is `QuadLinearPass` which I think is far from being of any use. Should not it be `StrMPoleSymplectic4Pass` ? Same question for dipoles…

swhite2401:
Ok for the change of default passmethods (already done in pyAT right?).

lnadolski:
I do agree *Symplectic4Pass for all elements.

> Since we are discussing default values, there is another one which is for me more important:
> 
> the default passmethod for quadrupoles is `QuadLinearPass` which I think is far from being of any use. Should not it be `StrMPoleSymplectic4Pass` ? Same question for dipoles…

this makes a lot of sense for all dipoles, quadrupoles and multipoles.

swhite2401:
I have updated this branch for both matlab and pyAT:
-`radiation_on()` default enables radiations in quadrupoles
-Move print/warning in terminal to the help in `set_cavity_phase()` / `atsetcavity()`
-Set default passmethods to *Symplectic4Pass for all elements

Change of version was mentioned in #257 , I believe we could think of merging this one in the new version too.

lfarv (approved):

PR #267, opened 2021-07-16
improved fastring

swhite2401:
Hello @lcarver, looking into beamloading issues, I have found several issues in the way fastring was generated.
Here is a temporary corrected implementation.
It can be significantly simplified after #257 is merged so I leave it on a separated branch for now.
Please take a look and let me know.


lcarver:
The merge_ring subfunction was not working well as it modified the all_rings[0] part of the lattice (pointers...). Some deepcopies and renaming of the variables made it work again.

lcarver (approved):

swhite2401:
Ok thanks, now that #257 is merged I will propose a new simpler version before the merge

lfarv:
@swhite2401: you are welcome to make it simpler ! It looks ugly right now…

swhite2401:
@lcarver, @lfarv here is a simpler version with a bit of cleaning up. I think this is good enough.
Let me know in case you see potential improvements.

swhite2401:
@lfarv, I have a question on change track: I did rebase the master in order to integrate the linopt6 changes but now it tells me there are 36 files changed, which is wrong, there are only 3. Do you know why it is doing this?

lfarv:
@swhite2401 : I already notices this behaviour, but I don't know why… I guess it records all modifications **since you forked** your branch, so rebased modules are counted. So now I avoid rebasing unless it's necessary. Otherwise, if very few files are modified, you can start a new branch and copy the really modified files from the old branch to the new one. By renaming the branches, maybe you can keep the pull request as it is ? I don't know…

lfarv (approved):
For me, that's OK. I did not test it, but if @lcarver approves, that's good enough. I'll need to modify the Matlab version in the future and I'll take a closer look then, but for the time being, go on !

lcarver:
Few comments:
1. it is much slower (24s vs 6s for previous).
2. It fails when there is more than one cavity frequency present. It fails when computing the chromaticity because of the functions get_rf_frequency and set_rf_frequency does not work in the presence of multiple frequencies.

swhite2401:
I have fixed these issues, all related to the detuning element:
-chromaticity is derived from a ring with `radiation_off()` to avoid uncertainty on RF
-the detuning elementis computed only once for the ring without radiation and then a copy is passed to the function that generates the fastring with radiations

lcarver:
Looks good to me.

lcarver (approved):

PR #270, opened 2021-08-11
get_optics automated selection

swhite2401:
This branch implements an automated method selection in `get_optics` in case method=None.

Few other changes in linear.py:
-avoid duplicated code
-ouptut same values in linopt and linopt4

Few observations that could be corrected in this branch:
-twiss_in option does account for mu0 -> this could be useful when matching part of a ring
-get_chrom, get_w functions disabled with twiss_in -> again useful for matching (especially the w function, in this case we need db/dp and da/dp to reconstruct the initial conditions, for now we only have access to W)

### DIFF
--- a/pyat/at/physics/linear.py
+++ b/pyat/at/physics/linear.py
@@ -656,7 +656,7 @@ def linopt6(ring, *args, **kwargs):
     return _linopt(ring, analyze, *args, **kwargs)
 
 
-def get_optics(ring, refpts=None, dp=None, method=linopt6, **kwargs):
+def get_optics(ring, refpts=None, dp=None, method=None, **kwargs):
     """Perform linear analysis of a fully coupled lattice
 
     elemdata0, beamdata, elemdata = get_optics(lattice, refpts, **kwargs)
@@ -672,8 +672,10 @@ def get_optics(ring, refpts=None, dp=None, method=linopt6, **kwargs):
                         3) a numpy array of booleans of maximum length
                            len(ring)+1, where selected elements are True.
     KEYWORDS
-        method=linopt6  Method used for the analysis of the transfer matrix.
-                        Can be at.linopt2, at.linopt4, at.linopt6
+        method=None     Method used for the analysis of the transfer matrix.
+                        Can be None at.linopt2, at.linopt4, at.linopt6
+                        None:       automatically selected depending on radation
+                                    and coupled flags
                         linopt2:    no longitudinal motion, no H/V coupling,
                         linopt4:    no longitudinal motion, Sagan/Rubin
                                     4D-analysis of coupled motion,
@@ -716,6 +718,13 @@ def get_optics(ring, refpts=None, dp=None, method=linopt6, **kwargs):
         chromaticity    Chromaticities
         damping_time    Damping times [s] (only if radiation is ON)
     """
+    if method is None:
+        if ring.radiation
+            method = linopt6
+        elif kwargs.pop('coupled', True)
+            method = linopt4
+        else:
+            method = linopt2
     return method(ring, refpts=refpts, dp=dp, **kwargs)
 
 
--- a/pyat/at/physics/linear.py
+++ b/pyat/at/physics/linear.py
@@ -87,6 +87,18 @@ def _closure(m22):
         return numpy.NaN, numpy.NaN, numpy.NaN
 
 
+# noinspection PyShadowingNames,PyPep8Naming
+def _tunes(ring, **kwargs):
+    """"""
+    if ring.radiation:
+        mt, _ = find_m66(ring, **kwargs)
+    else:
+        mt, _ = find_m44(ring, **kwargs)
+    _, vps = a_matrix(mt)
+    tunes = numpy.mod(numpy.angle(vps) / 2.0 / pi, 1.0)
+    return tunes
+
+
 def _analyze2(mt, ms):
     """Analysis of a 2D 1-turn transfer matrix"""
     A = mt[:2, :2]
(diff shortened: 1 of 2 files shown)


lfarv:
Hi @swhite: sorry to disappoint you, but I think the `None` method is dangerous, because it makes choices without the user's knowledge. Let me explain how I see the problem:

As it is, `get_optics` uses by default the `linopt6` method, because:
- it always works,
- for coupled beams it gives results which are in most cases easier to use or interpret, because they are expressed in the standard AT axis (radial, vertical and longitudinal). For instance, in the 4D case, if you know the emittances of the two independent modes, you can easily compute the beam sizes along the horizontal and vertical axes (SIGMAX = sqrt(EPS[0]*R[0,0,0] + EPS[1]*R[1,2,2])). With `linopt4`, it's much more complicated !
- it's as fast as `linopt4` in similar conditions

So in my opinion, the present `get_optics` is best suited to the majority of uses (maybe not matching, I'll come to that later…) because:
- you know which method it will use, and what output you'll get (or if you don't know, you'll get anyway correct results),
- you don't need to specify whether you have radiation or coupling, it just does what you ask for,
- the default behaviour is always correct, even if not optimal.

The reasons why you would select another method are very few. I just see:
- you know what the Sagan/Rubin method does, and you are interested in the A,B,C or gamma values, so you select `linopt4`,
- you want to save some time when you know there is neither radiation nor coupling, so you select `linopt2`.

An example of potential problem with a `None` method; you have a ring with H/V coupling, you look at it without longitudinal motion, you get some beta values. You turn a cavity ON (let's assume no radiation) and look again: you get different betas!!! The machine is the same, but by switching silently to `linopt6` instead of `linopt4`, you get another view of the same thing. Looking at both with `linopt6`, you'll get exactly the same result, no surprise.

So to come back to the particular case of matching, it's a case where speed is critical. So it's there where you would select the fastest method, for the specific lattice you enter. There I could understand a `None` method, selecting `linopt2` or `linopt6` as appropriate (I would skip `linopt4`, but  that's a matter of taste). Otherwise you still need to specify the right method if you want to fit a parameter specific to a given one.

To summarise, you should move the automatic choice of the best method to the matching module, because for another use, the best choice might be different.

lfarv:
Otherwise I think you are right to restore the output of A, B and C in `linopt4`. I removed them for speed reasons, because they are rarely used. But they may be useful in some cases, and it's cleaner as you put it. So OK for the reorganisation!

swhite2401:
Speed in my opinion matters for several applications running loops. Matching is one, the simulators (that is now running pyAT thanks to @simoneliuzzo btw!) is another and I can think of others...
I agree the default should always give the correct results though, how about keeping `method=linopt6` as default but creating another method `linoptauto` ? 
That would be a few lines in linear.py and would avoid bringing in some tests in the matching modules. Skipping linopt4 is in fact a good idea.

lfarv:
I rapidly looked at various uses of `get_optics` and situations where the default `linopt6` would not be optimal, and there are cases where `linopt4` would be interesting (`avlinopt` for instance). Following you proposal, one should then imagine different keywords like 'auto_for_speed', 'auto_for_xxx' and so on. Added to the fact that all this needs reintroducing at least the '`coupled`' keyword that has no nothing to do there except helping for the choice, and possibly other keys for different decision making schemes… Finally I am really convinced that for clarity, the choice of the method (other than `linopt6`) has to be made and documented where it belongs. There you can keep method=None, as the default or not, to indicate that the choice has to be made automatically. If the same scheme is to be used in different places, you could write a helper function for that, but it's so simple that I find more self-explanatory to expand it wherever needed (but that's a matter of taste, you may disagree).

In fact I am not fond of "automatic" things, in particular here where you have to help it with this 'coupled' thing… If you know that the ring is uncoupled, why don't you ask for 'linopt2' directly? If a user is unaware of what it means, he will get a slower (but correct) result, and if he looks for speed, that's an opportunity to learn what's behind `get_optics`. Doc strings are made for that, they just need be improved… Never trust a result if you don't know how it's obtained… But this approach means that there are protections against wrong manipulations. `linopt4` is protected against longitudinal motion but we are missing an error if `linopt2` is called with a coupled ring. A test on non-zero off-diagonal blocks of the transfer matrix is not sufficient, but it's fast and would detect the majority of cases. To be done!

So if you want to put an 'automatic' option in matching, do it there, but keep `get_optics` clean. It's true that
in this case, one could accept this kind of "black box" behaviour. Concerning the simulators, I think the author is aware enough of optics computation to choose explicitly the best method for each case, but you're free !

swhite2401:
I am afraid I disagree with you:
-I very much prefer a single function that is located in a well defined place, optics calculations belong to `linear.py` so this is where it should appear in my opinion
-automated functions are very useful in some instances, in fact you implemented one yourself in `find_orbit`, these are found in many codes and are in general much appreciated by the users
-the `coupled` keyword will remain in some form as long as we support a function that assumes uncoupled lattice, if we remove it from AT it will appear in the users scripts, I personally prefer to have it in AT 

Generally speaking AT is a beam dynamics code used by accelerator physicists, we can assume that they know what they are doing, however I agree that we show try to protect against wrong manipulations, but the most powerful tool for that is a well written documentation which is obviously missing from AT.

Also if we are going to be so "strict" for new developments, we should also try to clean-up the existing code, for instance why is GWSymplecticWigglerRadPass.c still distributed, while it has obvious bugs? This is probably the worst thing you can think of for a fresh user...

I will write the `helper`  function outside of get_optics to keep it clean as you suggest, the default will be linopt6 both in get_optics and matching so that it provides correct results in all configuration.
Switching to other methods is the users responsibility but it is also his responsibility to understand the consequences (described in the docstring).

swhite2401:
I believe this one is ready to merge

PR #271, opened 2021-08-12
Matching with linopt2/4/6

swhite2401:
This branch implements some modifications of matching function to integrate the developments from #257

**-globalfit.py:** 
-use the the function from linear.py
-compute the jacobian only once

**-matching.py:** 
-use `get_optics` and `find_orbit` functions
-Providing linopt and orbit constraint are now compatible with radiation_on() no ring copy is done to handle combinations of `EnvelopeConstraints` and other constraints
-In case `EnvelopeConstraints` is called for a ring without radiations, the decorator of `Ohmi_envelope` will throw an error

To be noted: this branch relies on #270 to select the proper optics method in case radiation is on

### DIFF
--- a/pyat/at/matching/globalfit.py
+++ b/pyat/at/matching/globalfit.py
@@ -4,19 +4,18 @@ such as the tune and the chromaticity
 """
 import numpy
 from at.lattice import set_value_refpts
-from at.physics import linopt
+from at.physics import get_tune, get_chrom
+from at.matching import matching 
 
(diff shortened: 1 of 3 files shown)


swhite2401:
@lfarv, it would be good to review this one with a bit more attention, thanks!
I believe the `get_chrom` flag needs to ON only to compute chromaticity now, right?

lfarv (approved):

PR #274, opened 2021-08-16
Web documentation

lfarv:
This is the basis for upgrading the AT Web site on GitHub Pages. This site is almost empty right now, and could include most of what has been prepared in the ATdoc project.
- the theme is changed to "minima", which is the only one to include a navigation bar on top of the pages. It's not so nice, but coolers will come later,
- the theme is upgraded to allow math equations, using markdown and LaTex in pages,
- the site was only referring to Matlab, there are now distinct branches for Matlab and python,

The AT web site is [here](https://atcollab.github.io/at/), but GitHub Pages generated the site from the master branch only. So you will not see the result until the branch is merged. I could only test it locally, so there is no guarantee that it will be nice on the first commit…


willrogers:
Good idea. I'm happy to merge, we can always improve it later.

lfarv:
Ok, it's at anyway better than it was. The idea is now to move here, step by step and after updating,  the material gathered in the sleeping "atdoc" project.

PR #275, opened 2021-08-16
Matlab documentation

lfarv:
The AT Documentation is now accessible in the Matlab help browser:

on "Documentation Home", there is now an item called "AT: Accelerator Toolbox" which give access to a preliminary AT documentation.

This documentation is generated with .m and .mlx files, and scripts grouped in a new "developer/mdoc" directory in the AT project. This directory has not to be in the Matlab path for users, it's only for developing the Matlab documentation. The doc is preliminary, but preparing it as .mlx files allows formatting including mathematical expressions.

In addition, the Git repository is now set as a Matlab project, which helps for validating the whole toolbox. It can be opened by opening the at.prj file with Matlab.

And finally, there is preliminary support for the distribution of AT in Matlab Central, the community repository for Matlab toolboxes. This will allow the **installation in one click** !


swhite2401:
Sounds good to me

swhite2401 (approved):

Issue #276, opened 2021-08-19
Inconsistent radon radoff between matlab and python


swhite2401:
In `matlab` element types are detected looking at attributes likes `BendingAngle`, `PolynomB` etc... while in python it checks the class.
This leads to issues when a lattice in `matlab` can be generated with radiations on in multipoles. This lattice if imported in `pyAT` cannot be used as `ring.radiation_off(`) does not handle multipoles.

It would to make sure the two implementations are consistent. As a temporary fix I will turn radiations in multipoles in `ring.radiations_off()`

willrogers:
Can this be closed?

swhite2401:
Yes I will close it, matlab cannot handle classes anyway

PR #277, opened 2021-08-19
handle multipole in radon/off

swhite2401:
This branch adds the capability to handle `at.Multipole` in `radiation_on()` and `radiation_off()` in pyAT

Default:
-`radiation_off()` acts on ALL elements including multipoles
-`radiation_on()` turns the radiation on in `Dipole` and `Quadrupole` only

Options allow to select which element classes to act on.

This solves the issue #276 on the `pyAT` side, a more general discussion on consistent detection algorithm in both codes is nevertheless necessary.


swhite2401:
Ok to merge?

PR #282, opened 2021-08-31
Small fix to matching

lcarver:
the matching script keeps failing with AtError('Unstable Ring'). Naturally in matching this is to be expected so needs to be handled. I made a quick bug fix but maybe some small additional modifications are needed to clean it up.


lcarver:
In fact this is a larger problem that requires more discussion. I will close this pull request and open an issue

PR #283, opened 2021-08-31
Matlab toolbox

lfarv:
#### AT is now distributed directly from the [Mathwork web site](https://www.mathworks.com/matlabcentral/fileexchange/).
This is now the recommended way of installation (instructions [here](https://atcollab.github.io/at/minstallation.html)): it is a one-click installation and avoids all problems with the Matlab path.

For this, we needed:

- a modification of the Matlab "project",
- an upgrade of the help sections of a few AT functions,
- an improvement of the documentation generation.


willrogers (approved):
This seems like a good thing in principle.

swhite2401 (approved):
ok

PR #284, opened 2021-09-03
At 2 elegant cavity and multipole conversion

simoneliuzzo:
the multipole conversion in AT_2_elegant was missing the *L for integrated strength. this is fixed in this branch. 
Also, the cavity parameters are taken from the lattice using atenergy


lfarv:
This should be an opportunity to change the function names, which are inconsistent:

`elegant2at` / `AT_2_Elegant`
`atfrommadx` / `AT_2_madX`

Up to now, the naming was like:
`atreadxx` / `atwritexx`
It can be something else, but reading and writing should be symmetrical…

PR #285, opened 2021-09-05
WIP: proposed module for DynamicAperture,  MomentumAcceptance, 6D Acceptance computation

simoneliuzzo:
features:
* DA computation
* Local momentum acceptance 
* 6D grid tracking
* plots
* computation at given location in lattice
* example/tests

known issues, missing features:
* module not exploiting multiple cores (may be easy to include for 6D Acceptance)
* potentially unwanted `if __name__ `
* reference to files not in repository


lfarv:
Hi @simoneliuzzo : importing matplotlib is forbidden outside the `at.plot` package, which has to be **explicitly** loaded (for performance reasons). So all plotting functions must belong to this package! The idea is that the computation functions are in the `at.physics` package (usually), and only the plotting belongs to the optional `at.plot`.

swhite2401:
Nice development, it would be interesting to include the possibility to track a gird (for 2D aperture) or a vector (1D) of particles to make use of multicore capabilities, either with `OpenMP` or `patpass`, you will have to send the full array of particle to your tracking function for that

Also, I am wondering whether we should start a new folder for lifetime and DA applications.

Cosmetic comments:
-it is better to use `lattice_pass` instead of `atpass` directly 
-we have been avoiding importing matplotlib by default

simoneliuzzo:
@swhite2401  I will try to make the parallelization changes as soon as the tests do not fail. It is already possible to define grids of 1, 2, 3, 4, 5 or 6 dimensions. the 2D grids have a somehow simplified input, to try to be user friendly

@lfarv I will try to minimize the use of  matplotilb  by moving the import of matplotlib.pyplot to the Acceptance6D.plot_base and compute method . I will in any case need the contour function from matplotlib, that is the simplest way I found to define the DA border. Matplotlib would then be imported only if a plot is required or an arbitrary dimension grid border is required. in the basic cases of dynamic_aperture, momentum_acceptance and off_energy_dynamic_aperture, it will not be imported unless the user wants the plot.

simoneliuzzo:
I close this pull request. the module has troubles to run for the HMBA lattice. I looked for several hours without understanding. It is also based for some options on the use of a matplotlib tricontour function to define bounduaries of arbitrary array of coordinates.  I do not think it is worth to pursue the work on this branch any more.

simoneliuzzo:
I reinstalled python3 (3.9) and pyAT from scratch (including several local software updates). Now I have a test script running for the hmba lattice in at/pyat/test_matlab/hmba.mat and other lattices. The use of matplotlib is still there, for the generic case. There is no need to load this module by default. I will try to do as for at.plot. I did not commit any change for now.

simoneliuzzo:
I am not good in testing tests. in pycharm, all is Green

Screenshot 2021-09-15 at 17 13 38

simoneliuzzo:
Screenshot 2021-09-15 at 17 21 20

lfarv:
@simoneliuzzo: the problem is that you need a big reorganisation of your code!

pyAT must run 100% without `matplotlib`. `matplotlib` is not in the package `requirements.txt`, so it is not installed with pyAT.  And the tests are run without it. An installation for a machine simulator, for instance, will also run without it. So there are some rules to optionally provide plotting in an interactive session:
- all plot modules are in a `at.load` package which is **not** imported by the main `at` package.
- to allow plotting the user needs to **explicitly** load `at.plot`. This automatically adds plotting methods to existing classes. Look:
```python
from at import Lattice
help(Lattice.plot_beta)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/var/folders/kc/97y2h1wj6sl5gp6cbqzgmyfc000105/T/ipykernel_71614/174098400.py in <module>
      1 from at import Lattice
----> 2 help(Lattice.plot_beta)

AttributeError: type object 'Lattice' has no attribute 'plot_beta'



import at.plot
help(Lattice.plot_beta)
Help on function plot_beta in module at.plot.specific:

plot_beta(ring, **kwargs)
    Plot beta functions and dispersion
    
    PARAMETERS
        ring            Lattice object
    
    KEYWORDS
        dp=0.0          Ignored if radiation is ON. Momentum deviation.
        dct=None        Ignored if radiation is ON. Path lengthening.
```
After importing `at.plot`, the `plot_beta` method appears in the `Lattice` object.
You can also access non-method function with `from at.plot import xxx`

- Finally, in `at/plot/__init__.py`, there is a try/except on importing all the contents which avoids crashing if `matplotlib` is absent. Instead, it just prints a message:
```python
import at.plot
No module named 'matplotlib'
Plotting is disabled

```


So to comply with this, you must completely separate computation and plotting. All the computation must be in .`…/at/physics/dynamic_aperture.py` and the plotting (as light as possible) goes to `…/at/plot/dynamic_aperture.py`. There you can freely `import matplotlib`, because it's protected by the outer package. Anyway, this clear separation if a good programming practice, so better following it. Try to put as much as possible in "computation" and keep plotting as light as possible. This allows to have the tests cover as much code as possible, because of course the plotting itself cannot be tested on GitHub…

Hint: you can get the reasons of the test failure by looking at the line with the red cross (others have been stopped before reaching the error). It's "3.9, ubuntu-latest" in your case. The "details' button of this line gives:
```
==================================== ERRORS ====================================
__________________ ERROR collecting test/test_dynap_momap.py ___________________
ImportError while importing test module '/home/runner/work/at/at/pyat/test/test_dynap_momap.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test/test_dynap_momap.py:2: in <module>
E   ModuleNotFoundError: No module named 'matplotlib'
```
showing that matplotlib is not installed.

simoneliuzzo:
Hi @lfarv, 

thanks for this reply! 

Infact I use `matplotlib` to get the border of the acceptance in a generic case. 
I found this solution nice cause it simply used existing code, but if `matplotlib` is forbidden even if not used for plots, I will have to write an equivalent function or simply live without it, simply returning to the user a vector of survived/not survived for example in the generic case. 
 
Do you know any python package that has this feature? In matlab I would have used `boundary`.

I tried also this without real success: 
https://stackoverflow.com/questions/50549128/boundary-enclosing-a-given-set-of-points

I will do the restructuring as you explained. May be I loose some "feature" (not really crucial), but at least the branch will pass the tests.

ciao
Simone

lfarv:
@simoneliuzzo: indeed using `matplotlib` for something else than plotting is misleading… Not strictly forbidden, but then you lock most of your code inside the optional `at.plot` package, so in interactive mode only, no tests…
I have no idea for the equivalent of "boundary". Did you explore `scipy`? It may help.

lfarv:
I now come to the contents:
I am afraid by the number of 'deepcopies': normally it should very rarely happen, this defeats the purpose of working with pointers. So I started to look why you are copying things.

10/12 copies are for coords. instead of 
```python
coord0 = {'x': 1e-6, 'xp': 0.0, 'y': 1e-6, 'yp': 0.0, 'delta': 0.0, 'ct': 0.0}
coord = copy.deepcopy(coord0)
```
you could write:
```python
coord = dict(x=1e-6, xp=0.0, y=1e-6, yp=0.0, delta=0.0, ct=0.0)
```
which creates a new directory each time.

2 other copies concern the lattice.  In principle, to compute a dynamic aperture, you need to track on a given lattice, there should be no change on any element during this tracking, so no need to make a copy of the lattice. The constructor already creates a shallow copy (`self.ring = at.Lattice(ring[start_index:] + ring[:start_index])`), that should be more than enough…

problem 1: you are switching the radiation state of the lattice. You should not! This is wrong:
```python
        da.ring.radiation_off()
        da.orbit = da.ring.find_orbit4(dp=da.dpp)  # dpp is added to orbit here

        # restore radiation state
        if ring_rad:
            da.ring.radiation_on()
```
To determine the 6D off-momentum closed orbit, you have to change the RF frequency by alpha * f0 * dpp, compute the 6D closed orbit and then move the 6 coordinates around this closed orbit. `orbit4` does not give you the proper 'synchronous phase' (ct) to start with. So if you are in 4D, stay in 4D and if you are in 6D, stay in 6D. The function `find_orbit` gives you the closed orbit in any condition and accepts either `dp` or `dct` in 4D. My favorite solution would be to process 'dp' or 'dct' in the constructor, to find the right frequency and closed orbit once for ever, and not change them any more later. New dpp => new Acceptance6D object.

problem 2: I need some more time…

simoneliuzzo:
Hi Laurent @lfarv,
Thank you again for looking more deeply into this module. 
I started removing some deepcopies. The radiation_off has remained since before the `linopt` upgrades. I will remove it. 
I will also try to correct the `dpp` usage. It is introduced for the off energy searches (`off_energy_dynamic_aperture` and `momentum_acceptance`). I though I copied the way it is done in matlab at `atdynap`, I must have done some errors... 

Hi @swhite2401,

 I introduced patpass and added examples folder with an example in it. The tests seems ok, while the example, ideally the same identical code, in my terminal loops forever... could you try it as well?

lfarv:
@simoneliuzzo: The use of `patpass` should be optional. It may conflict with openMP, unless we show that they can live together. I have a better experience with openMP than with `patpass`, and each user should evaluate the best in his case. However, if the use of `patpass` grows, we should think of having it check the presence of openMP. This is not obvious and is not the topic of this PR. Until this is clarified, the keep the default for using `patpass` to false.

lfarv:
Getting the boundary of the DA looks is a minor result, the real result is the full grid. Apert from plotting it, there is little you get from it. So you could very well put the boundary computation in the optional plot module and use `matplotlib` for that.

lfarv:
I now come to the `Acceptance6D` constructor.

#### mode parameter:

- **None**: the help says None = 6D (default). Then why not putting mode='6D' in the argument list?
- '**6D**': what is the '6D' mode ? Do you mean varying all 6 coordinates together? '6D' mode it is not handled by the if/elif sequence of the constructor, and at line 183 (`else:  # default for no mode = x-y`), it id converted to 'x-xp', as None is. But later, in `select_test_points_based_on_mode`, both are not handled at all!

  So it looks like the modes None and '6D' are not handled correctly.

#### dpp parameter:
For compatibility with other functions, it should be called `dp` instead of `dpp`. And a similar `dct` parameter should be available to specify the off-momentum in terms of path length (see `find_orbit`, `linopt`, `get_optics`…)

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,895 @@
...
+    planes = ('x', 'xp', 'y', 'yp', 'delta', 'ct')
+
+    # 2D combination of planes
+    modes = ('x-y',
+             'delta-x',
+             '6D',
+             'xp-yp',
+             'x-xp',
+             'y-yp',
+             'ct-delta')
```
lfarv, line 28: Apparently it's nor used anywhere

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,895 @@
...
+        if mode == 'x-y':
+            self.n_points = {'x': 13,
+                             'xp': 1,
+                             'y': 13,
+                             'yp': 1,
+                             'delta': 1,
+                             'ct': 1
+                             }
+        elif mode == 'delta-x':
+            self.n_points = {'x': 13,
```
lfarv, line 132: This could be achieved much easier with a dictionary:
in the class variables:
```python
    default_n_points = {'x': 13, 'xp': 1, 'y': 13, 'yp': 1, 'delta': 1, 'ct': 1 }
    mode_n_points = {
        'x-y':      {'x': 13, 'xp': 1, 'y': 13, 'yp': 1, 'delta': 1, 'ct': 1 },
        'delta-x':  {'x': 13, 'xp': 1, 'y': 1, 'yp': 1, 'delta': 13, 'ct': 1},
 #       '6D':       {'x': 13, 'xp': 1, 'y': 13, 'yp': 1, 'delta': 1, 'ct': 1 },
        'xp-yp':    {'x': 1, 'xp': 13, 'y': 1, 'yp': 13, 'delta': 1, 'ct': 1},
        'x-xp':     {'x': 13, 'xp': 13, 'y': 1, 'yp': 1, 'delta': 1, 'ct': 1},
        'y-yp':     {'x': 1, 'xp': 1, 'y': 13, 'yp': 13, 'delta': 1, 'ct': 1},
        'ct-delta': {'x': 1, 'xp': 1, 'y': 1, 'yp': 1, 'delta': 13, 'ct': 13},
    }
```
and in the constructor, a single line:
```python
self.n_points = mode_n_points.get(mode, default_n_points)
```

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,895 @@
...
+                            'delta': [],
+                            'ct': []
+                            }
+        self.survived = []
+
+        self.number_of_turns = n_turns
+        self.search_divider = search_divider
+
+        self.grid_modes = ['grid', 'radial']
+        self.grid_mode = self.grid_modes[0]
```
lfarv, line 120: Apparently useless: self.grid_mode is positioned in `test_points_grid` called at the end.

simoneliuzzo:
> Until this is clarified, keep the default for using `patpass` to false.

Done

simoneliuzzo:
> Getting the boundary of the DA looks like a minor result, the real result is the full grid. Apert from plotting it, there is little you get from it. So you could very well put the boundary computation in the optional plot module and use `matplotlib` for that.

Dear @lfarv, I think I disagree with this. The limit of the DA are the real useful output of the computation. They are used for optimizations for example or for comparison plots. I would actually like that the standard user completely forgets about grids and booleans. This is also the case for `atdynap`, the output is a couple of arrays of coordinates in 2D. Also the strongly simplified boundary condition search `Acceptance6D.get_border` (presently simply looking for a maximum of each column, definitely not generic) makes use of some of the class properties (as grid_mode), this is why I placed it in the Acceptance6D class.

simoneliuzzo:
>     * '**6D**': what is the '6D' mode ? Do you mean varying all 6 coordinates together? '6D' mode it is not handled by the if/elif sequence of the constructor, and at line 183 (`else:  # default for no mode = x-y`), it id converted to 'x-xp', as None is. But later, in `select_test_points_based_on_mode`, both are not handled at all!
>       So it looks like the modes None and '6D' are not handled correctly.

6D is meant for 6D acceptance, it is the same as None mode. I have fixed as you suggest the default and added a 6D case in the init.  select_test_points_based_on_mode is supposed to determine a 2D digest of the 6D coordinates. For 6D and None this is not relevant. I have added in `select_test_points_based_on_mode` a check for modes None and 6D.

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: now it is at least in one place! I added it to use it for user input check.

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: The class code at initialization is cumbersome, I agree. 
For the user it seemed simpler to define a mode (self.mode = 'x-y') and then define a number of points to scan setting self.n_points) . In fact there is presently no check that for a given mode the number of points are correctly corresponding. The user is in fact free to set `self.mode = 'x-y'` and `self.n_points['x'] = 1 ` and that will fail. 
I will think more about this

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: the grid_mode is used also by the function dynamic_aperture and Acceptance6D.get_border.
I have made a more consistent use of it. it was somewhere a function input and somewhere a class property.

simoneliuzzo:
I think I am missing something with patpass. @swhite2401 mentioned I needed to recompile with some option. Could you remind me? 

When I switch from atpass to patpass (Acceptance6D.parralel_computation = True) all the 4 cores of my mac are fully busy (fans turning fast), but the computation never ends. With atpass in few seconds the same is over.

swhite2401:
@simoneliuzzo , patpass doe snot work for MACOS, however it should return atpass and not freeze so there might be an issue there.
I'll add an explicit warning to make sure it is clear.

There is another way to parallelize computation, it is openMP, for this you need you rebuilt pyAT see #226 and #221 for usage

pyat/at/physics/dynamic_aperture.py
lfarv: I have nothing against the choice you made for inputs, my proposal is just a cleaner way of doing exactly the same thing !

lfarv:
@simoneliuzzo: Recall of the 2 possibilities for parallel computing:

1. OpenMP: It's activated at compile time, for python by setting an environment variable for `setup.py`:
  `OPENMP=1 python setup.py build` (or install or develop)
  **Pros**: robust, I use it all the time and never had a single problem. Transparent: it's confined in the C code, you never notice anything. Since parallelisation happens within each integrator, it can handle collective effects (no interest for dynamic aperture…). And most important: it transparently accelerates **everything** (transfer matrices,…), as soon as the number of particles is larger that a tunable threshold (10 by default).
  **Cons**: since the job is split in each integrator, more overhead than tracking each particle independently. No control of the behaviour of OpenMP, it'a all automatically handled. You may optionally specify a "desired" number of procs, but you don't know what happens in reality.
2. patpass: parallelism handled by python. Each particle is independent (no collective effect !).
  **Pros**: you can use it where you want. Efficient since the job is distributed only once for each tracking.
  **Cons**: portability problems; though it does work on Linux, MacOS and Windows, it has to be tuned for each platform so there may be problems with python of system updates. I am surprised by your problem, I'll check.

In terms of efficiency, it depends ! That's why we decided to propose both, but if has to be evaluated in each case…
I have no experience if both are activated together: you cannot expect any gain, but does it work?

lfarv:
Looking at compute_orbit:

First, to have harmonic cavities handled properly, you must access cavities using the methods and properties provided by the cavity_access module. Methods:
```
h=ring.get_harmonic_number(**kwargs)
frf=ring.get_rf_frequency(**kwargs),
ring.set_rf_frequency(frf, **kwargs),
```
properties:
```
h=ring.harmonic_number
frf=ring.rf_frequency
ring.rf_frequency=fry
```
Don't try to get the cavity location yourself ! In your case, to set the frequency, you must use the method with the copy=True keyword, to avoid affecting the original lattice: `ring.set_rf_frequency(frf, copy=True)`

second: pleas try to make the user's life easy: remove the use_radiation keyword and don't look for cavities. If the lattice has radiation ON, compute with it, otherwise, keep it as it is. I know that the naming is misleading, but radiation ON means "there is an active cavity",  whether there is radiation or not.

simoneliuzzo:
> cavity_access module

Dear @lfarv 

I did not know a cavity_access module exists! I will make use of it. thank you

lfarv:
Sorry to bother you again, but I looked at the way you search for cavities:
```python
elems = self.ring.get_elements(key=at.RFCavity('', 0.0, 0.0, 0.0, 0.0, 0.0))
self.ind_rf = [self.ring.index(elem) for elem in elems]
```
This does not work: the `index` method scans the whole list looking for identity. First, this is a lot of unnecessary work. Second, if you have several identical cavities, its will always return the 1st one! Third, we gave up with the definition of identity for `Elements`: it's too time consuming. So python uses the default behaviour: same memory address. Not what we expect.

So there are too ways to do this:
**1. Pure python way:**
  `[i for i, elem in enumerate(ring) if isinstance(elem, at.RFCavity)]` for a numerical index,
  `[isinstance(elem, at.RFCavity) for elem in ring]` for a boolean index,
  I like this way; it's clear for any python programmer, and efficient since other methods all rely on that,
**2. "AT" way:**
  `at.get_cells(ring, checktype(at.RFCavity)` gives a boolean index.
  There are also `checkname` and `checkattr` available test functions and you can provide yours. Some like this way, you're free to choose.

Finally, to get a list of selected elements (pointers to, in fact…), you 1st line is right, but could be simply be `ring.get_elements(at.RFCavity)`. No need to instanciate a fake element.

simoneliuzzo:
Yet an other unknown (to me) feature of pyAT!

I will modify as you say, and actually use at.get_cells also in the pyAT ring simulator DS!

lfarv:
I do hope that this PR will be merged ! But there are still a number of details to clean. In `recursive_search_of_directional_limit`:


```python
        # define initial coordinate
        coord0 = {'x': 1e-6, 'xp': 0.0, 'y': 1e-6, 'yp': 0.0, 'delta': 0.0, 'ct': 0.0}

        # define step in each plane
        step = {pl: direction[ip]*(self.dict_def_range[pl][1] - self.dict_def_range[pl][0])
                for ip, pl in enumerate(self.planes)}

        limit = [v for _, v in coord0.items()]

        for ir in range(number_of_recursions):
...
            if ir == 0:
                coord = {'x': 1e-6, 'xp': 0.0, 'y': 1e-6, 'yp': 0.0, 'delta': 0.0, 'ct': 0.0}

            if self.verbose:
...
```
Could be:
```python
        # define initial coordinate
        coord = {'x': 1e-6, 'xp': 0.0, 'y': 1e-6, 'yp': 0.0, 'delta': 0.0, 'ct': 0.0}

        # define step in each plane
        step = {pl: direction[ip]*(self.dict_def_range[pl][1] - self.dict_def_range[pl][0])
                for ip, pl in enumerate(self.planes)}

        limit = [v for _, v in coord.items()]

        for ir in range(number_of_recursions):
...

            if self.verbose:
```

(Removed `coord0` and moved the initialisation of `coord` to the top.

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,1008 @@
...
+            if self.verbose:
+                print('step back')
+                print(coord)
+
+            # last coordinates are lost and add an additional test point lost (for display and contour computation
+            if self.survived:
+                self.survived[-1] = False
+
+            coord_lost = copy.deepcopy(coord)
+            for pl in self.planes:
```
lfarv, line 333: Why 'deep' ? For a shallow copy, the `dict` object has a copy method more efficient that the generic copy. Better writing:

`coord_lost = coord.copy()`
or 
`coord_lost = dict(coord)`

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,1008 @@
...
+
+        # define initial coordinate
+        coord0 = {'x': 1e-6, 'xp': 0.0, 'y': 1e-6, 'yp': 0.0, 'delta': 0.0, 'ct': 0.0}
+
+        # define step in each plane
+        step = {pl: direction[ip]*(self.dict_def_range[pl][1] - self.dict_def_range[pl][0])
+                for ip, pl in enumerate(self.planes)}
+
+        limit = [v for _, v in coord0.items()]
+
```
lfarv, line 273: Simpler:
`limit = list(coord.values())`

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,1008 @@
...
+            coord_lost = copy.deepcopy(coord)
+            for pl in self.planes:
+                coord_lost[pl] += step[pl]
+                # update tested points
+                self.coordinates[pl].append(coord_lost[pl])
+            self.survived.append(False)
+
+            # define limit 6 element array
+            limit = [copy.deepcopy(v) for _, v in coord.items()]
+
```
lfarv, line 341: Why 'deep' ?
Why even copy ? All values are numerical (I think…)
Use:

`limit = list(coord.values())`

simoneliuzzo:
I still have no clear idea about how to handle p/atpass . 
Is there a way to know from pyAT if openMP is active? 
```
import os
os.environ['openMP']
```

Also, to get the operating system in python this 
```
import platform 
platform.system()

```
is ok ?

then the logic could be:
- if asked` (parallel_compuation=True)` use `patpass` (if not(openMP) or unix)
  - if openMP: use `atpass` (force parallel_compuation=False)
  - if mac or windows: use `atpass` (force parallel_compuation=False)

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: bacause I am scared not to get what I would get in matlab by this action. 
so to "save" the value of coord in coord_lost I used deepcopy. LAter on I change coord again, but I do not want coord_lost to change as well magically!

lfarv: So a shallow copy if that you need. Affecting a new valus to coord does not change coord__lost. Would you use a=deepcopy(42) ?

swhite2401:
I think @lfarv has added a function `isopenmp` function in at.c just for this purpose, I haven't looked at the details but I believe it is accessible in the python.

Have you tried from `at.tracking import isopenmp` ?

You don't need to worry about OS, it is already handled in `patpass`. The problem you had was not expected and needs to fixed, for now you can assume that this will be the case.

lfarv:
Looking at environment does not answer: the environment variable is used at compile time only. You have to look in C code to know. So I just introduced a check function in at.c in the last commit. Still to be tested

- Then the test for openmp should be done in `patpass` itself, not in your code
- I still do not know why hou have problems on MacOS. In tracking tests it works…

lfarv:
Same answer as @swhite2401; "Don't worry !'

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: done

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: done

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: done

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: I would! to make sure 42 is 42 and will be 42 also if b = a and b = 43!

simoneliuzzo:
> Don't try to get the cavity location yourself ! In your case, to set the frequency, you must use the method with the copy=True keyword, to avoid affecting the original lattice: `ring.set_rf_frequency(frf, copy=True)`

I had to switch to copy = False. or the frequency in ring would not change. is this ok?

pyat/at/physics/dynamic_aperture.py
lfarv: No ! Look:
```
a=object1
b=a
```
now if you **modify** `b`, like `append(b,42)`, of course `a` will be modified also. But if you assign something else to b:
`b=object2`
`a` will still pointing to `object1`, unmodified.

So for dicts:
```python
a['x'] = 0
b=a.copy()   # shallow copy, b is identical to a
b['x'] = 1   # a['x'] is unchanged
```
You assign 1 to b['x']. But a['x'] is still 0. That would be different you had put `b = a` instead of `b=a.copy()`.

A deep copy is a very complex operation, but is rarely necessary. And you never make copies of constants.

lfarv:
@simoneliuzzo : No, because then you modify the original lattice, which is not what we want.

`copy = True` replaces the cavity with a modified one, so the original lattice is unchanged
`copy = False` modifies the cavity, so the original lattice also sees the modification.

If this does not work, then it's a bug in `set_value_refpts`. Let me check…

lfarv:
@simoneliuzzo: it's my fault, the correct lines should be:

for an in-place modification (undesired)
`self.ring.set_rf_frequency(new_rf_frequency, copy=False)
`

For a local modification:
`self.ring = self.ring.set_rf_frequency(new_rf_frequency, copy=True)`

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,1012 @@
...
+        :return:
+        """
+        # disable parallel computation
+        init_parcomp = self.parallel_computation
+        self.parallel_computation = False
+
+        # relevant_planes = []
+        # if self.mode:
+        #     relevant_planes = self.mode.split('-', 2)
+        coord0 = {'x': 1e-6, 'xp': 0.0, 'y': 1e-6, 'yp': 0.0, 'delta': 0.0, 'ct': 0.0}
```
lfarv, line 365: `coord0` is unused

simoneliuzzo: done

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,1012 @@
...
+                            self.dict_def_range[two_planes[1]][1]]
+
+            if self.dict_def_range[two_planes[1]][0] == 0:
+                theta = np.linspace(0, math.pi, self.n_points[two_planes[0]])
+            else:
+                theta = np.linspace(0, 2*math.pi, self.n_points[two_planes[0]])
+
+            ip = 0
+
+            for ind_p, p in enumerate(self.planes):  # must loop all plane to define all 6D ranges (1 point)
```
lfarv, line 425: `ind_p` is not used, though I'm not sure it should not be used instead of `ip`. Please check and then, you could use;
`for p in self.planes:`

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,1012 @@
...
+
+            flat_dd = []
+            for k, val in d_.items():
+                flat_dd.append([item for sublist in val for item in sublist])
+                print('{} has {} elements'.format(k, len(flat_dd[-1])))
+
+            num_points = max([len(dd) for dd in flat_dd])
+
+            # define lists for DA scan
+            for ip, p in enumerate(self.planes):
```
lfarv, line 450: `ip` is not used:
`for p in self.planes:`

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,1012 @@
...
+                    # if d_ is not 0.0:
+                    #    d_.append(0.0);
+
+            # define mesh of points
+            xx, xpxp, yy, ypyp, deltadelta, ctct = np.meshgrid(d_[0],
+                                                               d_[1],
+                                                               d_[2],
+                                                               d_[3],
+                                                               d_[4],
+                                                               d_[5],)
```
lfarv, line 486: `xx, xpxp, yy, ypyp, deltadelta, ctct = np.meshgrid(*d_)`

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: done

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: done

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,1012 @@
...
+                              [coordinates['delta'] + self.orbit[0][4]],
+                              [coordinates['ct'] + self.orbit[0][5]]),
+                              axis=0)
+
+        # track coordinates for N turns
+        if not self.parallel_computation:
+
+            t, losses = at.atpass(self.ring,
+                                  copy.deepcopy(np.asfortranarray(rin)),
+                                  self.number_of_turns,
```
lfarv, line 536: A deepcopy of a numerical array does not make sense !
```python
            t, losses = at.atpass(self.ring,
                                  rin.copy(),
                                  self.number_of_turns,

```

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,1012 @@
...
+                                  refpts=np.array(np.uint32(0)),  # np.array(np.uint32([len(self.ring)])))
+                                  losses=True)
+
+        else:
+            print('parallel computation')
+
+            # track coordinates
+            t, losses = at.patpass(self.ring,
+                                   copy.deepcopy(rin),
+                                   self.number_of_turns,
```
lfarv, line 546: Same here:
```python
            t, losses = at.patpass(self.ring,
                                  rin.copy(),
                                  self.number_of_turns,

```

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: here I needed to add np.asfrotranarray. it is not needed?

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: done, testing

simoneliuzzo: actually what does *d_ do?

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: for this I did not need  np.asfortranarray

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: in these calls, I also have the flag losses = True. Is there a way to minimize the output in this flag? I remember that in the matlab version it is a rather big output, and here I just need an array of booleans.

simoneliuzzo: > here I needed to add np.asfrotranarray. it is not needed?

  File "/Users/liuzzo/beamdyn/at/pyat/at/physics/dynamic_aperture.py", line 529, in test_survived
    t, losses = at.atpass(self.ring,
ValueError: rin is not Fortran-aligned

swhite2401: The losses flag was meant to compute loss maps. 
In case you just need a vector of boolean I would suggest the following   

`mask = numpy.isnan(t[0,:,0,-1])`

Or something similar: a lots particle will have its first coordinate set to nan

simoneliuzzo: Dear @swhite2401 ,

thanks for your reply. Infact this is what I used to do, before reading that atpass has losses output. So, I thought it would have been a good choice at least for readibility, to use losses. 
Actually, in the matlab ringpass there are 3 outputs, [ particlescoordinates, loss_flag, losses ]= ringpass(...). 
In pyAT there is a ringpass equivalent? 

If it is faster/better/suggested I will do as you say:

`mask = numpy.isnan(t[0,:,0,-1])
`

where t[coordinate, particle, turn, index_in_lattice ]

I guess the t shape/dimensions will be the same when tracking a single particle or multiple particles

swhite2401: Yes `lattice_pass` also has the flag available, I ported everything expect  for the history buffer of lost particles (that was the big chunk). In fact activating the flag represent a negligible loss in performance.
I am not quite sure about your coordinate order... it seems the last one is the turn number which is why I did put -1 there.
No difference for multi-particle.

simoneliuzzo: ok. so t[coordinate, particle, index_in_lattice, turn_number]
I commit in few minutes the change

swhite2401: Here is the help, source is in tracking/patpass.py     

OUTPUT:
        (6, N, R, T) array containing output coordinates of N particles
        at R reference points for T turns.
        If losses ==True: {islost,turn,elem,coord} dictionnary containing
        flag for particles lost (True -> particle lost), turn, element and
        that survived

swhite2401: By the way if you have doubts about a function you can easily access the help in a python prompt by typing:
`help(at.tracking.patpass)` for example

simoneliuzzo: finally I have committed this:

notice the np.asfortranarray()

```

t = at.atpass(self.ring,
                                  np.asfortranarray(rin.copy()),
                                  self.number_of_turns,
                                  refpts=np.array(np.uint32(0)))

t = at.patpass(self.ring,
                       rin.copy(),
                       self.number_of_turns,
                       refpts=np.array(np.uint32(0)))

```

swhite2401: I think you can use the default `refpts=None` in case you want the coordinates at the start of the ring only.

For single core I would suggest to use `lattice_pass` instead (no need to care about ordering in this case), I will add the same check in patpass so that you don't need to worry about this anymore

simoneliuzzo: using ..., refpts=None) I get this error:

TypeError: argument 4 must be numpy.ndarray, not None

simoneliuzzo: removing refpts, it gives an errors when looking at the output tracking, the refpts dimension is empty

survived = [not s for s in np.isnan(t[0,:,0,-1])]
IndexError: index 0 is out of bounds for axis 2 with size 0

simoneliuzzo: so for now I keep it like it is

simoneliuzzo:
How can I disable this warning:
```
/Users/liuzzo/beamdyn/at/pyat/at/physics/orbit.py:391: AtWarning: In 6D, "dp" and "dct" are ignored
  warnings.warn(AtWarning('In 6D, "dp" and "dct" are ignored'))
```

pyat/at/physics/dynamic_aperture.py
swhite2401: Ah yes this is in fact an problem I have noticed in the past, using the default (removing refpts),
It probably needs fixing

swhite2401:
You can disable warning for a given function with:

```
import warnings
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    yourfunction()
```

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: if you noticed and I notice, may be we should open an issue?

lfarv: **You must not use `atpass` directly**:
the public interface is `lattice_pass`. The arguments are the same, but `lattice_pass` does all the argument checking that was too tricky to do in C. This answers both you questions concerning alignment and refpts.

lfarv: Coming back to Fortran alignment: It is not strictly necessary, since both lattice_pass and patpass will reorganise the memory if necessary. However it's more efficient if you set Fortran alignment in r_in, so that the memory layout is done once for ever. This is true for both `lattice_pass` and `patpass`.

In your particular case, this can be done easily by:
```
        # create 6xN matrix add orbit (with dpp) to each coordinate
        rin = np.asfortranarray(list(coordinates.values())) + self.orbit.reshape((6,1))
```

simoneliuzzo: Dear @lfarv 

I replaced this: 

```
 rin = np.concatenate(([coordinates['x'] + self.orbit[0][0]],
                              [coordinates['xp'] + self.orbit[0][1]],
                              [coordinates['y'] + self.orbit[0][2]],
                              [coordinates['yp'] + self.orbit[0][3]],
                              [coordinates['delta'] + self.orbit[0][4]],
                              [coordinates['ct'] + self.orbit[0][5]]),
                              axis=0)

```

with 

```

rin = np.array(list(coordinates.values()))

 p = list(rin.shape)
 p[0] = 1
 orb =  np.tile(self.orbit[0],p)

```

in order to make it work also in 2D.
Both options would be eventually followed by :

`rin = np.asfortranarray(rin)
`

However, I have the impression that the first version is actually more intuitive to read.

swhite2401: Are you sure this is the case for `patpass`? I am not convinced this is the case...to checked

simoneliuzzo: Dear @lfarv 

I replaced this: 

```
 rin = np.concatenate(([coordinates['x'] + self.orbit[0][0]],
                              [coordinates['xp'] + self.orbit[0][1]],
                              [coordinates['y'] + self.orbit[0][2]],
                              [coordinates['yp'] + self.orbit[0][3]],
                              [coordinates['delta'] + self.orbit[0][4]],
                              [coordinates['ct'] + self.orbit[0][5]]),
                              axis=0)

rin = np.asfortranarray(rin)

```

with 

```
rin = np.array(list(coordinates.values()))

p = list(rin.shape)
p[0] = 1
orb =  np.tile(np.array([self.orbit[0]]).T,p)

rin = np.asfortranarray(rin+orb)

```

in order to make it work also in 2D.

However, I have the impression that the first version is actually more intuitive to read.

swhite2401:
I believe adding a constants module in pyAT would be useful and rather easy.

pyat/at/physics/dynamic_aperture.py
lfarv: > Are you sure this is the case for `patpass`? I am not convinced this is the case...to checked

Yes, sure:
```python
args = [(ring, r_in[:, i]) for i in range(r_in.shape[1])]
```
` r_in[:, i]` consecutive implies Fortran ordering

swhite2401: Ok, but still it fails in case a collective `passmethod` is present or for one particle.
We may just force it in a similar fashion to what is done in `lattice_pass` to avoid problems

By the way, would you know any smart way to prevent people from using atpass directly? I have seen that AT users tend to do it a lot...

lfarv: > ```
> rin = np.array(list(coordinates.values()))
> 
> p = list(rin.shape)
> p[0] = 1
> orb =  np.tile(np.array([self.orbit[0]]).T,p)
> 
> rin = np.asfortranarray(rin+orb)
> ```
Can you explain what you are trying to do? `self.orbit `is supposed to have shape (6,): that is a 1D array. I couldn't find where a 2nd dimension appears.

And what is wrong with my version?

np.array(list(coordinates.values())) has shape (6,n)
self.orbit.reshape((6,1)) has shape (6,1)
So where is the problem?

lfarv:
> I believe adding a constants module in pyAT would be useful and rather easy.

What do you mean ? Replacing the SciPy module ?

swhite2401:
Well this I though we could have a module importing some relevant scipy constants that could later
be called simply with `at._clight` or `at._emass`, sometimes it can be cumbersome to look through
`scipy.constants.physical_constants[name]`

But this is just a suggestion to avoid importing the same things over and over again in various modules

pyat/at/physics/dynamic_aperture.py
lfarv: > Ok, but still it fails in case a collective `passmethod` is present or for one particle.
> We may just force it in a similar fashion to what is done in `lattice_pass` to avoid problems

Well, I don't think it's worth changing anything. This memory rearrangement is minor compared to the overhead of starting the thread… And forcing alignment earlier will not make it faster.

> By the way, would you know any smart way to prevent people from using atpass directly? I have seen that AT users tend to do it a lot...

May be:
1. renaming it to `_atpass`, showing clearly it's private
2. Instead of importing it in `__init__.py`, importing it where it's needed (`track.py` etc.)
3. Adding a dummy `atpass` raising an exception and telling what to do

Remember, I suggested `patpass` could be renamed as `plattice_pass` to better show the equivalence.

swhite2401: This is probably a good time to do this what do you think? 
We can leave the original methods with a deprecation warning until the next version increment

simoneliuzzo: > > ```
> > rin = np.array(list(coordinates.values()))
> > 
> > p = list(rin.shape)
> > p[0] = 1
> > orb =  np.tile(np.array([self.orbit[0]]).T,p)
> > 
> > rin = np.asfortranarray(rin+orb)
> > ```
> 
> Can you explain what you are trying to do? `self.orbit `is supposed to have shape (6,): that is a 1D array. I couldn't find where a 2nd dimension appears.
> 
> And what is wrong with my version?
> 
> np.array(list(coordinates.values())) has shape (6,n)
> self.orbit.reshape((6,1)) has shape (6,1)
> So where is the problem?

it says it cannot "broadcast" , so I tried several .T, tile, shape, reshape, till I got to the one I sent.

simoneliuzzo: > Ok, but still it fails in case a collective `passmethod` is present or for one particle.
> We may just force it in a similar fashion to what is done in `lattice_pass` to avoid problems
> 
> By the way, would you know any smart way to prevent people from using atpass directly? I have seen that AT users tend to do it a lot...

I did this! nothing forbid me, or warned me I was doing something not welcome. An _atpass would have discouraged me.

In case, this modification would be a different PR, right?

lfarv:
Ok, understood. Then I suggest adding a `constants` module to the `lattice` package: early enough so that it's available to all the following modules. We could even add a backup solution if `scipy.constants` is unavailable, to allow PyAt to run in a degraded mode without SciPy (I tried recently to run PyAT undef PyPy, and failed because I could not install SciPy).

pyat/at/physics/dynamic_aperture.py
lfarv: > > > ```
> > > rin = np.array(list(coordinates.values()))
> > > 
> > > p = list(rin.shape)
> > > p[0] = 1
> > > orb =  np.tile(np.array([self.orbit[0]]).T,p)
> > > 
> > > rin = np.asfortranarray(rin+orb)
> > > ```
> > 
> > 
> > Can you explain what you are trying to do? `self.orbit `is supposed to have shape (6,): that is a 1D array. I couldn't find where a 2nd dimension appears.
> > And what is wrong with my version?
> > np.array(list(coordinates.values())) has shape (6,n)
> > self.orbit.reshape((6,1)) has shape (6,1)
> > So where is the problem?
> 
> 
> it says it cannot "broadcast" , so I tried several .T, tile, shape, reshape, till I got to the one I sent.

Can you look where the shapes differ from what I expect and mentioned above?

simoneliuzzo: rin.shape
(6, 169)
self.orbit[0].shape
(6,)
rin + self.orbit[0]
  File "/Applications/PyCharm CE.app/Contents/plugins/python-ce/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
ValueError: operands could not be broadcast together with shapes (6,169) (6,)

lfarv:
By the way, I partially did that in `physics/__init__.py` for e_mass, Cgamma and Cq. It would be better the way I propose now.

pyat/at/physics/dynamic_aperture.py
lfarv: So you have to use `rin + self.orbit[0].reshape((6,1))`. Broadcasting prepends the missing dimensions, so reshapes orbit[0] as (1,6) instead of (6,1), and addition fails.

But I still don't see the shape of `self.orbit` itself, and why you use `self.orbit[0]` instead of `self.orbit`.

simoneliuzzo: I use self.orbit[0] due to this:
```

>> self.orbit
(array([-2.63484764e-09, -1.45845243e-10,  0.00000000e+00,  0.00000000e+00,
       -6.69657366e-06, -5.88436654e-02]), array([], shape=(0, 6), dtype=float64))
>> self.orbit[0]
array([-2.63484764e-09, -1.45845243e-10,  0.00000000e+00,  0.00000000e+00,
       -6.69657366e-06, -5.88436654e-02])
>> self.orbit[0]+rin   # this works
array([ 9.94730305e-07, -2.91690487e-10,  1.00000000e-06,  0.00000000e+00,
       -1.33931473e-05, -1.17687331e-01])
>> self.orbit+rin   # this gives error
<input>:1: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  File "/Applications/PyCharm CE.app/Contents/plugins/python-ce/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
ValueError: operands could not be broadcast together with shapes (2,) (6,) 


```

simoneliuzzo: This seems to work:

```
rin = np.array(list(coordinates.values()))

if len(rin.shape) == 1:
    rin = rin + self.orbit[0]
else:
    rin = rin + self.orbit[0].reshape(6,1)

```

lfarv: Now, I understand everything ! When calling `find_orbit`, you get 2 output arguments: `orbit0` and `orbit`. So the call should be:

`self.orbit, _ = find_orbit(...)`

As it is now, `self.orbit` is a tuple with 2 values. `orbit0` and an empty array. And I missed that !

lfarv: The final line should be:

`rin = np.asfortranarray(list(coordinates.values()).reshape(6, -1) + self.orbit.reshape(6, 1)`

By the way, `len(rin.shape)` is `rin.ndim`

lfarv:
> But this is just a suggestion to avoid importing the same things over and over again in various modules

I take care of this in a PR in progress.

pyat/at/physics/dynamic_aperture.py
simoneliuzzo: that's python! I do the modification

lfarv (commented):
Starting a new review

pyat/at/physics/dynamic_aperture.py
```diff
@@ -0,0 +1,997 @@
...
+            new_rf_frequency = self.rf_frequency + drf
+            if self.verbose:
+                print('dp = {dp}, rf set to f_RF0 + {rf} Hz'.format(dp=self.dp, rf=drf))
+                print('new RF = {rf:3.6f} MHz'.format(rf=new_rf_frequency*1e-6))
+
+            # set RF frequency to all cavities
+            try:
+                self.ring = self.ring.set_rf_frequency(new_rf_frequency, copy=True)
+            except Exception:
+                print('RF not set. Probably the lattice has no RFCavity.')
```
lfarv, line 231: Do you really want to continue after this error ? With radiation but no cavity, there is no hope to get a closed orbit. Don't catch the exception!

Note: in such a case, you should catch the most accurate exception (`AtError` in that case), to avoid catching any other one (memory fault...) that you are unable to handle.

simoneliuzzo: like this ? 

```
            try :
                self.ring = self.ring.set_rf_frequency(new_rf_frequency, copy=True)
            except AtError as exc:
                print('RF not set. Probably the lattice has no RFCavity.')
                raise(exc)
```

lcarver:
I have missed a lot of activity during my holidays. Is now a good time for my review?

simoneliuzzo:
for me yes! 

I think also @lfarv is going through a further review. 

A quick functionality check is to run pyat/at/examples/example_dynamic_aperture.py 
I have a few minor commits (functions help) that I will not push till you have given me further comments.

lfarv:
I start again from the beginning:

I repeat that the long if` mode==/elif mode ==`  in` __init__`, line 145 can be removed:

in the class variables, replace
```python
    modes = ('x-y',
             'delta-x',
             'xp-yp',
             'x-xp',
             'y-yp',
             'ct-delta',
             '6D')
```
with
```python
    modes = {
        'x-y':      {'x': 13, 'xp': 1, 'y': 13, 'yp': 1, 'delta': 1, 'ct': 1 },
        'delta-x':  {'x': 13, 'xp': 1, 'y': 1, 'yp': 1, 'delta': 13, 'ct': 1},
        'ct-delta': {'x': 1, 'xp': 1, 'y': 1, 'yp': 1, 'delta': 13, 'ct': 13},
        'x-xp':     {'x': 13, 'xp': 13, 'y': 1, 'yp': 1, 'delta': 1, 'ct': 1},
        'y-yp':     {'x': 1, 'xp': 1, 'y': 13, 'yp': 13, 'delta': 1, 'ct': 1},
        'xp-yp':    {'x': 1, 'xp': 13, 'y': 1, 'yp': 13, 'delta': 1, 'ct': 1},
       '6D':       {'x': 5, 'xp': 5, 'y': 5, 'yp': 5, 'delta': 5, 'ct': 5 },
    }
```
and in `__init__()`, a single line:
```python
self.n_points = modes.get(mode)
```
Replaces 143 lines, from 145 to 288

lfarv:
At line 139:
```
        self.grid_modes = ['grid', 'radial']
        self.grid_mode = grid_mode
        if not (self.grid_mode in self.grid_modes):
            raise('grid mode must be: grid or radial')
```
grid_modes does not need to be an instance variable. It can be a class variable.

lfarv:
Again on the same lines:

Could you change `grid  / radial`, to the clearer choice `cartesian / polar`?

lfarv:
Repeating my comments on `deepcopy()`:

line 341:
`            coord_lost = copy.deepcopy(coord)
`
You do not need a deep copy: the values are floating point numbers, so the new value replaces the old one in the copy, and the original is unchanged.
`            coord_lost = coord.copy()
`

line 779:
`    da = Acceptance6D(copy.deepcopy(sr_ring), start_index=start_index, n_turns=n_turns)
`
You do not need a deep copy, you even do not need a copy at all: The `Acceptance6D` constructor makes its own copy. So:
`    da = Acceptance6D(ring, start_index=start_index, n_turns=n_turns)
`

line 813:
`            coord = copy.deepcopy(coord0)
`
Same as before, `coord = coord0.copy()`

line 864: same as 779
line 933: same as 779

With that, we get rid of all these deep copies !

lfarv:
First look at the `plot` module:

Line 2:
`from at.physics.dynamic_aperture import Acceptance6D
`
This is correct, of course. But in the PyAT style, we consider that the distribution of modules within a package is an implementation detail, subject to changes. So within the `plot` package, you shouldn't make any assumption on where inside the `physics` package you function is located. So `from at.physics`  or `from ..physics` is safer.

Oherwise, if you want `plot`,  `plot_base` and `plot6d` to be method of the `Acceptance6D` object, you can put such lines at the end of the module:

`Acceptance6D.plot = plot
`

lfarv:
For the tests now:

- the tests as they are cannot stay in the default test sequence: they take much too long. So you should move then to another directory, `test_da` for instance, so that any user may run them upon request.
- You can remove all prints: you normally don't get any output from the tests!
- In fact, you do not check anything, there is no assertion anywhere. You just check that it does not crash… Is there a way to check the results?

lcarver:
Hi Simone,
Some thoughts and feelings. Sorry if I repeat something that has already been said. 

1. When you initialise the class Acceptance6D you specify a mode 'x-y' or 'xp-yp' and you then have lots of if statements later depending on which mode is specified. Is it not better to specify xaxis='x', yaxis='xp'. Then you can use these keys to more cleanly access the coordinates or the information in the dictionaries that you need. Then from ln 678 and on can be written in a much more generic and simple way. For example, you can reduce it from 

```
  if self.mode == 'x-xp':
      h = self.coordinates['x']
      v = self.coordinates['xp']

      sel = [a and b == 0 and c == 0 and d == 0 and e == 0 for a, b, c, d, e in zip(
          self.survived,
          self.coordinates['y'],
          self.coordinates['ct'],
          self.coordinates['delta'],
          self.coordinates['yp'])]
```

to something like
```
h = self.coordinates[xaxis]
v = self.coordinates[yaxis]
otherplanes =[]
[otherplanes.append(plane) for plane in planes if h!=plane and v!= plane]
sel = [a and b == 0 and c == 0 and d == 0 and e == 0 for a, b, c, d, e in zip(
    self.survived,
    self.coordinates[otherplanes[0]],
    self.coordinates[otherplanes[1]],
    self.coordinates[otherplanes[2]],
    self.coordinates[otherplanes[3]])]
```
This new version is way more generic and therefore does not need to be repeated for all the different configurations. 
Also then the specification of the number of points becomes a bit easier and the rangers become a bit easier. You don't have to provide a 6D dictionary for each parameter.

simoneliuzzo:
> EDIT: If you really want to keep the mode='x-y' thing you could even cheat and just reduce this function a lot by doing xaxis, yaxis = self.mode.split('-') before the second block of code. Then you can remove most of the function select_test_points_based_on_mode and it saves on lines.

Hi @lcarver 
thank you for your comment! indeed it would be clearer and simple. However the code is not restricted to 2D, and I kept the mode as a more generic way to state 6D mode or to specify more exotic modes (probably never used...) ? 

I will definitely take your code to simplify `select_test_points_based_on_mode` !

lfarv:
Sorry for the interruption: wrong manipulation !

swhite2401:
Hello all, I will try to revive this development and add the Touschek lifetime calculation that I have on my private space: more news in some time...

Issue #287, opened 2021-09-08
pyAT: memory leak in atpass


swhite2401:
Dear @lfarv , @willrogers , 
We have been facing issues with a memory leak in pyAT for our ringsimulators.
We were able to pin it down to the specific case were `atpass` is called withing a` radiation_on() / radiation_off()` block inside a loop. In this case the memory usage of `lattice_object` grows.

Your help would be most helpful. I attached the exemple code that we used to diagnose the issue.

```
import at
import numpy as np
import tracemalloc
import linecache
from at.lattice import get_refpts, get_elements


path = '/machfs/swhite/at_dev/at/pyat/test_matlab/'
filename = 'hmba.mat'
key = 'hmba'
latticef = path+filename
ring = at.load_mat(latticef)
#ring = at.load_lattice(latticef,key=key)
ring.radiation_off()


elem = get_elements(ring,at.Quadrupole)[0]
ring = at.Lattice([elem],energy=6.0e9)

tracemalloc.start()
snapshot1 = tracemalloc.take_snapshot()


for i in range(10):
    print('iter #'+str(i))
    ring.radiation_on()
    at.atpass(ring,np.zeros(6),nturns=1)
    ring.radiation_off() 
    snapshot2 = tracemalloc.take_snapshot()
    top_stats = snapshot2.compare_to(snapshot1, 'lineno')
    for stat in top_stats[:20]:
            if 'pyat' in stat.traceback[0].filename:
                print(stat.traceback[0].filename, stat.size, stat.count)
                frame = stat.traceback[0]
                line = linecache.getline(frame.filename, frame.lineno).strip()
                print(line)
    print('')
```

swhite2401:
This one is fixed I close it

PR #288, opened 2021-09-08
fixed memory leak in element.update

swhite2401:
@lfarv , @willrogers , @simoneliuzzo , this branch fixes (I think) the memory leak discussed in #287 
I have modified the behavior of element.update in case element.attr is a string. 
In this case the string is replace entirely, this is quick and dirty. I am quite sure you can come up
with more elegant solution, your help is more than welcome.


swhite2401:
In fact it does not work...

swhite2401:
I delete the branch this is not working

PR #289, opened 2021-09-09
memory leak on passmethod name in at.c

swhite2401:
The reference to `PyPassMethod` in `at.c` is properly deallocated to avoid memory leak.
This solves #287.


simoneliuzzo (approved):

lfarv (approved):
well done!

willrogers (approved):
Looks sensible to me.

Issue #296, opened 2021-09-16
Release of PyAT 0.2.0


willrogers:
I would like to use some of the newer functions in pyat, so I intend to release version 0.2.0. Does anyone have any comments before I do so?

I will try to make this compatible with Python 3.9 (but not Python 2).

Aside: I am leaving Diamond in October, so I may be making no further contributions to PyAT! @mjgaughran will pick up some of this work so you might see him ask questions or make contributions.

cc @lfarv @swhite2401 @simoneliuzzo for comments.

swhite2401:
Ok for me, although it would have been nice to merge #271 first to be consistent with the latest developments in `linopt`.
@lfarv did you take a look?

Good luck for the future and welcome to @MJGaughran!

lfarv:
I agree for #271. But can you wait one more day for another pull request related to linopt?

@willrogers: We'll miss you! Anyway good luck, and welcome @MJGaughran!

swhite2401:
I merged #271, ok for me to release the new version

lfarv:
Ok for version 0.2.0.
Concerning #290, up to now it corrects a small part of the problem, only in a PassMethod which is almost never used, so it can be merged, but it really does not matter.
@willrogers : you need to explain the procedure to create a new release in pip, and where this procedure is defined.

willrogers:
It is not straightforward to support Python 3.9 since our supported numpy version (1.14.3) does not work.

There are other mechanisms to handle this but they are not trivial, so I propose to release pyat-0.2.0 supporting 3.5 - 3.8 and consider Python 3.9 later.

lfarv:
PyAT runs smoothly on Python 3.9. Isn't it possible to define the numpy minimum version to what is necessary for Python 3.9?
Anyway, no problem to consider Python 3.9 later.

willrogers:
I've just experimented and it looks like we could support Python 3.5 to 3.9 with Numpy 1.16.6.

Just because it builds doesn't necessarily mean it will run... but I think we should be ok.

willrogers:
I ran all tests using numpy 1.16.6. There is one problem (one test fails) on MacOS for Python 3.8 (and probably 3.9) with that specific version - otherwise all tests pass. The problem goes away for newer versions of Numpy, which will usually be what is installed.

https://github.com/atcollab/at/runs/3649562724?check_suite_focus=true

[...]

PR #297, opened 2021-09-16
Linopt6 in plots and parameters

lfarv:
#### Matlab
`atlinopt6` is used instead of `atlinopt` in `atplot`. This allows plotting with radiation ON. The compatibility is ensured with existing plot helpers, unless they use atlinop4-specific data (very unlikely, I did not find any). However some helpers are modified for simplification. `atx` remains unchanged…
#### python
`get_optics` is used instead of `linopt` in `radiation_summary`, envelope `summary` and plotting functions, with the same benefit.


lfarv:
Ready for merging and for releasing PyAT 0.2.0 (#296)

willrogers (approved):
The Python part of this looks fine and helps me by removing the radiation check on `get_radiation_integrals()`.

The new API of `linopt2,4,6` seems to be a big improvement (assuming all the physics is correct!).

PR #298, opened 2021-09-19
Use numpy 1.16.6 as minimum Numpy version.

willrogers:
This appears to allow Python 3.5 to 3.9.


lfarv (approved):
numpy 1.16.6 is ok for me!

lfarv:
In the meantime, I could solve the memory leak problems. It's up to you to include it in the release or not…

PR #299, opened 2021-09-19
Fixed the memory leak in atpass

lfarv:
A memory leak mentioned in #290 was due to errors in the reference counting in `atpass`. This is now fixed.
While working on `at.c`, two other minor modifications:
- remaining lines for Python2 compatibility are removed,
- a new function `isopenmp` allows to test the presence of OpenMP


swhite2401:
Very nice!

swhite2401 (approved):

PR #300, opened 2021-09-20
Pyat docs

willrogers:
Drop Python 3.5 and improve documentation for making releases.

cc @mjgaughran


MJGaughran:
I think you also need to update docs/python.md with the Python versions

lfarv (approved):
I tried the release procedure and downloaded the generated wheel. It contains a version for PyPy. Is it supposed to work? I could not run it; impossibility to install SciPy!

PR #301, opened 2021-09-21
Test for use of keep_lattice in find_orbit6().

willrogers:
This is how I expect to use the `keep_lattice` keyword argument, and it appears to be failing in pyat-0.2.0.

The second call (with `keep_lattice=True`) is returning nan in each element of the array.

Is this related to #299 ?

cc @mjgaughran

### DIFF
--- a/pyat/at/physics/orbit.py
+++ b/pyat/at/physics/orbit.py
@@ -278,6 +278,8 @@ def _orbit6(ring, cavpts=None, guess=None, keep_lattice=False, **kwargs):
 
     if guess is None:
         _, dt = get_timelag_fromU0(ring, method=method, cavpts=cavpts)
+        if method is ELossMethod.INTEGRAL:
+            keep_lattice = False
         ref_in = numpy.zeros((6,), order='F')
         ref_in[5] = -dt
     else:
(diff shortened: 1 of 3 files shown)


willrogers:
I see that this test starts failing in 4f2195463b32bbcbc8d9c869bcb2b0c77345a42c.

Is this a valid test and something needs fixing? Or am I misunderstanding something?

swhite2401:
I think this is a problem that needs fixing, @lfarv, any idea?

lfarv:
@willrogers: what you try is exactly how it is supposed to work.

I see the problem, and strangely `find_orbit6` is the only affect one: `find_orbit4`, `find_m44`, `find_m66` work as expected.
So I am looking at the problem ! Thanks for the hint on the origin of the problem.

willrogers:
I have been debugging this morning.

The difference is calling `get_timelag_fromU0()` with `ELossMethod.TRACKING` instead of `ELossMethod.INTEGRAL`.

The problem is the call in `energy_loss.py` line 66:

```
        ringtmp = ring.radiation_off(dipole_pass=None,
                                     quadrupole_pass=None,
                                     wiggler_pass=None,
                                     sextupole_pass=None,
                                     octupole_pass=None,
                                     copy=True)
```

I think this call must be modifying the original lattice somehow.

willrogers:
I added a warning if you pass `keep_lattice` when it is not valid. That might be a bit noisy so I can remove that commit if you like.

I update the `find_orbit6()` test and added one for `find_orbit4()`.

lfarv:
I don't think the message is useful: `keep_lattice` is a hint given to try and save some time (not a lot, in fact…). How this is honoured in the function does not really matter since it does not impact the result at all. So I would remove the warning.

PR #302, opened 2021-09-22
changed atfitchrom.m deltaP to make fit work

simoneliuzzo:
Dear @lfarv and @swhite2401 ,

Did `atfitchrom` change recently (about <1year) ? 

I had to change `deltaP` from 1e-8 to 1e-6 in `atfitchrom` to be able to fit the EBS lattice model chromaticity (deterministically and correctly). 

The problem arised only in ESRF CTRM (grappa ubuntu20.04 + matlab2020b) and the symptom was a wrong fit of chromaticity leading to sextupole strengths basically random ("fitted chromaticity" = tens of units, incorrect and not deterministic values).

The fix proposed in this PR solves the issue for the fit of chromaticity in the EBS model at ESRF-CTRM. 

I am sure that testing in your local computers you will not see this problem. It is the same for me. 

thank you for your help

regards
Simone

### DIFF
--- a/atmat/lattice/atfitchrom.m
+++ b/atmat/lattice/atfitchrom.m
@@ -20,7 +20,7 @@ else
     dpp=0;
     [newchrom,famname1,famname2]=deal(varargin{:});
 end
-deltaP = 1e-8;
+deltaP = 1e-6;
 idx1=varelem(ring,famname1);
 idx2=varelem(ring,famname2);
 kl1=atgetfieldvalues(ring(idx1),'PolynomB',{3});


swhite2401:
It was probably changed to adapt to #257, @lfarv can probably confirm.
You may want to look at the python implementation that allows to set a
tolerance and number of iterations

lfarv:
No, `atfitchrom` did not change since long, it' still using `linopt`… But this is indeed the problem. Some time ago, I did a search for the best default step values, which resulted in  3e-8 for DXYStep and 3e-6 for DPStep. These have been set in both Matlab and python, but `atfitchrom` still does not use them. In Matlab, you get it with:

`[DPStep, varargin]=getoption(varargin, 'DPStep'];`

So you can overwrite it with a key/value argument pair.
Note that the default value itself may be modified with `setoption`. Look at `help getoption` for details and how to get the list of all predefined constants.

simoneliuzzo:
Dear @lfarv ,

I do not understand what to do. 
1) should I modify atfitchrom to use some predefined global defaults? In case I did not understand how to do it. 
Or 
2)should I modify atfitchrom for an additional varargin, and later add to each usage call to atfitchrom this additional input? 

I would like the first to be the case, just to simplify my life...

lfarv:
@simoneliuzzo:The solution is to change **atfitchrom** to:
- use the predefined value
- (not strictly necessary) add **optional** key/value arguments pairs to handle the DPStep and also all `linopt` options as well,
- (not strictly necessary) move to `atlinopt6`

So no modification to the calling code, but  moderate work on `atfitchrom`. Mainly converting the processing of arguments to `getflag`/`getoption`/`getargs` instead of the clumsy `varelem` nested function. For the time being do locally what you proposed to solve your problem.

lfarv:
See the solution in #327

lfarv:
Solved in #327

Issue #303, opened 2021-09-23
Unstable lattice in transfer line mode for linopt6


swhite2401:
It seems the transfer line mode is not properly handled in linopt6 as it fails on tune calculation and returns and "Unstable lattice". I believe, we should simply return the phase advance (or nan) in transfer line mode.
For now linopt2 or linopt4 can be used for matching etc...
I haven't looked at the matlab implementation so I don't if it suffers the same issue

swhite2401:
Solved in #304

PR #304, opened 2021-09-23
Avoid W and Chrom calculations in transfer line mode

swhite2401:
For now the possibility to compute Q' and W in transfer line mode is not available (a warning is issued in case the user enables them with `twiss_in is not None`)
However a test was missing to prevent to run through the code that computes them in this case, these we added in this branch.
This solves #303.

It would be good to introduce these computation in future developments.

### DIFF
--- a/pyat/at/physics/linear.py
+++ b/pyat/at/physics/linear.py
@@ -338,8 +338,10 @@ def _linopt(ring, analyze, refpts=None, dp=None, dct=None, orbit=None,
 
     # Perform analysis
     vps, dtype, el0, els = analyze(mxx, ms)
-    tunes = numpy.mod(numpy.angle(vps) / 2.0 / pi, 1.0) if twiss_in is None \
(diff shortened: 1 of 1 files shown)


lfarv (commented):
It works, but instead these tests spread around, I would very much prefer to set the flags to False at the beginning: at line 327, replace:

```python
    else:                       # Transfer line
        if get_chrom or get_w:
            warnings.warn(AtWarning("'get_chrom' and 'get_w' are ignored in "
                                    "transfer-line mode"))
```
by

```python
    else:                       # Transfer line
        if get_chrom or get_w:
            warnings.warn(AtWarning("'get_chrom' and 'get_w' are ignored in "
                                    "transfer-line mode"))
            get_chrom = False
            get_w = False
```

swhite2401:
Done

lfarv (approved):
OK !

PR #312, opened 2021-10-04
Check keys on load

swhite2401:
Instead of printing a message without explanation when loading a matfile with the wrong key, an AtError is raised that prints the list of available keys

### DIFF
--- a/pyat/at/load/allfiles.py
+++ b/pyat/at/load/allfiles.py
@@ -38,7 +38,10 @@ def load_lattice(filepath, **kwargs):
     Known extensions are:
     """
     _, ext = os.path.splitext(filepath)
-    return _load_extension[ext.lower()](filepath, **kwargs)
+    try:
+        return _load_extension[ext.lower()](filepath, **kwargs)
+    except KeyError:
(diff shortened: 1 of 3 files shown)


swhite2401:
Oups, I checked out this branch from another dev branch, please ignore the changes in orbit.py

pyat/at/load/allfiles.py
```diff
@@ -38,10 +38,7 @@ def load_lattice(filepath, **kwargs):
     Known extensions are:
     """
     _, ext = os.path.splitext(filepath)
-    try:
-        return _load_extension[ext.lower()](filepath, **kwargs)
-    except KeyError:
```
lfarv, line 43: Do you think that the `KeyError` message is clear enough ? If so, you can apply the same simplification to `save_lattice`.

swhite2401:
Ah ok now I understand. This was very confusing as it was also catching the `KeyError` coming from the unknown `mat_key` in `matfile.py`. Now this is handled and throws an `AtError` so it should be ok. I put back the message (a bit different).

lfarv (approved):
Very good !

Issue #305, opened 2021-09-24
Test failure in Matlab / Python comparison tests


willrogers:
A significant difference in A1 when using fastring, not that I know what A1 is specfically.

cc @MJGaughran 

```
$ python -m pytest test_matlab
============================= test session starts ==============================
platform linux -- Python 3.7.2, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
system: Linux 4.18.0-305.12.1.el8_4.x86_64
numpy version: 1.21.2
rootdir: /home/hgs15624/code/at/pyat
plugins: lazy-fixture-0.6.3
collected 51 items                                                             

test_matlab/test_cmp_atpass.py ........                                  [ 15%]
test_matlab/test_cmp_avlinopt.py ......                                  [ 27%]
test_matlab/test_cmp_linopt6.py .......                                  [ 41%]
test_matlab/test_cmp_physics.py ........F...                             [ 64%]
test_matlab/test_int.py .                                                [ 66%]
test_matlab/test_orbit_matrices.py .................                     [100%]

=================================== FAILURES ===================================
_____________________________ test_fastring[hmba] ______________________________

engine = <matlab.engine.matlabengine.MatlabEngine object at 0x7f010bfa9358>
lattices = (Lattice([Marker('S28d', Periodicity=32), RFCavity('RFC', 0.0, 187500.0, 352372212.4670127, 31, 6000000000.0, PassMeth...([[0.0,0.0]]), ...}, {'Class': 'Drift', 'FamName': 'DR_02', 'Length': 0.042552, 'PassMethod': 'DriftPass'}, ...], None)

    @pytest.mark.parametrize('lattices',
                             [pytest.lazy_fixture('hmba')])
    def test_fastring(engine, lattices):
        py_lattice, ml_lattice, _ = lattices
        # Python call
        ring, ringrad = physics.fastring.fast_ring(py_lattice)
        # Matlab call
        ring_ml, ringrad_ml = engine.pyproxy('atfastring',
                                             ml_lattice, nargout=2)
        # Comparison
        for r, rml, idq in zip([ring, ringrad], [ring_ml, ringrad_ml], [3, 4]):
            assert_close(r[0].Frequency,    rml[1]['Frequency'], rtol=1.0e-20)
            assert_close(r[0].Voltage,      rml[1]['Voltage'], rtol=1.0e-20)
            assert_close(r[1].I2,           rml[2]['I2'], rtol=1.0e-20)
[...]

PR #314, opened 2021-10-04
Fix on atfastring

lfarv:
Fix the bug reported in #305.

### DIFF
--- a/pyat/at/physics/fastring.py
+++ b/pyat/at/physics/fastring.py
@@ -68,7 +68,7 @@ def _fring(ring, split_inds=[], detuning_elem=None):
         fastring.append(qd_elem)
     except:
         pass
-    fastring = Lattice(fastring, energy=ring.energy)
+    fastring = Lattice(fastring, **vars(ring))
     fastring.radiation
     return fastring
 
--- a/atmat/lattice/atfastring.m
+++ b/atmat/lattice/atfastring.m
@@ -54,7 +54,7 @@ ibeg=[1 iend(1:end-1)];
 
 xm=0.001;
 zm=0.0005;
-GLOBVAL.E0=atenergy(ring0);
+[GLOBVAL.E0,nper]=atenergy(ring0);
 [lindata,tunes,xsi]=atlinopt(ring0,0); %#ok<ASGLU>
 gamma=(1+lindata.alpha.*lindata.alpha)./lindata.beta;
 
@@ -66,7 +66,7 @@ ringrad=atradon(ring);
 orbit4=zeros(6,sum(markers));
 orbit4(1:5,:)=findsyncorbit(ring,0,markers);
 orbit4=num2cell(orbit4,1);
-r1=detuning(ring,gamma,xm,zm,orbit4(:,1));
+r1=detuning(ring,gamma,xm,zm,orbit4(:,1))/nper;
 
 orbit6=num2cell(findorbit6(ringrad,markers),1);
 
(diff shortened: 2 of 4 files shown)


lfarv:
In fact, there was a bug on each side ! Though not optimised, it should be ok for the moment.

lfarv:
For the radiation version of the fast ring, the non-linear and diffusion element are swapped compared to the initial version and to the Matlab version. I think I remember this make a difference (on the equilibrium state) in the first tests we did a long time ago. Are you sure they can swapped ? Maybe @simoneliuzzo or @carmignani remember, I may be wrong…

lfarv:
@swhite2401 : line 61: what kind of error do you want to ignore ? It would be better to be more precise to avoid ignoring unexpected errors.

```python
     try:
         qd_elem = gen_quantdiff_elem(merged_ring)
         fastring.append(qd_elem)
     except:
         pass
```

pyat/at/physics/fastring.py
```diff
@@ -41,12 +39,9 @@ def _rearrange(ring, split_inds=[]):
 
 def _fring(ring, split_inds=[], detuning_elem=None):
     all_rings, merged_ring = _rearrange(ring, split_inds=split_inds)
-    ibegs = get_refpts(merged_ring, 'xbeg')
-    iends = get_refpts(merged_ring, 'xend')
-    markers = numpy.sort(numpy.concatenate((ibegs, iends)))
-    with warnings.catch_warnings():
```
swhite2401, line 47: I think this change is fine, however we generally use fond_orbit_sync for this kind of calculation without radiation. This is why I had suppressed the warning.

lfarv: Agreed, but in this case, `dct=0` is strictly equivalent to `dp=0` : it's the on-momentum orbit. And it's the default behaviour.

swhite2401:
> 
> 
> @swhite2401 : line 61: what kind of error do you want to ignore ? It would be better to be more precise to avoid ignoring unexpected errors.
> 
> ```python
>      try:
>          qd_elem = gen_quantdiff_elem(merged_ring)
>          fastring.append(qd_elem)
>      except:
>          pass
> ```

I think this was a simple way to manage the case where `ring.radiation=True` but damping is off -> cavity only

swhite2401:
> 
> 
> For the radiation version of the fast ring, the non-linear and diffusion element are swapped compared to the initial version and to the Matlab version. I think I remember this make a difference (on the equilibrium state) in the first tests we did a long time ago. Are you sure they can swapped ? Maybe @simoneliuzzo or @carmignani remember, I may be wrong…

Why would this be an issue?

lfarv:
> Why would this be an issue?

No idea. I just remember that the equilibrium phase-space when running a large population was different, but again I may be wrong. It could be tested again

lfarv:
> I think this was a simple way to manage the case where `ring.radiation=True` but damping is off -> cavity only

Ok. Do you know what class of error this raises ? If this is not clear, I'll put `except Exception:` to remove the warning…

swhite2401:
> 
> 
> > I think this was a simple way to manage the case where `ring.radiation=True` but damping is off -> cavity only
> 
> Ok. Do you know what class of error this raises ? If this is not clear, I'll put `except Exception:` to remove the warning…

I am not quite sure, I think it is the cholesky transform complaining (ValueError?), I can check that later today, but what you propose is fine

lfarv:
Well this tool is not so stupid: `except Exception:` is still too broad…

swhite2401:
> 
> 
> Well this tool is not so stupid: `except Exception:` is still too broad… Anyway, that's good enough.

haha ok then I check later today and let you know

lfarv:
The except clause is now more strict. OK for merging ?

PR #316, opened 2021-10-06
Cleanup Lattice and cavity access

lfarv:
The Lattice object has new methods an properties:
```
get_slip_factor
get_revolution_frequency
```
and corresponding read-only properties:
```
slip_factor
revolution_frequency
```
The cavity access module is improved, particularly on frequency: it is possible to set the frequency to the "nominal" value, optionally specifying an off-momentum.

All this is now explained in a [new page of the Web site](https://atcollab.github.io/at/pcavitycontrol.html). To allow previewing this page, the site is temporarily directed to this development branch. Please look at it, in particular to the frequency control, and give your comments, either on the code or on the documentation.

NB: I had to change the name of frequency controls, to clearly distinguish `revolution_frequency` and `rf_frequency`. If this is not acceptable, I can revert to initial (`frequency` in short).

### DIFF
--- a/pyat/at/__init__.py
+++ b/pyat/at/__init__.py
@@ -1,8 +1,10 @@
 """Python port of the Accelerator Toolbox"""
 
 # Make all functions visible in the at namespace:
+import sys
 from .lattice import *
 from .tracking import *
 from .physics import *
 from .load import *
 from .matching import *
+sys.modules['at.constants'] = sys.modules['at.lattice.constants']
--- a/pyat/at/lattice/__init__.py
+++ b/pyat/at/lattice/__init__.py
@@ -4,7 +4,6 @@ Helper functions for working with AT lattices.
 A lattice as understood by pyAT is any sequence of elements.  These functions
 are useful for working with these sequences.
 """
-from .constants import *
 from .options import DConstant
 from .elements import *
 from .utils import *
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -16,7 +16,7 @@ import numpy
 import math
 import itertools
 from warnings import warn
-from at.lattice import clight
+from at.lattice.constants import clight
 from at.lattice import AtError, AtWarning
 from at.lattice import uint32_refpts as uint32_refs, bool_refpts as bool_refs
 from at.lattice import refpts_iterator, refpts_len
--- a/pyat/at/physics/energy_loss.py
+++ b/pyat/at/physics/energy_loss.py
@@ -6,7 +6,7 @@ from scipy.optimize import least_squares
 from at.lattice import Lattice, Dipole, Wiggler, RFCavity
 from at.lattice import check_radiation, AtError
 from at.lattice import checktype, set_value_refpts, get_cells, refpts_len
-from at.lattice import clight, e_mass, Cgamma
+from at.lattice.constants import clight, e_mass, Cgamma
 from at.tracking import lattice_pass
 
 __all__ = ['get_energy_loss', 'set_cavity_phase', 'ELossMethod',
--- a/pyat/at/physics/linear.py
+++ b/pyat/at/physics/linear.py
@@ -5,7 +5,8 @@ import numpy
 from math import sqrt, pi, sin, cos, atan2
 import warnings
 from scipy.linalg import solve
-from ..lattice import DConstant, get_s_pos, clight
+from ..lattice.constants import clight
+from ..lattice import DConstant, get_s_pos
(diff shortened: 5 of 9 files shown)


swhite2401:
This looks good, I have question though: why are you storing the particle mass in `constats.py` while other quantities are initiated in the` __init__.py` of `physics`?

Also variable names like `clight` are used by many, are not afraid that this gets over-written by mistake? Don't you want to call these variables `atclight` etc... instead?

lfarv:
> Reading the new help of `atsetcavity` I would set the Harmonic number to `HarmNumber/ncells` even for a the full ring with `periodicity=1`, this is not good.... 

if periodicity (or cells) is 1, it sets the Harmonic number to HarmNumber. Is not it what you want ?

I think you are misunderstanding the meaning of `periodicity`. `periodicity` is the number of times the lattice description is repeated to get a full ring. So here are a few hints:
- if you are dealing with non-periodic machines, you describe the full ring, set periodicity to 1, and AT does exactly what you ask for: all input are set unchanged to the lattice elements,
- it often happens that in the early design stage, we consider a ring as periodic. Then you describe only one cell, and set periodicity to 32 (for instance). Then hundreds of details break the periodicity, you describe the full ring, and set periodicity to 1. Thanks to `periodicity`, both machines have the same circumference, tunes, revolution frequency, harmonic number etc.
- If in spite of this, you still don't like it, always set periodicity to 1! If you describe 1 cell, you get 1 cell, if you describe the full ring, you get the ring. It's up to you to convert the parameters between both.

It's not a question of speed. It just makes the description of periodic machines easier by describing only one cell and say "repeat it n times".

Of course it may happen that periodicity is not properly taken into account in the code. The solution is to repair it, not to remove it.

lfarv:
2 questions about `constants.py`:

1. > Well I often write `clight=2.998e8` in my scripts because my mind doesn't seem to be able to remember more than 3 digits, in which case it will affect the results...

   So there is an easy way to update your script: remove the line `clight=2.998e8` and add `from at import clight`. All the rest of the code does not change and you get a more accurate result. Nice, isn't it ? True, there is a problem if you have both lines, but why ?

2. At the moment, all the constants are in the `lattice` namespace, and consequently in the `at` namespace, so that you can get `clight` (for instance) as one of:
   ```
   from at import clight
   from at.lattice import clight
   ```
   We can isolate them a bit more by adding only the constants module to the  lattice namespace. Then, to access `clight`, you would need to use: 
   ```
   from at.lattice.constants import clight
   ```
   Do you have a preference ?

swhite2401:
> 
> 
> 2 questions about `constants.py`:
> 
>     1. > Well I often write `clight=2.998e8` in my scripts because my mind doesn't seem to be able to remember more than 3 digits, in which case it will affect the results...
>        
>        
>        So there is an easy way to update your script: remove the line `clight=2.998e8` and add `from at import clight`. All the rest of the code does not change and you get a more accurate result. Nice, isn't it ? True, there is a problem if you have both lines, but why ?
> 
>     2. At the moment, all the constants are in the `lattice` namespace, and consequently in the `at` namespace, so that you can get `clight` (for instance) as one of:
>        ```
>        from at import clight
>        from at.lattice import clight
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        We can isolate them a bit more by adding only the constants module to the  lattice namespace. Then, to access `clight`, you would need to use:
>        ```
>        from at.constants import clight
>        from at.lattice.constants import clight
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        Do you have a preference ?

What you have at the moment is fine, still I think your implementation is not safe and cannot guaranty that clight (or other constants) are not overwritten. Providing we cannot declare constants in python I would strongly advocate for a naming convention that prevents interference with users scripts.
I would be an error to have both lines as you mention but this can happen and we should make sure that we are robust (as much as possible) against this.

swhite2401:
Sure none of these are good but we may try to make it a bit safer by naming our constants `_clight` for instance. It costs nothing and there is less probability that the internal AT constants are overwritten by mistake ( `clight` is obvious but what about `mu0` or `eps0` for example?).

I also found this (I haven't tried myself), but this could be a solution:
```
class _const:
    class ConstError(TypeError): pass
    def __setattr__(self,name,value):
        if self.__dict__.has_key(name):
            raise self.ConstError, "Can't rebind const(%s)"%name
        self.__dict__[name]=value
import sys
sys.modules[__name__]=_const()
```

lfarv:
> I think keeping constants outside of the `at` namespace is a reasonable design decision regardless.

This is done. I think we are now ready to merge. Any other comment ?

PR #317, opened 2021-10-15
Cavity pass accounting for frequency shift frequency

swhite2401:
`CavityPass` was updated to account for changes in RF frequency in all configurations.
This also allows to simplify `find_orbit6` method and account for the frequency shift in tracking simulation.
The time coordinate behavior is not affected as in `RFCavityPass`.
The path length can be set by the user manually or is derived in `at.c` using the element lengths cumulative sum.
The harmonic number is not required anymore as it is derived internally in `CavityPass.c`.

This `CavityPass` could replace the present `CavityPass+RFCavityPass`.

Modifications were applied on both python and matlab implementation, however they were tested only for python: help would be appreciated for the matlab part.

Let me know what you think.

### DIFF
--- a/pyat/test/test_physics.py
+++ b/pyat/test/test_physics.py
@@ -67,6 +67,7 @@ def test_find_orbit4_with_two_refpts_with_and_without_guess(dba_lattice):
 
 
 def test_orbit_maxiter_warnings(hmba_lattice):
+    set_rf_frequency(hmba_lattice, Frf.NOMINAL)
     hmba_lattice_rad = hmba_lattice.radiation_on(copy=True)
     with pytest.warns(AtWarning):
         physics.find_orbit4(hmba_lattice, max_iterations=1)
@@ -89,6 +90,7 @@ def test_find_m44_returns_same_answer_as_matlab(dba_lattice, refpts):
 
 @pytest.mark.parametrize('refpts', ([145], [20], [1, 2, 3]))
 def test_find_m66(hmba_lattice, refpts):
+    set_rf_frequency(hmba_lattice, Frf.NOMINAL)
     hmba_lattice = hmba_lattice.radiation_on(copy=True, quadrupole_pass=None)
     m66, mstack = physics.find_m66(hmba_lattice, refpts=refpts)
     assert_close(m66, M66_MATLAB, rtol=0, atol=1e-8)
@@ -134,6 +136,7 @@ def test_find_orbit6(hmba_lattice):
 
 
 def test_find_orbit6_produces_same_result_with_keep_lattice_True(hmba_lattice):
+    set_rf_frequency(hmba_lattice, Frf.NOMINAL)
     hmba_lattice = hmba_lattice.radiation_on(quadrupole_pass=None, copy=True)
     orbit0, _ = physics.find_orbit6(hmba_lattice)
     # Technicality - the default arguments to find_orbit6 mean that
@@ -272,6 +275,7 @@ def test_nl_detuning_chromaticity(hmba_lattice):
 
 
 def test_quantdiff(hmba_lattice):
+    set_rf_frequency(hmba_lattice, Frf.NOMINAL)
     hmba_lattice = hmba_lattice.radiation_on(copy=True)
     dmat = physics.radiation.quantdiffmat(hmba_lattice)
     lmat = physics.radiation._lmat(dmat)


swhite2401:
I think you did not understand the interest of this `passmethod`: it is compatible with ALL AT functionalities and replaces both `CavityPass` and `RFCavityPass`, this is a major simplification to a long standing issue, that was in fact requested.
The problem with non relativistic particles is easily solved and in any case RFCavityPass suffers the same issue.

It is not true that `RFCavityPass` does not need any modification to other files: `at.c` needs to record the machine length and turn number to pass it to the function in `params`. Other functions (like `find_orbit6`) are not modified but they are simply not compatible with `RFCavityPass` due to the incremental 5th coordinate.
I do not see what is complicated there...for me this is rather a simplification and also works for any combination of cavities ( I tested it). The formula is as simple as `RFCavityPass` (it is in fact the same...).

So I can see many advantages:
-a single `passmethod` for RF cavities that behaves as expected in ALL cases: you change the RF frequency the energy changes, isn't that how it should be?
-the particles 5th coordinate oscillates around 0, as opposed to `RFCavityPass` for which a post-processing script was needed to recover the standard behavior
-removes any ambiguity on the harmonic number vs frequency, however the harmonic number can still be used to set the frequency to its "nominal" value
-The path length can be a users input: you may easily take into account misalignments if you want to
-`find_orbit6` is much simpler and does no require any assumption on the numbers of cavities, which to take into account etc...
-there is still a request to handle cleanly multiple RF systems (for now `find_orbit6` and `cavity_access` don't) and this will make it simpler

I understand that the `CavityPass` is a critical AT element and we have to be very careful, but if this works and gives exact results, I think this is really an improvement for users.
The only "issue" is that now the users needs to make the frequency is properly set otherwise find_orbit6 will return a dispersive orbit, but I very much prefer it this way as this is what we expect from a real machine.

lfarv:
I see see your point, but I want first to check what `RFCavityPass` can do, and what is missing, if any. So lets's take 2 points for the moment:

### a single passmethod for everything:
`RFCavityPass` can (and should) replace both `CavityPass` and `RFCavityPass`. It is compatible with ALL AT functionalities. Here is a Jupyter notebook:
```python
import at
import numpy as np
```

Take the hmba test lattice and turn radiation on:


```python
ring=at.load_lattice('/Users/famille/dev/libraries/at/pyat/test_matlab/hmba.mat').radiation_on(copy=True)
```

Set the PassMethod of all cavities to `RFCavityPass`:


```python
at.set_value_refpts(ring,at.checktype(at.RFCavity),'PassMethod','RFCavityPass')
```

Set the frequency to get 1% off-momentum:


```python
ring.set_rf_frequency(dp=0.01)
```

Compute the closed orbit:


```python
o0,_=at.find_orbit(ring)
```

Let's track 2 particles:
- the closed orbit delayed by 3 mm
- the closed orbit


```python
rin=np.stack((o0+[0,0,0,0,0,0+0.003],o0), axis=0).T
rin
```




    array([[ 2.06716284e-05,  2.06716284e-05],
           [-2.67035110e-09, -2.67035110e-09],
           [ 0.00000000e+00,  0.00000000e+00],
           [ 0.00000000e+00,  0.00000000e+00],
           [ 9.81790180e-03,  9.81790180e-03],
           [-5.68280966e-02, -5.98280966e-02]])



The frequency shift is computed from &eta;<sub>c</sub> , so the off-momentum value is not exactly the one specified: 0.98% instead of 1%. That's good enough for our test

Now, track over 9100 turns:


```python
rout=at.lattice_pass(ring,rin.copy(),9100)
```

Plot the test particle


```python
plt.plot(rout[5,0,0,::50],rout[4,0,0,::50],'.')
plt.xlabel('${\Delta}s$')
plt.ylabel('$\delta$')
```

    Text(0, 0.5, '$\\delta$')

output_14_1

- The energy oscillates around the selected off-momentum value
- the path lengthening drifts. This is normal: the path of the off-momentum closed orbit is longer than the reference path, so the path lengthening of the closed orbit monotonously increases, and the test particle oscillates around this monotonous drift.

So let's take the difference of both path lengthenings (test particle - closed orbit):


```python
plt.plot(rout[5,0,0,::50]-rout[5,1,0,::50],rout[4,0,0,::50],'.')
plt.xlabel('${\Delta}s - {\Delta}s_{co}$')
plt.ylabel('$\delta$')
```

    Text(0, 0.5, '$\\delta$')

output_16_1
    
    
The momentum oscillates around the off-momentum value, the path length oscillates around the closed orbit path length, as expected.
So I claim that the tracking is correct, and find_orbit works as it should. RFCavityPass can replace both present methods.

### handling of multiple RF systems:
Now that I am convinced that the tracking, and particularly the 6<sup>th</sup> component is right, this is true for as many different cavities as you want. Each cavity has its own frequency, voltage and phasing. `cavity_access` is a convenience to deal all the cavities of the same system, each system being defined by the location of its cavities. But you can access cavities individually of you like. I keep in mind your remark about find_orbit6: here there is something to clarify.

More to come…

lfarv:
About `HarmNumber`:

Here I think I agree with you: `HarmNumber` should not appear in the attributes of a cavity. It should be an attribute of the lattice, useful since it defines the nominal RF frequency. Historically it has been put in the cavity element because there was no better place to put it, like for `energy`. `energy` is now a lattice attribute (python) or stored in the RingParam element (Matlab). The same should be done for `HarmNumber`. 
The first thing to do is to get rid of it in passmethods. It does not appear in `CavityPass`, and is easy to remove from `RFCavityPass` (one more PR).
For the rest, no urgency in my opinion.

more to come…

lfarv:
I come now to the heart of the problem: the tracking.

Comparing the results of `RFCavityPass` and the new method, they diverge on the 6<sup>th</sup> coordinate. As the definition of this coordinate is not ambiguous, one method has to be wrong (or possibly both…).

It comes out that `RFCavityPass` is right, and that the new method is wrong. Or to say it more softly, the new method introduces its own the definition of the  6<sup>th</sup> coordinate, different from the AT one. It’s no more the path lengthening of the particle. The fundamental problem is that a thin element (the cavity) changes the path lengthening, which is physically impossible.

So I have two questions (at least) before thinking of changing AT's coordinate system:

1. Can we give a clear definition of this modified 6<sup>th</sup> coordinate? I recall the present definition :
"Path lengthening with respect to the reference particle (fictitious particle travelling on the theoretical axis without radiation or acceleration)".
  Missing this, any derivation based on this coordinate, or more generally on the longitudinal tracking becomes unclear

2. Such a change has unpredictable consequences on existing code. You pointed out that it affects find_orbit6, but what about derivations of the 6x6 transfer matrix? It may be transparent if the orbit subtractions cancel out the problem, but it may be not. Going through all AT to identify all problems is already a huge job, but AT has been around for years, and there are many user projects based it, so that we cannot take the risk of breaking anything.

So the new method, as it is now, cannot be the default (as long as it breaks the path lengthening). 
First, we can safely adapt `at.c` to provide the machinery allowing the method to work. And make it available optionally as an alternate method for cavities. Knowing that further than tracking, any result must carefully examined.

Next, I’ll have comments on `find_orbit6`. I agree about it’s present limitations.

lfarv:
Hi @swhite2401: what are you doing in front of a computer on such a nice day ? I'm still locked at home, so I take a few minutes to answer you:

>the reference path is just a convention.

True, but changing it is a very delicate thing…

>We may well use the off-momentum path to prevent the drift (as proposed in this PR), this is in fact more consistent with real data, the user should at least be given the choice.
>-There are (at least) 2 application that will fail due to this drift: harmonic analysis of the longitudinal motion, tracking with impedance (slicing will fail in multiturn mode: PR to come), however one may argue that these are generally not done off-momentum but who knows..

Yes, but to remove the off-momentum drift, you have to compute it properly. And subtracting it also answers the question of harmonic analysis. No idea for impedance…


>-I insist that these lines in `find_orbit6` are really hurting and can be avoided. In particular, I do not like at all that the closed orbit depends on the harmonic number...:
> 
> ```
>     f_rf = cavities[0].Frequency
>     harm_number = cavities[0].HarmNumber
>     theta = numpy.zeros((6,))
>     theta[5] = clight * harm_number / f_rf - l0
> ```
>

I fully agree, it has to be improved. Still don't know how. 

> To be honest I do not see much interest in simply replacing `CavityPass` with `RFCavityPass` without answering these issues... 

Just because `CavityPass` is wrong if the frequency is not the nominal one. For me, that's a good reason!

>On the other hand, the solution proposed in this PR solves all of the above in a simple way and seems to perform correctly (at least for all the tests I have done, matlab not checked) but of course if you see fundamental issues or would like to propose more elegant solutions I am very interested in hearing them!

I'm still trying to understand the implications of your proposal involves. I think that a rigorous computation of the path lengthening is necessary and have the fear (maybe unjustified) that simply hiding it is not a solution: to avoid a real, though inconvenient, behaviour, we should not cheat with the measurement instrument to hide it.
And fundamentally, I think we cannot abruptly change the definition of the 6th coordinate. You cannot figure out where it is used in user code. Plus the fact that I still don't understand what is this new 6th coordinate and how it can be used.

swhite2401:
I fact I just realized `RFCavityPass` has a bug...no additional term is applied for turn=0...so this is why you were getting identical results in `find_orbit6` that uses only the first turn.
With the bug fixed you get an offset corresponding exactly to the `theta` applied in `find_orbit6`, I guess this is good news because it allows to remove these doubtful lines in `orbit.py`, but requires more work for verification etc...to be followed up.

lfarv:
Hi @swhite2401: I know how to get rid of `HarmNumber` in `find_orbit6`, it's directly inspired by your pass method: let's take the lower frequency among all active cavities. Then look for the closest multiple of the revolution frequency. The difference of these two frequencies directly gives the necessary path lengthening. It's the value to put in the right member of the equation! That's it.

Now for cavities having different frequencies:
- if all frequencies are separated by an integer times the revolution frequency, the phasing of all cavities along the second turn will be identical, so the solution is correct over the 2nd turn and the next ones. This includes:
  - "modulation" cavities : h-1 or h+1
  - "harmonic" cavities: n * h
- if not, then the second turn will be different, and there is **no closed orbit** ! You have to wait a full period of the beating between frequencies to get back to identical conditions. The orbit closes one over many, many turns. Your solution does not help in this case: it is also based on a single turn, and must give the same result as the one I propose.

3rd PR! With this, I think we cover most of the problems. Not impedance, but I did not figure out yet what is the problem. That's a first positive result of this pull request.

swhite2401:
> 
> 
> > I fact I just realized `RFCavityPass` has a bug...no additional term is applied for turn=0...
> 
> It's not a bug. What has to enter in the RF sine wave is the **absolute** path length. This is equal to:
> 
> [the reference path (nturns * circumference)] + [path lengthening]
> 
> Along the 1st turn, you just have the path lengthening !

Are you really sure about this? The path length should offset as soon as you cross a dipole regardless on which turn you are no?

swhite2401:
In fact the problem with impedance might not be so severe, the main issue is to handle fixed slicing -> if the bunch is drifting the slices have to drift as well...
Then there might be other issue because everything assumes the origin of time is the zero crossing of the RF voltage but I have not looked at it in details. Anyway, it will work if the frequency is adjusted.

Setting the time origin to the zero crossing of the RF voltage is what I was trying to achieve: I believe this is a clean well defined reference, but let's keep it in mind for later. Offsetting the path length in cavities is not nice I have to come up with a better solution.  The intermediate solution you proposed is ok for now

lfarv:
> But this additional term in the energy kick in RFCavityPass is equivalent to a non-zero TimeLag...except it is incrementing turn after turn instead of being static

It's not a time lag, it's just time flowing. To avoid values increasing too much, we just subtract from the absolute time an integer number of RF periods, which does not change the sine value. This integer does not have to be the "harmonic number", any integer is ok. Taking the optimum one is the way we will upgrade RFCavityPass to get rid of "harmonic number".

> Mmm... but what is the reference path? Isn't it the path of the synchronous particle on-momentum?

No ! The reference path is a fictitious particle with the nominal energy travelling around the ring without radiating. At any time it defines the origin the coordinate system at its location, both transverse and longitudinal. Consequently it also defines the time origin of cavities as the instant when if crosses the cavity on the 1st turn.

The on-momentum synchronous particle is delayed with respect to the reference because it needs some voltage to compensate the losses. Since it follows the same path as the reference, its path length is ~equal to the circumference, its delay(or path lengthening) is constant.

Now let's look at the off-momentum closed-orbit particle. Its path is longer that the reference, so the "lengthening" increases at each dipole, you are right. On the first turn, it should cross the cavity at the same time as the on-momentum synchronous particle (+/- its lengthening over 1 turn) and then monotonously deviate. To get the absolute path length at a point along the ring, you would add the reference path (turns * circumference + fractional portion of circumference) and the lengthening at that point (6th coordinate). But for the arrival time in the cavity, remember that the time origin of the cavity is not the beginning of the first turn, by the passing time of the reference particle on 1st turn. So that the fractional part of the circumference from origin to cavity must be ignored. That's what is in the argument of the sine in RFCavityPass.

The system is fully consistent, though not easy to figure out. But you have to take it as a whole, you cannot cheat!

PR #318, opened 2021-10-18
Remove the dependency of find_orbit6 on the harmonic number

lfarv:
`find_orbit6` does not depend on the harmonic number any more. The RF frequency is  deduced from the cavity having the lowest frequency in the ring. The harmonic number is deduced from the RF frequency and revolution frequency.

### DIFF
--- a/atmat/atphysics/Orbit/xorbit_6.m
+++ b/atmat/atphysics/Orbit/xorbit_6.m
@@ -10,21 +10,15 @@ function Ri = xorbit_6(RING,varargin)
 [Ri,varargs]=getargs(varargs,Ri,'check',@(arg) isnumeric(arg) && isequal(size(arg),[6,1])); %#ok<ASGLU>
 
 cavities=RING(atgetcells(RING,'Frequency'));
-
 if isempty(cavities)
     error('AT:NoFrequency', 'The lattice has no Cavity element')
 end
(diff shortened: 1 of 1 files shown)


lfarv:
@carmignani pointed out [here](https://github.com/atcollab/at/pull/319#issuecomment-946792784) a potential problem for large frequency deviations. Looking at the equation for 6D closed orbit (see `help(find_orbit6)`):
```
    (X, PX, Y, PY, DP, CT2 ) = M (X, PX, Y, PY, DP, CT1)

    with M the one-turn 6D transfer matrix,
    and constraint  CT2 - CT1 = C*HarmNumber(1/Frf - 1/Frf0)
```
the harmonic number does play a role. For large deviations, taking the nearest harmonic might be wrong. So the initial code is right and its dependence on harmonic number is correct. Seeing this, I was tempted to close this pull request and keep the initial version.

But looking closer, here is what may happen: let's increase the frequency. The beam gets off-momentum and deviates from the central trajectory. If we reach a frequency deviation of one revolution frequency (which looks very optimistic…), the beam may very well jump back on axis and on momentum, running on a different harmonic number. At least if we re-inject, that's what will happen. I do not know when the beam may jump, is it at h + 1/2 ? However, the modified version of `find_orbit6` seem to agree with what may happen in reality: the beam may stabilise around the nearest harmonic number, possibly not the design one.

So I tend to keep this version…

PR #319, opened 2021-10-18
Remove the dependency of RFCavityPass on harmonic number

lfarv:
Going on with removing harmonic number where it is not necessary, we now remove it from `RFCavityPass`.

### DIFF
--- a/atintegrators/RFCavityPass.c
+++ b/atintegrators/RFCavityPass.c
@@ -8,6 +8,7 @@
 #include <math.h>
 #include "atelem.c"
 #include "atlalib.c"
+#include "driftkickrad.c"
 
 #define TWOPI  6.28318530717959
 #define C0  	2.99792458e8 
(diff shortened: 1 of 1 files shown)


jbengtsson:
I think that this also requires an update of/change in ring_parameters.py:

  voltage = ring.voltage -> ring.rf_voltage
 
https://github.com/atcollab/at/blob/eb3fe845d341511c7b95af44e681dd560a371e9e/pyat/at/physics/ring_parameters.py#L103

jbengtsson:
As you probably know, the core beam dynamics C code for AT, is a manual translation of the "numerical engine" for the beam dynamics Pascal library, aka Tracy-2, that implemented 1990; having been tasked with providing an on-line model to guide the commissioning of ALS. Hence, as a side effect, I could also calibrate the model; vs. a modern synchrotron light source.

Instead, for the SLS commissioning M. Böge & J. Chrin made a machine translation of it (with p2c) & implemented it as a client/server architecture:

  M. Böge, J. Chrin “A Corba Based Client-Server Model for Beam Dynamics Applications at the SLS” ICALEPS 1999.

Bottom line, clearly, the harmonic number is a property of the RF Cavity object/structure/class.

jbengtsson:
Roger that. (from radio comm: *received & understood*)
Thanks.

lfarv:
> In RFCavityPass the harmonic number is needed in the tracking. The proposed modification is fine, because you compute h from the Frequency and the circumference, but I would keep it as optional field. In case of very off-energy simulations, the computation could fail by one unit in the harmonic cavities.

No, it's not necessary ! In the formula quoted by @jbengtsson 

>ct -> ct - harm_no / f_rf * c

harm_no can be any integer, it does not change the result of the sine function. It's only there to avoid numerical overflow. I do not compute the harmonic number from the frequency, I computed the integer which minimises the argument of the sine function.

> the harmonic number is a *property* of the RF cavity (not something being
computed; i.e., nonsense to me)

OK, so we can call this computed integer with another name, no problem.

What @carmignani mentions is valid for `find_orbit6`: there it's true, the path lengthening is computed with respect to the the nearest harmonic number, which for very large frequency deviations may be different form the real harmonic number.

jbengtsson:
True.

However, for the original implementation (1990), for simplicity for the
beam line (Pascal) structure/record, I took a LEGO approach (something that
I grew up with); which hasn't changed to date.
In other words, each element has only local properties/attributes, which I
then assemble into a ring; or beam line, collider, re-circulator, etc., for
that matter.
I.e., the internal numerical engine makes no assumption on whether its
periodic or single pass system.
Hence, e.g. the *orbit finder* is on the *user level*, i.e., depends
on the *use
case*.

Clearly, the harmonic number is given by the circumference & number of
buckets; i.e., in the case of a periodic system.
And the former is a number that I can obtain from e.g. the RF engineer; for
a local description of their subsystem/component/LEGO piece.
And, if there wasn't an FP overflow issue with longterm tracking, the RF
cavity object only has local parameters: length, voltage, frequency, and
phase; the latter defined relative to the design orbit/co-moving frame.

jbengtsson:
Dear All,

Nice conclusion & summary.

Bottom line, what's essential (I think) is to (strategically) make a clear
distinction between local vs. global properties of the structure/lattice;
i.e., a *LEGO approach*.
As for the implementation in which object to store the *harmonic number*:
the RF cavity or Ring; *to do or not to do...* I think is a matter of
convenience; depending on the *use cases*.
Fundamentally, as already observed/stated/concluded, it's not part of the
local beam dynamics model/integrator for the component (RF cavity).

Best regards.

carmignani:
> No, it's not necessary ! In the formula quoted by @jbengtsson
> `ct -> ct - harm_no / f_rf * c`
> harm_no can be any integer, it does not change the result of the sine function. It's only there to avoid numerical overflow. I do not compute the harmonic number from the frequency, I compute the integer which minimises the argument of the sine function.

Yes, that is true. Then you are right and we can remove the harmonic number.

Issue #325, opened 2021-11-09
Integral tolerances in Touschek Piwinski lifetime in matlab


carmignani:
I did some Touschek lifetime simulations varying different parameters for the ESRF storage ring and I noticed that with the default parameters often we get not precise results, at least for very large vertical emittances.
 
I attach here a plot of the Touschek lifetime computed with TouschekPiwinskiLifeTime as a function of vertical emittance.
The blue is using the method "quad", that is not recommended by matlab, but it gives the good results. 
The other ones are instead using the "integral" method, but changing the relative and absolute tolerances. The red curve is with the default values of tolerances of "integral", which is what we get by running TouschekPiwinskiLifeTime with default parameters.

I think the best is to allow the user to give the absolute and relative tolerances as input (now you have to modify the function to change the tolerances) and to give a smaller number as default. It seems like 1e-16 and 1e-12 are small enough for the case I studied.

![Screenshot 2021-11-09 at 15 47 37](https://user-images.githubusercontent.com/14995243/140945924-27d6a43c-e662-4ca9-9a84-fe955e06e1e3.png)

simoneliuzzo:
I see only benefit in this change.

swhite2401:
I agree. Why is quad not recommended? Looking at these curves I would just set it as default...
Btw, I did an implementation in python using `integrate.quad` that I still need to push at some point that seems to work very well

carmignani:
I don't know exactly why quad is not recommended: this is what they write in the documentation.
![Screenshot 2021-11-10 at 09 12 51](https://user-images.githubusercontent.com/14995243/141075032-66838715-8e37-4572-bbcd-ddf5b3e0e837.png)

lfarv:
See #326 for an answer to this!

lfarv:
Solved in #326

PR #326, opened 2021-11-14
Added optional tolerances in "TouschekPiwinskiLifeTime"

lfarv:
@carmignani pointed out in #325 a convergence problem in `TouschekPiwinskiLifeTime`. This pull request adds optional name-value argument pairs to specify the tolerances on the call to `integral()`.
For example, for an absolute tolerance of 1.0e-16, TouschekPiwinskiLifeTime` can be called as:
```Matlab
[Tl,contributionsTL]=TouschekPiwinskiLifeTime(...,'AbsTol',1.e-16);
```
Or for Matlab >= R21a:
```Matlab
[Tl,contributionsTL]=TouschekPiwinskiLifeTime(...,AbsTol=1.e-16);
```
The default values are set as `AbsTol = 1.0e-16, RelTol = 1.0e-12`.

It is still unclear why Matlab discourages using the "quad" integration method, which seems indeed to give good results…

### DIFF
--- a/atmat/atphysics/TouschekPiwinski/TouschekPiwinskiLifeTime.m
+++ b/atmat/atphysics/TouschekPiwinski/TouschekPiwinskiLifeTime.m
@@ -72,7 +72,7 @@ end
 emitx=pa.modemittance(1);
 emity=emitx./2;
 
-[emitx,emity,integrationmethod,sigp,sigs,rest] = ...
+[emitx,emity,integrationmethod,sigp,sigs,~] = ...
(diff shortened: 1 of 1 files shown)


lfarv:
@carmignani: I am missing data to validate this version. Can you check?

simoneliuzzo (approved):
ok for me!

carmignani:
I tried the function running the same tests of the issue #325. The default values give correct results and the change of tolerances from the input works well, so for me we can merge.

carmignani (approved):

PR #327, opened 2021-11-15
Fit tunes and chromaticities

lfarv:
@simoneliuzzo reported in #302 a problem in `atfitchrom`, where the `dp` step could not not modified. Here is the solution:
- the `dp` step can be modified with the name/value pair "`'DPStep',step`". Example:
  `newring=atfitchrom(ring,...,'DPStep',1.e-6)` or
  `newring=atfitchrom(ring,...,DPStep=1.e-6)`
- the default value is the `DPStep` option value (see `getoption`/`setoption`). It is by default 3.0e-6, which is optimum in most cases.
- `atfitchrom` now uses `atlinopt6`. However it is still restricted to radiation OFF because the chromaticity computation uses fixed dp/p values.

Simultaneously, `atfittune` is also modified to use `atlinopt6`, and can now be used with radiation ON.

### DIFF
--- a/atmat/lattice/atfitchrom.m
+++ b/atmat/lattice/atfitchrom.m
@@ -20,6 +20,7 @@ check_radiation(ring,false);
 [dpp,varargin]=getargs(varargin,0.0,'check',@(arg) isscalar(arg) && isnumeric(arg));
 [newchrom,famname1,famname2]=deal(varargin{:});
 
+idx1=varelem(ring,famname1);
 idx2=varelem(ring,famname2);
 kl1=atgetfieldvalues(ring(idx1),'PolynomB',{3});
 kl2=atgetfieldvalues(ring(idx2),'PolynomB',{3});
--- a/atmat/lattice/atfitchrom.m
+++ b/atmat/lattice/atfitchrom.m
@@ -1,8 +1,10 @@
 function newring=atfitchrom(ring,varargin)
-%ATFITTUNE fits chromaticites by scaling 2 sextupol families
+%ATFITTUNE Fit chromaticites by scaling 2 sextupole families
 % NEWRING = ATFITCHROM(RING,NEWCHROM,SEXTFAMILY1,SEXTFAMILY2)
+% NEWRING = ATFITCHROM(RING,DPP,NEWCHROM,SEXTFAMILY1,SEXTFAMILY2)
 %
 %RING:          Cell array
+%DPP:           Optional momentum deviation (default 0)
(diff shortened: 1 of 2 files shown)


atmat/lattice/atfitchrom.m
```diff
@@ -13,19 +15,16 @@
 %   Numeric array: list of selected elements in RING
 %   Cell array: All elements selected by each cell
```
simoneliuzzo, line 17: may also the optional inputs could be in the help

atmat/lattice/atfittune.m
```diff
@@ -22,41 +22,39 @@
...
 end
 
 idx1=varelem(ring,famname1);
 idx2=varelem(ring,famname2);
 
 kl1=atgetfieldvalues(ring(idx1),'PolynomB',{2});
 kl2=atgetfieldvalues(ring(idx2),'PolynomB',{2});
-delta = 1e-6;
-if ~UseIntegerPart
+delta = 1e-6;   % Step on quadrupole strengths
```
simoneliuzzo, line 38: This could be an optional input defaulted to this value?

atmat/lattice/atfitchrom.m
```diff
@@ -13,19 +15,16 @@
...
+[dpp,varargin]=getargs(varargin,0.0,'check',@(arg) isscalar(arg) && isnumeric(arg));
+[newchrom,famname1,famname2]=deal(varargin{:});
+
 idx1=varelem(ring,famname1);
 idx2=varelem(ring,famname2);
 kl1=atgetfieldvalues(ring(idx1),'PolynomB',{3});
 kl2=atgetfieldvalues(ring(idx2),'PolynomB',{3});
-if true
+%if true
     deltaS = 1e-5; % step size in Sextupole strngth
```
simoneliuzzo, line 28: Also this deltaS could be treated as the DPStep? set as optional input and defaulted to the present value

simoneliuzzo (approved):
I highly appreciate this update

atmat/lattice/atfittune.m
lfarv: Done

atmat/lattice/atfitchrom.m
lfarv: Done

atmat/lattice/atfitchrom.m
lfarv: Done

lfarv:
Excellent suggestions. All applied!

carmignani (approved):

PR #330, opened 2021-11-19
pypassmethods examples and tests adpated to new rin format

### LINKED ISSUE #328: rin format changed in python passmethods
@lfarv in PR #321 I think you have changed the format of the input of the python passmethods going form `(6*n,)` to `(6,n)`  by using `rin` instead of drin. The concerned source in `at.c` is:

```
            if (*pyintegrator) {
                #old implementation
                if (!*kwargs) *kwargs = Buildkwargs(*element);
                *kwargs = pyIntegratorPass(drin, *pyintegrator, *kwargs, num_particles);
                if (!*kwargs) return print_error(elem_index, rout);       /* trackFunction failed */
                #new implementation
                PyObject *res = PyObject_CallFunctionObjArgs(*pyintegrator, rin, *element, NULL);
                if (!res) return print_error(elem_index, rout);       /* trackFunction failed */
                Py_DECREF(res);
```

A user complained his code is broken... however, my feeling is that the present implementation is more inline with the python convention for particles that is `(6,n)`. I tempted to keep this new format and ask him to modify his code accordingly.
We will need to update the example passmethods and tests that should include several particles in order to avoid this kind of issue to re-appear.

What do you think?

swhite2401:
Fixes #328

### DIFF
--- a/pyat/at/integrators/pyDriftPass.py
+++ b/pyat/at/integrators/pyDriftPass.py
@@ -13,8 +13,8 @@ The element creation is as follows:
 drift = elements.Drift('drift', 1.0, PassMethod='pyDriftPass')
 
 To be noted:
-The particles are described by a one dimensional vector
-6*num_particles in the C tracking engine.
+The particles are described by a 2 dimensional vector
+(6,num_particles)
 In case a python passmethod and a c passmethod with the same name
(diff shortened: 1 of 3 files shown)


lfarv (approved):
OK

PR #334, opened 2021-11-29
RF cavity tuning: introduce fundamental frequency

swhite2401:
`cavity_access.py` is modified to give more flexibility, 3 modes are introduced:
-`RFMode.UNIQUE`
present implementation

-`RFMode.ALL`
returns array of all cavities

-`RFMode.FUNDAMENTA`L (proposed new default)
uses the lowest frequency cavities as reference, the getters returns attributes of the fundamental cavities. The setters affect all cavities maintaining the following relations between the fundamental and harmonics.
-`frequency`: keep constant ratio
-`voltage`: keep constant ratio (apply scale factor based on variation of total voltage)
-`timelag`: keep constant offset

This provides a a realistic and flexible way of tuning cavities and is fully compatible with complex systems involving multiple (accelerating) RF systems and solves some of the issues of the present implementation that was allowing to tune only one harmonic at the time. `cavpts` can still be used to exclude bunch lengthening cavities. 

`revolution.py` is improved to allow to set the rf frequency based on `orbit6` search: the rf frequency can be set to cancel the average orbit as is generally done in the control room

### DIFF
--- a/docs/p/CavityControl.md
+++ b/docs/p/CavityControl.md
@@ -2,73 +2,106 @@
 title: Controlling the RF cavities
 ---
 A lattice may contain multiple RF cavities, grouped according to different RF systems:
-main cavities, harmonic cavities…
-<br />
+main cavities, harmonic cavities…\
 AT provides simple tools to tune them, with methods and properties of the
 `Lattice` object.
 
 ### Lattice methods for cavity control
 
-The values handled by these methods concern the full ring (`periodicity` $$\times$$ superperiod).
+All methods have `array` and `cavpts` keyword arguments, used to select the cavities concerned by the command.
 
-All methods have a `cavpts` argument, used to select the cavities concerned by the command.
+- if `array is True`, the output attribute value is an array as long as the number of selected cavities,
+  and the input argument must be broadcastable to the number of cavities,
+- if `array is False` (default), the input and output are scalars. The scalar value
+applies to the set of cavities with the lowest frequency within the selection.
+The other cavities are ignored in `get_*` methods. For `set_*` methods, the other
+cavities are scaled as explained if the specific method description.
+
 The `cavpts` argument is used as follows:
-- `cavpts` is a "refpts" type (integer, integer or boolean array, callable): it is used to select the cavities,
-- `cavpts` is `None` (default value), and the `Lattice` object has a `cavpts` attribute: the lattice attribute is used to select the cavities,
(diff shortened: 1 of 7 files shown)


lfarv:
Hi @swhite2401 . I'll be back home and have more time to look at that next week, but I have a first comment on `revolution.py`:

For a ring where the energy varies along the circumference, the definition of the machine energy, and consequently of the nominal frequency is somewhat arbitrary: is the machine energy (dp/p = 0) the average one, the minimum or the maximum one ? the nominal frequency will be different…

At the moment, I see 2 options:
- the AT definition: the nominal frequency corresponds to the nominal circumference of the ring: the energy deviation is equally distributed between positive and negative values, the nominal machine energy is the average along the ring.
- the "control room" definition: the average horizontal closed orbit is zero. For a perfect machine, it is very similar to the AT definition, you can say that the energy deviation **as seen by the BPMs** is equally distributed between positive and negative values.S, it depends a bit on the location of the BPMs and probably on the error distribution.

Here you are proposing a 3rd definition: the machine energy (dp/p = 0) is the energy at the 1st point of the lattice. This definition as a drawback that makes it unacceptable: the nominal frequency depends on the choice of the starting point of the lattice ! That cannot be, the nominal frequency must be independent of the choice of the starting point. We could accept any other definition satisfying this requirement, but not the one you propose…

Note that the AT definition also ensures that the nominal frequency in independent of the location of the cavities: practically, the "nominal" machine energy (not the local one) does not change if you change to voltage distribution between cavities along the ring.

swhite2401:
Ah right, thanks @lfarv. In fact I don't really care about the notion of nominal frequency, for me it is defined as the corresponding to the nominal circumference but does not mean much in realistic simulations with errors etc.... This is why it can be useful for some application where the path length does not match the circumference to adjust the RF frequency. I agree this one is not good, I'll propose something else soon.

swhite2401:
Late night post-shift thinking: I did have this in my scripts to cancel initial offset for impedance simulations and generate a beam at dp,ct=0,0. Clearly not appropriated here, I will remove this from this branch, the most important part is all the rest in any case.

lfarv:
Ok, perfect. Give me some time to look at that, not before next week…

swhite2401:
Sure no rush. I will implemented the average orbit cancellation that could be of some interest to simulate control room conditions

lfarv:
New version with a single mode of operation (always "fundamental mode").

Issue #338, opened 2021-12-16
Relativistic  particles


swhite2401:
Presently only relativistic particle are allowed with `rest_energy=0`
This causes a problem in all calculations using `ring.gamma` which is coming out as `inf`

For example `ring_parameters.radiation_parameters` returns many `nans` and `inf` which is
clearly an issue.
We probably have to find a better way to handle this, I would suggest to move this to `ring.beta`
that could have a boolean keyword ultra-relativistic for example, default particle could then be electrons

lfarv:
Initially, all calculations assumed electrons, except tracking which assumes &beta;=1. The present version brings consistence between tracking and processing. But as long as the non-relativistic tracking is not validated, AT is consequently blocked on "relativistic particle".
To solve the problem, I propose to switch back to a default of "electron" (temporarily non-modifiable), which will turn all the outputs back to what they were before, but to force the tracking to &beta;=1. This way:
- Everything will be back to what it was before,
- Using 1 instead of the electron &beta; should not make any visible difference.

swhite2401:
Ok for me

swhite2401:
However @simoneliuzzo expressed some interest in modeling ultra-relativitic hadrons, to allow this we just have to set beta=1 regardless of the particle mass.
We may issue a warning is the "real" beta differs too much from 1.

lfarv:
Corrected in #339

PR #339, opened 2021-12-17
Set the default particle to "electron"

lfarv:
SInce its origin, AT has been using a value of &gamma; for electrons but a &beta; strictly equal to 1.
With the introduction of the particle object, the default (and only allowed) particle was 'relativistic', so that &beta;=1 was in agreement with the tracking. But this induced problems with the computation of some analytic parameters, because of infinite &gamma;. We now switch the default particle to 'electron', so that the analytic computations are correct, and identical to the historical values. The use of the real &beta; instead of 1 is still not applied everywhere, but the difference is negligible.

&beta;=1 is still assumed for the computation of the revolution frequency and 6D closed orbit, as it was up to now. It is necessary to be compatible with the tracking.


swhite2401 (approved):
OK!

lfarv:
Fixes #338

PR #341, opened 2021-12-23
Fix problems reading .mat files

### LINKED ISSUE #340: File save/load not possible
Several issues experienced with saving and loading files:

-Sometimes the first element (ringparam) is initialized with FamName='', this prevents the lattice from loading
properly. This is consecutive to a save_mat()
-When the Particle attribute of this first element is not a particle object and does not have the proper attributes
loading also fails

I add @simoneliuzzo in this thread because he found the problem generating a lattice from scratch (I think?), he can provide more information if useful

lfarv:
This fixes #340.
- Workaround a bug (?) in scipy.io.loadmat where empty strings from .mat files disappear,
- Fix problems with particle objects in .mat, .m and .repr files

### DIFF
--- a/atmat/lattice/atparticle.m
+++ b/atmat/lattice/atparticle.m
@@ -1,11 +1,12 @@
 classdef atparticle
     %ATPARTICLE Particle definition for AT
+    %
     %   A particle is defined by its name, rest energy and charge
 
     properties(SetAccess=private)
-        name
-        rest_energy
-        charge
+        name        % Name of the particle
+        rest_energy % Particle rest energy [eV]
+        charge      % Particle charge [elementary charge]
     end
 
     properties(Constant, Hidden)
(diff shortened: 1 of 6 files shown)


simoneliuzzo:
save/load of a mat file from python does not give errors

simoneliuzzo (approved):

simoneliuzzo:
sorry I have to take it back. I tried an other file.

a file saved in python has the RF cavities with DriftPass, and gives this error:

AttributeError: Error in element 136: PassMethod DriftPass is not compatible with Class RFCavity.
{'FamName': 'RFC', 'Length': 1e-12, 'PassMethod': 'DriftPass', 'Voltage': 9000000.0, 'Frequency': 352371625.9975046, 'HarmNumber': 979.0, 'Energy': 6000000000.0, 'TimeLag': 0.0}

lfarv:
@simoneliuzzo : this is a different problem. The "coherence" test when reading a file is too strict. We'll change that: I agree that DriftPass may be use for any class. For the moment, you can skip the test by adding "check=False" to the arguments of the load function (`load_lattice` or `load_mat`).

lfarv:
@simoneliuzzo: side remark: you should NEVER use lengths like 1.0E-12.
- for cavities, it's not dramatic: it just adds a half-long drift (0.5E-12) on each side of the the (thin) cavity,
- for multipoles it's much worse: apart from being about 20 times slower than a thin multipole (slices, drifts and kicks) thus increasing the numerical errors, it may be wrong because some computations like sin(x)/x may be inaccurate for small arguments.

lfarv:
The coherence test is relaxed so that 'DriftPass' is allowed for any `Element` class.

simoneliuzzo:
it works for me!

PR #342, opened 2021-12-28
Fix problems with transfer line mode

lfarv:
The transfer-line mode of `linopt*` has some problems:
- in Matlab, the initial orbit is not correctly taken into account (pointed out by @simoneliuzzo)
- in 6D, the initial dispersion is not correctly reproduces (both Matlab and python)

Both are fixed in this pull request.

### DIFF
--- a/atmat/atphysics/LinearOptics/atlinopt.m
+++ b/atmat/atphysics/LinearOptics/atlinopt.m
@@ -1,5 +1,11 @@
 function [lindata, varargout] = atlinopt(RING,dp,varargin)
-%ATLINOPT Performs linear analysis of the COUPLED lattices
+%ATLINOPT Performs 4D linear analysis of the COUPLED lattices
+%
+%ATLINOPT only works for CONSTANT energy. So no PassMethod should:
+%   1. change the longitudinal momentum dP (cavities, magnets with radiation)
+%   2. have any time dependence (localized impedance, fast kickers etc)
+%
+%For a more general computation, see ATLINOPT6
 %
 % LinData = ATLINOPT(RING,DP,REFPTS) is a MATLAB structure array with fields
 %
--- a/atmat/atphysics/LinearOptics/atlinopt6.m
+++ b/atmat/atphysics/LinearOptics/atlinopt6.m
@@ -86,14 +86,18 @@ if isempty(twiss_in)        % Circular machine
     is6d=check_radiation(ring);
     [orbs,orbitin]=findorbit(ring,refpts,'dp',dp,'orbit',orbitin,varargs{:});
     [vps,ms,mu,ri,ai]=build_1turn_map(ring,dp,refpts,orbitin);
+    o1P=[];
+    o1M=[];
 else                        % Transfer line
-    if isempty(orbitin), orbitin=zeros(6,1); end
+    [orbitin,sigma,d0]=build_sigma(twiss_in,orbitin);
     orbs=linepass(ring,orbitin,refpts);
-    sigma=build_sigma(twiss_in);
+    dorbit=0.5*DPStep*[d0;1;0];
     nv=size(sigma,1);
     dms=nv/2;
     is6d=(dms>=3);
     [vps,ms,mu,ri,ai]=build_1turn_map(ring,dp,refpts,orbitin,'mxx',sigma*jmat(dms));
+    o1P=orbitin+dorbit;
+    o1M=orbitin-dorbit;
 end
 
 tunes=mod(angle(vps)/2/pi,1);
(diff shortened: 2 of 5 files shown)


simoneliuzzo:
Dear @lfarv and @swhite2401,

thank you for this fix! 

It works when using atlinopt6.

if I keep my script unmodified and use atlinopt, it gives this error.

Screenshot 2022-01-04 at 08 12 54

when I block the script at that line here is the size of the orbitin variable:

Screenshot 2022-01-04 at 08 08 58

in case this is an intended feature, then I would suggest modifying the help of atlinopt as well, for example triggering an error when computing transfer line optics, saying, please use atlinopt6, and giving ad example of "conversion"
use:
[a,twi] = atlinopt6(THERING,ind);
instead of:
THERING =atradoff(THERING);
twi = atlinopt(THERING,0,ind);

I also noted a 10* in the code for the vertical emittance (line 260 of linear.py). This is may be dangerous? I think about my lattices with errors, would I get all time 10 times larger vertical emittance?

lfarv:
@simoneliuzzo : You are still running `atlinopt` with radiation on ! Its output is 4D, and you enter it a 6D processing, so the dimensions are wrong. The test preventing the use of atlinopt with radiation ON is ready since long, just waiting for you ! The foreseen error message will mention atlinopt6 as a possible workaround.

simoneliuzzo:
Hi Laurent,

I tried to add an atradoff line, but the message is still the same, dimensions not ok. 

atplot transferline mode is working only with 6D-radiation-on-atlinopt6 optics? What about transferlines?

where can I try the test that you mention? There is a branch for it probably.

Indeed this radiation-atlinopt-atlinopt6 combos are rather hard to get to my simple user mind, I never get the correct combination! 
If I get it correctly, the message is: never use atlinopt again, use atlinopt6. May be we could add a deprecation message to annoy lazy/messy users as me?

thank you!

best regards
Simone

lfarv:
@simoneliuzzo: Concerning your last sentence, you are basically right: it's usually better now to use `atlinopt6`. `atlinopt` may still be used if you are interested in the Sagan/Rubin formalism for H/V coupling. In this respect it is not "deprecated", but I can add a sentence in the help recommending `atlinopt6` for general purpose.

But keep in mind that dispersion and &beta;-functions are 2D notions. Even if `atlinopt6` can deal with radiation ON by kind of "annihilating" the longitudinal motion to extract them, you get the same results (except closed orbit effects) faster by switching radiation OFF. `atlinopt6` with rad. OFF is slightly faster than `atlinopt`, but with rad. ON it is significantly slower.

Concerning transfer lines:
- The `twiss_in` data has to be generated with the right "dimension": 4x4 without radiation, 6x6 with radiation. So you must use the same mode to produce the data and to enter it at the entrance of the transfer line (even with `atlinopt6`).
- There may still be problems using `atlinopt` data in `atlinopt6` input or vice-versa, because of the different formalism. The point you mention comes from different dimensions of the closed orbit (4x1 or 6x1). I'll solve this one, and try to get simple cases (no H/V coupling) work. But it will never be perfect.

simoneliuzzo:
Dear Laurent, 
since this pull request acts on atplot, (somehow) I am facing an other small problem. The field ElemIndex in the lindata structure is missing. 
Calls such as atplot(ring,@plotWdispP) and atplot(ring,@plB0curlyh) are failing with this error: Unrecognized field name "ElemIndex".
thank you for your help!
ciao
Simone

lfarv:
> Calls such as atplot(ring,@plotWdispP) and atplot(ring,@plB0curlyh) are failing with this error: Unrecognized field name "ElemIndex".

I do not get this error. Am I missing something?

lfarv:
Now, with radiation OFF, the `test_da_elettra` test runs with either `atlinopt` or `atlinopt6`.

simoneliuzzo:
> > Calls such as atplot(ring,@plotWdispP) and atplot(ring,@plB0curlyh) are failing with this error: Unrecognized field name "ElemIndex".
> 
> I do not get this error. Am I missing something?

I do not know.
here a picture
Screenshot 2022-01-04 at 18 11 12

lfarv:
You must be in a wrong branch: neither `master` nor `fix_twiss_in` have this line !

simoneliuzzo:
> You must be in a wrong branch: neither `master` nor `fix_twiss_in` have this line !

You are right Laurent! the branch is correct, actually the whole path is different. This is an ESRF based issue (computing cluster home not equal to the standard site user home). sorry for spamming. 


This branch works also for me, both with atlinopt and atlinopt6.

simoneliuzzo (approved):

simoneliuzzo:
I have not looked at the python version

Issue #315, opened 2021-10-05
pyat: pip install . does not work


willrogers:
As reported by @jbengtsson. This is something to do with the file copying process not working.

This is listed in the documentation as a way to install pyat.

```
$ pip3 install .
Processing /tmp/atbuild/at/pyat
  Installing build dependencies ... done
Requirement already satisfied: numpy>=1.16.6 in /tmp/atbuild/venv/lib/python3.7/site-packages (from accelerator-toolbox==0.2.1) (1.21.2)
Requirement already satisfied: scipy>=0.16 in /tmp/atbuild/venv/lib/python3.7/site-packages (from accelerator-toolbox==0.2.1) (1.7.1)
Installing collected packages: accelerator-toolbox
  Found existing installation: accelerator-toolbox 0.2.1
    Uninstalling accelerator-toolbox-0.2.1:
      Successfully uninstalled accelerator-toolbox-0.2.1
  Running setup.py install for accelerator-toolbox ... error
    Complete output from command /tmp/atbuild/venv/bin/dls-python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-req-build-2b9wngi0/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-record-xkbnkd43/install-record.txt --single-version-externally-managed --compile --install-headers /tmp/atbuild/venv/include/site/python3.7/accelerator-toolbox:
    running install
    running build
    running build_py
    creating build
    creating build/lib.linux-x86_64-3.7
    creating build/lib.linux-x86_64-3.7/at
    copying at/__init__.py -> build/lib.linux-x86_64-3.7/at
    creating build/lib.linux-x86_64-3.7/at/lattice
    copying at/lattice/lattice_object.py -> build/lib.linux-x86_64-3.7/at/lattice
    copying at/lattice/cavity_access.py -> build/lib.linux-x86_64-3.7/at/lattice
    copying at/lattice/elements.py -> build/lib.linux-x86_64-3.7/at/lattice
    copying at/lattice/utils.py -> build/lib.linux-x86_64-3.7/at/lattice
    copying at/lattice/__init__.py -> build/lib.linux-x86_64-3.7/at/lattice
    copying at/lattice/options.py -> build/lib.linux-x86_64-3.7/at/lattice
    creating build/lib.linux-x86_64-3.7/at/tracking
[...]

PR #346, opened 2022-01-10
Python packaging

lfarv:
This PR is a partial move to the new python packaging standards:
- most of the packaging configuration is moved from `setup.py` to `setup.cfg`,
- `setup.py` only keeps the configuration of the compilation of C and C++ extensions and the copy of C sources, which have to be dynamic,
- more configuration of the packaging process is moved to `pyproject.toml`,
- the up-to-date build tools `python -m pip` and `python -m build` can now be used instead of `python setup.py`.

Versioning is handled fully automatically by `setuptools_scm`. The AT version is determined on each build by comparison to the last git tag of the form "pyat-x.y.z".

The version number of an 'official' release is the "x.y.x" string defined by the tag which triggered the build. The version number of intermediate builds is in the form "x.y.t.dev000+g11111" where:
- x.y.t is the value of the last git tag incremented by 1 on the last component,
- 000 is the number of commits since the tag,
- 111 is the commit identifier.

Git tags not in the form "pyat-x.y.z" are ignored by the versioning.

The version number is accessible at run time in` at.__version__`.

### DIFF
--- a/.github/workflows/build-python-wheels.yml
+++ b/.github/workflows/build-python-wheels.yml
@@ -6,6 +6,7 @@ on:
   push:
     tags:
       - pyat-*
+  workflow_dispatch:
 
 defaults:
   run:
@@ -20,40 +21,51 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-
-      - uses: actions/setup-python@v2
-        name: Install Python
         with:
-          python-version: '3.7'
+          # Necessary to fetch tags and allow setuptools_scm
+          # see: https://github.com/pypa/setuptools_scm/issues/480
+          fetch-depth: 0
 
-      - name: Install cibuildwheel
-        run: python -m pip install cibuildwheel
+      - name: Install Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
 
-      - name: Prepare source
-        run: |
-          pip install -e .
-
-      - name: Remove built files
-        if: (! contains(matrix.os, 'windows'))
-        run: rm -rf build
+      - name: Set build configuration
+        run: cp githubproject.toml pyproject.toml
 
       - name: Build wheels
-        run: python -m cibuildwheel --output-dir wheelhouse
-        env:
-          # We don't yet support Python 3.10.
-          # Pypy does not have Scipy so we cannot support it.
-          CIBW_SKIP: cp310* pp*
-          CIBW_BUILD_VERBOSITY: 1
-
-      - name: Build sdist
-        run: python setup.py sdist
+        uses: pypa/cibuildwheel@v2.3.1
+        with:
+          package-dir: pyat
 
       - name: Upload wheels
         uses: actions/upload-artifact@v2
         with:
           name: wheels
-          path: ./pyat/wheelhouse/*.whl
+          path: ./wheelhouse/*.whl
           if-no-files-found: error
+
+  build_sdist:
+    runs-on: ubuntu-18.04
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          # Necessary to fetch tags and allow setuptools_scm
+          # see: https://github.com/pypa/setuptools_scm/issues/480
+          fetch-depth: 0
+
+      - name: Install Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+
+      - name: Install build tools
+        run: python -m pip install build
+
+      - name: Build sdist
+        run: python -m build --sdist
 
       - name: Upload sdist
         uses: actions/upload-artifact@v2
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -35,7 +35,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install flake8
-        python -m pip install -r requirements.txt
+        python -m pip install pytest
+        python -m pip install pytest-lazy-fixture
         python -m pip install pytest-cov
 
     - name: Build and install at
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -32,15 +32,10 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install flake8
-        python -m pip install pytest
-        python -m pip install pytest-lazy-fixture
-        python -m pip install pytest-cov
+      run: python -m pip install --upgrade pip
 
-    - name: Build and install at
-      run: python -m pip install -e .
+    - name: Build and install at with tests
+      run: python -m pip install -e ".[dev]"
 
     - name: Lint with flake8
       run: |
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['3.6', '3.7', '3.8', '3.9']
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
         os: [macos-latest, ubuntu-latest, windows-latest]
 
     steps:
@@ -31,11 +31,8 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install dependencies
-      run: python -m pip install --upgrade pip
-
     - name: Build and install at with tests
-      run: python -m pip install -e ".[dev]"
+      run: python -m pip install ".[dev]"
 
     - name: Lint with flake8
       run: |
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -32,7 +32,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Build and install at with tests
-      run: python -m pip install ".[dev]"
+      run: python -m pip install -e ".[dev]"
 
     - name: Lint with flake8
       run: |
--- a/pyat/developers.rst
+++ b/pyat/developers.rst
@@ -19,6 +19,12 @@ Install the Python development package for your OS. For example, using yum:
 
 * ``sudo yum install python3-devel``
 
+Source download
+---------------
+Download the latest version of AT:
+    ``$ git clone https://github.com/atcollab/at.git``
+
+
 Installation (all platforms)
 ----------------------------
 
@@ -30,9 +36,18 @@ It is easiest to do this using a virtualenv. inside pyat:
 
 Then:
 
-* ``source venv/bin/activate  # or venv\Scripts\activate on Windows``
-* ``pip install -r requirements.txt``
-* ``pip install -e .``
+* activate the virtual environment:
+
+  ``source venv/bin/activate  # or venv\Scripts\activate on Windows``
+* make sure you have a recent pip installer:
+
+  ``pip install --upgrade pip``
+* Go to the pyat root directory:
+
+  ``cd <atroot>/pyat``
+* install AT:
+
+  ``pip install -e .``
 
 Finally, you should be able to run the tests:
 
@@ -58,17 +73,17 @@ recompiled.  To force recompilation, remove the build directory:
 Any changes to .py files are automatically reinstalled in the build, but to
 ensure any changes to .c files are reinstalled rerun:
 
-* ``python setup.py develop``
+* ``pip install -e .``
 
-If you get strange behaviour even after running setup.py develop again, then
+If you get strange behaviour even after running pip install develop again, then
 running the following, inside pyat, should fix it:
 
 * ``rm -rf build``
 * ``find at -name "*.pyc" -exec rm '{}' \;``
 * ``find at -name "*.so" -exec rm '{}' \;``
-* ``python setup.py develop``
+* ``pip install -e .``
 
-N.B. setup.py develop needs to be run with the same version of Python (and
+N.B. ``pip install -e .`` needs to be run with the same version of Python (and
 numpy) that you are using to run pyAT.
 
 Releasing a version to PyPI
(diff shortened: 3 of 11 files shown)


MJGaughran:
This reminds me of #315 which hasn't yet been done, but is relatively straightforward. We need to fix that here as well in order for our automated release process to work.

Is it okay if I submit a pull request for #315 and then we can work on the changes here?

lfarv:
@MJGaughran : I think that this PR solves all the problems mentioned in #315. `pip install .` and `pip wheel .` work, as does `python -m build`, which is now the preferred method for generating source and wheel distributions. I removed all the references to python `setup.py install | develop` from the documentation, and kept only `pip install`.

MJGaughran:
Okay, yes, let's just use this pull request. I'll go through it in more detail in a bit.

I think we still need to revert the changes in #306 to the GitHub workflows, no? The version of numpy needs to be pinned at the earliest version for compatibility reasons.

lfarv:
@MJGaughran: by the way, it seems that I do not have the maintainer status on PyPI. Can you set this ?

MJGaughran:
> @MJGaughran: by the way, it seems that I do not have the maintainer status on PyPI. Can you set this ?

I've just made the invite.

lfarv:
@MJGaughran : Ok for PiPI, thanks. Do you have any comment on the pull request?

lfarv:
@MJGaughran: I do not agree:
The NumPy version declared in `pyproject.toml` is the one used for **building**: the one necessary to compile of the C extensions, used by the `import numpy` in **setup.py**. It is used either on GitHub for building the wheel, or at the users's when installing from source.
The version declared in **setup.cfg** is the one used for **installation**: it is the one automatically downloaded by `pip install`, both from wheel and from source.

Both must agree, and the minimum required value is 1.16.6, for the moment.
But the installation should NEVER install an older version than the one previously installed by the user, if any. So we need the >= in both cases.

The present setup guaranties a safe installation for users, either from source or from wheel.

It does not check that pyat can run with 1.16.6. The GitHub tests use the most recent one. If this has to be checked, it requires a dedicated test, but that's another point.

But for packaging, this is the right definition.

lfarv:
Additional remark: as set here, the wheel files are compiled with the latest numpy release. That's a good point: the most recent, the better !

MJGaughran:
> Additional remark: as set here, the wheel files are compiled with the latest numpy release. That's a good point: the most recent, the better ! The version installed at the users' for running pyat may be older, any between 1.16.6 and the most recent one. This does not matter, the compiled C extension will work. That would not be true in the other way.

The most recent is definitely not better. Numpy code is guaranteed to be forwards compatible only.

MJGaughran:
There appears to be some confusion around the use of the terms 'forward' and 'backward' compatible. We use the definition as described here: https://numpy.org/devdocs/user/depending_on_numpy.html

lfarv:
> The most recent is definitely not better. Numpy code is guaranteed to be forwards compatible only.

See above: the ABI is forwards compatible, so better compiling with an old one. The NumPy package is backwards compatible: better running with the latest one.

MJGaughran:
I accept your correction on the first point (I can't even claim it made sense to me, it was just sloppy thinking as I haven't looked at how `at` is compiled), but some of your statements are directly contradictory:

> Additional remark: as set here, the wheel files are compiled with the latest numpy release. That's a good point: the most recent, the better ! The version installed at the users' for running pyat may be older, any between 1.16.6 and the most recent one. This does not matter, the compiled C extension will work. That would not be true in the other way.

> See above: the ABI is forwards compatible, so better compiling with an old one. The NumPy package is backwards compatible: better running with the latest one.

Does the above mean you have changed your mind and now agree that this pull request needs to be modified to pin `numpy == 1.16.6` as part of the wheel-building process? Because I can't see anything that does this. Specifying `numpy >= 1.16.6` in `pyproject.toml` breaks Will's (incorrect) method for doing this.

MJGaughran:
I see we might also just be getting confused when I say things like 'numpy code is forwards compatible'. I meant that the ABI is forwards compatible, and agree with pretty much everything you wrote in your second to last comment.

Your clarifying points are definitely welcome.

MJGaughran:
I've deleted my comments. I'm honestly getting confused by this thread and I don't want to make things worse. I agree with most of your clarifying points (and apologise for the sloppy thinking in my initial bullet point).

Do we now agree on the following:

We still need to pin the version of numpy used as part of building the Python wheels in GitHub Actions. This merge request breaks the existing mechanism for doing that. You will provide a commit that fixes this issue.

lfarv:
@MJGaughran: Exactly. And the way to do that is to add a line in `build-python-wheels.yml` installing the oldest NumPy version (but >= 1.16.6) compatible with each python version. So that the build itself will not need to download a more recent version.

This will keep the present requirements for a user build as they are: >= 1.16.6, because it's the only thing AT needs.

MJGaughran:
Okay, that sounds good to me.

But just to be clear, the reason Will specified `numpy == 1.16.6` was because 1.16.6 is the oldest version we found that was compatible all the way up to Python 3.9; this decision was not based on the actual requirements of AT / pyat as you are suggesting above. See the comments in #298 for example.

The fact that this causes issues with 3.10 is a good motivation to find a method that allows us to specify different versions of numpy for different versions of Python.

lfarv:
> But just to be clear, the reason Will specified `numpy == 1.16.6` was because 1.16.6 is the oldest version we found that was compatible all the way up to Python 3.9; this decision was not based on the actual requirements of AT / pyat as you are suggesting above. See the comments in #298 for example.

Yes, that's what I understood. I do not know what is the exact version providing everything needed by AT, and I do not want to change from 1.16.6.

But I think we can now separate the needs  of AT, specified in the "`requires:`" statements, and the fact that we want to provide compiled wheel files working in any users' configuration. This is better handled in `build-python-wheels.yml`. I'll stick to 1.16.6 for python <= 3.9, and look for the adequate version for 3.10.
Practically, we could reasonably choose the numpy version which was current at the time python was released: it's very unlikely someone would downgrade a fresh python install… But I'll try to go as early as possible. If it ever happens that the wheel file does not work, installation from the sdist is always possible.

lfarv:
Updated to pin the numpy version when building wheels. This is done by using a dedicated "pyproject.toml" when building wheels. The numpy version is set to 1.21.5 for python 3.10 ( I could not install anything older).

MJGaughran:
Thanks.

I have a couple of questions about the `githubproject.toml` file:

- Why do you have a separate option under `tool.cibuildwheel` for `cp310`? Is there no equivalent brackets formula that accepts multiple characters?
- Can you document the difference in git describe between that file and the default `pyproject.toml`?

lfarv:
> * Why do you have a separate option under `tool.cibuildwheel` for `cp310`? Is there no equivalent brackets formula that accepts multiple characters?

I'm not expert in regular expressions, it looked to me easier this way…

> * Can you explain the difference in `git describe` between that file and the default `pyproject.toml`?

When copying `githubproject.toml `to `pyproject.toml`, the working directory diverges from the committed version. With the standard command,` git describe` will add a "dirty" indicator, which translates into a "`.dyyyymmdd`" suffix appended to the version number. Removing the `--dirty` flag avoids that. HEAD is the default target (HEAD might be omitted, in fact, as in the initial command).

On the other hand, with the` pyproject.toml `version, if a user locally installs from a "dirty" working directory, he will get the suffix, indicating an installation from a non-commited working directory.

.github/workflows/python-tests.yml
```diff
@@ -35,7 +35,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install flake8
-        python -m pip install -r requirements.txt
```
MJGaughran, line 38: You've deleted `requirements.txt` which included `pytest` and other dev packages. While you've added them here, there is no obvious way for a user to find which packages are required to install for testing.

There may be a better way, but the following is my understanding of how you should add these to your setup.cfg:

```
[options.extras_require]
dev =
    pytest>=2.9
    pytest-lazy-fixture
```

It is probably sensible to add `flake8` etc to these dev packages as well.

MJGaughran:
> I'm not expert in regular expressions, it looked to me easier this way…

Did the following not work?
`["cp3{6,7,8,9}*", "cp310*"]` -> `["cp3{6,7,8,9,10}*"]`

[Their docs](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip) claims it uses bracex expansions; I don't see any reference to a regex:
```
>>> bracex.expand(r'cp3{6,7,8,9,10}')
['cp36', 'cp37', 'cp38', 'cp39', 'cp310']
```

And thanks for your explanation regarding `git describe`.

.github/workflows/python-tests.yml
MJGaughran: You can install the dev packages alongside the module with, for example:
`pip install -e .[dev]`

lfarv: Done. Nice improvement, `python-test.yml `is cleaner

.github/workflows/python-tests.yml
```diff
@@ -31,15 +31,8 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
```
MJGaughran, line 36: Did you remove the `pip` upgrade because we're now using Python 3.9 to build the wheels?

.github/workflows/build-python-wheels.yml
```diff
@@ -20,41 +21,52 @@ jobs:
...
 
       - name: Upload wheels
         uses: actions/upload-artifact@v2
         with:
           name: wheels
-          path: ./pyat/wheelhouse/*.whl
+          path: ./wheelhouse/*.whl
           if-no-files-found: error
 
+  build_sdist:
```
MJGaughran, line 49: What is required to fully test the GitHub Actions jobs?

Would you mind pushing the `pyat-test` tag to at least ensure that process runs correctly.

pyat/setup.py
```diff
@@ -81,40 +79,30 @@ def select_omp():
 if not sys.platform.startswith('win32'):
     cflags += ['-Wno-unused-function']
```
MJGaughran, line 80: Commenting on this bit so I can have a thread, but this is for line 31.

I find I need to use:
`cflags = ['--std=c99']`

Otherwise, I get an error about C99 `for` loops. This was not a problem in 0.2.1.

pyat/developers.rst
```diff
@@ -58,17 +73,17 @@ recompiled.  To force recompilation, remove the build directory:
...
 
-If you get strange behaviour even after running setup.py develop again, then
+If you get strange behaviour even after running pip install develop again, then
 running the following, inside pyat, should fix it:
 
 * ``rm -rf build``
 * ``find at -name "*.pyc" -exec rm '{}' \;``
 * ``find at -name "*.so" -exec rm '{}' \;``
-* ``python setup.py develop``
+* ``pip install -e .``
```
MJGaughran, line 84: Is it possible to explain why I'm only able to install this module in editable mode during development?

It's not an obvious constraint. I'm fairly certain it was explained to me before, but I have already forgotten.

It's clearly attempting to compile the code (that's why I get a compiler error unless I add the `--std=c99` compiler flag), yet it still fails to properly install the module.

.github/workflows/python-tests.yml
lfarv: cibuildwheel does its own upgrade of build tools (it even installs another python on MacOS).

MJGaughran: Okay, great.

.github/workflows/build-python-wheels.yml
lfarv: I tested the wheels by running the action "on push". Wheels look correct, though I could not test on Windows.

pyat/developers.rst
lfarv: Installing in editable mode allows changing python code and just restarting python without needing to run pip again. Otherwise you need to run `pip install` again after each modification. Of course, when modifying C code, or at 1st installation, you have to compile again.

MJGaughran: At least for me, it simply doesn't work outside of editable mode.

There's an error about missing modules, which I believe are compiled from C.

MJGaughran: `at.tracking.atpass` is missing, for example

pyat/setup.py
lfarv: Very strange… I does not happen in GitHub actions. Which platform are you using ?

.github/workflows/build-python-wheels.yml
MJGaughran: How did you run the action 'on push'?

I've looked at the jobs that were run 2 days ago, and none of them have produced any artifacts. I'm looking for something like this: https://github.com/atcollab/at/actions/runs/1282945133

It appears the last time this was run for the `python_packaging` branch was 12 days ago.

pyat/developers.rst
lfarv: Ah, it's a different problem. Outside of editable mode, you have to work from a working directory different from pyat. The "current directory" is in your python path before "site-packages", so python tries to import from the local "at" instead of the standard "site-packages", and does not find the compiled modules. Try cd somewhere else, you'll see the installation is OK.

.github/workflows/build-python-wheels.yml
lfarv: I just created a temporary "python_pack" branch, changed the "on:" sequence in the .yml file and pushed. Then I downloaded the artifacts and tested them on linux and mac. Apparently all this disappeared when I deleted the branch.

pyat/setup.py
MJGaughran: The following fails on RHEL7:
`gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC)`

The following works on RHEL8:
`gcc version 8.5.0 20210514 (Red Hat 8.5.0-4) (GCC) `

As far as I can tell, it is a valid complaint:
```
    integrator-src/WakeFieldPass.c:132:9: error: ‘for’ loop initial declarations are only allowed in C99 mode
             for(int i=0;i<3;i++){
             ^
    integrator-src/WakeFieldPass.c:132:9: note: use option -std=c99 or -std=gnu99 to compile your code
```

pyat/developers.rst
MJGaughran: Okay, good point

.github/workflows/build-python-wheels.yml
MJGaughran: Okay, that sounds reasonable.

I think the `pyat-test` tag was made for this purpose, and is probably easier to mess around with.

pyat/setup.py
lfarv: I see. WakeFieldPass.c was recently introduced, so you had no problem before. Apparently, all recent compilers are by default compliant with c99, only your very old one is not (gcc 4 !!). Once it's understood, I have no objection to add -std=c99.

.github/workflows/build-python-wheels.yml
lfarv: Pushing the tag would have worked only once the branch was merged into master. And when it is, it will now be easier to trigger it from the GitHub interface, no need for a fake tag any more.

lfarv:
Added a "-std=c99" flag for C compilation. As this is rejected for C++, it required separating C and C++ extensions.

MJGaughran:
Okay, I think you've fixed everything now

PR #348, opened 2022-01-11
Cavity phase offset arg

swhite2401:
This branch adds an optional phase offset argument to the cavity passmethods.
This will facilitate the upcoming beam loading development by giving an independent input to set the tuning angle without having to overwrite Timelag that is in principle fixed.

No impact on the standard use of these passmethods.

### DIFF
--- a/atintegrators/CavityPass.c
+++ b/atintegrators/CavityPass.c
@@ -18,10 +18,10 @@ struct elem
     double Frequency;
     /* optional fields */
     double TimeLag;
-    double Phi0;
+    double PhaseLag;
 };
 
-void CavityPass(double *r_in, double le, double nv, double freq, double lag, double phi0, int num_particles)
+void CavityPass(double *r_in, double le, double nv, double freq, double lag, double philag, int num_particles)
 /* le - physical length
  * nv - peak voltage (V) normalized to the design enegy (eV)
  * r is a 6-by-N matrix of initial conditions reshaped into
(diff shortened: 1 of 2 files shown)


lfarv:
@swhite2401: I think we discussed already that. In fact, you are just adding both attributes.
- should it be called PhaseLag ?
- Why is the sign convention opposite to TimeLag: &omega;(t - TimeLag) + PhaseLag ?
  Maybe it more logical like that, I do not know…

swhite2401:
@lfarv  yes we did and yes it is just more convenient like this, no physics behind.

- Good idea, PhaseLag is more consistent with TimeLag
- Sign is arbitrary, but right again let's remain consistent with timelag

I make the changes

swhite2401:
Done, ready to merge

lfarv:
Sorry, I was not asking to change the sign, just pointing the fact that it must be carefully chosen;
- \+ is more intuitive because of the usual expression sin(&omega;t + &phi;),
- \- is consistent with TimeLag, and possible with the term "lag" in PhaseLag

So make you choice, but document it, please…

swhite2401:
minus sign is fine, where is a good place to document passmethods? I saw some things in element.py where timelag is defined...however in atrfcavity TimeLag is not even defined in the help so I am not sure , any suggestions?

I can see CavityPass is still the default btw, when are we making the switch?

PR #350, opened 2022-01-12
atenergy computes the harmonic number from the RF frequency

lfarv:
`atenergy` does not rely any more on the `HarmNumber` field of cavity elements. It computes the harmonic number from the frequency of the 1st RF cavity.

### DIFF
--- a/atmat/atphysics/Radiation/atenergy.m
+++ b/atmat/atphysics/Radiation/atenergy.m
@@ -20,7 +20,10 @@ function [energy,nbper,voltage,harmnumber,U0]=atenergy(ring)
 %
 %  NOTES
 %    1. voltage and harmnumber assume a single RF system: fundamental
-%       frequency, no harmonic system.
+%       frequency, no harmonic system. More prcisely:
+%
+%       voltage is the sum of the voltages of all cavities,
+%       harmnuber is computed with the frequency of the 1st cavity in the ring
 %
 %  EXAMPLES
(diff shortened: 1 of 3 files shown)


carmignani:
Hello, I think the harmonic number is stored in the ringParam element, if that exists. So maybe atenergy should first check there, and after it can computes from the frequency of the first cavity if there is no ringParam element.

lfarv:
@carmignani : excellent remark. I'll do that.

simoneliuzzo:
may also the radiation state of the ring could be added to atGetRingParam ?

simoneliuzzo:
and may be also the dp/p and RF shift of the the ring?

lfarv:
@simoneliuzzo: in principle, I agree.But practically, "RingProperties" is an image of what is stored in the RingParam element. It is difficult to store there information which may be modified without going through `atSetRingProperties`: the stored information would the become wrong. In your examples, one could imagine that `atradon/off` would call `atSetRingProperties` to update the information. For dp/p or RF frequency it's more difficult…
Let's keep this for future.

lfarv:
The upgrade is now more ambitious: `atenergy` now calls `atGetRingProperties`. This eliminates duplicated computations and has some benefits: for voltage and harmonic number, they are no more computed from the 1st cavity, but from the set of cavities with the lowest frequency (considered as "fundamental", even if it's not always true). This choice can be modified by setting the optional "`cavpts`" ring property.

simoneliuzzo (approved):

PR #351, opened 2022-01-18
Lifetime and DA calculations for pyAT

swhite2401:
This branch allows the compute the dynamic aperture and lifetime of a pyAT lattice.
Below a script with example usage.
To be noted: `use_mp=True` has crashed in the past with MacOS, to be tested again with the new `patpass` #347 

```
import at
from at.plot import plot_acceptance
import multiprocessing
import time
import numpy
from at.physics import chromaticity

path = '../lattice/'
filename = 'S28F.mat'
key = 'LOW_EMIT_RING_INJ'
latticef = path+filename
ring.radiation_off()

'''
a,b,c = chromaticity(ring, method='linopt', dpm=0.02, npoints=11, order=3)
plt.plot(b,c)
plt.show()
'''


planes = ['x','y']
npoints = [21,21]
amplitudes = [15e-3,0.006]

# example computing 1D acceptance using multi-processing
# *vertical* and *momentum* exist, first argument is desired precision
#default with is Grid (or vector in 1D)
t0 = time.time()
b,s,g = ring.get_horizontal_acceptance(0.1e-3, amplitudes[0], use_mp=True)
print(time.time()-t0)

# example computing 1D acceptance using single processor
# *vertical* and *momentum* exist, first argument is desired precision
#default with SP is recursive search
t0 = time.time()
b2,s2,g2 = ring.get_horizontal_acceptance(0.1e-3, amplitudes[0], grid_mode=at.GridMode.RECURSIVE, use_mp=True)
print(time.time()-t0)

print(b, b2)


# example computing 1D acceptance using multi-processor
# default is radial grid
bf,sf,gf = ring.get_acceptance(planes, npoints, amplitudes, use_mp=True)

# same thing with a plot at the end, warning: plot is non blocking
ring.plot_acceptance(planes, npoints, amplitudes, use_mp=True)

#  Now example lifetime calculation for 1 cell 
emity = 10e-12
bunch_curr = 200e-3/992
zn = 0.7
ids = ring.get_refpts('ID*')
quads = ring.get_refpts(at.Quadrupole)
dips = ring.get_refpts(at.Dipole)
ncells = 1
refpts = numpy.sort(numpy.concatenate((ids[:ncells], 
                                       quads[quads<ids[ncells]],
                                       dips[dips<ids[ncells]])))
tlt, momap, refpts = ring.get_lifetime(emity, bunch_curr, zn = zn, refpts=refpts,
                                       use_mp=True, verbose=True)
print(tlt/3600)


# some figures
plt.figure()
plt.plot(*gf,'.')
plt.plot(*sf,'.')
plt.plot(*bf)

plt.figure()
plt.plot(momap[:,0])
plt.plot(momap[:,1])

plt.show()
```

TODO:
1-~~different radii distribution~~ Done
2-~~divider in recursive search (defaults to 2 now)~~ Done
3- ~~search for max range in 2D -> use as starting point~~ Tested: no performance gain
4-cluster submission (separate PR)


swhite2401:
I convert this to draft since their might be many further iterations before converging to the final version

lfarv:
@swhite2401: works nicely as expected, but there is a problem:

Instead of your awful refpts combination:
```python
ids = ring.get_refpts('ID*')
quads = ring.get_refpts(at.Quadrupole)
dips = ring.get_refpts(at.Dipole)
ncells = 1
refpts = numpy.sort(numpy.concatenate((ids[:ncells], 
                                       quads[quads<ids[ncells]],
                                       dips[dips<ids[ncells]])))
```
I'm using a simpler combination with boolean refpts:
```python
ncells = 1
ids = ring.get_cells(at.checkname('ID*'))
quads = ring.get_cells(at.checktype(at.Quadrupole))
dips = ring.get_cells(at.checktype(at.Dipole))
# Combine references
refpts= ids | quads | dips
# Set cells > ncells to False
cellstart = ring.uint32_refpts(ids)
refpts[cellstart[ncells]:] = False
```
Using booleans is usually more efficient, but `get_acceptance` crashes with boolean refpts.

At least at line 98 of `acceptance.py`:
```rp = numpy.atleast_1d(refpts)```
should be replaced by
``` rp = ring.uint32_refpts(refpts)```

But there may be other problems…

**Comment by lfarv** [comment]
```python
class GridMode(Enum):
    """"
    Class to defined the grid mode use when searching
    for the boundary
    """
    RADIAL = 0
    GRID = 1
    RECURSIVE = 2
```

CARTESIAN instead of GRID ?

lfarv:
By analogy with `get_optics`, the processing of `dp` should be:
- if ring is 6D, take the ring as it is, dp is ignored,
- if ring is 4D, dp is honoured.

It makes things simpler:
- in *boundary_search, use `dp=None` as the default value. None is interpreted as 0 in 4D
- set_ring_orbit becomes simpler: forget the frequency, forget the warnings. 
```python
  def set_ring_orbit(ring, dp, refpts, orbit):
      """
      Returns a ring starting at refpts and initial
      closed orbit
      """
      if refpts is not None:
          assert numpy.isscalar(refpts), 'Scalar value needed for refpts'
          newring = newring.rotate(refpts)
      if orbit is None:
              orbit, _ = ring.find_orbit(dp=dp)
      return orbit, newring
  ```
Now, if `orbit` is the initial closed orbit of the **non-rotated** ring, why computing it again for each starting point? It could be done only once. Or should there be a `newring.find_orbit` instead of `ring.find_orbit`? Or could you compute once for ever the orbit at all refpts points, and rotate this orbit array simultaneously with ring?
Just looking for minimising the closed orbit searches, though is very fast compared to the rest…

**Comment by lfarv** [comment]
By the way, I get no problems with `patpass` (on MacOS), unlike what happens on GitHub…

**Comment by swhite2401** [comment]
I could not reproduce the problems with `patpass` either...I am not sure how to move further with something I cannot reproduce

**Comment by swhite2401** [comment]
> @swhite2401: works nicely as expected, but there is a problem:
> 
> Instead of your awful refpts combination:
> 
> ```python
> ids = ring.get_refpts('ID*')
> quads = ring.get_refpts(at.Quadrupole)
> dips = ring.get_refpts(at.Dipole)
> ncells = 1
> refpts = numpy.sort(numpy.concatenate((ids[:ncells], 
>                                        quads[quads<ids[ncells]],
>                                        dips[dips<ids[ncells]])))
> ```
> 
> I'm using a simpler combination with boolean refpts:
> 
> ```python
> ncells = 1
> ids = ring.get_cells(at.checkname('ID*'))
> quads = ring.get_cells(at.checktype(at.Quadrupole))
> dips = ring.get_cells(at.checktype(at.Dipole))
> # Combine references
> refpts= ids | quads | dips
> # Set cells > ncells to False
> cellstart = ring.uint32_refpts(ids)
> refpts[cellstart[ncells]:] = False
> ```
> 
> Using booleans is usually more efficient, but `get_acceptance` crashes with boolean refpts.
> 
> At least at line 98 of `acceptance.py`: `rp = numpy.atleast_1d(refpts)` should be replaced by ` rp = ring.uint32_refpts(refpts)`
> 
> But there may be other problems…

Found the same issue in Touschek, shoul dbe ok now

**Comment by swhite2401** [comment]
@lfarv, All tests passed with the new push without any change to `patpass` that is the same as for the master branch

**Comment by swhite2401** [comment]
> @lfarv, All tests passed with the new push without any change to `patpass` that is the same as for the master branch

Edit: I take it back, this is very long for unbuntu 3.9 and macos 3.7

**Comment by swhite2401** [comment]
> By analogy with `get_optics`, the processing of `dp` should be:
> 
> * if ring is 6D, take the ring as it is, dp is ignored,
> * if ring is 4D, dp is honoured.
> 
> It makes things simpler:
> 
> * in *boundary_search, use `dp=None` as the default value. None is interpreted as 0 in 4D
> * `set_ring_orbit` becomes simpler: forget the frequency, forget the warnings.
> 
> ```python
>   def set_ring_orbit(ring, dp, refpts, orbit):
>       """
>       Returns a ring starting at refpts and initial
>       closed orbit
>       """
>       if refpts is not None:
>           assert numpy.isscalar(refpts), 'Scalar value needed for refpts'
>           newring = newring.rotate(refpts)
>       if orbit is None:
>               orbit, _ = ring.find_orbit(dp=dp)
>       return orbit, newring
> ```
> 
> Now, if `orbit` is the initial closed orbit of the **non-rotated** ring, why computing it again for each starting point? It could be done only once. Or should there be a `newring.find_orbit` instead of `ring.find_orbit`? Or could you compute once for ever the orbit at all refpts points, and rotate this orbit array simultaneously with ring? Just looking for minimising the closed orbit searches, though is very fast compared to the rest…

Modifed as suggested, there was indeed a bug with `newring`, concerning the `orbit` everything is considered a scalar `refpts` in `boundary.py`, the loop takes place in `acceptance.py`, the orbit could be pre-computed there.
However I would like to keep `boundary.py` as is because it will be more convenient to distribute it on the cluster by `refpts` in this way (basically replacing the loop in `acceptance.py` by cluster submission).

File: pyat/at/acceptance/acceptance.py [hunk 1]
```diff
@@ -0,0 +1,264 @@
+import at
```
lfarv, line 1: `import at` not used

pyat/at/acceptance/boundary.py
```diff
@@ -0,0 +1,407 @@
+import at
+from at.lattice import AtError, AtWarning
+from enum import Enum
+import numpy
+import warnings
```
lfarv, line 5: `AtWarning` and `warnings` not used

pyat/at/acceptance/boundary.py
```diff
@@ -0,0 +1,407 @@
...
+                              use_mp=False, verbose=True):
+    """
+    Recursively search for the boundary in a given plane and direction (angle)
+    """
+    def get_r(arr):
+        r = numpy.linalg.norm(arr)
+        return numpy.around(r, decimals=9)
+
+    def search_boundary(ring, planesi, angles, rtol, rstep, nturns,
+                        offset, method):
```
lfarv, line 297: The `ring` parameter is not used in the function

pyat/at/acceptance/boundary.py
```diff
@@ -0,0 +1,407 @@
...
+    result = search_boundary(ring, config.planesi, angles, rtol, rstep,
+                             nturns, offset, method)
+    if verbose:
+        print('Calculation took {0}'.format(time.time()-t0))
+    return result
+
+
+def boundary_search(ring, planes, npoints, amplitudes, nturns=1024,
+                    refpts=None, dp=None, offset=None, bounds=None,
+                    grid_mode=GridMode.RADIAL, use_mp=False, verbose=True):
```
lfarv, line 392: I would rename the `refpts` argument, as it is not the usual refpts type (integer,  array of integers or array of booleans). Here it is an index in the lattice.

lfarv (commented):
I think `plot_acceptance` could be in its own module, because it has nothing in common with the rest of `specific.py`. But that's a matter of taste, put it where you like!

swhite2401:
Unused imports removed, `refpts` changed to `obspt` in `boundary.py` and `standalone.py` module added to at.plot.

Request from @simoneliuzzo:
1-different radii distribution (not uniform in recursive mode?)
2-divider in recursive search (defaults to 2 now)
3- search for max range in 2D -> use as starting point
4-cluster submission (separate PR other functions may profit form cluster access)

pyat/at/plot/standalone.py
```diff
@@ -0,0 +1,78 @@
...
+
+    units = {'x':'[m]', 'xp':'[rad]','y':'[m]', 
+             'yp':'[rad]', 'dp':'', 'ct':'[m]'}
+
+    refpts = kwargs.pop('refpts',None)
+    block = kwargs.pop('block',False)
+    if len(numpy.atleast_1d(refpts))>1:
+        print('Multiple refpts provided for acceptance plot {0}'.format(refpts))
+        print('Using only the first one {0}'.format(refpts[0]))
+        kwargs.update('refpts',refpts[0])
```
lfarv, line 58: 2 problems here:
- does not accept boolean refpts
- dict.update does not accept this syntax: should be  kwargs.update(refpts=kwargs.refpts[0]), or kwargs['refpts']=kwargs['refpts'][0]

lfarv (commented):
There are still references to` at.GridMode.GRID` in doc strings (plot_acceptance for instance).

lfarv:
I have a naive question on the `offset` keyword: what is the use of this orbit offset which is applied identically at all locations around the ring?

simoneliuzzo:
I tested these lines in this branch (clean compile) in MACOS 12.1

` import at
    import numpy
    from at.tracking import patpass
    import time
    import multiprocessing as mp

    mp.set_start_method('spawn')

    path = './'
    filename = 'EBS_SR_lattice.mat'
    key = 'betamodel'
    latticef = path + filename

    ring.radiation_off()
    parts = numpy.zeros((6, 200))

    t0 = time.time()
    patpass(ring, numpy.asfortranarray(parts), nturns=1024)
    print(time.time() - t0)
`

After about 2minutes of intensive CPU use:

Screenshot 2022-02-01 at 12 54 57

there is **NO error message**. 

Screenshot 2022-02-01 at 12 55 39

pyat/at/plot/standalone.py
swhite2401: Done, I force a scalar input similar as in `boundary.py`, plotting makes no sense for several `refpts`

swhite2401:
> I have a naive question on the `offset` keyword: what is the use of this orbit offset which is applied identically at all locations around the ring?

It is a user input replacing the closed orbit, I agree that for multiple `refpts` it is not clear I will allow the possibility of giving an `array(len(refpts))` as well

swhite2401:
Thanks @simoneliuzzo looks like we are to issue raised in #347

lfarv:
@simoneliuzzo : That looks very strange, since I never got any problem on MacOS 12.2 (or 12.1 before), neither starting the script form a terminal window, nor from a Jupiter notebook.

How do you run your job, and from which working directory ?

Or to start from scratch, which python are you using, and how did you install PyAT ?
- `pip install . ` (from the pyat directory)
- `pip install -e .`
- `python setup.py install ?`

simoneliuzzo:
> @simoneliuzzo : That looks very strange, since I never got any problem on MacOS 12.2 (or 12.1 before), neither starting the script form a terminal window, nor from a Jupiter notebook.
> 
> How do you run your job, and from which working directory ?
> 
> Or to start from scratch, which python are you using, and how did you install PyAT ?
> 
>     * `pip install . ` (from the pyat directory)
> 
>     * `pip install -e .`
> 
>     * `python setup.py install ?`

Dear @lfarv ,

I run the job using pycharm, pressing the play button.

I installed pyAT by doing the following:
git clone of atcollab 
cd pyat
git checkout master
git pull
rm -rf build/
rm -rf accelerator-toolbox.egg-link 
rm -rf accelerator_toolbox.egg-info
python3.9 setup.py develop --install-dir=./

I do not know remember how I installed python.

I tried your commands and the first one gives this message:
Simones-MacBook-Pro-3:pyat liuzzo$ pip install .
-bash: /usr/local/bin/pip: /usr/local/opt/python/bin/python3.7: bad interpreter: No such file or directory

lfarv:
Hi @simoneliuzzo.
From what I see, you are using a python 3.9 installed with Homebrew. I have the same installed, giving no problem.

But it seems that you do not work in a virtual python environment, which can lead to problems. In particular, apparently this forces you to use `--install-dir=./`, which is a non-standard location. Could you try the method given in the `developer.rst` document ?

The 1<sup>st</sup> thing is to create a virtual environment. This avoids a lot of problems of version compatibility, file permissions and more…. For instance, your attempt to use `pip` refers to` /usr/local/opt/python/bin/python3.7` instead of `/usr/local/bin/python3.9`: interpreter conflict ! So

* ``$ python3.9 -m venv <some_location>``

Then:

* activate the virtual environment:

  ``$ source <some_location>/bin/activate``
  From that moment, all commands like "python", "python3", "pip" etc. refer to the specific version within this environment, and you start with a fresh new python installation (no additional package installed yet).

Practically, you should ALWAYS work in a virtual environment.

* make sure you have a recent pip installer:

  ``$ pip install --upgrade pip``
* Go to the pyat root directory:

  ``cd <atroot>/pyat``
* install AT:

  ``$ pip install -e .``
  `pip` is now the preferred installation tool, rather that `setup.py`

The last point is to have PyCharm use this new interpreter. At the bottom right of the PyCharm window, it shows the current interpreter. By clicking on it, you can select "Add interpreter...". Then you select "Virtualenv Environment" on the left, "Existing environment" on the right, browse to `<some_location>/bin/python` (the one you created previously) for the "Interpreter" field, and check the box "Make available to all projects". Now PyCharm uses your environment for checking the syntax and running scripts. With the same menu, you can switch to any other interpreter, it's similar to the "activation" of the virtual environment shown above.

With that, @swhite2401's test should run without problems.

simoneliuzzo:
@lfarv and @swhite2401,

I will follow the list of commands and let you know.

My personal feeling is that a list of actions that we HAVE to do should be clearly stated in the code documentation.
This list seems growing every day.

regards
Simone

simoneliuzzo:
@lfarv and @swhite2401 ,

I followed step by step. I have the following error: 

Simones-MacBook-Pro-3:~ liuzzo$ cd ~/beamdyn/
Simones-MacBook-Pro-3:beamdyn liuzzo$ l
-bash: l: command not found
Simones-MacBook-Pro-3:beamdyn liuzzo$ ls
EBSlattices			at_ringpara
MDT				fitandresponseclasses
OPTICS				fitandresponseclasses_V0
Radia				matlab
at				response-matrix-processing
at_master			work
Simones-MacBook-Pro-3:beamdyn liuzzo$ 
Simones-MacBook-Pro-3:beamdyn liuzzo$ python3.9 -m venv .
Simones-MacBook-Pro-3:beamdyn liuzzo$ source ./bin/activate
(beamdyn) Simones-MacBook-Pro-3:beamdyn liuzzo$ ls
EBSlattices			fitandresponseclasses
MDT				fitandresponseclasses_V0
OPTICS				include
Radia				lib
at				matlab
at_master			pyvenv.cfg
at_ringpara			response-matrix-processing
bin				work
(beamdyn) Simones-MacBook-Pro-3:beamdyn liuzzo$ pip install --upgrade pip
Requirement already satisfied: pip in ./lib/python3.9/site-packages (21.2.4)
Collecting pip
  Downloading pip-22.0.3-py3-none-any.whl (2.1 MB)
     |████████████████████████████████| 2.1 MB 4.6 MB/s 
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 21.2.4
    Uninstalling pip-21.2.4:
      Successfully uninstalled pip-21.2.4
Successfully installed pip-22.0.3
(beamdyn) Simones-MacBook-Pro-3:beamdyn liuzzo$ cd at
at/          at_master/   at_ringpara/ 
(beamdyn) Simones-MacBook-Pro-3:beamdyn liuzzo$ cd at/pyat/
(beamdyn) Simones-MacBook-Pro-3:pyat liuzzo$ pip install -e .
Obtaining file:///Users/liuzzo/beamdyn/at/pyat
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting numpy>=1.16.6
  Using cached numpy-1.22.1-cp39-cp39-macosx_10_9_x86_64.whl (17.7 MB)
Collecting scipy>=0.16
  Downloading scipy-1.7.3-cp39-cp39-macosx_10_9_x86_64.whl (33.2 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 33.2/33.2 MB 49.3 MB/s eta 0:00:00
Installing collected packages: numpy, scipy, accelerator-toolbox
  Running setup.py develop for accelerator-toolbox
    error: subprocess-exited-with-error
    
    × python setup.py develop did not run successfully.
    │ exit code: 1
    ╰─> [60 lines of output]
        running develop
        /private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/command/easy_install.py:157: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
          warnings.warn(
        /private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
          warnings.warn(
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2037, in try_parse
            return packaging.version.Version(name)
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/_vendor/packaging/version.py", line 266, in __init__
            raise InvalidVersion(f"Invalid version: '{version}'")
        pkg_resources.extern.packaging.version.InvalidVersion: Invalid version: '/Users/liuzzo/beamdyn/at/pyat/accelerator_toolbox'
        
        During handling of the above exception, another exception occurred:
        
          File "<string>", line 2, in <module>
          File "<pip-setuptools-caller>", line 34, in <module>
          File "/Users/liuzzo/beamdyn/at/pyat/setup.py", line 151, in <module>
            setup(
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/__init__.py", line 155, in setup
            return distutils.core.setup(**attrs)
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/core.py", line 148, in setup
            return run_commands(dist)
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/core.py", line 163, in run_commands
            dist.run_commands()
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/dist.py", line 967, in run_commands
            self.run_command(cmd)
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/dist.py", line 985, in run_command
            cmd_obj.ensure_finalized()
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/cmd.py", line 107, in ensure_finalized
            self.finalize_options()
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/command/develop.py", line 52, in finalize_options
            easy_install.finalize_options(self)
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/command/easy_install.py", line 343, in finalize_options
            self.local_index = Environment(self.shadow_path + sys.path)
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 981, in __init__
            self.scan(search_path)
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 1013, in scan
            for dist in find_distributions(item):
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2083, in find_on_path
            for dist in factory(fullpath):
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2083, in find_on_path
            for dist in factory(fullpath):
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2083, in find_on_path
            for dist in factory(fullpath):
          [Previous line repeated 970 more times]
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2079, in find_on_path
            path_item_entries = _by_version_descending(filtered)
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2049, in _by_version_descending
            return sorted(names, key=_by_version, reverse=True)
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2047, in _by_version
            return [try_parse(part) for part in parts]
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2047, in <listcomp>
            return [try_parse(part) for part in parts]
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2039, in try_parse
            return packaging.version.Version('0')
          File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/_vendor/packaging/version.py", line 269, in __init__
            self._version = _Version(
          File "<string>", line 1, in <lambda>
        RecursionError: maximum recursion depth exceeded while calling a Python object
        [end of output]
    
    note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× python setup.py develop did not run successfully.
│ exit code: 1
╰─> [60 lines of output]
    running develop
    /private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/command/easy_install.py:157: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
      warnings.warn(
    /private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
      warnings.warn(
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2037, in try_parse
        return packaging.version.Version(name)
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/_vendor/packaging/version.py", line 266, in __init__
        raise InvalidVersion(f"Invalid version: '{version}'")
    pkg_resources.extern.packaging.version.InvalidVersion: Invalid version: '/Users/liuzzo/beamdyn/at/pyat/accelerator_toolbox'
    
    During handling of the above exception, another exception occurred:
    
      File "<string>", line 2, in <module>
      File "<pip-setuptools-caller>", line 34, in <module>
      File "/Users/liuzzo/beamdyn/at/pyat/setup.py", line 151, in <module>
        setup(
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/__init__.py", line 155, in setup
        return distutils.core.setup(**attrs)
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/core.py", line 148, in setup
        return run_commands(dist)
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/core.py", line 163, in run_commands
        dist.run_commands()
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/dist.py", line 967, in run_commands
        self.run_command(cmd)
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/dist.py", line 985, in run_command
        cmd_obj.ensure_finalized()
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/_distutils/cmd.py", line 107, in ensure_finalized
        self.finalize_options()
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/command/develop.py", line 52, in finalize_options
        easy_install.finalize_options(self)
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/setuptools/command/easy_install.py", line 343, in finalize_options
        self.local_index = Environment(self.shadow_path + sys.path)
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 981, in __init__
        self.scan(search_path)
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 1013, in scan
        for dist in find_distributions(item):
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2083, in find_on_path
        for dist in factory(fullpath):
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2083, in find_on_path
        for dist in factory(fullpath):
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2083, in find_on_path
        for dist in factory(fullpath):
      [Previous line repeated 970 more times]
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2079, in find_on_path
        path_item_entries = _by_version_descending(filtered)
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2049, in _by_version_descending
        return sorted(names, key=_by_version, reverse=True)
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2047, in _by_version
        return [try_parse(part) for part in parts]
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2047, in <listcomp>
        return [try_parse(part) for part in parts]
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/__init__.py", line 2039, in try_parse
        return packaging.version.Version('0')
      File "/private/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/pip-build-env-qfph36do/overlay/lib/python3.9/site-packages/pkg_resources/_vendor/packaging/version.py", line 269, in __init__
        self._version = _Version(
      File "<string>", line 1, in <lambda>
    RecursionError: maximum recursion depth exceeded while calling a Python object
    [end of output]

note: This error originates from a subprocess, and is likely not a problem with pip.
(beamdyn) Simones-MacBook-Pro-3:pyat liuzzo$

swhite2401:
@liuzzo you have to start from a clean at folder:

```
rm -rf accelerator*
rm -rf build
find at -name "*.pyc" -exec rm '{}' \;
find at -name "*.so" -exec rm '{}' \;
```

Then start again

lfarv:
@simoneliuzzo:
I forgot to say: in the command

`python3.9 -m venv <some_location>`

<some_location> **must not exist**. `venv` will create it and fill it with fresh info. Then all that you install will go into this directory (in `<some_location>/lib/python3.9/site-packages`), so that all environments are independent. In your case, it's already full of "garbage", in particular the `bin` and `lib` sub-directories.

Better creating you virtual environments up in your home directory (and not in a common area). Example:

```
home
    python
        port37
        sys38
        brew39
```
which I created with
```
$ /opt/local/bin/python3.7 -m venv ~/python/port37
$ /usr/bin/python3 -m venv ~/python/sys38
$ /usr/local/bin/python3.9 -m venv ~/python/brew39
```
with different python versions. I can activate one with for instance:
`$ source ~/python/sys38/bin/activate`

simoneliuzzo:
I have no idea of what is the effect of the actions you request me to do.
I blindly obey to all MUST DOs listed above in an order that I figured out after several trials/errors. 
I would prefer to run simulations rather than installations.

Let me know if there is anything I did not do as I should have done.
The "minimal list" for installation I used is:

```
cd ~
/usr/local/bin/python3.9 -m venv ~/python/brew39
source ~/python/brew39/bin/activate
pip install --upgrade pip
cd <atroot>/pyat
rm -rf accelerator*
rm -rf build
find at -name "*.pyc" -exec rm '{}' \;
find at -name "*.so" -exec rm '{}' \;
pip install -e .

```

You probably will not agree, but I do not find this really "user friendly".



(liuzzo) Simones-MacBook-Pro-3:~ liuzzo$ cd ~
(liuzzo) Simones-MacBook-Pro-3:~ liuzzo$ ls
AllCellsS28F.xlsx		Dropbox				Pictures			machfs				tmp14days
Applications			Library				Public				machfs2				venv
Desktop				Movies				PycharmProjects			matlab_crash_dump.3597-1
Documents			Music				beamdyn				matlab_crash_dump.95516-1
Downloads			NoMachine			include				ownCloud
(liuzzo) Simones-MacBook-Pro-3:~ liuzzo$ /usr/local/bin/python3.9 -m venv ~/python/brew39
(liuzzo) Simones-MacBook-Pro-3:~ liuzzo$ 
(liuzzo) Simones-MacBook-Pro-3:~ liuzzo$ source ~/python/brew39/bin/activate
(brew39) Simones-MacBook-Pro-3:~ liuzzo$ 
(brew39) Simones-MacBook-Pro-3:~ liuzzo$ pip install --upgrade pip
Requirement already satisfied: pip in ./python/brew39/lib/python3.9/site-packages (21.2.4)
Collecting pip
  Using cached pip-22.0.3-py3-none-any.whl (2.1 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 21.2.4
    Uninstalling pip-21.2.4:
      Successfully uninstalled pip-21.2.4
Successfully installed pip-22.0.3
(brew39) Simones-MacBook-Pro-3:~ liuzzo$ cd ~/beamdyn/at/pyat
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$ pip install -e .
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$ rm -rf accelerator*
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$ rm -rf build
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$ find at -name "*.pyc" -exec rm '{}' \;
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$ find at -name "*.so" -exec rm '{}' \;
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$ 
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$ pip install -e .
Obtaining file:///Users/liuzzo/beamdyn/at/pyat
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: scipy>=0.16 in /Users/liuzzo/python/brew39/lib/python3.9/site-packages (from accelerator-toolbox==0.2.3.dev35+gbdd50875) (1.7.3)
Requirement already satisfied: numpy>=1.16.6 in /Users/liuzzo/python/brew39/lib/python3.9/site-packages (from accelerator-toolbox==0.2.3.dev35+gbdd50875) (1.22.1)
Installing collected packages: accelerator-toolbox
  Running setup.py develop for accelerator-toolbox
Successfully installed accelerator-toolbox-0.2.3.dev35+gbdd50875
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$ 
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$

swhite2401:
@simoneliuzzo , yes this may seem complicated but once you have set your `venv` etc.. only the last command `pip install -e .` is needed

lfarv:
@simoneliuzzo, @swhite2401 : As far as I can see, your installation is now correct. Your installation of `matplotlib` is ok (you could also have done it with `pip install matplotlib`).

I tried exactly what you did (including running from PyCharm), and I could never reproduce the problem…

Capture d’écran 2022-02-03 à 16 54 18

I'm running out of ideas…

swhite2401:
Me too...you are running python 3.9 with the same macos version?

simoneliuzzo:
I update to macos 12.2 and try again the whole sequence

lfarv:
I tried on 3 different Macs, including an old one running MacOS HighSierra (10,13), with python versions 3.6, 3.7, 3.8, 3.9:
impossible to reproduce the bug. Running on GitHub also succeeds on test_patpass, at least most of the time (there is another problem). I'm convinced that it has to do with Simone's setup, but I do not see what it may be…

simoneliuzzo:
may be I need some special XCode developper tools?

simoneliuzzo:
I updated to MACOS 12.2

the I did the following (compulsory order and actions, so 2 lines less then the "first time" list)

```
cd ~
source ~/python/brew39/bin/activate
pip install --upgrade pip
cd <atroot>/pyat
rm -rf accelerator*
rm -rf build
find at -name "*.pyc" -exec rm '{}' \;
find at -name "*.so" -exec rm '{}' \;
pip install -e .
restart pycharm
```

the I tested the example_tltda.py and the error persists as before.

swhite2401:
@carmignani can you run the example script as well?

lfarv:
@simoneliuzzo: I don't have much hope it may help, but could you try 2 things?
Still within your virtual environment (repeat `source ~/python/brew39/bin/activate` if you are in a new shell):

1.  rather that starting from PyCharm, just run from the terminal:
```
   cd "/Users/liuzzo/beamdyn/OPTICS/New Booster/DBA_long"
   python example_tltda.py
```
2. Rather than an "editable" installation, try a "normal" one: just remove the "-e" flag in the `pip` command:
   ```
   cd <atroot>/pyat
   pip install .
   ```
   (you get forget all intermediate lines, they are not necessary).
   This completely disconnects the installation from the "beamdyn" directory, all the code is moved to the virtual environment. "editable" mode is made for development: you do not need to install again after making modifications, just restart python. In "normal" mode, you need to re-install again after each modification, but it's safer…

You can of course combine both…

lfarv:
@swhite2401, @simoneliuzzo: I don't have any new idea for Simone's problem, ideally I would try and have a look at the Mac. But while keeping the problem in mind, I would merge the branch. Do you still intend to modify it?

swhite2401:
@lfarv, @simoneliuzzo, this is really a pity.... I have made the start_method available as input so the solution is to set `start_method='fork'`.

I personally think that `'fork'` should be the default for MacOS, it works fine and it solves the problem... however since @lfarv doesn't like it I am afraid I have nothing better to propose than the present situation.

I have a few requests from @simoneliuzzo to implement before merging, besides that I think it is fine

lfarv:
@simoneliuzzo, @swhite2401: another check for Simone: can you check that you have no variable in your environment with a name containing "PYTHON", like for example PYTHONPATH or PYTHONSTARTUP ?

simoneliuzzo:
dear @lfarv and @swhite2401 ,

I will be able to test only next Tuesday. I hope it is not too late.

Also, if the problem is just my installation I have no objections to merge. Actually we could ask @carmignani  to test as well, he has a mac.

best regards
Simone

swhite2401:
No rush! and yes I would be very curious to see if @carmignani as the same issue

swhite2401:
I have fixed the conflicts and implemented the suggestions by @simoneliuzzo, I think this one is ready for merging: should I add examples in `at/pyat/examples`? I don't think adding tests is appropriate since the calculation is too long

simoneliuzzo:
do we still have time to comment?
I am sorry, but I will not have much time to test the code before tomorrow.

For the test, in the declined DA module I proposed a few month ago I did include them. I think they are useful indicator of something globally wrong. A computation based on few turns would not be long to run in the test.

Since it was crucial for the last acceptation/refusal of a proposed similar development, I would request that also @lcarver is involved in testing and reviewing of this new version.

swhite2401:
Sure, all the testing by everyone is very useful! Plenty of time to comment as well

Ok you are right, I will add a test module with "fast" calculations

swhite2401:
Tests for 1D and 2D acceptance added

lfarv:
@swhite2401: 2.5 second for the three tests! It's perfectly acceptable! Nice.

simoneliuzzo:
Dear @swhite2401 and @lfarv 

I refollowed the AT3.0  installation/update procedure:

cd ~
source ~/python/brew39/bin/activate
pip install --upgrade pip
cd <atroot>/pyat
rm -rf accelerator*
rm -rf build
find at -name "*.pyc" -exec rm '{}' \;
find at -name "*.so" -exec rm '{}' \;
pip install -e .
restart pycharm

and run the example_tltda.py test. 
It still gives me errors.

Should I modify the installation above or may be simply try the example script with use_mp=False? 

best

lfarv:
@simoneliuzzo : Did you check environment variables ? If nothing appears, I would really like to have a look at you Mac configuration. There must be something unusual. It's a pity that you could not use fast DA computation while you are a major user! @swhite2401's solution will work, but I'm afraid it hides a tricky problem…

simoneliuzzo:
> @simoneliuzzo, @swhite2401: another check for Simone: can you check that you have no variable in your environment with a name containing "PYTHON", like for example PYTHONPATH or PYTHONSTARTUP ?

Dear @lfarv 
here below a list of variables. 
(brew39) Simones-MacBook-Pro-3:pyat liuzzo$ printenv
TERM_PROGRAM=Apple_Terminal
SHELL=/bin/bash
TERM=xterm-256color
TMPDIR=/var/folders/pk/h5y3jmlj5nzcwfztgzk51v000000gn/T/
TERM_PROGRAM_VERSION=443
OLDPWD=/Users/liuzzo
TERM_SESSION_ID=B467E0B4-F8F8-4654-BC64-9DCAAAFC1ECB
USER=liuzzo
SSH_AUTH_SOCK=/private/tmp/com.apple.launchd.iY8CVhClqi/Listeners
VIRTUAL_ENV=/Users/liuzzo/python/brew39
PATH=/Users/liuzzo/python/brew39/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin:/opt/X11/bin
LaunchInstanceID=D12C9D64-9F92-45F6-8AF3-6B9173800B1A
__CFBundleIdentifier=com.apple.Terminal
PWD=/Users/liuzzo/beamdyn/at/pyat
XPC_FLAGS=0x0
PS1=(brew39) \h:\W \u\$ 
XPC_SERVICE_NAME=0
SHLVL=1
HOME=/Users/liuzzo
LOGNAME=liuzzo
LC_CTYPE=UTF-8
DISPLAY=/private/tmp/com.apple.launchd.ZUuJ1arXaa/org.macosforge.xquartz:0
SECURITYSESSIONID=186b7
_=/usr/bin/printenv

lfarv:
@simoneliuzzo: it looks clean. Any file or folder in your home directory with a name like `.py*`?

simoneliuzzo:
yes, there is `.python_history`

lfarv:
> yes, there is `.python_history`

No problem with that one…

Concerning this PR, if @swhite2401 and you are happy with it, I propose to merge it, and move the Mac discussion to private.

simoneliuzzo:
I would like to do a few more tests, and read the code in more detail. Also I think the feedback of @lcarver and @carmignani is important, even if it is simply "all ok".

simoneliuzzo:
no problem to move the mac installation MPI discussion elsewhere.

simoneliuzzo:
I am running some test. I appreciate the input dp. 
I would like to use the output of get_acceptance to plot 3 curves with the DA on and off energy (a classic).
The help helped me, but I may not immediately guess the shape/type/format of the tree outputs. They are lists. is this sufficient information in the python word? I will stop the code in pycharm at that point to look at the shape in order to produce my plot.

simoneliuzzo:
I launched this command 
tlt, momap, refpts = booster_ring.get_lifetime(emity, bunch_curr, zn=zn, refpts=refpts,
                                               use_mp=True, verbose=True , start_method='fork')
but it does not seem to use my 4 cores

swhite2401:
Thanks, in fact the help is missing some details on the output and the format is not clear, I will improve this.
For your case only the first output is interesting, the 2 others contains the survived particles and the full grid

-> for a single refpts you get a numpy.array((2,i)), that you can plot using plt.plot(*out)
-> for n refpts you get a list of n numpy.array((2,i)), you can plot any refpts using plt.plot(*out[i])

I had to use a list in this case because in recursive mode I cannot predict the grid size...

Concerning the dp input, I had matched the until now logic to ignore the dp input in 6D, however to be consistent with #364 I will re-introduce the automated frequency shift in 6D

swhite2401:
> I launched this command
> 
> ```
> tlt, momap, refpts = booster_ring.get_lifetime(emity, bunch_curr, zn=zn, refpts=refpts,
>                                                use_mp=True, verbose=True , start_method='fork')
> ```
> 
> but it does not seem to use my 4 cores

Ah no? It was working fine on grappa, strange...

simoneliuzzo:
the outputs have to be used as in the example_tltda.py script?
ax.plot(*bf)

Also, description of the units for the input and output could be of use

swhite2401:
> the outputs have to be used as in the example_tltda.py script? ax.plot(*bf)
> 
> Also, description of the units for the input and output could be of use

Yes, or you may use the built-in function:
bf,sf,gf = ring.get_acceptance(planes, npoints, amplitudes, use_mp=True)
ring.plot_acceptance(planes, npoints, amplitudes, use_mp=True, acceptance=(bf, sf, gf))
If acceptance is provided it is used, otherwise the other inputs are used to recompute it

Btw I liked the plot from your version it looked better so I might steal it

simoneliuzzo:
> > the outputs have to be used as in the example_tltda.py script? ax.plot(*bf)
> > Also, description of the units for the input and output could be of use
> 
> Yes, or you may use the built-in function: bf,sf,gf = ring.get_acceptance(planes, npoints, amplitudes, use_mp=True) ring.plot_acceptance(planes, npoints, amplitudes, use_mp=True, acceptance=(bf, sf, gf)) If acceptance is provided it is used, otherwise the other inputs are used to recompute it
> 
> Btw I liked the plot from your version it looked better so I might steal it

Ok, is the * that is not trivial to me, but this is due to my limited python skill. Also I need to make a figure out of 3 computations. I run 3 plot/get_acceptance and use the outputs to show the data.

simoneliuzzo:
> > I launched this command
> > ```
> > tlt, momap, refpts = booster_ring.get_lifetime(emity, bunch_curr, zn=zn, refpts=refpts,
> >                                                use_mp=True, verbose=True , start_method='fork')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but it does not seem to use my 4 cores
> 
> Ah no? It was working fine on grappa, strange...

sorry, it was not your command, but my old version

swhite2401:
> > > the outputs have to be used as in the example_tltda.py script? ax.plot(*bf)
> > > Also, description of the units for the input and output could be of use
> > 
> > 
> > Yes, or you may use the built-in function: bf,sf,gf = ring.get_acceptance(planes, npoints, amplitudes, use_mp=True) ring.plot_acceptance(planes, npoints, amplitudes, use_mp=True, acceptance=(bf, sf, gf)) If acceptance is provided it is used, otherwise the other inputs are used to recompute it
> > Btw I liked the plot from your version it looked better so I might steal it
> 
> Ok, is the * that is not trivial to me, but this is due to my limited python skill. Also I need to make a figure out of 3 computations. I run 3 plot/get_acceptance and use the outputs to show the data.

You can use * to "unfold" an ndarray -> a1,a2 = *a

You can do:

b1,_,_ = ring.get_acceptance(...)
b2,_,_ = ring.get_acceptance(...)
b3,_,_ = ring.get_acceptance(...)

plt.plot(*b1,label='case1')
plt.plot(*b2,label='case2')
plt.plot(*b3,label='case3')
plt.legend()
plt.show()

simoneliuzzo:
> Thanks, in fact the help is missing some details on the output and the format is not clear, I will improve this. For your case only the first output is interesting, the 2 others contains the survived particles and the full grid
> 
> -> for a single refpts you get a numpy.array((2,i)), that you can plot using plt.plot(*out) -> for n refpts you get a list of n numpy.array((2,i)), you can plot any refpts using plt.plot(*out[i])
> 
> I had to use a list in this case because in recursive mode I cannot predict the grid size...
> 
> Concerning the dp input, I had matched the until now logic to ignore the dp input in 6D, however to be consistent with #364 I will re-introduce the automated frequency shift in 6D

infact I start seeing this worrying "warning" 
Screenshot 2022-02-17 at 13 00 13

simoneliuzzo:
I tried the different grid modes. recursive seems a bit more optimistic on the y=0 axis. grid mode seems to skip the point at x<0 and y=0.

Screenshot 2022-02-17 at 13 27 54

The other test I would like to do are comparisons with the same computations in matlab-at.

swhite2401:
> I tried the different grid modes. recursive seems a bit more optimistic on the y=0 axis. grid mode seems to skip the point at x<0 and y=0.
> 
> Screenshot 2022-02-17 at 13 27 54
> 
> Screenshot 2022-02-17 at 13 33 13
> 
> The other test I would like to do are comparisons with the same computations in matlab-at.

For the recursive mode, you may use the optional  input `divider` to make a finer search...maybe it missed a lost point?

The grid mode boundary search has some issues it seems, I had never considered a case where the grid is smaller than the DA...I will fix that thanks! On the other hand, I think you have found a weakness, I have done many tests to remove these isolated point and thought I managed but it seems not... maybe my algorithm doesn't work...

swhite2401:
> > Thanks, in fact the help is missing some details on the output and the format is not clear, I will improve this. For your case only the first output is interesting, the 2 others contains the survived particles and the full grid
> > -> for a single refpts you get a numpy.array((2,i)), that you can plot using plt.plot(*out) -> for n refpts you get a list of n numpy.array((2,i)), you can plot any refpts using plt.plot(*out[i])
> > I had to use a list in this case because in recursive mode I cannot predict the grid size...
> > Concerning the dp input, I had matched the until now logic to ignore the dp input in 6D, however to be consistent with #364 I will re-introduce the automated frequency shift in 6D
> 
> infact I start seeing this worrying "warning" Screenshot 2022-02-17 at 13 00 13
> 
> and the consequent "results" Screenshot 2022-02-17 at 13 09 35
> 
> I wait to do any further check/test

I fix this now

swhite2401:
I have fixed the case when the boundary is larger than the DA, recursive still find the correct boundary as expected

image
image

```
import at
import numpy

path = '../lattice/'
filename = 'S28F.mat'
key = 'LOW_EMIT_RING_INJ'
latticef = path+filename
ring.radiation_off()

planes = ['x','y']
npoints = [21,21]
amplitudes = [8e-3,0.006]


br,_,_ = ring.get_acceptance(planes, npoints, amplitudes, grid_mode=at.GridMode.RECURSIVE, use_mp=True)
bg,sg,gg = ring.get_acceptance(planes, npoints, amplitudes, grid_mode=at.GridMode.CARTESIAN, use_mp=True)
bc,_,_ = ring.get_acceptance(planes, npoints, amplitudes, grid_mode=at.GridMode.RADIAL, use_mp=True)

plt.figure()
plt.title('Radiations OFF')
plt.plot(*bg, label='Cartesian')
plt.plot(*gg,'.', label='Grid')
plt.plot(*sg,'.', label='Survived')
plt.legend()

plt.figure()
plt.title('Radiations OFF')
plt.plot(*br, label='Recursive')
plt.plot(*bg, label='Cartesian')
plt.plot(*bc, label='Radial')
plt.legend()
plt.show()
```

simoneliuzzo:
Dear Simon,

here is what I obtain for the ESRF long booster DBA

Screenshot 2022-02-18 at 15 15 46
.

I attach here the lattice

[booster_ring_for_DA.mat.zip](https://github.com/atcollab/at/files/8097860/booster_ring_for_DA.mat.zip)

lfarv:
@swhite2401, @simoneliuzzo:
I looked briefly at your tests, and I think there is a problem: you have been running multiturn tracking with radiation ON, and a frequency different from nominal: these are the exact conditions where `CavityPass` is wrong, and `RFCavityPass` should be used.
I'll immediately open a pull request to set `RFCavityPass` as default, in both versions. This is too dangerous. I do not want to remove `CavityPass` because it has been there from the beginning and probably widely used. I recall the condition where `CavityPass` is wrong:
frequency different from nominal **AND** more that 1 turn (it means it's correct for orbit and matrix computations).

I also noticed that the cavity length is 1.e-10m, which is a very bad thing to do. You should set it to 0.

simoneliuzzo:
@lfarv : the code I use to generate the lattice is here: https://gitlab.esrf.fr/BeamDynamics/newbooster/DBA_long
I add this 1e-10 m to cavities since so long that I do not remember any more why I do it. It will take me time to restore this memory. Could we wait for a few weeks before changing the default? I would like to try to recall why I add this 10^-10 meters.

lfarv:
@simoneliuzzo: of course we'll wait for you. But this should not have anything to do with the pass method: both handle the 1.e-10 m length the same way: adding a 0.5e-10m long drift on each side of a thin cavity ! Which is not a very useful computation…
Don't forget that CavityPass is completely wrong in your configuration. I consider this as an AT bug, and this time, changing the default pass method will change the results of existing code ! Would you consider this bug fix as a backward compatibility problem?

lfarv:
@swhite2401 : The "dp" processing in set_ring_orbit is intricate, though it seems correct. But:

- when catching the warnings, you catch not only the one you expect, but also any other warning, which can be dangerous. You should catch only the specific one: `simplefilter('ignore:In 6D, "dp" and "dct" are ignored:AtWarning')`
- better: if you know that "dp" will trigger a warning, why not removing it rather that silencing the warning? Once you set the frequency, dp is meaningless.

However, you know that @simoneliuzzo and @carmignani are pushing for an error instead of a warning. I'm not pushing too much for that, but I opened #373 for discussion. This may have an impact.

In the same function:
`newring = ring.copy()`
The copy is useless, `newring = ring` would work, newring is used only tracking.

lfarv:
I have another thought about your processing of `dp`:
As it is done, since you changed the RF frequency, the particles you throw for transverse acceptance (1D or 2D) do not oscillate longitudinally (except a small oscillation from the path lengthening with transverse amplitude). So the 6D option does not bring anything, either in 4D or in 6D the particles stay at constant off-momentum `dp`. And the results are identical, so no interest in 6D tracking.

Would not it be more interesting to keep the lattice unchanged, and to add the `dp` to the 5th coordinate of the thrown particles:` r_in = closed_orbit + variable_transverse_amplitude + dp`. (Can be applied for both 4D and 6D).
Then you would see the difference between a particle staying at constant off-momentum (4D), and a particle which oscillates and is possibly damped (6D).

In other words, the way it is done now, you look at the performance for on-energy particles, in a ring tuned off-momentum. The other way, you would look at particles experiencing a momentum kick.

lfarv:
Another thought: as it is done now, the `dp` processing would better done with a python decorator applied to `get_acceptance`. It's a 10 lines function, which can later decorate any function. It can be done in a next step !

swhite2401:
> I have another thought about your processing of `dp`: As it is done, since you changed the RF frequency, the particles you throw for transverse acceptance (1D or 2D) do not oscillate longitudinally (except a small oscillation from the path lengthening with transverse amplitude). So the 6D option does not bring anything, either in 4D or in 6D the particles stay at constant off-momentum `dp`. And the results are identical, so no interest in 6D tracking.
> 
> Would not it be more interesting to keep the lattice unchanged, and to add the `dp` to the 5th coordinate of the thrown particles:` r_in = closed_orbit + variable_transverse_amplitude + dp`. (Can be applied for both 4D and 6D). Then you would see the difference between a particle staying at constant off-momentum (4D), and a particle which oscillates and is possibly damped (6D).
> 
> In other words, the way it is done now, you look at the performance for on-energy particles, in a ring tuned off-momentum. The other way, you would look at particles experiencing a momentum kick.

You are right `dp` is a fixed offset, the interest in having it in 6D is to be able to include radiation damping and RF: results are not the same when these are turned on. This is generally what is done in real life when measuring off-momentum DA.

What you propose can already be done using the (wrongly named) `offset` argument, in case its value is not `None` it overwrites the initial closed orbit: you may add whatever quantity to the 5th coordinate there to get a synchrotron oscillation.
I can rename this argument to `init_coord` maybe?

swhite2401:
> Dear Simon,
> 
> here is what I obtain for the ESRF long booster DBA
> 
> Screenshot 2022-02-18 at 15 15 46
> .
> 
> I attach here the lattice
> 
Thanks, looks like I broke something while fixing another, these wiggles are not normal...

lfarv:
> You are right dp is a fixed offset, the interest in having it in 6D is to be able to include radiation damping and RF: results are not the same when these are turned on. This is generally what is done in real life when measuring off-momentum DA.

I fully agree on the interest of having longitudinal motion in DA computation ! My only concern is that there are 2 different ways of interpreting  the fixed `dp` argument: one is to look at the performance of lattice tuned off-momentum (as it is done now), the other one is to throw the test particles with a momentum deviation from the machine tuning. If this is can be done with `offset`, that's perfect. May be it deserves a page on the documentation web site (apart from the function doctoring)?

lfarv (approved):

swhite2401:
> > You are right dp is a fixed offset, the interest in having it in 6D is to be able to include radiation damping and RF: results are not the same when these are turned on. This is generally what is done in real life when measuring off-momentum DA.
> 
> I fully agree on the interest of having longitudinal motion in DA computation ! My only concern is that there are 2 different ways of interpreting the fixed `dp` argument: one is to look at the performance of lattice tuned off-momentum (as it is done now), the other one is to throw the test particles with a momentum deviation from the machine tuning. If this is can be done with `offset`, that's perfect. May be it deserves a page on the documentation web site (apart from the function doctoring)?

Yes, this and collective effects definitely need documentation: once we have the final version I will work on it

swhite2401:
> Another thought: as it is done now, the `dp` processing would better done with a python decorator applied to `get_acceptance`. It's a 10 lines function, which can later decorate any function. It can be done in a next step !

I do not really see the interest in a decorator here...this very specific no? But maybe I am missing something, I removed the warning filter

swhite2401:
> Dear Simon,
> 
> here is what I obtain for the ESRF long booster DBA
> 
> Screenshot 2022-02-18 at 15 15 46
> .
> 
> I attach here the lattice
> 
> [booster_ring_for_DA.mat.zip](https://github.com/atcollab/at/files/8097860/booster_ring_for_DA.mat.zip)

Wiggles fixed, for the CARTESIAN GRID I am afraid it will very difficult to handle isolated lost points inside the acceptance...any idea?

simoneliuzzo:
I have updated and re-run my test script. I have this strange message now, 
```
booster_ring[135]

```
PyDev console: starting.
RFCavity('RFC', 1e-12, 9000000.0, 352371625.9975046, 979, 6000000000.0)


```
bf, sf, gf = booster_ring.plot_acceptance(['x', 'y'],
                                                  [31, 31],
                                                  [0.05, 0.03],
                                                  nturns=turns,
                                                  use_mp=True, start_method='fork')

8 cpu found for acceptance calculation
Multi-process acceptance calculation selected...
The estimated load for grid mode is 120.125
The estimated load for recursive mode is 40.0
GridMode.RECURSIVE is recommended
  File "/Applications/PyCharm CE.app/Contents/plugins/python-ce/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<string>", line 1, in <module>
  File "/Users/liuzzo/beamdyn/at/pyat/at/plot/standalone.py", line 80, in plot_acceptance
    boundary, survived, grid = ring.get_acceptance(*args, **kwargs)
  File "/Users/liuzzo/beamdyn/at/pyat/at/acceptance/acceptance.py", line 126, in get_acceptance
    b, s, g = boundary_search(ring, planes, npoints, amplitudes,
  File "/Users/liuzzo/beamdyn/at/pyat/at/acceptance/boundary.py", line 401, in boundary_search
    result = grid_boundary_search(ring, planes, npoints, amplitudes,
  File "/Users/liuzzo/beamdyn/at/pyat/at/acceptance/boundary.py", line 271, in grid_boundary_search
    offset, newring = set_ring_orbit(ring, dp, obspt,
  File "/Users/liuzzo/beamdyn/at/pyat/at/acceptance/boundary.py", line 75, in set_ring_orbit
    newring = ring.set_rf_frequency(dp=dp, copy=True)
  File "/Users/liuzzo/beamdyn/at/pyat/at/physics/revolution.py", line 99, in set_rf_frequency
    * ring.harmonic_number
  File "/Users/liuzzo/beamdyn/at/pyat/at/lattice/lattice_object.py", line 397, in harmonic_number
    raise AttributeError('harmonic_number undefined: '
AttributeError: harmonic_number undefined: No cavity found in the lattice


```

swhite2401:
@lcarver , @simoneliuzzo , where do we stand with this one? I would like to merge if nothing is blocking on your side

swhite2401:
> I have updated (git pull, rebuild install) and re-run my test script. I have this strange message now,
> 
> ```
> booster_ring[135]
> ```
> 
> PyDev console: starting. RFCavity('RFC', 1e-12, 9000000.0, 352371625.9975046, 979, 6000000000.0)
> 
> ```
> bf, sf, gf = booster_ring.plot_acceptance(['x', 'y'],
>                                                   [31, 31],
>                                                   [0.05, 0.03],
>                                                   nturns=turns,
>                                                   use_mp=True, start_method='fork')
> 
> 8 cpu found for acceptance calculation
> Multi-process acceptance calculation selected...
> The estimated load for grid mode is 120.125
> The estimated load for recursive mode is 40.0
> GridMode.RECURSIVE is recommended
>   File "/Applications/PyCharm CE.app/Contents/plugins/python-ce/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
>     exec(exp, global_vars, local_vars)
>   File "<string>", line 1, in <module>
>   File "/Users/liuzzo/beamdyn/at/pyat/at/plot/standalone.py", line 80, in plot_acceptance
>     boundary, survived, grid = ring.get_acceptance(*args, **kwargs)
>   File "/Users/liuzzo/beamdyn/at/pyat/at/acceptance/acceptance.py", line 126, in get_acceptance
>     b, s, g = boundary_search(ring, planes, npoints, amplitudes,
>   File "/Users/liuzzo/beamdyn/at/pyat/at/acceptance/boundary.py", line 401, in boundary_search
>     result = grid_boundary_search(ring, planes, npoints, amplitudes,
>   File "/Users/liuzzo/beamdyn/at/pyat/at/acceptance/boundary.py", line 271, in grid_boundary_search
>     offset, newring = set_ring_orbit(ring, dp, obspt,
>   File "/Users/liuzzo/beamdyn/at/pyat/at/acceptance/boundary.py", line 75, in set_ring_orbit
>     newring = ring.set_rf_frequency(dp=dp, copy=True)
>   File "/Users/liuzzo/beamdyn/at/pyat/at/physics/revolution.py", line 99, in set_rf_frequency
>     * ring.harmonic_number
>   File "/Users/liuzzo/beamdyn/at/pyat/at/lattice/lattice_object.py", line 397, in harmonic_number
>     raise AttributeError('harmonic_number undefined: '
> AttributeError: harmonic_number undefined: No cavity found in the lattice
> ```

Sorry I had not seen this one! This is strange but independent I think, can I see you lattice?

simoneliuzzo:
is the same ring that I shared before, no change.

simoneliuzzo:
Also, I would like to run other 3 tests: 1) DA/TL for EBS, 2) DA/TL with errors in the lattice, 3) comparison of the 2 before with the results in matlab AT. 

It will take some time (end of next week?), as now we have Machine time.

simoneliuzzo:
I would also appreciate if @lcarver and @carmignani try the software as well, before it is merged.

lcarver:
So maybe I am doing something stupid, but I want to take a look at the longitudinal acceptance for a ring with radiation on, and I hope to recover the nice bucket profile that I would expect. I run the following:

```
import at

ring = at.load_mat('/machfs/carver/Repo/EBSlattices/AT/S28F.mat', key='LOW_EMIT_RING_INJ')

npoints = [15,15]
amplitudes = [-0.9, 1]
ring.radiation_on()
bg,sg,gg = ring.get_acceptance(['ct','dp'], npoints, amplitudes, grid_mode=at.GridMode.CARTESIAN, use_mp=True, nturns=20*1024, bounds=((-0.45,0.45),(-10, 10)))
plt.figure()
plt.title('Radiations ON')
plt.plot(*bg, label='Cartesian')
plt.plot(*gg,'.', label='Grid')
plt.plot(*sg,'.', label='Survived')
plt.legend()
plt.show()

```

and I get out the following:

image

I am trying to increase the range of the y-scale by playing with the amplitudes[1] and the bounds but it always returns with the same y scale. How can I change it?

pyat/at/acceptance/touschek.py
```diff
@@ -0,0 +1,215 @@
...
+                        GridMode.RADIAL/RECURSIVE: ((0,1),(pi,0))
+        grid_mode       mode for the gird default GridMode.RADIAL
+        use_mp=False    Use python multiprocessing (patpass, default use
+                        lattice_pass). In case multi-processing is not
+                        enabled GridMode is forced to
+                        RECURSIVE (most efficient in single core)
+        verbose=True    Print out some inform
+        epsabs, epsrel  integral absolute and relative tolerances
+
+    OUTPUT
```
lcarver, line 125: the output of the help is wrong

lcarver:
I am trying to do a simple easy calculation of the Touschek LT using the function `get_lifetime`.

First I tried to run this, on grappa with MPI build of pyat.

`bg,sg,gg = at.get_lifetime(ring, 10e-12, 10e-3, use_mp=True)`
but it took an enormous amount of time (>2 hours) so I abandoned.
I tried again with
`bg,sg,gg = at.get_lifetime(ring, 10e-12, 10e-3, use_mp=True, nturns=10)`
and it was again slow to compute anything. So I looked into the code and found:
```

    if refpts is None:
        refpts = range(len(ring))
    else:
        refpts = ring.uint32_refpts(refpts)

```
ok, no problem. If refpts is undefined, it computes the momentum acceptance for every element. So I I try to just do something quick to see it working, I want to take refpt of 0

```
bg,sg,gg = at.get_lifetime(ring, 10e-12, 10e-3, use_mp=True, nturns=10, refpts=[0])

>>> bg,sg,gg = at.get_lifetime(ring, 10e-12, 10e-3, use_mp=True, nturns=10, refpts=[0])
  File "/machfs/carver/pyat_dev/at/pyat/at/acceptance/touschek.py", line 191, in get_lifetime
    dpp = ma[:, i]
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed

```
Looking at the code I am a bit confused by this treatment of momap and refpts.

pyat/at/acceptance/touschek.py
```diff
@@ -0,0 +1,215 @@
...
+        espread = ep.sigma_e
+    if zn is None and bunch_curr is None:
+        bl = bl0
+    elif zn is None or bunch_curr is None:
+        raise AtError('Please provide both current and Z/n for bunch '
+                      'length calculation')
+    else:
+        vrf = ring.get_rf_voltage()
+        h = ring.harmonic_number
+        etac = ring.get_slip_factor()
```
lcarver, line 46: If I give a ring with radiation on, and a bunch current, and a zn to get_lifetime, i return an error here because ring.get_slip_factor needs rad off.

lcarver: replace with         `etac = at.get_slip_factor(ring.radiation_off(copy=True))`

simoneliuzzo:
Here the few tests I run

Screenshot 2022-03-26 at 15 43 41

where the lattice with errors on quad and sext looks like this:
Screenshot 2022-03-26 at 15 43 50
Screenshot 2022-03-26 at 15 44 02

simoneliuzzo (approved):

swhite2401:
> EDIT: So it's clear now. I can not use refpts with len = 1 or len = 2. It has to be more. When it runs the line:
> ` ma, rp = momap[:-1], refpts[:-1]` it reduces a list of len 2 into one of len 1, and then fails at ` dpp = ma[:, i]`

I have added an assertion with error message to clearly require this condition

lcarver (approved):
fine for me (once it passes all the checks)

swhite2401:
@lfarv I had to exclude again the python 3.7 windows test but I saw it was commented on the master...how do you want to handle this?

lfarv:
@swhite2401 : the master branch has modifications in at.c and some .py files to allow the tests again.
I'm not happy with these since apparently they are linked to specific problems with the Github Windows runners which may change without notice (this happened last week).

For the moment, take the `python-test.yml` form master (or revert to the previous version): the tests will fail, but you will not disturb the master.

In the future, I would like to remove the workarounds in the pyat code a keep a clean version. With that, the test failed for Windows/3.7 in the old Github configuration, and would fail for Windows/3.6 in the present one (so exclusion in the test sequence). But the AT code would be clean.

swhite2401:
> For the moment, take the `python-test.yml` form master (or revert to the previous version): the tests will fail, but you will not disturb the master.

Ok restore

lfarv:
All the tests are now running ? very strange… Apparently GitHub changed again. And maybe corrected the bugs. I'll look at that. Anyway it looks ok for merging.

lfarv:
@swhite2401 : now I understand: you merged the changes in master, which allows the tests to run. But the problem is still there, like for instance in the `wake_factory_method` branch…

swhite2401:
Final verification regarding lifetime calculation for @simoneliuzzo .
I used the ESRF lattice and compute the momentum aperture over 2 cells with 248 turns, this gives:

image

Then I compute the lifetime using the default method in matlab, only one method is provided in python, results are:

Python MA + integral: 36.41h
Matlab MA + python integral : 38.51h
Matlab MA + Matlab integral : 38.51h

For me this is good, small difference in MA can be explained by difference in the algorithm and default tolerances.

**To be noted:**
In python I had originally implemented the lifetime as the weighted sum or each element where the weight was defined as the distance to the next element
However in matlab the weight is the length of the element.

I have corrected the python to match the matlab but I am not convinced this is correct, naively I would have thought that the distance between elements makes more sense as it implicitly assume constant MA between elements. Adding elements would refine this approximation.
On the other hand using the length of the element does not seem very logical to me...

@simoneliuzzo , @carmignani , @lfarv you insight would be much appreciated on this.

Besides that I think all functionalities are validated and this is ready for the merge.

swhite2401:
> @swhite2401 : now I understand: you merged the changes in master, which allows the tests to run. But the problem is still there, like for instance in the `wake_factory_method` branch…

Ah yes you are right! I made a mess at some point...sorry for the confusion

lfarv:
@swhite2401 
```
    #spos = numpy.diff(ring.get_s_pos(refpts))
    spos = numpy.array([e.Length for e in ring[refpts]])
```
These are equivalent if `refpts` points to all the elements. If not, I agree that your 1<sup>st</sup> idea is correct (specially if you select only 0-length elements like markers!).

However, the MA and beam dimensions are computed at the entrance of the selected element, and the values are applied over the following length. Wouldn't it be better to compute the weight (or length) as going from half-way from the previous element to half-way to the new element?

swhite2401:
> @swhite2401
> 
> ```
>     #spos = numpy.diff(ring.get_s_pos(refpts))
>     spos = numpy.array([e.Length for e in ring[refpts]])
> ```
> 
> These are equivalent if `refpts` points to all the elements. If not, I agree that your 1st idea is correct (specially if you select only 0-length elements like markers!).
> 
> However, the MA and beam dimensions are computed at the entrance of the selected element, and the values are applied over the following length. Wouldn't it be better to compute the weight (or length) as going from half-way from the previous element to half-way to the new element?

Yes you are right, this would even be better! (selection of zero length elements is properly discarded in matlab, I have not done it because it was not necessary with my initial idea...)

So how would you like to proceed? This has consequences for the matlab that everyone is using...and from the test I have done results in underestimated lifetime. This certainly goes beyond this PR.
I can propose to merge the python with the correct version, and then we may proposed this bugfix for matlab in a separate PR?
What do you think?

lfarv:
@swhite2401 : 

- ok for reverting python to the "correct" version and merging,
- ok for a bug fix in Matlab (with the usual consequences that "the results will change"). Who volunteers?


Concerning the possible improvement, up to you to see if it's worth going to that in another PR.

swhite2401:
I am clearly more comfortable with python, but I can give it a try for matlab (on a separate PR)
Concerning the improvement, I have to think about boundary conditions... but I would like this to be the final version.

swhite2401:
Ok so discussing with @carmignani we thought of a better solution that is to use the optics calculated at each element + interpolated momentum aperture (MA calculation is the time consuming part) and distance between interpolation points.

This is what we think would be the less approximated solution. I will make a proposal for the python in this direction.

PR #358, opened 2022-01-30
Fix bug in CavityPass.c

lfarv:
Fixes a bug preventing the compilation of `CavityPass.c` (regression in #348)


swhite2401:
Sorry about this, I have to remember to compile the matlab as well before merging.

PR #361, opened 2022-02-04
Haissinski solver for short range wake

lcarver:
Short range wake solver. Example file using esrf.m from machine_data is also included.

One thing that needs to be done: The paper the equations are based on use -s notation for the wake. This is taken into account for the input but not the output. Needs to be done cleanly so as to avoid issues.

Also if you want a distribution for a high current, it may not converge from a gaussian as it may be too different. I need to add the ability to solve the distribution in current steps.

### DIFF
--- a/pyat/at/collective/Haissinski.py
+++ b/pyat/at/collective/Haissinski.py
@@ -55,9 +55,9 @@ class Haissinski(object):
 
 
         #negative s to be consistent with paper and negative Wz
-        s = wake_element.WakeT
+        s = wake_element._wakeT
         self.ds = numpy.diff(s)[0]/self.sigma_l
-        self.wtot_fun = interp1d(-s/self.sigma_l, -wake_element.WakeZ) 
+        self.wtot_fun = interp1d(-s/self.sigma_l, -wake_element._wakeZ) 
 
         self.s = -s/self.sigma_l
 
--- a/pyat/at/collective/Haissinski.py
+++ b/pyat/at/collective/Haissinski.py
@@ -211,6 +211,8 @@ class Haissinski(object):
         '''
         self.conv = (numpy.sum(numpy.abs(self.phi_1 - self.phi)))/numpy.sum(numpy.abs(self.phi))
 
+    def set_output(self):
+        self.res = (self.phi_1.copy())[::-1]
 
     def solve(self):
         for it in numpy.arange(self.numIters):
@@ -218,6 +220,7 @@ class Haissinski(object):
             self.dFi_dphij()
             self.compute_new_phi()
             self.convergence()
+            self.set_output()
             print('Iteration: ', it, ', Delta: ', self.conv)
             if self.conv < self.eps:
                 print('Converged')
--- a/pyat/at/collective/Haissinski.py
+++ b/pyat/at/collective/Haissinski.py
@@ -29,7 +29,6 @@ class Haissinski(object):
     eps is the convergence criteria.
 
     Future developments of this class:
-        When a high current is provided, it is better to solve for intermediate currents and use this solution as start for higher current otherwise it may not converge
         Adding LR wake or harmonic cavity as done at SOLEIL. Needs to be added WITH this class which is just for short range wake.
     '''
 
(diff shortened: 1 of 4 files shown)


lcarver:
Can now be reviewed.

lfarv:
Sorry, @lcarver, I thought that since the wake components in the wake element were just a redistribution of the components of the wake object, they could be considered as "private". if needed, they could be retrieved from the wake object itself. But it appears that they are needed from the wake element, so you should make them public again. Or safer, add properties to extract them (read-only). Adding properties avoid problems with setting the wakes, and does not require modifying the C passmethod!

pyat/at/collective/Haissinski.py
```diff
@@ -0,0 +1,257 @@
...
+    def __init__(self, wake_element, ring, m=12, kmax=1, current=1e-4, numIters = 10, eps = 1e-10):
+
+        self.circumference = ring.circumference
+        self.energy = ring.energy
+
+        envelpars = envelope_parameters(ring.radiation_on(copy=True))
+        self.f_s = envelpars.f_s
+        self.nu_s = self.f_s / (clight/self.circumference)
+        self.sigma_e = envelpars.sigma_e * self.energy
+        self.sigma_l = envelpars.sigma_l
```
lfarv, line 46: `f_s`, `sigma_e` and `sigma_l` are available in radiation_parameters (analytically computed), which does not need to turn radiation on.

pyat/at/collective/Haissinski.py
```diff
@@ -0,0 +1,257 @@
...
+
+        self.eta = get_mcf(ring.radiation_off(copy=True))
+        self.ga = self.energy/e_mass
+        self.betrel = numpy.sqrt(1.0-1.0/self.ga/self.ga)
```
lfarv, line 50:
```python
self.eta = ring.radiation_off(copy=True).slip_factor
self.ga = ring.gamma
self.betrel = ring.beta
```

lfarv:
@lcarver: Why do you use for input a lattice element (wake_elem)? This objet is not related at all to tracking, so the logical input should be a wake object rather that a lattice element, don't you think?
(that does not kill the idea of adding properties to the wake element, but it's not so crucial…).

swhite2401:
@lcarver, I tend to agree with @lfarv `WakeElement `is meant for tracking in an AT structure, in case you do not intend to track `wake_object `should be used

I will nevertheless need to be able to modify the wake components on the fly (tuning loop of cavity with beam loading -> resonator frequency needs to change) so I will integrate these properties as suggested in a future PR

lcarver:
I understand your points but I was trying to be more efficient by using the LongResonatorElement function which does everything for me in few lines. I agree that the Element is only used for tracking, which this is not doing, so it can be changed. But I wonder if it should be made to work with either wake_object and wake_element.

lfarv:
@lcarver : if you want one-line functions to generate simple Wake objects, you can create factory functions, implemented as static methods of the Wake object, for instance: `Wake.LongResonator(srange, frequency, qfactor, rshunt, **kwargs) ` will return the desired Wake object. Much cleaner than relying on a wake element! And these factory functions can also be used to initialise wake elements.

swhite2401:
@lcarver, I agree with @lfarv, it seems more logical to have the specific constructors in the wake_object module. I think we have seen a number of potential improvements, maybe we can treat these in a new PR?

lfarv:
@lcarver and @swhite2401 : I think that factory functions are more appropriate than a too sophisticated constructor. Now or later, it's as you like!

lcarver:
So how should I proceed?

I propose that we merge this PR as it is as there are no issues with the Haissinski part of the code, and I will make a new PR with the modifications to the wake object (for the static functions) and also to allow safe access of the wake components of the wake object.

lfarv:
@lcarver: for me, ok as you propose.

lcarver:
As a comment, the f_s computed from radiation_parameters is slightly different (lower by 5Hz) the f_s computed from envelope_parameters.

Issue #360, opened 2022-02-04
Memory allocation error when changing Nslice in WakefieldPass


lcarver:
When defining a wake element, a value for Nslice is given. For example
```
from at.collective.wake_elements import LongResonatorElement
freq=10e9; qfactor=1; Rs=1e4;
welem = LongResonatorElement('wake', ring, srange, freq, qfactor, Rs, Nslice=100)
```
If the value of Nslice is then changed i.e.
```
welem.Nslice=300
```
Then when you are tracking it will return a memory allocation error and give a segmentation fault. This is because the buffers 'TurnHistory*' are defined for the original value of Nslice and not the new one. 

Possible fixes: 
When Nslices is changed then the buffers must be updated. This will have a knock on effect to some other parameters too.
or Error catching and informative error messages.

lfarv:
Corrected in #362

PR #362, opened 2022-02-05
Fix WakeElement

lfarv:
Quick fix for #360.
The WakeElement has some dangerous features: most of its attributes cannot be modified without introducing troubles. So:
- `Nslice` and `Nturns` are now properties, so that their modification rebuilds the `TurnHistory` buffer,
- most of the other attributes are marked as "private" by renaming them with a leading "_". The only "public" ones are now
  - `Intensity`
  - `Nslice` (property)
  - `Nturns` (property)
  - `Circumference` (can probably be removed)
  - `NormFact`
  - `Wakefact` (can probably be removed)
  - `Zcuts`
- WakefieldPass is modified to take into account the renamed attributes,
- attributes are now type-checked,
- as leading underscores are not allowed in Matlab field names, a conversion is set so that the corresponding Matlab field names use a trailing underscore.

However, there is a more deep design problem: having `ring` as argument of the constructor is a bad thing (normally you need the elements first to build the ring…). And `ring` is not necessary: the integrator has the circumference available in the common parameters. I'll look for a better version in another PR.


swhite2401:
@lfarv, thanks for cleaning this up, I had a quick look and could not think of a better solution, maybe we could think of addition a consistency check in the pass method to avoid segmentation faults?

Also, I think it would be better to use `ring.beta` instead of recomputing it all the time.

One comment thought, for these "special" elements that do not represent any physical hardware `ring` is very often needed to initialize some parameters. Beam loading, IBS scattering, for example would require ring as input because you need to know the optics, u0 etc...

However, I agree it should be avoided when possible so if you have a good solution it is more than welcome.

lfarv:
By the way, 'beta' is a problem: up to now, the tracking is performed with **beta strictly equal to one**. So the wake should also be computed with beta=1 (until non-relativistic tracking is implemented). So, using `ring.beta` leads to another problem: recently we switched the "default" (and only) particle from "relativistic" to "electron". I now think it was a mistake, because internally we must use tricks to stick to beta=1 for coherence with tracking (like in find_orbit or computation of the revolution frequency, or here with impedances). I am now thinking of reverting to 'relativistic' so that all the computations are consistent, and solving the problem of parameter output by locally "assuming" electrons when no real particle is declared.

swhite2401:
> By the way, 'beta' is a problem: up to now, the tracking is performed with **beta strictly equal to one**. So the wake should also be computed with beta=1 (until non-relativistic tracking is implemented). So, using `ring.beta` leads to another problem: recently we switched the "default" (and only) particle from "relativistic" to "electron". I now think it was a mistake, because internally we must use tricks to stick to beta=1 for coherence with tracking (like in find_orbit or computation of the revolution frequency, or here with impedances). I am now thinking of reverting to 'relativistic' so that all the computations are consistent, and solving the problem of parameter output by locally "assuming" electrons when no real particle is declared.

The main issue for me was ring.gamma (that I started to use in some places) was returning infinity, then of course ring_parameters was returning nans...would you assume electrons in ring.gamma as well in case you revert to "relativitic"?

lfarv:
> The main issue for me was ring.gamma (that I started to use in some places) was returning infinity, then of course ring_parameters was returning nans...would you assume electrons in ring.gamma as well in case you revert to "relativitic"?

In every places where `beta=ring.beta` is commented at the moment (find_orbit, revolution frequency, impedances…), I would put it back again. And in the few places where gamma infinite is a problem (parameters, radiation effects…), I would assume electrons (gamma and beta), because it's what AT silently does now. You cannot get everything consistent at the moment, but I prefer keeping everything related to tracking internally correct (and ready for non-relativistic), and cheat only where necessary.

Even when non-relativistic tracking is implemented, I think the present situation should be maintained by default: tracking electrons with a beta very slightly less that 1 would degrade significantly the performance with no obvious benefit.

swhite2401:
> > The main issue for me was ring.gamma (that I started to use in some places) was returning infinity, then of course ring_parameters was returning nans...would you assume electrons in ring.gamma as well in case you revert to "relativitic"?
> 
> In every places where `beta=ring.beta` is commented at the moment (find_orbit, revolution frequency, impedances…), I would put it back again. And in the few places where gamma infinite is a problem (parameters, radiation effects…), I would assume electrons (gamma and beta), because it's what AT silently does now. You cannot get everything consistent at the moment, but I prefer keeping everything related to tracking internally correct (and ready for non-relativistic), and cheat only where necessary.
> 
> Even when non-relativistic tracking is implemented, I think the present situation should be maintained by default: tracking electrons with a beta very slightly less that 1 would degrade significantly the performance with no obvious benefit.

Wouldn't it be simpler that `ring.beta` returns 1 and `ring.gamma` returns the value assuming electrons? infinite gamma is really a problem not only in AT but also in other codes depending on AT...

swhite2401:
You may even add a properties of you particles as follows:

```
    _known = dict(
        relativistic=dict(rest_energy=0.0, charge=-1.0, relativistic=True),
        electron=dict(rest_energy=e_mass, charge=-1.0, relativistic=False),
        relativistic_electron = dict(rest_energy=e_mass, charge=-1.0, relativistic=True),
        positron=dict(rest_energy=e_mass, charge=1.0, relativistic=False),
        proton=dict(rest_energy=p_mass, charge=1.0, relativistic=False)
    )
```

That would determine whether to force the beta to 1 or not

lfarv:
> Wouldn't it be simpler that ring.beta returns 1 and ring.gamma returns the value assuming electrons

Having together gamma finite and beta=1 cannot be ! You have to chose one or the other!
My idea is to make a different choice depending on what you are computing, but at least keep all the tracking part correct. Which can be done for instance by providing two sets of properties:
- (`ring.gamma`, `ring.beta`) for the main part
- (`ring.electron_gamma`, `ring.electron_beta`) where appropriate

swhite2401:
> > Wouldn't it be simpler that ring.beta returns 1 and ring.gamma returns the value assuming electrons
> 
> Having together gamma finite and beta=1 cannot be ! You have to chose one or the other! My idea is to make a different choice depending on what you are computing, but at least keep all the tracking part correct. Which can be done for instance by providing two sets of properties:
> 
>     * (`ring.gamma`, `ring.beta`) for the main part
> 
>     * (`ring.electron_gamma`, `ring.electron_beta`) where appropriate

Yes I agree but something returning infinity is useless...we very often need the particle gamma so it would be useful to return the correct value maybe introduce `ring.particle.gamma`? `particle` could be anything define in the particle class and `ring.gamma` is as you suggest the ring property that is always ultra-relativistic used in most calculations?

swhite2401:
> Another question: since I was going through this impedance implementation, I looked at the "`srange`" definition: combining short and long range looks smart, and seems to be very well fitted for multibunch effects. But while putting several bunches in `r_in` is perfectly feasible, and even practically workable by reducing the number of particles in each bunch, the wake computation as it is if not suitable: for instance it would slice the whole gap between bunches. On the other hand, long range with single bunch seems to me unrealistic: one can imagine a wake covering several bunches (a few meters), but not hundreds of meters (the circumference). Do the cavity resonators (the longest range) have any effect on single bunch (apart from the short range contribution)? Or do I miss something?
> 
> So for now, I would keep everything as it is, but just set the default `Nturns` value to zero to reduce computation.
> 
> Last comment, related to the example you provided and that I'm using for test: specially for that kind of computation, I urge **not to touch TimeLag**, so that you can see impedance effects.
> 
> Here is a modified [version](https://github.com/atcollab/at/files/8017483/test_resonator-2.pdf) running with radiation excitation and damping: you see the additional delay coming from the real part of the impedance on top of the radiation delay.

For beam loadaing simulation with uniform filling you can reduce the problem to a single 200mA bunch (at ESRF) but you need the multi-turn wake for that, this whole implementation started from there.

The real multi-bunch implementation is on my TODO list, in principle everything is set to add it. As you said the only problem is the slicing: I need to find a smart way to discriminate bunches in rin, any ideas more than welcome!

Of course MPI+cluster is strongly recommended for multi-bunch simulation, in the past I used 1 CPU/bunch, in fact in this configuration the slicing issue is solved by itself.

I do no see at all how a non zero `TimeLag` prevents you from seeing current dependent phase shift...

swhite2401:
Have you tested `Nturns=0`? Are you sure it works? I sort of remember that the lowest allowed value is 1...

swhite2401:
> Have you tested `Nturns=0`? Are you sure it works? I sort of remember that the lowest allowed value is 1...

Looking at `atimplib.c`, it seems in fact that `nturns` has to be one at least, I think I assumed that the trailing slices were part of the first turn, however I realized now that this may not have been the most logical things. We may put the default to 0 but we have to adapt the rest of the code as well in this case.

lfarv:
I thought I used NTurns=0 (called `wturn` in the example), but in fact it was ignored. And you are right, it should not work. It looks like a question of definition, so we keep 1.

lfarv:
I removed the `Circumference` attribute. @swhite2401: can you check the multiturn ? I don't have any example for that.

swhite2401:
@lcarver would you have a meaningfull test case for that? I only have debug scripts

lcarver:
Yes I will fix it and add it to EXAMPLES today

lcarver:
The damping rate from longitudinal resonator is the same for the old wakefield passmethod as with this one.

lfarv:
@swhite2401 and @lcarver

First: I'm worried by two confusing attribute names: `Intensity` and `Current`. `Intensity` is not really an intensity, but a charge (or a number of charges). Shouldn't we change the name?

Second: the docstring of the WakeElement says:
```
            NormFact    (default=[1,1,1]) normalization for the 3 planes,
                        to account for beta function at the observation
                        point for example
```
For me, this does not have to do with the observation point, but rather the location of the wake element: the wake is multiplied by this factor before computing the kick. Also, in the resonator elements, one could indicate the unit for rshunt: &Omega; for longitudinal and &Omega;/m for transverse. Is this right?

swhite2401:
@lfarv, generally the total wake = Sum(beta_i*W_i) it is already weighted by the beta at the location of the impedance element. However to apply the proper kicker at the observation point you need to normalize by 1/beta_obs.
`Normfact` can be used for more general purpose if needed.

Concerning intensity, you are right this is a number of charge (for info this name is inherited from the collider community that always calls this intensity), we may rename it to `NumParticles` or 'NumCharge'. I prefer `NumParticles` because for ions, in case we ever want to implements this it is clearer you just have to modify `wakefact` accounting for Z and A.

Units are correct.

lcarver:
Also be careful. In wakefield_pass, num_particles is used for number of particles in bunch. It should probably be, num_macroparticles and num_particles (or N_particles or N_macroparticles) so that it is clear.

lfarv:
> (for info this name is inherited from the collider community that always calls this intensity)

This would be fine if there were no `Current` defined simultaneously…

So `Intensity` is changed to `NumParticles`. Is it OK for you now? Ready for merging?

Note: `_wakefact` is now "private", and should disappear in a near future when energy and beta are available in the C code

lcarver (approved):
Fine for me.

PR #367, opened 2022-02-13
Default particle

lfarv:
The default particle is set back to 'relativistic', meaning "relativistic electron", and its definition is modified as follows:
- its rest energy is zero,
- its charge is -1,
- its relativistic &beta; is 1,
- its relativistic &gamma;, **as reported by the `Lattice.gamma` property** is the one of an electron.

This is physically inconsistent (see discussion [here](https://github.com/atcollab/at/pull/362#issuecomment-1031689042)), but it describes the historical behaviour of AT.
It allows to generalise the use of the lattice properties `Lattice.gamma` and `Lattice.beta` everywhere, while being consistent with tracking.

In addition, the energy and particle properties are made available to the C integrators, in preparation for non-relativistic tracking. The "global" parameters transmitted to the C integrators are now:
- ring length,
- ring revolution time,
- turn number,
- energy,
- particle rest energy,
- particle charge,
- s coordinate

The new parameters must not be used until the equivalent is made for Matlab!

### DIFF
--- a/atintegrators/attypes.h
+++ b/atintegrators/attypes.h
@@ -12,6 +12,10 @@ struct parameters
   int nturn;
   double RingLength;
   double T0;
+  double s_coord;
+  double energy;
+  double rest_energy;
+  double charge;
 };
 
 #endif /*ATTYPES_H*/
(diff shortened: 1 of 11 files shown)


swhite2401 (approved):
Ok for me

PR #368, opened 2022-02-15
Energy and particle available to C integrators

lfarv:
This is the equivalent for Matlab of #367 for python:
The energy and particle properties (rest energy and charge) are made available to all the C integrators. The list of global parameters is now:
- ring length,
- ring revolution time,
- turn number,
- energy,
- particle rest energy,
- particle charge,
- s coordinate

Since the parameters are now available in both python and Matlab, they can be used from now on in the integrators.
But for the moment, only relativistic tracking is still available (only `RFCavityPass` is involved in that). The behaviour is:
- in python, the default particle is 'relativistic', but if it is modified, the particle properties are transmitted to the integrators (which at the moment ignore it). So a warning is issued when changing the particle type.
- in Matlab, the same warning is issued, <strike>but the particle type is not yet transmitted to the C. So the integrators always see the properties of the 'relativistic' particle. The reason is the following: ring properties in Matlab are stored in the `RingParam` element. If it is absent, gathering the properties takes long, so that `atpass` (`ringpass` and `linepass`) would be too slow. A better solution is still to be found, possibly by providing the properties optionally , when the are necessary…</strike>

#### Update:
A better solution is found for Matlab: if a `RingParam` element is found (this is very fast, 0.2 ms if found, 4 ms if not), its properties are sent to `atpass.c`. Otherwise, nothings sent. For comparison, building the properties from the lattice takes 80 ms).
- a lattice without RingParam therefore assumes 'relativistic': nothing is changed,
- a lattice with a RingParam element is able to set energy and particle, ready for future use.

#### Conclusion:
Nothing new appears with this PR, but it is an important and necessary step before working on non-relativistic tracking. Additionally, having the energy available will also remove the necessity of having "Energy" as an element attribute.
**This branch (and #367) bring significant modifications of  `atpass.c`**, and start minimising the differences between the Matlab and python versions.

Side remark: while working on C code, I saw that the detection of OpenMP presence which was still missing in Matlab. It is now operational:
```Matlab
>> getoption()

ans = 

  atoptions with properties:

             XYStep: 3.0000e-08
             DPStep: 3.0000e-06
     OrbConvergence: 1.0000e-12
         OrbMaxIter: 20
    omp_num_threads: 0
        WarningDp6D: 1
             openmp: 0

>> 
```

### DIFF
--- a/atmat/atmexall.m
+++ b/atmat/atmexall.m
@@ -85,6 +85,7 @@ alloptions=[atoptions varargs];
 % atpass
 cdir=fullfile(atroot,'attrack','');
 compile([alloptions, {passinclude}, LIBDL, ompoptions], fullfile(cdir,'atpass.c'));
+compile([atoptions, ompoptions],fullfile(cdir,'coptions.c'))
 
 [warnmess,warnid]=lastwarn; %#ok<ASGLU>
 if strcmp(warnid,'MATLAB:mex:GccVersion_link')
--- a/atmat/attrack/atpass.c
+++ b/atmat/attrack/atpass.c
@@ -47,26 +47,29 @@ typedef double mxDouble;
 #define POSTHOOK prhs[6]
 #define LHIST prhs[7]
 #define NUMTHREADS prhs[8]
+#define RINGPROPERTIES prhs[9]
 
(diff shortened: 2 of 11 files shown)


lfarv:
The behaviour now the expected one: the ring properties are available in the C integrators, if available.

Issue #356, opened 2022-01-27
Time-out in tests


lfarv:
For a few days, we have been suffering from time-outs in the GitHub tests. From a 1st analysis:
- it happens only for Ubuntu and MacOS, never for Windows,
- it's not always for the same configuration (platform/python version). Re-running the tests fails differently,
- it's always hanging in `test_patpass.py` (parallel python)

So the questions are:
- is it related to the last modification of `patpass` (which succeeded all tests when it was merged)?
- should we take out this test until we understand ? It's penalising all other developments…

swhite2401:
Ok for me to remove it from tests, when it is hanging do you get an error message?

lfarv:
@swhite2401: no error, no message, it's just hanging until the job is killed (after 6 hours !). You can have a look at all the recent tests…

I would also skip the test, but for sure it needs investigations, this makes `patpass` unsafe.

swhite2401:
Ok i'll take lok I agree this is worrying

lfarv:
A temporary workaround is set in #357, but the problem is still there. So I keep the issue open.

Could it be due to some change on GitHub? It started abruptly 2 days ago (25/01/2022, around 12:00), while `patpass` was modified 9 days ago…
Simultaneously, there seems to be a lack of MacOS runners, macOS jobs wait for long before getting an available one…

PR #369, opened 2022-02-15
Test patpass re-enabled + flake8 clean-up

swhite2401:
This PR re-enables the `patpass` tests: 12 commits done without a single failure...the problem from #356 is not understood
Doing this many flake8 errors noted: fixed as many as possible, remaining mostly related to `__init__.py` files


lfarv:
Very surprising ! This test failure happened suddenly, and disappeared as well ?

Perfect for re-enabling the `patpass` test!

On the other hand, 34 modified files is a disaster… I'll have to go through all my on-going modifications to solve manually all conflicts… Ideally one should try to limit modifications in each PR to as few files as possible to limit the risk of conflicts in parallel development.

But ok, it's done, and it's cleaner now…

By the way, about development, I would suggest 2 points (from my experience):
- when merging a branch into master, use the "Squash and merge" option rather than "Create a merge commit" or "Rebase and merge". This makes the history cleaner in my opinion,
- On the other hand, to re-inject the up-to-date master branch into a development branch, better doing a rebase ("Rebase my_branch onto master"). This avoids creating useless commits and keep the list of modified files unchanged on Github. If there are no conflicts, it's the easiest way: it just "replays" all your work on the present master, rather than on where it was when you branched. No change, no commit, and you are up-to-date. The only trick is to do a "Force push" to GitHub after the rebase succeeded locally. If there are many conflicts, it's nightmare… If you do not really need to incorporate to master in your branch, don't do it.

lfarv (approved):
Now, better merging as soon as possible, not to block so many files…

lfarv:
@swhite2401 : It was not so terrible, I'm mostly on Matlab at the moment  ! Go on, it's a nice cleanup

lfarv:
A last warning removed… (invalid escape sequence).

swhite2401:
@lfarv, I think we may take this occasion to remove all warnings and maybe enforce stricter checks (errors) of pep code style, what do you think?

lfarv:
@swhite2401 : I remember that Will Rogers tried to add checks with flake8 on GitHub, but he gave up. I think code editors like PyCharm give the developer enough information to write clean code. For me, it eliminates all flake8 warning, except trailing white space. But there are some cases where I want to deviate from the PEP rules, for instance when the usual name of a physical quantity starts with a capital letter, I like the variable to do so. Can we just ask developers to eliminate all PEP8 warnings? flake8 should be reasonably happy!

Concerning the remaining warnings `(__init__.py`), the best would be to silence them, the code is correct.

For the time being, I would like to merge this one, so the I can go on!

swhite2401:
Ok please merge , we can silence all `__init__.py` warning with `#noqa`, same for too complex functions but this can be part of another PR

PR #371, opened 2022-02-16
Custom elements

### LINKED ISSUE #370: Impossible to import lattice with custom elements in pyAT
Presently, when the class cannot be defined (from attributes or passmethod) it defaults to `at.Marker`
However, when the lattice is "sanitized", `at.Marker `accepts only `'IdentityPass'`.

This makes it impossible to import lattices with custom passmethod for which a class cannot not be defined.

A temporary fix, to avoid users to be blocked, is proposed in the branch `custom_elements` by remove the check on at.Marker which now accepts any passmethod.
Alternatively we may use at.Element as the default when the class cannot be defined and the `passmethod != 'IdentityPass'`

swhite2401:
This branch fixes #370


lfarv (approved):
No problem. We may come back later to review the checks

PR #373, opened 2022-02-19
Raise an Error if dp or dct is specified for closed orbit computation of a 6D lattice

lfarv:
Recent discussions concerning the Matlab version raised a question about a warning;
`In 6D, "dp" and "dct" are ignored`
issued in closed orbit computation of 6D lattices. This warning is similar in the python version, and the question was:

"if dp or dct is ignored, the result is wrong, so an error should be triggered."

I do not consider this as an error: for a 6D lattice, the orbit is unique, entirely determined by the RF frequency and independent of anything else you may specify, so the returned value is correct.

It is like asking for `find_orbit(ring, color=blue)`. The result is not wrong because the orbit is not blue, it's correct ! The color specification makes no sense, and it is ignored, with a warning (or should be…).

However we may consider that accepting any unexpected keyword should not be allowed, so this is a proposal for raising an error instead of a warning.

### DIFF
--- a/atmat/atphysics/Orbit/findorbit.m
+++ b/atmat/atphysics/Orbit/findorbit.m
@@ -34,7 +34,8 @@ function [orbs,orbitin]  = findorbit(ring,varargin)
 if isempty(orbitin)
     if check_radiation(ring)    % Radiation ON: 6D orbit
         if isfinite(dp) || isfinite(dct)
-            warning('AT:linopt','In 6D, "dp" and "dct" are ignored');
+%           warning('AT:linopt','In 6D, "dp" and "dct" are ignored');
+            error('AT:linopt','In 6D, "dp" and "dct" are nor allowed');
         end
         orbitin=xorbit_6(ring,varargs{:});
     elseif isfinite(dct)
--- a/pyat/at/physics/orbit.py
+++ b/pyat/at/physics/orbit.py
@@ -388,7 +388,7 @@ def find_orbit6(ring, refpts=None, orbit=None, dp=None, dct=None,
     See also find_orbit4, find_sync_orbit.
     """
     if not (dp is None and dct is None):
-        raise AtError('In 6D, "dp" and "dct" cannot be specified')
+        raise AtError('In 6D, "dp" and "dct" are not allowed')
     if orbit is None:
         orbit = _orbit6(ring, keep_lattice=keep_lattice, **kwargs)
         keep_lattice = True
--- a/pyat/at/physics/orbit.py
+++ b/pyat/at/physics/orbit.py
@@ -3,7 +3,7 @@ Closed orbit related functions
 """
 import numpy
 from at.constants import clight
-from at.lattice import AtWarning, check_radiation, DConstant
+from at.lattice import AtError, AtWarning, check_radiation, DConstant
 from at.lattice import Lattice, get_s_pos, uint32_refpts
 from at.tracking import lattice_pass
 from at.physics import ELossMethod, get_timelag_fromU0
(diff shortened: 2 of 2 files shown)


carmignani:
Hello @lfarv,

why don't we allow the dp also in findorbit6 and findorbit in 6D?
I don't think it was extremely needed to allow the dp in atlinopt6, but once we accept there we should maybe accept the dp everywhere, don't you think so?

lfarv:
@carmignani: everywhere ? Possibly, but it will take some time to go through all functions… Anyway, you're right, as far as I could see, the only place where `dp` is used  is in finding the closed orbit, and that's where the check is done. `findorbit` is called by most of the optics functions, and if the RF frequency has to be changed, I think that for safety it has to be done before entering the high level optics function. It would be dangerous that the lattice is changed by `findorbit` in the middle of the function. So the wrapper wraps the whole optics function, not `findorbit`. And `findorbit` is (or will be) always called correctly. And I would not like to duplicated the wrapping…

Note that I do not push for the modification, it's based on your comments, and the PR may be closed if it is not necessary.

swhite2401:
> @carmignani: everywhere ? Possibly, but it will take some time to go through all functions… Anyway, you're right, as far as I could see, the only place where `dp` is used is in finding the closed orbit, and that's where the check is done. `findorbit` is called by most of the optics functions, and if the RF frequency has to be changed, I think that for safety it has to be done before entering the high level optics function. It would be dangerous that the lattice is changed by `findorbit` in the middle of the function. So the wrapper wraps the whole optics function, not `findorbit`. And `findorbit` is (or will be) always called correctly. And I would not like to duplicated the wrapping…
> 
> Note that I do not push for the modification, it's based on your comments, and the PR may be closed if it is not necessary.

I agree that find_orbit should not modify the lattice otherwise it becomes close to impossible to handle

swhite2401:
My personal feeling is that it may not be very appropriate to throw errors in `find_orbit()` , the warning is well suited at this level and I would keep it as is. `find_orbit()` is used all over the place and should be "non-blocking".

I understood the problem is mostly in `atplot()`, in which case this function (or other problematic functions) could be throwing errors instead. This has no impact on the rest of the code or existing scripts and is relatively easy to integrate.

simoneliuzzo:
Dear @lfarv @carmignani and @swhite2401 ,

there use to be several functions for orbit, including findorbit4 (4D, 4D + dp) and findorbit6 (6D, no dp)  

If find_orbit replaces findorbit6 I have no objections, a part the fact that I would personally have to replace all the calls to findorbit6 to calls to find_orbit

if find_orbit is supposed to work in any case, then if the input lattice is 4D (no radiation) I would like to be able to input dp or that the an error message tells: "4D lattice: please use findorbit4 for this purpose or use atradon").

I would prefer not to be a reviewer, so I approve this pull request.  

best regards
Simone

simoneliuzzo (approved):

swhite2401:
@liuzzo, all the functions remain available and valid as you describe, `find_orbit()` is choosing automatically for you depending on lattice properties but you do not have to use it or change any of your code

Basically:
RAD OFF : find_orbit() = find_orbit4() (dp input available)
RAD ON :  find_orbit() = find_orbit6() (dp input not available)

You may still call explicitly find_orbit6 or find_orbit4 if you prefer this will not change

lfarv:
Finally, it seems that there is no real interest in replacing the warning by an error.
- find_orbit will stay as it is, and its output is correct.
- The handling of dp (depending on the presence of an RF cavity) should be done in the upper levels

So I close the pull request.

PR #377, opened 2022-02-25
New Matlab function atloadlattice

lfarv:
This PR introduces the `atloadlattice` function mentioned [here](https://github.com/atcollab/at/pull/364#issuecomment-1043234224). This function makes it easier to load lattice from files by combining several actions in one command:
- it handles .mat and .m files (and more in the future),
- it extracts the desired variable from a .mat file which may contain several:
  - if there is a single variable, it takes it,
  - otherwise it looks for 'RING', for compatibility with python,
  - it RING does not exist, the variable name must be provided.
- it creates or updates the `RingParam` element which accelerates the access to ring properties.

Matlab help for `atloadlattice`:
```
>> help atloadlattice
 atloadlattice  load a lattice from a file
 
 LATTICE=atloadlattice(FILEPATH)
    Create an AT lattice (cell array of AT elements) from FILEPATH
 
 LATTICE=atloadlattice(FILEPATH,'matkey',VARIABLE_NAME,...)
    Give the name of a variable containing the lattice in .mat files
    containing several variables. Defaults to 'RING'.
 
 LATTICE=atloadlattice(FILEPATH,KEY,VALUE...)
    (Key,value) pairs are added to the RingProperties of the lattice or
    overload existing ones.
    Standard keys include: FamName, Energy, Periodicity, HarmNumber, Particle
    Custom properties are allowed
 
 Allowed file types:
    .mat    Binary Matlab file. If the file contains several variables,
            the variable name must be specified using the 'matkey' keyword.
 
    .m      Matlab function. The function must output a valid AT structure.
```


swhite2401 (approved):
Looks good to me

simoneliuzzo:
Dear @lfarv ,

I did a few checks. It seems ok.

The default variable name "RING" is in my opinion not the most widely used in the matlab AT community. THERING would be in my opinion a more user friendly choice. 

I loaded also a lattice with mistuned ringparam element (I set manually HarmNumber to 100), and atloadlattice did not modify it (ok). It also did not complain about something being potentially wrong.

The RingParam Length parameter does not seem to be mandatory (IdentityPass) but it is ardscreated for the RingParam element.

best regards
Simone

simoneliuzzo (approved):

lfarv:
@simoneliuzzo: I agree with your remark about the default variable name `RING`. This is the name used in PyAT, because the examples provided in the distribution use this variable name (I don't know where they come from). So I think we have to keep this for compatibility, but we can easily search for another name in addition. `THERING` is reserved for the global variable, but do you have any of a "usual" variable name for keeping lattices? Probably `ring` (small letters) or `lattice` are common ones… We can look for any of these.

lfarv:
In a .mat file, look for a `ring`, `lattice` and `RING` variable in order.

PR #379, opened 2022-02-28
Wake object develop

lcarver:
Added get_srange and get_wake to wake_object to access the private properties.
Added functions LongResonator, TransResonator and ResWall to wake_object so it can easily be generated without needed to use the wake_element. 
Updated Haissinski files to work with wake_object.

### DIFF
--- a/pyat/Examples/CollectiveEffects/LongDistribution.py
+++ b/pyat/Examples/CollectiveEffects/LongDistribution.py
@@ -2,7 +2,7 @@ import numpy
 import at
 import matplotlib.pyplot as plt
 from matplotlib import cm
-from at.collective.wake_object import build_srange
+from at.collective.wake_object import build_srange, Wake
 from at.collective.wake_elements import LongResonatorElement
 from at.collective.Haissinski import Haissinski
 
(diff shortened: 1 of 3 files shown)


swhite2401:
@lcarver, I am not sure I understand the motivation for these additional `get_*` function... you have already properties defined for `s_range` and all the wake components

lcarver:
I misunderstood a previous comment from @lfarv . In fact is in the wake_element that there should be functions to access the private components of the wake. For the wake_object it is fine as it is. Nonetheless, this PR still contains useful additions for the LongResonator and TransResonator functions that are needed for the Haissinski.

lfarv:
@lcarver: The new methods are more or less duplicating the "add" method. I would better understand static methods acting as factory functions: create the wake, fill it and return the wake object. You get your simple wake in one line, and it is still possible add other components later.

lfarv:
Hi all: I'll be away for 3 weeks, starting to-morrow. Back on March 23rd !

swhite2401:
@lfarv, I started working on your proposal on a separate branch. Happy vacations (I am just back)!

PR #383, opened 2022-03-03
Wake factory methods

swhite2401:
This branch modifies the `collective` sub-packages:
-factory methods are provided to build predefined wake objects (resonator, resistive wal, etc...) in a single line
-wake component of a `WakeElement` object are accessible as read-only
-the wakefield of a WakeElement can be redefined using the `rebuilt` function
-For `WakeElement` based on analytical wakefield, it is possible to recompute it on the fly by changing pre-defined parameters (resonator frequency for example)


pyat/at/collective/wake_object.py
```diff
@@ -165,3 +165,49 @@ def reswall(self, wcomp, length, rvac, conduct, beta, yokoya_factor=1):
                                       length, rvac, conduct, beta)
         else:
             raise AtError('Invalid WakeComponent: {}'.format(wcomp))
+
+
+def res_object(srange, wakecomp, frequency, qfactor, rshunt, beta,
```
lcarver, line 170: I like the fact that it allows for multiple resonators. When the WakeComponent is the same for all resonators (for example I make a WakeZ which sums many resonators of different Q, Rs, fr) the interface requires [WakeComponent.Z]*10 which is not so clean. Maybe it can be added that if WakeComponent is not an array, then it is assumed for all resonators. 

```
numRes = 10
p = res_object(srange, [WakeComponent.Z]*numRes, np.arange(1,numRes+1)*1e9, np.ones(numRes)*5000, np.arange(numRes)*1e4, 1, yokoya_factor=1)

```

pyat/at/collective/wake_object.py
```diff
@@ -165,3 +165,49 @@ def reswall(self, wcomp, length, rvac, conduct, beta, yokoya_factor=1):
...
+        yokoya_factor = numpy.broadcast_to(yokoya_factor, (len(wakecomp), )) 
+    except ValueError:
+        raise AtError('Wake object inputs should be either scalars '
+                      'or with shape (len(wakecomp), )')
+    for wc, fr, qf, rs, yk in zip(wakecomp, frequency, qfactor, rshunt, yokoya_factor):
+        wake.add(WakeType.RESONATOR, wc, fr, qf, rs, beta, yokoya_factor=yk)
+    return wake
+
+
+def longres_object(srange, frequency, qfactor, rshunt, beta):
```
lcarver, line 190: We have res which is short for resonator, then reswall which is short for resistive. I propose to change reswall_object to rw_object

lcarver (approved):
Made a few tests with the wake objects. All seems fine to me.  Examples are updated and working. Haissinski still needs to be reorganised but can be done in a separate PR.

pyat/at/collective/wake_elements.py
```diff
@@ -67,6 +73,36 @@ def set_normfactxy(self, ring):
...
+            return None
+        
+    @property
+    def WakeT(self):
+        return self._wakeT
+        
+    @property
+    def WakeZ(self):
+        return self.get_opt_attr('_wakeZ')
+
```
lfarv, line 90:
```python
def WakeZ(self):
         return getattr(self, '_wakeZ', None)
```
The `get_opt_attr` method is inefficient!
Same comment for the similar properties.

lfarv, line 90:
```python
def WakeZ(self):
         return getattr(self, '_wakeZ', None)
```
The `get_opt_attr` method is inefficient!
Same comment for the similar properties.

pyat/at/collective/wake_object.py
```diff
@@ -165,3 +165,49 @@ def reswall(self, wcomp, length, rvac, conduct, beta, yokoya_factor=1):
...
+    """
+    Method to build a longitudinal resonator wake object
+    """
```
lfarv, line 193: I would prefer to implement this as a static method of the `Wake` class rather than an independent function:
```python
@staticmethod
def long_resonator(srange, frequency, qfactor, rshunt, beta):
```
A longitudinal resonator object would be created with:
```python
long_res = Wake.long_resonator(srange, frequency, qfactor, rshunt, beta)
```
This cleans the global namespace and allows better names.
And the same for the `res_object` and `reswall_object`.

lfarv, line 193: I would prefer to implement this as a static method of the `Wake` class rather than an independent function:
```python
@staticmethod
def long_resonator(srange, frequency, qfactor, rshunt, beta):
```
A longitudinal resonator object would be created with:
```python
long_res = Wake.long_resonator(srange, frequency, qfactor, rshunt, beta)
```
This cleans the global namespace and allows better names.
And the same for the `res_object` and `reswall_object`.

lfarv (commented):
To avoid conflicts with the factory static methods, could the methods

- `resample` (line 131)
- `readwakefile` (line  140)
- `resonator` (line 148)
- `reswall` (line 159)

be renamed as private methods ?

pyat/at/collective/wake_elements.py
swhite2401: Can you propose an alternative for optional attributes?

lfarv: Just the line I proposed above !

lfarv: Just the line I proposed above !

pyat/at/collective/wake_object.py
swhite2401: Again I am a bit confused, these methods return a `Wake` instance, does it make sense for a class to initialize itself in a static method? 

Other topic: why is python3.7 test in windows systematically failing?

lfarv: > Again I am a bit confused, these methods return a Wake instance, does it make sense for a class to initialize itself in a static method?

Of course, it makes sense. it's much cleaner ! That's factory methods.

Concerning the tests, the failure is due to a problem on GitHub actions for windows runners. They changed the way python 3.7 is built on windows, so that the file suffix for compiled extension is now `.pyd` instead of `.cp37-win_amd64.pyd`. It's probably a bug on their side, but the way PyAT gets the suffix has to be changed. The modification concerns 2 files:  `pyat/at.c` and `pyat/at/load/utils.py`. I have a working solution in the `turn_number` branch, but since it looks like a GitHub bug, I would rather disable the test on windows/python 3.7. For this, modify the `at/.github/python-tests.yml` file: replace

```
    strategy:
      matrix:
        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
        os: [macos-latest, ubuntu-latest, windows-latest]
```
with
```
    strategy:
      matrix:
        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
        os: [macos-latest, ubuntu-latest, windows-latest]
        exclude:
          - os: windows-latest
            python-version: '3.7'
```

swhite2401: Sorry but I do not see it... maybe you mean a `@classmethod`? this I could understand but a static method no... you have to show me an example of what you have in mind

lfarv: NO, not a class method ! You need a static method ! Just move your factory functions inside the class, without  any change, and prefix them with "@staticmethod". They still behave as a normal functions, but  are encapsulated in the class for cleanliness. For instance, look at the` Lattice.load` static method (last lines of `load/allfiles.py`):

`ring = Lattice.load(filepath)` calls such a factory method (a static method of the Lattice class) to create a lattice object from a file name.

Here you would call
```python
long_res = Wake.long_resonator(srange, frequency, qfactor, rshunt, beta)
```
to create a wake object. With this, you do not introduce a new function in the global namespace, and the "Wake" prefix clearly shows the kind of object you create.

A class method would have a "class" 1<sup>st</sup> argument which is useless for this purpose.

swhite2401: Well a class method would result in exactly the same syntax and cleanliness but also handle inheritance no? So it seems cleaner to me while I do not see any drawback.

Anyway it is just a question of taste, I can implement what you suggest

swhite2401:
Suggestions by @lfarv implemented, examples are running.
@lcarver could validate the new implementation when you have time? You may take a look at the examples for usage.
Thanks!

pyat/at/collective/wake_object.py
swhite2401: @lcarver I understand your point but I would have to think about this one. I am not quite sure this can be easily and cleanly handled

lfarv:
@swhite2401 : A few minor remarks:

- I still have some PEP8 warnings (indentation, line length…) in `wake_object.py` and `wake_functions.py`.
- In `wake_functions.py`, line 5:
  ```python
  from at.lattice.constants import clight
  ```
  It's better to refer to `at.constants` rather than `at.lattice.constants`. Constants are not related to lattice, and have been moved to the top-level (though `at.lattice.constants` still works).
- The function `build_srange`, which is tightly linked to wake objects, could also be a static method of the Wake class. But it's a matter of taste, keep it as it is if you like.

swhite2401:
> @swhite2401 : A few minor remarks:
> 
>     * I still have some PEP8 warnings (indentation, line length…) in `wake_object.py` and `wake_functions.py`.
> 
>     * In `wake_functions.py`, line 5:
>       ```python
>       from at.lattice.constants import clight
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       It's better to refer to `at.constants` rather than `at.lattice.constants`. Constants are not related to lattice, and have been moved to the top-level (though `at.lattice.constants` still works).
> 
>     * The function `build_srange`, which is tightly linked to wake objects, could also be a static method of the Wake class. But it's a matter of taste, keep it as it is if you like.

All done + excluded windows test for python 3.7 as suggested

pyat/at/collective/wake_object.py
swhite2401: @lcarver, so I have thought about it and what I can provide is an optional `nelems=1` argument. All input arguments would therefore have to be broadcastable to `(nelem,)`
This also presents any conflicts with `wake_elements.py` where the default of nelems=1 would be used

swhite2401:
@lcarver when you have time could you look at this one? I would like to merge it

lfarv:
It's ok for me. Again, take `python-tests.yml` from master to solve the conflicts.

swhite2401:
> It's ok for me. Again, take `python-tests.yml` from master to solve the conflicts.

Done

lfarv:
Anything new here ? Still pending ?

swhite2401:
No just waiting for @lcarver approval

lcarver:
so build_srange is now a static method of the Wake class. But in order to initialise the Wake class, I need to provide an srange. So if I am doing multibunch, then I need to create a fake srange, initialise the Wake with wob = Wake(srange_fake) then I have access the build_srange.

lcarver:
I am happy with the rest, just this question of srange that I find confusing.

PR #384, opened 2022-03-07
Beamloading

swhite2401:
This PR introduces a beamloading element that combines a `LongResonatorElemen`t and `RFCavity` elements.
This allows to automatically adjust the cavity and resonator parameters when changing the Current property of the `BeamLoadingElement`.

`An example usage is provided in examples/CollectiveEffects/Beamloading.py`

This PR remains draft until @lcarver has validated it and #383 has been merged as it is a prerequisite.


swhite2401:
#191 could be replaced by this method, however I would suggest to still keep the passmethod available in the distribution as it could be of some interest for some applications, for example usage with multi-processing or paralelization

PR #385, opened 2022-03-07
correct fundamental cavity mask 

swhite2401:
This PR fixes a bug in `cavity_access.py`:
In case all cavity have the same frequency `_fundmask` returns `None`, however is `mask=None` the voltage is applied on all cavities which is wrong because it leads to `ring.rf_voltage = desired_voltage*ncavities`.
The `mask` is now returned in all cases which fixes this issue.


lfarv (approved):
Right !

lfarv:
@swhite2401 : please merge !

PR #393, opened 2022-03-30
Energy loss tracking

swhite2401:
I has been brought to my attention that in some instances particles do not make it through one turn if cavities are off.
In this case `at.ELossMethod.TRACKING` cannot be used.
it is presently the default in `find_orbit()` and is therefore a serious issue.

This PR solves this issue by calculating the energy/element in case the particle is lost in less than a turn. This algorithm can be refined by smart iterative splitting of the ring based on loss index.

This also raises the question whether TRACKING should be the default, especially in find_orbit() for which it is used only to determine the initial guess: I would personally suggest to use INTEGRAL but this can be discussed.

@lfarv what do you think? Any better ideas?

### DIFF
--- a/pyat/at/physics/energy_loss.py
+++ b/pyat/at/physics/energy_loss.py
@@ -70,9 +70,16 @@ def get_energy_loss(ring, method=ELossMethod.INTEGRAL):
                                      sextupole_pass=None,
                                      octupole_pass=None,
                                      copy=True)
-        o0 = numpy.zeros(6)
-        o6 = numpy.squeeze(lattice_pass(ringtmp, o0, refpts=len(ringtmp)))
-        return -o6[4] * ring.energy
+
+        o6 = numpy.squeeze(lattice_pass(ringtmp, numpy.zeros(6), refpts=len(ringtmp)))
+        if numpy.isnan(o6[0]):
+            dp = 0
+            for e in ringtmp:
+                ot = numpy.squeeze(lattice_pass([e], numpy.zeros(6)))
+                dp += -ot[4] * ring.energy
+            return dp
+        else:
+            return -o6[4] * ring.energy
 
     if isinstance(method, str):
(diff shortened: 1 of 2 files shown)


lfarv:
@swhite2401 : the problem with the `INTEGRAL` method in `find_orbit6` is that it will always take all the dipoles/wigglers into account, while you can call `find_orbit6` without dipole radiation (non-radiating machine), or with only some of them (why?). The `TRACKING` method will consider exactly what the user selects. For orbit guess, `TRACKING` is safer because it always matches the conditions for which the orbit is searched.

One could condition the 'integral' contribution of each dipole with its passmethod, but it would break the possibility of computing the energy loss independently of the radiation state of the lattice… It's nice to have all the lattice parameters available whatever the radiation state.

So given your patch, I would leave the defaults as they are. For marginal cases like the one you are mentioning, the user can still change the method: that's why the choice is proposed. Maybe a clear error message if the tracking fails would be helpful.

lfarv:
@swhite2401 : the problem with the `INTEGRAL` method in `find_orbit6` is that it will always take all the dipoles/wigglers into account, while you can call `find_orbit6` without dipole radiation (non-radiating machine), or with only some of them (why?). The `TRACKING` method will consider exactly what the user selects. For orbit guess, `TRACKING` is safer because it always matches the conditions for which the orbit is searched.

One could condition the 'integral' contribution of each dipole with its passmethod, but it would break the possibility of computing the energy loss independently of the radiation state of the lattice… It's nice to have all the lattice parameters available whatever the radiation state.

So given your patch, I would leave the defaults as they are. For marginal cases like the one you are mentioning, the user can still change the method: that's why the choice is proposed. Maybe a clear error message if the tracking fails would be helpful.

swhite2401:
So, this one is ok for merging?

swhite2401:
Ok I merge

PR #394, opened 2022-04-02
Matlab AT release 2.3

lfarv:
**No change in AT**

The PR only includes updates to the Matlab documentation and project structure, to define the upcoming 2.3 release.

The Matlab section of the Web site is also modified in accordance


swhite2401:
@lfarv, do you plan to increment only MATLAB? 

I think python would deserve a new version number as well, in which case is it possible to wait for the final version of the collective effects, and lifetime/DA calculation?
I would like to write some documentation for these as well before we release a new version.

lfarv:
@swhite: up to now, Matlab and python version are not synchronised. I am planning a Matlab version because the change of RF cavity passmethod is an important point, together with the various but fixes, and because there is no on-going changes. The previous release (2.1.003) was in September 2021 (~6 months old), so a new one makes sense.

For python, I am indeed looking at the ongoing developments, mainly from you or @lcarver, so I am waiting for you to finalise the present ones, and give me an indication when you think we could make a new release. The previous release (0.2.2) was in January, a little more that 2 months ago. So no urgency, though PyAT evolves more rapidly ! So tell me when you think it would make sense.

lfarv:
Some tests of the upcoming Matlab version revealed a conflict with the "Simulink Control Design" toolbox (duplicate function name). So rebase is postponed, until this is fixed.

PR #397, opened 2022-04-07
add collimator element

swhite2401:
This PR adds a Collimator element class that inherits from Aperture.

### DIFF
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -211,6 +211,10 @@ class Aperture(Element):
         super(Aperture, self).__init__(family_name, Limits=limits, **kwargs)
 
 
+class Collimator(Aperture):
+    """pyAT aperture element"""
+
+
 class Drift(LongElement):
     """pyAT drift space element"""
 
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -211,10 +211,6 @@ class Aperture(Element):
         super(Aperture, self).__init__(family_name, Limits=limits, **kwargs)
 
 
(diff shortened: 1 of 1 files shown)


lfarv (approved):
No specific property for this element ?

lfarv:
Side note:
It's very easy for any user to derive AT elements for his/her specific goals. You can imagine elements for valves, bellows, girder limits, and so on. This does not imply that these elements need to be integrated in the core of AT.

If such developments are worth sharing, there is the possibility of creating a dedicated repository for "user contributions". Long ago, the "pubtools" directory of AT was intended for that. However its contents have not been properly evaluated, it has a lot of missing references and is largely unusable. If this proves useful, it should be started again from scratch.

swhite2401:
@lfarv this was requested by users, mostly to mimic was is done in MADX and facilitate conversion.
It think it makes sense, a Collimator does interfere with the beam as opposed to the other elements you mention.

I have to look, but thinking about it I think the aperture has to be forced to an `RAperture`, an elliptical aperture makes no sense for a colimator.

Concerning users contribution, I personally do not like it so much... `pubtools` turned out to become just some place to store just anything an in some cases unusable code, but that is my personal taste, I don;t have strong objection if you think this is useful

swhite2401:
Changed collimator to drift with `RApertures `required

lfarv:
> - Aperture allows for elliptical restriction (on choice)

`Aperture` is a rectangular aperture, `EAperture` is an elliptical aperture: an element does one thing, and does it right, as usual !

lfarv (approved):
Simple enough !

PR #400, opened 2022-04-20
Avoid problems with GitHub Windows runners

lfarv:
This is a workaround for a problem with GitHub Windows runners used for tests. The problem is the following:
The file extensions for the python extensions compiled from C are accessible from the python "`sysconfig`" module. This module is imported in 2 locations:
- `at.c`, where the C extensions are dynamically loaded,
- `load/utils.py`, where the presence of the extension is checked when loading a Matlab lattice description.

Unfortunately, the `sysconfig` module returns a wrong value on GitHub for python 3.6 and 3.7 on Windows: `.pyd` instead of  `.cp36-win_amd64.pyd`. Even worse, the file extension for python 3.7 changed twice without notice in the last few weeks, resulting in failing tests. 

Up to now the solution was to introduce a Windows-specific workaround in the 2 files mentioned above, by importing the deprecated `distutils.sysconfig` module instead of `sysconfig`. But this workaround has to be modified each time the GitHub configuration changes…

The proposed solution here is:
1. Remove all Windows-specific workarounds in the PyAT code. `sysconfig` is always used,
2. disable the tests on Windows for python 3.6 (always fails up to now) and python 3.7 (succeeds or fails, it depends on GitHub good will).

Consequences:
- cleaner code
- the compatibility of PyAT with python 3.6 and 3.7 is still checked on Linux and MacOS
- the problem on Windows looks specific to GitHub, and should not harm normal Windows users, though I cannot check,
- the problem is restricted to really old python versions.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -22,13 +22,13 @@ jobs:
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
         os: [macos-latest, ubuntu-latest, windows-latest]
-#       exclude:
-#         - os: windows-latest
-#           python-version: '3.7'
+        exclude:
+          - os: windows-latest
+            python-version: ['3.6', '3.7']
 
     steps:
 
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/python-tests.yml
```diff
@@ -22,13 +22,15 @@ jobs:
...
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
         os: [macos-latest, ubuntu-latest, windows-latest]
-#       exclude:
-#         - os: windows-latest
-#           python-version: '3.7'
+        exclude:
+          - os: windows-latest
+            python-version: '3.6'
+          - os: windows-latest
+            python-version: '3.7'
```
MJGaughran, line 29: Can you combine these values as above? i.e. `python-version: ['3.6', '3.7']`

lfarv: This was my 1st trial ! It does not work, both tests were still active.

Issue #391, opened 2022-03-23
Harmonic number not initialized


swhite2401:
@liuzzo has found that in some cases, even though an RF system is defined, the `harmonic _number` properties of the `Lattice` is not initialized preventing some functionalities of AT to work properly.
The error returned is that the ring as no cavity which is obviously wrong.

In case `frequency` or `harmonic_number` are not provided to the `Lattice` constructor would it be possible to get the `frequency` from the cavity element, and only if this element is not found return an error?

lfarv:
Matlab or python ?
Can you give an example showing the problem?

swhite2401:
@lfarv, this is python, here is a piece of code to make fail:

```
import at
ring = at.load_lattice('../lattice/hmba.mat')
ring = list(ring[1:])
ring = at.Lattice(ring, energy=6.0e9)
print(ring.harmonic_number)
```

lfarv:
@swhite2401 : in the particular case, there is indeed no cavity. So this behaviour is normal. Such a lattice is perfectly suitable for all 4D optics computations, and this is checked by the test sequence by using the `dba.mat` test file which has no cavity.
On the other hand, in the 1<sup>st</sup> comment, you refer to a lattice with a cavity : this is a problem.

When thinking about it, the error message may be misleading in some cases: if the lattice is initialised without cavity (no harmonic number attribute), and if you manually insert one, there is no way to generate the harmonic number, you have to set it explicitly also. In that case, the message is wrong. One could instead make it more general: "harmonic_number is undefined".

lfarv:
The rule to derive harmonic_number is the following (theoretically…):

- if found in a `RingParam` element, take it (only when reading `.m` or `.mat` files)
- otherwise if it is defined in the lowest frequency cavity, take it,
- otherwise use the lowest cavity frequency and ring circumference,
- otherwise it's undefined.

This happens at lattice initialisation. Later on, it has to be defined explicitly.

swhite2401:
Why do you say this ring has no cavity?


```
ring = at.load_lattice('../lattice/hmba.mat')
cavs = ring.get_elements(at.RFCavity)
print(cavs)
```

Gives:
`[RFCavity('RFC', 0.0, 187500.0, 352372212.4670127, 31, 6000000000.0, PassMethod='IdentityPass')]`

swhite2401:
Ah sorry, matlab/python confusion... in fact I remove the cavity

swhite2401:
[...]

PR #401, opened 2022-04-20
Initialisation of the harmonic number

lfarv:
@simoneliuzzo and @swhite2401 pointed out in #391 a problem with the `harmonic_number` Lattice attribute. It is properly initialised when creating a lattice, but not when a lattice is modified: if the attribute does not exist initially (no cavity in the lattice), it is not added if a cavity is added to the lattice. This concerns the following actions:
- `insert`, `append` and `extend` methods,
- `+` and `+=` operators.

This is corrected here, as well as updating properly the `radiation` property in the same events.


swhite2401 (approved):
Ok for me

PR #403, opened 2022-04-29
Update Collective.md

swhite2401:


swhite2401:
@lcarver, I have started this branch for the documentation on collective effects, you can edit directly the `Collective.md` on the github page using the markdown syntax and see a preview.
To rebuild the webpage it has to be merged onto the master branch, we will do this at the end.
Please populate this page when you have time.

swhite2401:
Outdated I closed it

lfarv:
@swhite2401: are you sure you do not want to keep the `Collective.md` file ? It displays nicely with Sphinx, but I don't know if it is valid… If you really want to discard it, please also delete the branch, otherwise keep it for a future update;

swhite2401:
It is valid yes, but is more or less a repetition of what is already in the docstring so I think we can discard it, we will come with something better with formulas etc... for now one can refer to the API doc
I delete the branch

PR #405, opened 2022-05-11
Sphinx docs

swhite2401:
Alternative to doxygen, here for testing -> this is much better and is now fully deployed

The source to the sphinx webpages are located in:
`docs/sphinx/source`
and the build in:
`docs/sphinx/build`

In order to rebuild the API documentation locally you need run:
`sphinx-apidoc -M -f -o ./source ../../pyat/at ../../pyat/at/integrators`

The last argument allow to ignore the compiled integrators.
To build the html page:
`make clean` and then `make html`

The webpage is hosted on [readthedocs.org](readthedocs.org), it can be seen at [https://at.readthedocs.io/en/latest/](https://at.readthedocs.io/en/latest/)

A webhook was added such that the page is automatically rebuilt from source every time a push is made on this branch.
Once this is merged we can move this to the master.
All the link from the present documentation are now pointing to the readthedocs webpage.

**TODO**: modify all docstring to comply to google style (only done for the acceptance package for now)


lfarv:
Hi @swhite2401, I have no way to look at this at the moment, but from my experience, Sphinx is much nicer than doxygen. I'll look when I'm back.

swhite2401:
@ lfarv, this one look much better and modern than doxygen and is very easy to customize.
The API documentation is automatically generated using [sphinx](https://www.sphinx-doc.org/en/master/). The docstrings will need to be reworked in order to be consistent with [google style docstring for sphinx](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
I have edited the `acceptance` package so you can see the final version, I think it is worth the effort. The page is published in [read the docs](https://readthedocs.org/), you can access it directly at [at docs webpage](https://at.readthedocs.io/en/latest/)

Unfortunately the `markdown` files don't come out so nicely but I think this can be fixed, I ll take a look.

You may look at [pytango webpage](https://pytango.readthedocs.io/en/stable/) to see a finished product using the same engines.

swhite2401:
It is incomparable! The only drawback is that markdown are not rendered properly and our `docstrings` will need to be adapted.
So it will take a bit of effort but in my opinion this is really worthwhile and after that everything is automatically generated.

With the user community growing we are really lacking documentation...

lfarv:
Hello @swhite2401. This indeed looks nice ! I have a few questions:

- you choose to store the html docs on a 3rd party site (readthedocs). Is there a good reason not to keep them in Github  Pages?
- In this case, is it necessary to have git manage the `docs/sphinx/build` directory? Apparently it is filled when running sphinx from `docs/sphinx/source`. Or maybe store only  the directory itself but not its contents?
- the `docs/sphinx/p` directory now looks useless, its contents has been moved to `docs/sphinx/source/howto`. Consequently the `python_sidebar.yaml ` file is not used any more, apparently.

But there is another option: keep the hand-written documents (how-tos and general information) in the AT web site where they are, and link to the sphinx-generated module and function docs from the python sidebar.

swhite2401:
Welcome back @lfarv, I found `readthedocs` very convenient as there is basically nothing to do, the page is built automatically on push. `Github` pages looked a bit more complicated to me without obvious advantages. I can give you admin right to the `readthedocs` project.

In fact you are right, using `readthedocs` you do not need the build directory. I have added it while testing (you can look at the page directly opening `index.html` with firefox) but this does not have to be tracked on git.

Yes, sphinx does not like `markdown` files, I had to translate them to `rst` to make them look nice, if we decide to adopt this we can clean up everything

I would rather have a single page for everything, already I realized resources are spread all over the place which I find difficult to follow.

lfarv:
@swhite2401: I just made a test using only GitHub Pages (without readthedocs): it as easy, with an identical result, but no ads. The only difference is that you need to do a "make html" in the `docs/sphinx` directory to generate the sphinx documentation. Then you commit and Jekyll does everything as usual.

The only thing I changed to do the test is the Jekyll `_config.yaml` file !

Coming back to the hand-written documentation, I compared side-by-side the "GitHub Pages" and "readthedocs" versions of the "Cavity control" page: Jekyll has nicer layout features (coloured boxes, in-line maths, and much more that I did not explore). So for these, I would stick to the markdown version.

swhite2401:
Ah good, so if you found a good way that is nice, did you deploy it? I tried and ended up on the `readthedocs` page.

swhite2401:
btw feel free to update this PR as you want, I just set this up for testing

lfarv:
> So I haven't looked at details myself but quickly browsing through the web, people seem to think rst has more capabilities that markdown... maybe my quick translation of the `.md` files was not good? (I used an online translator)

In fact, the features I was referring to are Jekyll features (more precisely what is called "liquid"), rather then markdown features.

I'll prepare a test so that you can look !

lfarv:
I created a temporary new branch `sphinx-doc-jekyll` derived from this one. Only 3 files are modified:
- `docs/_data/sidebars/python_sidebar.yml`
- `docs/_data/topnav.yml`
- `docs/_config.yml`

Then when running "make html"  in `docs/sphinx`, the `sphinx/build `directory is regenerated (another set of modified files).

Finally I pushed the branch on GitHub and switched the pages to this new branch. That's all. 
The result is very close the the previous one.

In this configuration, the AT Web site is entirely hosted on GitHub Pages, but nothing changed for the behaviour on "readthedocs": it should still produce the same web site!

Concerning the comparison of Jekyll .md files and sphinx .rst files, after looking around different sites, I agree that sphinx can also produce about anything. In this setup, we have the choice.

Look at [https://atcollab.github.io/at/]( https://atcollab.github.io/at/) and check that everything is there. If you are happy with it, I'll introduce my modifications in this branch.

You did a nice job in structuring the sphinx generation. The scheme I propose can still be improved by rearranging the source and destination directories in `at/docs`, and in refining both the Jekyll and sphinx themes, but I think it's already worth merging.

swhite2401:
Very nice! I think we can get rid of the `readthedocs` page, I do not see it as very useful with the solution you propose.

On the other hand I like that everything is hosted on the same page as it is also easier to maintain if we have a single source. What I would propose is:

- keep all the matlab docs as they are now
- host the python doc on the sphinx webpage, make sure all links point to this page (welcome github page, top navigation bar, etc...)

What do you think?

Concerning the merge, if we decide to adopt the sphinx page, I think we have to make the effort to adapt all `docstrings` before we do it to start from a clean situation. I did it for the `acceptance` package and started with `collective`.

I can continue doing it once you have committed your changes.

lfarv:
1. I agree, it's better to keep the whole python documentation in a single system (sphinx),
2. For now, we keep the top level and the Matlab doc as they are (Jekyll markdown). In the medium term, it may be possible to move Matlab to sphinx also, but it's another job,
3. About merging, I would not wait for all the docstrings to be updated:
   - the modification of too many files in a single pull request is dangerous and make the review very painful,
   - as it is, it's better than nothing: we don't remove anything, but just add new chapters. The present function documentation is not very nice, but it's understandable. Changing the theme may also make it better,
   - there are still a few points I would like to look at, so that we could propose a style guide.
   
   So I would propose to:
   - starting from now, adapt all **new** or **modified** modules to ReST (your "acceptance" example is very valuables for that),
   - modify other modules step-by-step

I'll commit my modifications, put back the top-level python doc in sphinx (trying to put it at the same level as the Jekyll one) and propose a clearer theme. Then we may decide about merging.

lfarv:
I modified the directory structure to clean the` at/docs` directory, used by GitHub pages. The sphinx source directory taken out and is now  `at/docsrc`.

The whole python branch of the Web site is now handled by sphinx.

The Web site is back to the sphinx-doc branch for control.

swhite2401:
Ok sounds good to me! Thanks for helping and cleaning. You may remove the `readthedocs` config file too (.readthedocs.yaml)!

lfarv:
I modified `at/readthedocs.yaml` and `at/docsrc/requirements.txt` according to the new structure. Normally, readthedocs should still work. Does it still work? Should we remove it now, or keep it for a while(it's only 2 same files)?

lfarv:
The sphinx-book-theme looks much better! It's highly configurable, bit the default behaviour looks excellent to me.

swhite2401:
In fact I disagree the look and feel of the RDT template looked better to me, especially the API docs with colored boxes etc

lfarv:
@swhite2401 : Ha, colors… 
I find the "rtd" theme really old-fashioned. The "book" theme has now been adopted by Matplotlib, numpy, scripy… It looks much cleaner to me and is likely to become a standard. But I switched back to "rtd" for the moment and let you think again about it !

Otherwise I'm still experiencing a lot:
- I tried the complete `autosummary` solution: a single `index.rst` file is enough for the whole site. To avoid breaking your structure, I derived a new `sphinx-docs-new` branch, which is currently displayed on the official site. This solution has many advantages, in particular a "summary" section at the beginning of each page. And full automation, of course. But it has a severe drawback: if a module has the same name as its containing package (or a function the same name as its containing module), sphinx gets lost. This is perfectly correct in python, but has indeed a problem: if you refer to` at.matching`, it may mean either the `at.matching` package, or the `at.matching.matching` module. In fact, the module name hides the package name, so "autosummary" ignores the package and documents only the module. I'm looking for a solution avoiding renaming modules (which is still possible since module names are considered private),
- I tried a minimal CSS customisation (font sizes: I don't like huge fonts). Some themes are easier for that (better structured), but it's easy enough. 
- I would still keep open for the moment the possibility of using "readthedocs". Could you try to apply it to the sphinx-docs-new branch?
- For evaluating the task of "adapting" the existing docstrings, I revisited entirely the `lattice` package, including introducing python type hints. It's indeed a considerable job. We should not wait for everything before merging…

Work in progress… Thanks for feedback!

swhite2401:
Ah yes I saw that, in fact this is a problem: `at.acceptance.acceptance` is showing as `at.acceptance.accept` which is quite bad...

Well maybe you are right, I just like the color boxes around the function instead of just plain text, I thought it was more readable, maybe we can play with font size and color as they did for matplotlib instead. 
So ok for your proposal, since all major python scientific API use it we can adopt it, no objection!

'readthedocs' needs the rdt-scheme I think so it is a complication if we adopt another one, why would you like to keep it?

Lots of work indeed...I agree we can merge and adapt as we go, I am very busy with IPAC papers these day unfortunately, once this is out of the way I'll have more time to dedicate to this.

lfarv:
> 'readthedocs' needs the rdt-scheme I think

Are you sure ? I added `sphinx-book-theme` to the `docsrc/requirements.txt` file, so I do not see why it would not work. But since I reorganised the directory structure, I'm not sure I did not forget something in `.readthedocs.yaml`.
The reason I'd like to try is that if at some point we could also move the Matlab doc to sphinx (there is an extension for that, to be tried), we could move the top-level of the site to sphinx. And then it's more complicated for GitHub Pages than just handling a subset. I read that it's possible, but I didn't understand how. This is hypothetic future, but if we can keep the compatibility with readthedocs for almost free, why not?

If you tell me how to run it myself, I'll do it! Don't waste your time on that!

Concerning the "book" theme, I do think we have a lot of freedom for fonts and colors. I saw also on matplotlib a possibility to link to source code which looks very interesting ! To be explored

swhite2401:
By the way it automatically builds on push

swhite2401:
Looks good to me! There are small differences between the github page and the readthedocs pages: no scroll down menu on the side bar, no sub-packages on the front pages, etc...
Basically readthedocs has forced the rtd theme but with the colors of the sphink-book it seems.

Anyway, this is good enough for me, we can probably merge this one and then adapt the docstrings progressively, maybe one PR per subpackage?

lfarv:
> There are small differences between the github page and the readthedocs pages

Can you check again ? Did you look at the right version of readthedocs ? I cannot see any difference !

Anyway, I created a new branch and pull request #408 , to start with a clean repository. When it's merged, I agree with your schedule of one PR per sub package.

swhite2401:
> > There are small differences between the github page and the readthedocs pages
> 
> Can you check again ? Did you look at the right version of readthedocs ? I cannot see any difference !
> 
> Anyway, I created a new branch and pull request #408 , to start with a clean repository. When it's merged, I agree with your schedule of one PR per sub package.

Yes here is the RTD webpage:
image

And here is the github webpage
image

lfarv:
Hi @swhite2401, you are looking at the "latest" version on ReadTheDocs ! You have to switch to the "sphinx-documentation" version, either in the URL ([https://at.readthedocs.io/en/sphinx-documentation/](https://at.readthedocs.io/en/sphinx-documentation/)), or using the little menu at the very bottom of the sidebar !

swhite2401:
Right all god for me then!

swhite2401:
Properly implemented in #413 , I closed this one too

PR #408, opened 2022-06-04
Sphinx documentation

lfarv:
This PR replaces and updates #405. It adds a `docsrc` directory containing the source of the python documentation handled by [sphinx](https://www.sphinx-doc.org/en/master/index.html).

This directory contains ReStructeredText or Markdown files used as sources for the python part of the AT website. The top-level and Matlab parts are unchanged.

**The sphinx  documentation must be compiled before pushing to GitHub**, so that the web site is updated. This operation converts ReST or markdown pages to html and extracts the docstrings from python objects to automatically generate the API documentation. The compilation process is explained in the `docsrc/README.rst` document. This manual operation is necessary:

- after adding or modifying  pages,
- to include the new or modified docstrings of the python source code.

A clone of the python documentation is also installed on [readthedocs.org](https://at.readthedocs.io/en/sphinx-documentation/). This one is automatically compiled on each push to GitHub (nothing to do).

For now, both GitHub Pages and readthedocs.org are based of this `sphinx-documentation` branch. After validation, they will of course be directed to the `master` branch.


lfarv:
- It seems that there is a way to automate the sphinx compilation using GitHub actions. We can keep this in mind for a future improvement,
- Having the top-level handled by sphinx is possible, if we need it,
- I tested the generation of Matlab documentation with sphinx. The problem is that it expects either a ReST syntax (then the internal Matlab help becomes unreadable), or a "Google-style" syntax, closer to Matlab-style doc, but which needs rewriting a lot of files… Not straightforward

MJGaughran:
For one of our DLS repositories, it appears that code compilation isn't needed for our Sphinx generation. You can see our current GitHub Actions here; you may find it useful in generating your own:
https://github.com/dls-controls/pythonSoftIOC/blob/1eb9405cbd0c659ba8b38b15a992a3fb4aabe871/.github/workflows/docs.yml

I'm happy for you to hold off on adding GitHub Actions for this pull request, but I think it's more than just a nice-to-have.

Can I ask why you are releasing documentation to both GitHub Pages and readthedocs? We are moving away from readthedocs at Diamond as it's easier to use the single login. We have no repositories which push to both, and it could cause confusion as to which is the 'primary' source of information on the repository.

lfarv:
@MJGaughran

> Can I ask why you are releasing documentation to both GitHub Pages and readthedocs? 

I agree with you to keep GitHub Pages as the documentation root (see https://github.com/atcollab/at/pull/405#issuecomment-1130483778). So you can forget readthedocs, all links still point to GitHub. But it's easy to keep the configuration ready in case we need it at dome time.

Concerning GitHub actions, I'm not expert enough to get it working rapidly, and I would prefer merging this PR first, so that we have at least a preliminary doc for the python API. Looking at your example, apparently it has a different setup, by putting the documentation in a separate branch `gh_pages` while we use the `/docs` directory of the `master` branch. If you see a way to adapt this example, that will be a nice improvement.

swhite2401 (approved):
Ok for me

MJGaughran:
I am still keen to support only GitHub Pages or readthedocs. I'm not sure being indecisive on that point will provide any benefit, but I'm happy to leave that for a second merge request.

Issue #411, opened 2022-06-08
GitHub action for running Sphinx


lfarv:
The AT documentation on GitHub Pages is now relying on [Sphinx](https://www.sphinx-doc.org/). As it is now, running sphinx to extract the documentation from the python code, or from  hand-written ReST documents, has to be done manually. Though this is rather simple, it would be safer to have this step automatically run by GitHub actions. @MJGaughran mentioned that this is possible (https://github.com/atcollab/at/pull/408#issuecomment-1147594393), but it does not look so simple:

in our configuration, the documentation source is in the `/docs` directory of the `master` branch. Ideally, a push to this branch should trigger an action updating the `/docs` directory. But this result must be pushed again so that GitHub pages are updated. This looks like an endless loop…

Unless I miss something, we are probably obliged to change the configuration and host the documentation source in a another branch (typically `gh_pages`). And then setup a GitHub action to run the whole process. Any expert (@MJGaughran ?) is welcome to open a PR to experiment this!

lfarv:
Solved in #413

PR #412, opened 2022-06-13
Remove Jekyll

lfarv:
In this PR, the (limited) Matlab documentation is now handled by Sphinx instead of Jekyll. No change in the documentation itself, except for style. This allows to get rid of Jekyll, the last step before using GitHub actions to automate the Sphinx compilation (in a next PR).

Some comments about style:
- the new colors come from an update of the `pydata-sphinx-theme` used for AT. This is related to a new "dark" mode available. Reverting to the previous colors (or other colours) is very easy, if necessary
- The Matlab pages are designed for the internal Matlab help system, and are generated with Matlab. Their style is different, and they are included in HTML "iframes".

As usual, the GitHub pages are temporarily set on this branch for review.


MJGaughran:
GitHub Pages has support for Jekyll [in some manner](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/about-github-pages-and-jekyll), but I didn't look much into how it all would fit together.

swhite2401:
The dark mode for the matlab page does not show well: text remains black for AT primer for example.
Also the links do not work, see `atsetcavity` for example

lfarv:
@MJGaughran : Jekyll is the native static page generator in GitHub Pages. That's why it was used until now. But Sphinx has much more possibilities, like automatic extraction of documentation from source files (python and Matlab). The python part of the doc was moved to Sphinx some time ago, but it appear that mixing both systems is delicate. Now the whole site in handled by Sphinx, and Jekyll is disabled with an empty `.nojekyll` file. Ihave now a solution for full automation of the doc with GitHub actions, so I'll close this PR and open a new one with full automation.

MJGaughran:
I was going to look at setting this up myself as you asked in #411, and this was the only detail that made the config unlike the one I previously linked. I like that you have transitioned to using Sphinx.

I will wait for your PR as you have apparently already made good progress with #411. I was going to spend a bit of time on it next week, so thanks.

Did you follow the method using peaceiris/actions-gh-pages?

lfarv:
@swhite2401 : the present problem with Matlab pages is that they are taken from the internal Matlab help system. I found useful to have the global overview (installation and example of use) accessible from the Web site, but they are drawbacks:

- these Matlab pages are put in "frames": it's a web page inside another one. The outer page (Sphinx) has no knowledge of what's inside the inner one. That's why the change of theme has no effect on the inner one.
- The links to Matlab individual functions are not http links, but Matlab links: "`matlab:doc('atSetRingProperties')`" rather that "`https://...`". They cannot work outside Matlab.

Of course, these few pages could be rewritten in markdown, but keeping two version is difficult to maintain. Or one could have the whole documentation generated by Sphinx, but up to now there are problems with the style of help strings: Sphinx (or Google style) is in contradiction with the Matlab rules… I'm looking at that.

However, for the time being, I made a significant improvement by automating the  Sphinx compilation with GitHub actions, so I'll close this PR and open another one (same documentation contents as this one)

lfarv:
@MJGaughran: yes, it's based on the  peaceiris/actions-gh-pages action, thanks for the tip. While I'm still working on details, you can have a look at this branch: `sphinx_automation`, the workflow is called `docs.yml`.

swhite2401:
I agree this is excellent! Please go ahead and merge!

PR #413, opened 2022-06-13
Sphinx automation

lfarv:
This PR replaces #412.

The AT web site, hosted on GitHub Pages, is now entirely generated by Sphinx. Jekyll is no more used. The documentation is automatically rebuilt on each push on the master branch. This is done by a new workflow for GitHub actions which will run Sphinx on each push, or upon explicit request.

The documentation source is in the `/docs` directory.

The compiled documentation is pushed to the new `gh-pages` branch, declared as the source for GitHub Pages. This branch only contains the html documentation, at its top-level (nothing else, no code…), and should not be modified.

The process for writing documentation is unchanged: create a new branch, add/modify ReST or markdown documents in the `/docs` directory and open a pull request. The web site will be updated when the branch is merged.

This process is briefly described in the README document  in the /docs directory.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Build and install at
+      run: python -m pip install ".[plot, doc]"
+      working-directory: pyat
+
+    - name: Compile docs
+      run: make html
+      working-directory: docs
+
+    - name: Publish Docs to gh-pages
+      uses: peaceiris/actions-gh-pages@v3
```
MJGaughran, line 33: As described in another example [here](https://github.com/dls-controls/scanspec/blob/2136794f6898f19ec6cccb819e31ca4149191b03/.github/workflows/docs.yml), it is [recommended ](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions)that third-party actions are referenced by commit hash instead of tag. It is only a recommendation, but it seems reasonable.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,36 @@
+name: Docs CI
+
+on:
+  push:
+    branches:
```
MJGaughran, line 5: Do we want to run this for tags as well?

Speaking to the one who sorted this out for our own docs, it seems that by default there is no option in GitHub Pages to easily select other tags, as there is for readthedocs. He made a change to the Sphinx template (not sure how) to add this, but we are using a different template here. If this interests you, I will talk to him about how we can 'port' this over.

You can probably navigate to the documentation for other tags even without the above change.

MJGaughran:
I am still keen to remove readthedocs configuration from the repository, unless there is a good reason to have to maintain and test the two separate configurations. I can understand leaving it as-is for this pull request and making a separate one for that.

.github/workflows/docs.yml
lfarv: Hi @MJGaughran: I see the point, but I don't think we should be too much concerned about that: if we were thinking of future compatibility problems, we should rather freeze versions on numpy, matplotlib, scipy… These would cause real problems like AT crashes or wrong results. Here, such an hypothetical compatibility breaking would have negligible consequences: stopping the update of documentation for a few hours!

So no problem to set a hash instead, but I would not put the "safety" comment before, which gives the impression that we put a strong emphasis on details rather then important points

.github/workflows/docs.yml
lfarv: For now, I am not yet thinking of keeping several versions of the documentation. It is so late behind the code that the main point is to improve it and keep it up-to-date with the code evolution. So updating on push is perfect, a tag is not related to any modification of the sources. Note that in the final version, only the `master` branch should be kept, the `sphinx_automation` branch is only there to check that it's properly working.

For the future, one could think of keeping a doc version corresponding to each PyAT release, so your ideas are welcome ! For now, the idea is that whatever PyAY release or commit you are running, the later documentation is the best.

lfarv:
@MJGaughran : readthedocs gives an easy way to check the behaviour of a branch under development, while the GitHub Pages are updated only once the merge is done  (will be, in fact, once this PR is merged). In this preliminary stage of development, it's helpful. The other alternative is of course to run Sphinx locally. Bit since there is no risk of having diverging versions of the doc (there is only one source for both), I keep it for the moment.

.github/workflows/docs.yml
lfarv: `peaceiris/actions-gh-pages` now referenced with hash `068dc23d9710f1ba62e86896f84735d869951305` (last version)

MJGaughran: Late response, but hash reference is good.

This is not about future compatibility, but instead potential to run compromised action as described in the recommendation.

PR #414, opened 2022-06-15
Improvement of RF frequency control

lfarv:
The `get_acceptance` function is an example of a function which may internally modify the ring RF frequency, to satisfy its `dp` or `dct` argument. The new `frequency_control` decorator avoids this and offers a cleaner solution to handle `dp` and `dct` in future similar functions.

This decorator can be applied to any function like `func(ring, *args, **kwargs)`. Example:

```python
@frequency_control
def get_acceptance(ring, planes, npoints, amplitudes, nturns=1024,
                   refpts=None, dp=None, offset=None, bounds=None,
                   grid_mode=GridMode.RADIAL, use_mp=False, verbose=True,
                   start_method=None, divider=2, shift_zero=1.0e-9):
```

It looks at the 1st argument of the decorated function, which must be a Lattice object (`ring`)

- if `ring.radiation` is `True` **and** `dp` or `dct` is specified, it makes a copy of `ring` with a modified RF frequency, removes the `dp` and `dct` keywords and call the decorated function with the copy of ring,
- otherwise, it directly calls the decorated function with the initial arguments.

So the decorated function does not need to care about RF frequency: It uses `ring` as it is, the `dp` or `dct` arguments will be present only when they are allowed (in 4D only).

In addition, `get_revolution_frequency` is improved when `dp` is specified: rather that relying on the momentum compaction factor, it uses closed orbit computation, giving a more accurate result.

### DIFF
--- a/atmat/atphysics/LongitudinalDynamics/atsetcavity.m
+++ b/atmat/atphysics/LongitudinalDynamics/atsetcavity.m
@@ -92,9 +92,12 @@ if isempty(varargs)             % New syntax
             if isfinite(dct)
                 frev=frev * lcell/(lcell+dct);
             elseif isfinite(dp)
-                [~,ringrad]=check_radiation(ring,false,'force');
-                etac=1/gamma0^2 - mcf(ringrad);
-                frev=frev + frev*etac*dp;
+                % Find the path lengthening for dp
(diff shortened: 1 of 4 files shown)


swhite2401 (approved):
Seems like a very nice improvement

PR #421, opened 2022-06-23
Silent atpass

swhite2401:
An optional option `silent` is added to `atpass`.
In case it is activated only the last turn is returned. This allows to track large number of particles without allocating excessive memory and is a preparation for the implementation of multi-bunch collective effects simulations.

### DIFF
--- a/pyat/at/tracking/patpass.py
+++ b/pyat/at/tracking/patpass.py
@@ -95,6 +95,7 @@ def patpass(ring, r_in, nturns=1, refpts=None, pool_size=None,
           keep_counter is True. The turn number is necessary to compute the
           absolute path length used in RFCavityPass.
         losses (Optional[bool]):        Boolean to activate loss maps output
+        silent (Optional[bool]):        Boolean to activate silent output
         pool_size (Optional[int]):      number of processes. If None,
           ``min(npart,nproc)`` is used
         start_method (Optional[str]):   This parameter allows to change the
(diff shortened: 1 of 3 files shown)


lfarv:
@swhite: there is no need for that: just set `refpts = []`. No allocation, the values at the end of tracking are in` r_in`, as usual!

swhite2401:
Are you sure about this? 
If I tracking 1e6 particles for 1e6 turns with `refpts=[]` I thought `at` would allocates an array with shape `(6,1e6,0,1e6)` in C and return it.
Instead with this PR `(6,1e6,refpts,1)` is allocated whatever the input, there is the same in matlab actually.

In fact I crashed the cluster doing something similar...am I missing something?

swhite2401:
Ok fair enough, I'll roll back and update the documentation in this PR to make this clear

swhite2401:
Ok, this is now just a documentation update, anything to add while we are at it?

pyat/at/tracking/patpass.py
```diff
@@ -126,6 +126,15 @@ def patpass(ring, r_in, nturns=1, refpts=None, pool_size=None,
           flag for particles lost (True -> particle lost), turn, element and
           coordinates at which the particle is lost. Set to zero for particles
           that survived
+          
+    .. note::
+
+       * For multiparticle tracking with large number of turn the size of ``r_out``
```
lfarv, line 132: "number of **turns**", may be ?

lfarv (approved):

lfarv:
Ok for me !

swhite2401:
Do I have to build the sphinx documentation or is it done automatically on merge? (sorry I haven't followed very closely your latest developments)

PR #422, opened 2022-06-23
Update of setuptools_scm

lfarv:
The versioning of pyat is handled by `setuptools_scm`. Up to now, a bug prevented storing the generated version number in a python file `_version.py` (just because pyat is not at the root of the git repository). This is now corrected. As a consequence, at run time, we can now import the version number from the `_version` module, rather than extracting it from the metadata of the package. This makes the import of `at` simpler and faster.

There is still a trick to handle python 3.6. Another question is: how long do we still need to support python 3.6?

### DIFF
--- a/pyat/at/__init__.py
+++ b/pyat/at/__init__.py
@@ -1,15 +1,6 @@
 """Python port of the Accelerator Toolbox"""
 
-import sys
-if sys.version_info.minor < 8:
-    from importlib_metadata import version, PackageNotFoundError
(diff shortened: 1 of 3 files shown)


swhite2401:
The python default at ESRF is 3.7, on our side it is ok to drop support for python 3.6

lfarv:
@swhite2401 : I' m just asking the question, it's not urgent but it'll allow some simplifications.

PR #429, opened 2022-06-30
New Lattice.reduce method

lfarv:
The `Lattice.reduce` method is similar to the `atreduce` Matlab function: it removed all elements with an `IdentityPass` passmethod, and then merges compatible consecutive elements.

A keep keyword allows to keep selected elements even if they have an `IdentityPass` passmethod. By default, `Lattice.reduce` keeps `Monitor` and `RFCavity` elements.

### DIFF
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -152,8 +152,9 @@ class Element(object):
     def merge(self, other) -> None:
         """Merge another element"""
         if not self.is_compatible(other):
+            badname = getattr(other, 'FamName', type(other))
             raise TypeError('Cannot merge {0} and {1}'.format(self.FamName,
-                                                              other.FamName))
+                                                              badname))
 
 
 class LongElement(Element):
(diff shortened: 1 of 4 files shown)


swhite2401 (approved):
This is very nice!

lfarv:
@simoneliuzzo : there was a bug when comparing `PolynomA/B` (missing the last term, `PolynomB[MaxOrder]`). Your example now seems to run. Can you check ?

simoneliuzzo (approved):
I have tested with two lattice files. One EBS SR lattice, with few elements to reduce and an ARC cell of 10000 elements that has been correctly reduced to 141 elements. The optics after reducing are identical a part the last few digits of some parameters (numerical).

lfarv:
@simoneliuzzo: thanks for testing, I rely on you for that !

Did you try "`keep=None`" which should be even more agressive by removing `Monitor` and `RFCavity` elements, or any other  "`keep`" combination ?

Issue #363, opened 2022-02-07
Compiling PyAT with MPI


lfarv:
The last PyAT release uses the new python packaging method based on `pyproject.toml`. With this setup, the requirements for **building** the project (not **installing**) must be declared in `pyproject.toml`. So at the moment, at attempt for building with MPI using:
```
$ MPI=1 pip install .
```
fails because `mpi4py` is not declared as a build dependency in `pyproject.toml`. This even if `mpi4py` is already installed, because the build occurs in a virtual env. And `mpi4py` cannot be declared unconditionally, which would make it mandatory for all builds. The installation requirement can be declared in `setup.cfg` in the `[options.extras_require]` section, but not the build one…

So at the moment, the way to build with MPI is to use the old method:
```
$ MPI=1 python setup.py install
```
which now triggers a deprecation warning.
I do not know how the build requirements can be adjusted depending on an environment variable, or any other selection method…

swhite2401:
@lfarv, I see that you have removed requirements.txt, why so? It is very convenient to have a file storing all dependencies required to run the code that you can use to build your venv with a single pip command

lfarv:
In the new packaging system, the requirements are handled by `pyproject.toml`, so `requirements.txt` is now deprecated. Then `pip install` handles both build requirements when building from source, and install requirements when building from a wheel file. So still a single command to install from scratch!

swhite2401:
What about pip install -e . , I think it calling develop, will it work in this case?

lfarv:
@swhite2401: yes, it works. It's the "editable" mode, you have to build either from the source distribution (called "sdist", a tar.gz file), or from a local directory.
See [this](https://packaging.python.org/en/latest/tutorials/packaging-projects/).

swhite2401:
btw I still think having the requirements file is useful as you can call it with pip -r, also many people use conda environments (default at ESRF) for which you need the requirements file -> conda install --file requirements.txt

Why not keep it? It doesn't arm no?

lfarv:
@swhite2401: the problem with this new system is that it privileges a "declarative" approach, in config files like pyproject.toml and setup.cfg, trying to remove all "executable" files (*.py files) for safety reasons: avoid potentally malicious installation procedures. `setup.py` still exists because it's still necessary to build extensions (compiled C and C++) until something better is defined. But its functionality is more restricted.

swhite2401:
Ok , but why remove the `requirements.txt` file? As I said I remains useful in some case and does no arm

[...]

PR #431, opened 2022-07-02
Fix a build problem with numpy 1.23

lfarv:
#### Problem:
The recent release of `numpy` 1.23 (available for python 3.8 to 3.10) introduced a problem for users of python 3.8+ building from source.` pip install` runs in its own environment, and installs there the most recent numpy version, to compile the C-extensions of AT (so `numpy` 1.23 now). At run time, the code links with the release installed by the user in its environment, possibly an older one. It appears that `numpy` 1.23 changes something so that `at` fails to import:
```python
RuntimeError: module compiled against API version 0x10 but this version of numpy is 0xe
```
The only solution for the user is to upgrade `numpy`. Though usually this can be done without side effects, this should not be requested!

#### Proposed solution:
The solution proposed here is to select for building with each python version the oldest compatible `numpy` release. The version used at runtime will then always be more recent, ensuring full compatibility (except for documented deprecations). The following table is used:

| Python version | numpy release |
| :-------------: | :-------------: |
|                     3.6 |                 1.16.6 |
|                     3.7 |                 1.16.6 |
|                     3.8 |                 1.17.3 |
|                     3.9 |                 1.19.3 |
|                    3.10 |                1.21.3 |

This the version used for **building only** (compilation of C extensions), configured in the `[build-system]` section of `pyproject.toml`. The requirement for **running** PyAT (`install_requires` in `setup.cfg`) stays as it is: numpy > 1.16.6, to get the necessary features. Practically, it will be the current version at the time of 1st installation, but the user may upgrade as he wants.

#### Possible side effects:

- `python setup.py install` does not run in a separated environment. Instead, it will downgrade numpy in the user's environment. So **always use "`pip`" for installation !**
- This may prevent the C extensions to use the most recent features of `numpy`. However the AT C code does not call any numpy function, it just needs access to the internal structure of a `ndarray`. It's very unlikely that any new feature may improve (accelerate) this significantly. So this is not relevant.

### DIFF
--- a/pyat/pyproject.toml
+++ b/pyat/pyproject.toml
@@ -1,7 +1,9 @@
 [build-system]
 requires = [
-    "numpy >= 1.16.6;python_version<'3.10'",
-    "numpy >= 1.21.5;python_version>='3.10'",
+    "numpy ~= 1.16.6;python_version<'3.8'",
+    "numpy ~= 1.17.3;python_version=='3.8'",
+    "numpy ~= 1.19.3;python_version=='3.9'",
+    "numpy ~= 1.21.3;python_version=='3.10'",
     "setuptools >= 45",
     "setuptools_scm >= 7;python_version>='3.7'",
     "setuptools_scm >= 6.4.2;python_version<'3.7'",
--- a/pyat/setup.cfg
+++ b/pyat/setup.cfg
@@ -23,8 +23,7 @@ packages = find:
 zip_safe = False
 install_requires =
     importlib-resources;python_version<'3.9'
-    numpy>=1.16.6;python_version<'3.10'
-    numpy>=1.21.5;python_version>='3.10'
+    numpy>=1.16.6
     scipy>=0.16
 
 [options.package_data]


swhite2401:
> * `python setup.py install` does not run in a separated environment. Instead, it will downgrade numpy in the user's environment. So **always use "`pip`" for installation !**

@lfarv in #363 you showed that for the moment we cannot build with MPI using `pip`, does this mean that we will not be able to use the latest `numpy`, even worse that the environment that might be used for other applications will be affected?

In which case I would suggest that either #363 is solved (I do not see much problem in adding `mpi4py` to the requirements) or that another solution is found here.

lfarv:
@swhite2401 : in this situation, you could still upgrade `numpy` again after the installation is done… But I agree, that's not nice. Anyway `python setup.py install` should soon stop working completely, so we'll need to find another solution.

For the moment, I think the problem you and @simoneliuzzo encountered recently, and that everybody running python 3.8+ will encounter, is much worse.

So I'm looking at solutions for building with MPI.

lfarv:
@swhite2401 : here is a solution to build with MPI using `pip`:

`MPI=1 pip install --no-build-isolation .`

with the `--no-build-isolation` flag, pip will ignore the build specs from `pyproject.toml` and use the installed packages as they are. You just have to ensure yourself that `numpy` and `setuptools_scm` have been previously installed (which should be the case if you already ran` pip install` normally).

There is still a hope to find a cleaner solution, but does this allow to merge and solve this critical problem?

swhite2401:
Yes ok for me.
Not necessarily very urgent, but it would be nice to document this and other build option such as OpenMP on the github pages.

swhite2401 (approved):

simoneliuzzo (commented):
I haven't got sufficient python skills to approve or not this pull request

MJGaughran:
How does this affect the building of wheels as part of our release process?

lfarv:
@MJGaughran : if we were to produce a wheel now, the C integrators for python3.8+ would use the numpy 1.23 ABI, so would be incompatible with any older version. As mentioned above, this would force the users to upgrade their numpy to 1.23.

With this PR, wheels will be compiled versus the numpy version in the table above, that is the oldest numpy compatible with the python version. Consequently they will be compatible with any newer numpy version.

We had no problem until now because apparently the numpy ABI was stable since long. Now we can profit from the fact that with the new packaging method (pyproject.toml), we can have different requirements for building (freeze a version) and running (allowing any version recent enough).

PR #432, opened 2022-07-03
Rn Matlab tests in GitHub actions

lfarv:
AT has since long a test sequence comparing the python and Matlab results. However, it has to be run locally because of a problem in GitHub actions: the "Matlab engine for python" fails to start because of a license problem. A reply from Mathworks is pending…

This PR proposes another solution: instead of starting Matlab from python, it runs python under Matlab. The test sequence is described in the Matlab `atmat/attests/pytest.m` file and covers at the moment:
| python | Matlab |
| ------ | -------- |
| lattice_pass | linepass |
| find_orbit4 | findorbit4 |
| find_orbit6 | findorbit6 |
| find_sync_orbit | findsyncorbit |
| find_m44 | findm44 |
| find_m66 | findm66 |
| linopt6 | atlinopt6 |
| get_radiation_integrals | ringpara |
| lattice properties | atGetRingProperties |

The test is restricted at the moment to a `macos-latest` runner because:
- when running on `ubuntu-latest`, `linopt6` (and all derived functions…) fail for  `"Intel MKL ERROR: Parameter 11 was incorrect on entry to DGEEV."`. This may result from a conflict between the Matlab and python LAPACK implementations. A solution is under investigation, but a subset excluding  the `linopt6` tests works.
- when running on  `windows-latest`, `atmexall` fails to compile the C++ extensions (ExactHamiltonianPass, which is not used…). That can also be solved.

However, this already gives a good check of the consistence of both implementations.

### DIFF
--- a/.github/workflows/matlab-tests.yml
+++ b/.github/workflows/matlab-tests.yml
@@ -44,16 +44,9 @@ jobs:
       uses: matlab-actions/run-command@v1
       with:
         command: pwd,path
-#
-#    - name: Install test dependencies
-#      run: |
-#        python -m pip install --upgrade pip
-#        python -m pip install pytest
-#        python -m pip install pytest-lazy-fixture
-#        python -m pip install pytest-cov
-#
-#    - name: Build and install at
-#      run: python -m pip install -e .
-#
-#    - name: Test with pytest and coverage
-#      run: python -m pytest test_matlab --cov-report term-missing --cov=at
+
+    - name: Build and install at with tests
+      run: python -m pip install -e "./pyat[dev]"
+
+    - name: Test with pytest
+      run: python -m pytest pyat/test_matlab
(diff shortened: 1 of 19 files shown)


lfarv:
Now runs on MacOS and Windows

swhite2401:
This is a very nice and much needed development.
This `intel` error you are getting sounds familiar, I think I ran into similar issues try to interface a python code with matlab.
The work around was to preload some libraries such that matlab would not use its own, not sure that this is possible here.

lfarv:
> The work around was to preload some libraries such that matlab would not use its own

Interesting ! Any hint on how to do this is welcome !

swhite2401:
> > The work around was to preload some libraries such that matlab would not use its own
> 
> Interesting ! Any hint on how to do this is welcome !

I can't remember exactly and cannot find my old startup script.

However, could this help?
https://fr.mathworks.com/help/releases/R2019b/matlab/matlab_external/out-of-process-execution-of-python-functionality.html

PR #434, opened 2022-07-04
fix empty orbit in atlinopt

swhite2401:
This PR fixes an issue with `atlinopt` with `twiss_in` activated. 
When orbit_in was not provided an empty list was returned instead of 6 zeros.
~~The implementation could be further improved by using twiss_in.ClosedOrbit instead of zeros(6,1), to be discussed~~
If present the default is  `twiss_in.ClosedOrbit`

### DIFF
--- a/atmat/atphysics/LinearOptics/atlinopt.m
+++ b/atmat/atphysics/LinearOptics/atlinopt.m
@@ -114,7 +114,7 @@ if isempty(twiss_in)        % Circular machine
     [orbitP,o1P]=findorbit4(RING,dp+0.5*DPStep,REFPTS,orbitin,'XYStep',XYStep);
     [orbitM,o1M]=findorbit4(RING,dp-0.5*DPStep,REFPTS,orbitin,'XYStep',XYStep);
 else                        % Transfer line
-    if ~isempty(orbitin)
+    if isempty(orbitin)
         orbitin=zeros(6,1);
     end
     try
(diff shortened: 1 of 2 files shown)


lfarv:
> The implementation could be further improved by using twiss_in.ClosedOrbit instead of zeros(6,1)

This is what is done in `atlinopt6`, I just copied the 4 lines !

lfarv (approved):
OK

swhite2401:
Thanks, I checked and this is what is done in `pyAT` as well so I think we are all good

simoneliuzzo (approved):

lfarv:
@swhite2401 : I think this is not the right place for the fix. `atlinopt6` is designed to accept twiss input in any format, so what happens here is a bug in `atlinopt6`, not `atplot`. Let me have a look.

swhite2401:
Well if the rmatrix is missing it will fail... the other option is to provide a function that derives the rmatrix from atlinopt output but do you really want to go in this direction?

swhite2401:
By the way the same problem is in pyAT

lfarv:
Of course yes: if `R` is missing, `atlinopt6` (`build_sigma`) takes `alpha` and `beta` from the twiss input. But something must be missing… `atlinopt6` has not been tested enough in transfer line mode, thanks @simoneliuzzo. Could you separate the two problems, rollback to commit aa1fd9e and open another PR for `atplot`?

lfarv:
> By the way the same problem is in pyAT

Another good reason to open a dedicated PR !

swhite2401:
Ok I roll back the last commit and merge this one.

Issue #446, opened 2022-07-18
atcavityoff does not work anymore


lnadolski:
atcavityoff(ring) does not switch the cavity off

this a alias to [ring,~,ATCavityIndex]=atradoff(ring,'auto','','','','');

The issue is there : 
cavitiesIndex=atgetcells(ring,'PassMethod',@(elem,pass) endsWith(pass,'CavityPass'));

Of course CavityPass is not valid anymore if RF is switch off
Why not using the Class field instead ?

PR #447, opened 2022-07-19
Fix atcavityoff

lfarv:
atcavityoff did not correctly return the index of cavities (see #446). This is corrected here.

### DIFF
--- a/atmat/atcavityoff.m
+++ b/atmat/atcavityoff.m
@@ -15,6 +15,7 @@ function [ring, ATCavityIndex]=atcavityoff(ring)
 
 %% Written by Laurent S. Nadolski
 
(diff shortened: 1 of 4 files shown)


lfarv:
Note that there is a difference of meaning of `cavitiesIndex` between `atradon`/`atradoff` and `atcavityoff`.
- in `atradon`/`atradoff`, the cavities index points to **active** cavities. Since both functions allow the modification of some elements while keeping others intact, this is a valuable information,
- in `atcavityon`/`atcavityoff` (original version), the cavities index points to all cavities, independently of their state. This behaviour is restored here. The problem might even have occurred in `atcavityon` if the user for some reason, would set cavipass to a non-cavity passmethod.

lnadolski (approved):
That makes the job
Thanks

PR #448, opened 2022-07-20
Fix help on get_cells

lfarv:
Corrected errors in the help for the `get_cells` function, concerning the output length and the examples


simoneliuzzo (approved):

simoneliuzzo:
Dear @lfarv 
the online help still show len(ring) 
the +1 is not "converted" to the web doc.

PR #451, opened 2022-07-26
PyAT: improved timelag solver

swhite2401:
In `get_timelag_from_U0` the numerical synchronous phase solver was failing in some cases in the presence of multiple rf systems, such as higher order cavities.
This PR proposes a simpler more robust version.

### DIFF
--- a/pyat/at/physics/energy_loss.py
+++ b/pyat/at/physics/energy_loss.py
@@ -74,9 +74,6 @@ def get_energy_loss(ring: Lattice,
                                      sextupole_pass=None,
                                      octupole_pass=None,
                                      copy=True)
-        diff_elem = ringtmp.get_elements('Diffusion')
-        for de in diff_elem:
-            ringtmp.remove(de)
 
         o6 = numpy.squeeze(lattice_pass(ringtmp, numpy.zeros(6),
                            refpts=len(ringtmp)))
@@ -143,8 +140,7 @@ def get_timelag_fromU0(ring: Lattice,
     freq = numpy.array([cav.Frequency for cav in ring.select(cavpts)])
     rfv = numpy.array([cav.Voltage for cav in ring.select(cavpts)])
     tl0 = numpy.array([cav.TimeLag for cav in ring.select(cavpts)])
-    if u0 > numpy.sum(rfv):
-        raise AtError('Not enough RF voltage: unstable ring')
+
     try:
         frf = singlev(freq)
         tml = singlev(tl0)
@@ -170,6 +166,8 @@ def get_timelag_fromU0(ring: Lattice,
         ts = -r[numpy.argmin(res)].x[0]
         timelag = ts+tl0
     else:
+        if u0 > numpy.sum(rfv):
+            raise AtError('Not enough RF voltage: unstable ring')
         vrf = numpy.sum(rfv)
         timelag = clight/(2*numpy.pi*frf)*numpy.arcsin(u0/vrf)
         ts = timelag - tml
--- a/pyat/at/physics/energy_loss.py
+++ b/pyat/at/physics/energy_loss.py
@@ -5,7 +5,7 @@ from typing import Optional, Tuple
 import numpy
 from scipy.optimize import least_squares
 from at.lattice import Lattice, Dipole, Wiggler, RFCavity, Refpts
-from at.lattice import check_radiation, AtError
+from at.lattice import check_radiation, AtError, AtWarning
 from at.lattice import checktype, set_value_refpts, get_cells, refpts_len
 from at.constants import clight, Cgamma
 from at.tracking import lattice_pass
@@ -74,6 +74,9 @@ def get_energy_loss(ring: Lattice,
                                      sextupole_pass=None,
                                      octupole_pass=None,
                                      copy=True)
+        diff_elem = ringtmp.get_elements('Diffusion')
+        for de in diff_elem:
+            ringtmp.remove(de)
 
         o6 = numpy.squeeze(lattice_pass(ringtmp, numpy.zeros(6),
                            refpts=len(ringtmp)))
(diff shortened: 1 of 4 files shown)


lcarver:
I tested 2 cases. Single and multi harmonic. For both cases when the returned timelag is set on the cavities using
`
at.set_value_refpts(ring, cavpts, 'TimeLag', tl, copy=False)

`
then the equilibrium position for ct was on the order of 1e-5. 

All fine for me.

swhite2401:
> On closer inspection, I realised that I am not able to trigger the warning seen on line 163 of energy_loss.py. I put a 4th harmonic cavity with very high voltage (6.5e6/1.5) which gives multiple possible phis but it doesn't seem to catch the warning.

Could you send a figure of the wave form you got for this one? the initial points for the 2 solutions are at +/-lambda/4 around TimeLag so it is possible that if a solution is far off these initial values it is not found

lcarver:
image

it finds a solution, but it doesn't find the others so doesn't throw a warning

lcarver:
It's ok now. If I keep the harmonic TimeLag=0 then it triggers the warning. For TimeLag=0.1 (what you see in the figure) then it doesn't. So the warning can be triggered. but I guess the question now is why it doesn't find the second solution on the left of the main one?

swhite2401:
> It's ok now. If I keep the harmonic TimeLag=0 then it triggers the warning. For TimeLag=0.1 (what you see in the figure) then it doesn't. So the warning can be triggered. but I guess the question now is why it doesn't find the second solution on the left of the main one?

Right, can you send me your lattice?

swhite2401:
> > It's ok now. If I keep the harmonic TimeLag=0 then it triggers the warning. For TimeLag=0.1 (what you see in the figure) then it doesn't. So the warning can be triggered. but I guess the question now is why it doesn't find the second solution on the left of the main one?
> 
> Right, can you send me your lattice?

There was a logic issues settings the boundaries and initial guesses, this is fixed. In addition the range is now divided in 8 initial guesses over the full wake length on the main RF, instead of just 2 before. This should mitigate the issue with solutions not found

pyat/at/physics/energy_loss.py
```diff
@@ -74,6 +74,9 @@ def tracking(ring):
                                      sextupole_pass=None,
                                      octupole_pass=None,
                                      copy=True)
+        diff_elem = ringtmp.get_elements('Diffusion')
+        for de in diff_elem:
+            ringtmp.remove(de)
```
lfarv, line 79: `remove` should not be used here:
- its 2nd argument must be an Element, not a list of Elements
- it remove only the 1st Element which is equal to the 2nd argument. There may be several diffusion elements,
- it has to check each element in the lattice for equality with the second argument. It's long, and the quality between Elements is not well defined; it's identity, in fact.
A more efficient way to do that is:
```python
diff_elem = get_cells(ringtmp, checkattr('PassMethod', 'QuantDiffPass')
ringtmp=ringtmp(~diff_elem)
```

That beeing said, I think this has to be done in `Lattice.radiation_off` (called h=just before). Such elements still modify the momentum, so should be turned off in `radiation_off` (set to 'IdentityPass')

pyat/at/physics/energy_loss.py
```diff
@@ -123,40 +128,50 @@ def singlev(values):
...
     if cavpts is None:
-        cavpts = get_cells(ring, checktype(RFCavity))
-    u0 = get_energy_loss(ring, method=method) / ring.periodicity
+        cavpts = ring.get_cells(checktype(RFCavity))
+    u0 = ring.get_energy_loss(method=method) / ring.periodicity
     freq = numpy.array([cav.Frequency for cav in ring.select(cavpts)])
     rfv = numpy.array([cav.Voltage for cav in ring.select(cavpts)])
     tl0 = numpy.array([cav.TimeLag for cav in ring.select(cavpts)])
+    if u0 > numpy.sum(rfv):
+        raise AtError('Not enough RF voltage: unstable ring')
```
lfarv, line 147: This test is better located where it was before (in the "Else" clause): summing the voltages only makes sense if all frequencies and timelags are identical. And the "missing Voltage" error is already handled in the "Except" clause.

pyat/at/physics/energy_loss.py
swhite2401: Moved to `radiaton_off()` as suggested, I also introduced `Element.QuantDiffElement` to simplify detection

pyat/at/physics/energy_loss.py
swhite2401: Done

lfarv:
Please put the modifications of `lattice_object.py` and `elements.py` in a different pull request. They don't affect your initial PR (unless there is a diffusion element in the lattice). It avoids mixing unrelated things, makes validation easier and makes history easier to read (think of listing the changes in the release notes of the next version)… Thanks!

swhite2401:
As request I have created #453 to propose the `QuantDiffElement` and the modified `radiation_on/off()`

lfarv:
For the computation, I trust @lcarver and you. But:

- Why changing the default method for `TRACKING` ? It's decoupled from the topic of this PR: get a better solution for the equation V(c&tau;)= U<sub>0</sub>, whatever U<sub>0</sub>. `INTEGRAL` is robust, fast, works with or without radiation. Even if it may be less accurate, it's probably good enough in most cases as an initial guess for `find_orbit`. Why not keeping `TRACKING` for the cases where `INTEGRAL` is not accurate enough? 
- Why changing `math.pi` to `numpy.pi` ?  I have no opinion on which is best, apart from matching its type (float of numpy scalar array) to the type of the rest of the expression… Here it's mixed.

swhite2401:
> * UPDATE: `find_orbit` forces the default to `TRACKING`. So my remark on orbit search is irrelevant. But remains: why changing something without justification?

Exactly for that reason: we have to be consistent in the default we use and stick to the decision made in other places.
TRACKING is more precise and will give the correct result in any condition. I questioned the choice of TRACKING for `find_orbit` once because I thought integral was more robust but you replied exactly this to me.

swhite2401:
> * Why changing `math.pi` to `numpy.pi` ?  I have no opinion on which is best, apart from matching its type (float of numpy scalar array) to the type of the rest of the expression… Here it's mixed.

I have no opinion on that. I always use `numpy` rather than `math`, but this is just an habit. Isn't the expression mixed anyway because of `numpy.arcsin()`. Again we have to be consistent an use the same everywhere...shall we add the preferred one to `constants.py`?

lfarv:
> I have no opinion on that. I always use `numpy` rather than `math`, but this is just an habit. Isn't the expression mixed anyway because of `numpy.arcsin()`. Again we have to be consistent an use the same everywhere...shall we add the preferred one to `constants.py`?

So no reason to change anything, let's keep the expression type decide, and in this particular case it's mixed. I was just curious about the reason for the change !

So it's ok for me.

lfarv (approved):
ok

swhite2401:
Ok I merge then

PR #452, opened 2022-07-27
Fill pattern

swhite2401:
Beam characteristics are added to the `Lattice` object:

- `fillpattern`: contain information on the filled buckets and relative amplitude
- `bunch_spos`: bunch spatial distribution
- `bunch_currents`: bunch current distribution
- `nbunch`: number of bunches
- `beam_current`: total beam current (mean to replace the current properties in all collective effects `passmethods`)

These attributes are then accessible to the c `passmethods`. This is a pre-requisit for the implementation of multi-bunch tracking implementation that will be introduced in a future PR.

Existing pyAT functionalities are not modified by this PR.


pyat/at/lattice/particle_object.py
```diff
@@ -75,3 +92,59 @@ def rest_energy(self) -> numpy.ndarray:
...
+        return self._weights
+
+    @weights.setter
+    def weights(self, weights):
+        assert len(weights) == len(self._fillpattern), \
+            'Weights and fill pattern must have the same length'
+        self._weights = numpy.array(weights)
+
+    @property
+    def nbunch(self):
```
lcarver, line 143: this is a float but should be an int

pyat/at/lattice/particle_object.py
```diff
@@ -75,3 +92,59 @@ def rest_energy(self) -> numpy.ndarray:
...
+
+    @fillpattern.setter
+    def fillpattern(self, bunches):
+        if (self._harmn is None):
+            warn(UserWarning("Harmonic number not set in Beam(), "
+                             "nbunch and fillpattern set to 1"))
+            fp = numpy.ones(1)
+        elif numpy.isscalar(bunches):
+            if bunches == 1:
+                fp = numpy.ones(1)
```
lcarver, line 117: the output of np.ones is a float (1.0). Is it not more appropriate for filling pattern to be a mask? In which case you could use np.ones(992, dtype=bool). It is still compatible when multiplied with weights as True*2 = 2

lfarv:
Hi @swhite2401. As you suspected, I think that duplicating the harmonic number in the `Particle` object is a very bad thing. But I am not sure that this is necessary. If I look at `at.c`, I see that you just need:
- a total current,
- a number of bunches,
- an array of "weights" as long as the bunch number.

I understand that the bunch are equally spaced. And the harmonic number does not appear anywhere.

So why do you need more in the `Particle` object? Store the  same information (even the number of bunches is redondant: the length of the "weight" array is enough).

The only point is to check the validity of the number of bunches. So let the `Particle` constructor create a `Particle` with 0 current and 0 bunch.

To set bunches, you **must** call a method to which you provide the weight array, and the lattice. Then you can check that the `Lattice.harmonic_number` is a multiple of the number of bunches.

Or, may be better, you add a method to the `Lattice` object, which will set the bunches of its **Particle** attribute, through a private method of the `Particle`. A standalone Particle cannot have bunches, but that makes sense.

Well, may be I missed something, but that's an idea…

swhite2401:
> I understand that the bunch are equally spaced. And the harmonic number does not appear anywhere.

No bunch distribution can be arbitrary, in which case a vector `(harmn,)` is needed and the bunch spacing is required for long range wake field calculations, I'll look into integration some function into the `lattice_object`. This is what I started with but I thought this was already overloaded...but you are right it is cleaner

lfarv:
Hi @swhite2401: I was not clear enough in my description. I think that you can efficiently describe any arbitrary bunch combination with 2 variables;
1. the total current (float)
2. an array of floats (`weights`, that I would rather call `filling_pattern`), such that the harmonic number is a multiple of its length. It defines equidistant buckets, and the values are the fraction of the total current: 0.0 means empty bucket. The most general case is when the length of the array is equal to the harmonic number, where non-zero values indicate filled bunches. But shorter arrays may define common "few bunch" distributions.

So the harmonic number is only necessary in the property setter to check the length of the `filling_pattern` array. As you proposed, a scalar value could be interpreted as a number of equidistant and equally filled bunches.

It `at.c`, I see;
```C
struct parameters
{
  ...
  double beam_current;
  double nbunch;
  double bunch_spacing;
  double *bunch_currents;
};
```
These 2 attributes are sufficient to fill `beam_current`, `nbunch`, `bunch_current`. `bunch_spacing` is computed in C as `param.bunch_spacing = param.RingLength/param.nbunch;`, so if these variables are sufficient for the multi-bunch tracking, you don't need more than these two properties.

Now I think that `total_current` and `filling_pattern` could be properties of the `Lattice` object rather that `Particle`. This is simpler, `total_current` can be a normal attribute, `filling_pattern` must be a property, and the property setter can also take care of normalisation so that `sum(filling_pattern) = 1.0`. It's also possible that it converts a short array into a full (nharm,) array, if it makes further processing easier.

This means a very small modification of the original `lattice_object.py`:
- in the constructor, initialise `total_current` at 0 and `filling_pattern` as empty array,
- a property setter for the `filling_pattern` array.
- preferably resetting at empty `filling_pattern` when `harmonic_number` is changed.

Almost nothing to modify in `at.c` compared to your version: take properties from the lattice rather than particle.

I have a question concerning the future tracking: do you intend to track a single macro-particle in each bunch ? Then how to ensure that `r_in` is consistent with the declared filling_pattern (`r_in` particles in the right bunch)?

pyat/at/lattice/particle_object.py
swhite2401: `Particle.fillingpattern` now return a mask

pyat/at/lattice/particle_object.py
swhite2401: I don't think this really matters right?

swhite2401:
@lfarv, ok I get the point you mean `param.bunch_spacing = param.RingLength/nbuckets;` where `nbuckets` is the length of the `fillingpattern` vector, this would work yes but it really doesn't make any difference.
I have considered what you proposed, i.e. a single attribute for everything, but I found that separating the weights and the empty/filled buckets array was more user friendly.
Of course internally this can be reduced to a single array passed to `at.c`, I'll look into it.

It ok for me to move everything to the `Lattice` object, but I had in mind to start to introduce a beam object that we could use for tracking which I think would be a very useful development...anyway I'll do as you suggest we can keep this for later

swhite2401:
Now concerning your question it is a bit complicated, but I have a solution to handle multiple particles per bunch.
I am happy to discuss anytime and curious to have your opinion it will be more efficient than on git.

pyat/at.c
```diff
@@ -289,6 +291,15 @@ void set_energy_particle(PyObject *lattice, PyObject *energy,
...
                 Py_DECREF(pcharge);
             }
+            if (bcurrent != NULL) {
+                param->beam_current = PyFloat_AsDouble(bcurrent);
+                Py_DECREF(bcurrent);
+            }
+            if (bcurrents != NULL) {
+                param->bunch_currents = PyArray_DATA(bcurrents);
+                param->nbunch = PyArray_SIZE(bcurrents);
+                Py_DECREF(bcurrents);
```
swhite2401, line 301: @lfarv, this seems to work but I have a doubt, should I DECREF here? Isn't the pointer now pointing to an unreferenced array? In this case why is this working...?

swhite2401:
@lfarv, @lcarver following the discussions above I have completely refactored this PR to integrate the change to the lattice_object module rather than the particle_object, could you please give it a second look?

lfarv:
It now looks much simpler! Good! I have to go more into detail, in particular for `at.c`. Here are my first comments:

- the property setter for `beam_current` just sets the `_beam_current` attribute. Nothing else. So `beam_current` and `_beam_current` are synonyms. Why using a property? A simple attribute `beam_current` is enough,
- the default could be "no bunch". A single bunch with zero current does not make too much sense,
- I have difficulty to understand the relationship between `bunches` and `weights`. What happens if you put a weight where bunches is False ? Or alternately put a weight 0 on a True bunch? A boolean to say "there is current" and a corresponding float to tell how much is redundant. In a complex case, it could be a nightmare to put both in coincidence. It would be much simpler to give a single array where 1.0 (or any non-zero float in a more complex case) puts current and 0.0 defines an empty bucket. No ambiguity, it's more intuitive and you immediately figure out what it means. No need to care about normalisation, you just put numbers representing the relative current between bunches, the property setter will normalise it.
  For the code, it avoids all consistence checks between both, it's also simpler.
  (I keep apart the case of a scalar input for equidistant and equally filled bunches).

lfarv:
Even better: you can also accept an array of bools as the single `bunches` input: convert it to float, and you get the equally charged bunches. The `None` input in `weights` is no more necessary. To summarise:
- if `bunches` is an array of bools, it declares a list of equally charged bunches
- if `bunches` is an array of floats, it gives in addition the respective charge of each bunch.

swhite2401:
Ah, there is something I was unsure about: should I add these to the copied attributes list?

pyat/at/lattice/lattice_object.py
```diff
@@ -436,6 +454,70 @@ def particle(self, particle: Union[str, Particle]):
...
+                     the bunch distribution.
+                     For scalar input, equidistant bunches are assumed.
+                     ``ring.harmonic_number`` has to be a multiple of
+                     ``bunches``.
+                     For array input the condition
+                     ``len(bunches)==ring.harmonic_number`` is required.
+                     (default=1, single bunch configuration).
+        """
+
+        if numpy.isscalar(bunches):
```
lfarv, line 477: I would rather test:
```
if isinstance(bunches, int);
```

pyat/at/lattice/lattice_object.py
```diff
@@ -436,6 +454,70 @@ def particle(self, particle: Union[str, Particle]):
...
+                raise AtError('Harmonic number has to be a '
+                              'multiple of the scalar input '
+                              'bunches')
+        else:
+            assert len(bunches) == self.harmonic_number, \
+                'bunches array input has to be of shape ({0},)' \
+                .format(self.harmonic_number)
+            assert numpy.all(bunches >= 0.0), \
+                'bunches array can contain only positive numbers'
+            fp = numpy.ones(self.harmonic_number)[bunches>0.0]
```
lfarv, line 492: If you want to keep the bunch charge, you need:
```
fp = bunches[bunches > 0]

File: pyat/at/lattice/lattice_object.py [hunk 6]
```diff
@@ -436,6 +454,70 @@ def particle(self, particle: Union[str, Particle]):
...
+            if self.harmonic_number % bunches == 0:
+                fp = numpy.ones(bunches)
+                bs = self.circumference/bunches
+                fs = bs*numpy.arange(bunches)
+            else:
+                raise AtError('Harmonic number has to be a '
+                              'multiple of the scalar input '
+                              'bunches')
+        else:
+            assert len(bunches) == self.harmonic_number, \
```
lfarv, line 487: Here I would cast to float:
```
bunches=bunches.astype(dtype=float,casting='safe',copy=False)
```

pyat/at/lattice/lattice_object.py
```diff
@@ -436,6 +454,70 @@ def particle(self, particle: Union[str, Particle]):
         else:
             self._particle = Particle(particle)
 
+    def set_fillpattern(self, bunches=1):
```
lfarv, line 457:
```
def set_fillpattern(self, bunches: Union[int, numpy.ndarray] = 1):
```

pyat/at.c
```diff
@@ -291,6 +291,23 @@ void set_energy_particle(PyObject *lattice, PyObject *energy,
...
+          }
+          if (bspos != NULL && fpattern != NULL) {
+              param->bunch_spos = PyArray_DATA(bspos);
+              param->nbunch = (int)PyArray_SIZE(bspos);
+              param->bunch_currents = malloc(param->nbunch*sizeof(double));
+              double *fill_pattern = PyArray_DATA(fpattern);
+              for(int i=0; i<param->nbunch; i++){
+                  param->bunch_currents[i]=param->beam_current*fill_pattern[i];
+              }
+              Py_DECREF(fpattern);
```
lfarv, line 309: This gives a memory leak:  `param->bunch_currents` is not released.

The easier solution is to do the multiplication in python (in `lattice_pass`), and send the preprocessed bunch currents in a additional keyword argument. The question of Py_DECREF also disappears.

pyat/at/lattice/lattice_object.py
swhite2401: Don't we want to allow for floats as well?

pyat/at/lattice/lattice_object.py
swhite2401: Done

pyat/at/lattice/lattice_object.py
swhite2401: Done

pyat/at.c
swhite2401: I did not like so much to change ``lattice_pass`` so I integrated a new attribute ``_bunch_currents`` to ``Lattice``. A private method updates it when needed.
This should fix memory issues. As far as I understood ``PyObject_GetAttrString`` returns a borrowed reference, and ``PyArray_DATA`` has no effect on reference counting, could you confirm?

pyat/at/lattice/lattice_object.py
lfarv: For a number of bunches ? If so, you would need to check that the decimal part is zero. I think it's not worth the effort…

lfarv:
Hello @swhite2401
For the Lattice object, this seems to work. I'll put anyway a few comments:
- By using 4 attributes (`_beam_current`, `_fillpattern`, `_bunch_spos`, `_bunch_currents`), you are storing a lot of redundant information. In addition to this inelegant design of the lattice object, this forces to introduce correlations: when you modify one, you need to update some others. This is error prone.
- On the other hand, by squeezing the fill pattern to keep only the filled bunches, you "almost" loose the bucket number information (yes, you could get it back indirectly with `_bunch_spos`…). These may be useful.

A simpler solution is to keep only 2 attributes:
1. `beam_current`, as a normal attribute instead of a property,
2. the "non-squeezed" normalised fill pattern, as property + private attribute

These two contain the full information, and they are orthogonal: modifying one does nor influence the other. They represent exactly what you defined as the public interface in the present design. They are also the only ones needed in a copy.

This results in a much cleaner lattice object. Of course, you can still offer a number of read-only properties for informational purpose: number of bunches, bunch numbers, bunch locations…

The 2 parameters useful for tracking (`_bunch_current` and `_bunch_spos`, if I understand well, and possibly more) can be easily computed when they are needed, that is when interfacing with `atpass`. Possibly using the properties mentioned above. This precisely what `lattice_pass` and `patpass` are meant for.
The interface to `atpass` is considered as private (apart form the basic use of tracking through a `list`, not a `Lattice`). So you are still free to work on `lattice_pass` and `atpass`, while the `Lattice` object will be stable.

swhite2401:
> Hello @swhite2401 For the Lattice object, this seems to work. I'll put anyway a few comments:
> 
>     * By using 4 attributes (`_beam_current`, `_fillpattern`, `_bunch_spos`, `_bunch_currents`), you are storing a lot of redundant information. In addition to this inelegant design of the lattice object, this forces to introduce correlations: when you modify one, you need to update some others. This is complicated and error prone.
> 
>     * On the other hand, by squeezing the fill pattern to keep only the filled bunches, you "almost" loose the bunch number information (yes, you could get it back indirectly with `_bunch_spos`…). These may be useful.
> 
> 
> A simpler solution is to keep only 2 attributes:
> 
>     1. `beam_current`, as a normal attribute instead of a property,
> 
>     2. the "non-squeezed" normalised fill pattern, as property + private attribute
> 
> 
> These two contain the full information, and they are orthogonal: modifying one does nor influence the other. They represent exactly what you defined as the public interface in the present design. They are also the only ones needed in a copy.
> 
> This results in a much cleaner lattice object. Of course, you can still offer a number of read-only properties for informational purpose: number of bunches, bunch numbers, bunch locations…
> 
> The 2 parameters useful for tracking (`_bunch_current` and `_bunch_spos`, if I understand well, and possibly more) can be easily computed when they are needed, that is when interfacing with `atpass`. Possibly using the properties mentioned above. This precisely what `lattice_pass` and `patpass` are meant for. The interface to `atpass` is considered as private (apart form the basic use of tracking through a `list`, not a `Lattice`). So you are still free to work on `lattice_pass` and `atpass`, while the `Lattice` object will be stable.

I think we are now down to a question of taste... I prefer it this way I think it is more readable / understandable. With what you propose I have to modify the interface of 2 functions. Anyway I will do it because I need this branch merged... I hope after that it will be ok.

pyat/at/lattice/lattice_object.py
```diff
@@ -144,11 +161,13 @@ def __init__(self, *args,
         kwargs.setdefault('name', '')
         periodicity = kwargs.setdefault('periodicity', 1)
         kwargs.setdefault('_particle', Particle())
+        kwargs.setdefault('_fillpattern', numpy.array([1.0]))
```
lfarv, line 164: Should not `_fillpattern` be an array (harm number,) of floats?
Since it looks difficult to set a fill pattern from the constructor keyword arguments, you may just call set_fillpattern() at the end of `__init__`

pyat/at/lattice/lattice_object.py
```diff
@@ -144,11 +161,13 @@ def __init__(self, *args,
...
         # Remove temporary keywords
         frequency = kwargs.pop('_frequency', None)
         cell_length = kwargs.pop('_length', None)
         cell_h = kwargs.pop('_harmnumber', math.nan)
         ring_h = kwargs.pop('harmonic_number', periodicity*cell_h)
+        self.beam_current = kwargs.pop('beam_current', 0.0)
```
lfarv, line 170: Useless: the attribute will be set in `self.update(kwargs)`, line 179

pyat/at/lattice/lattice_object.py
```diff
@@ -436,6 +455,71 @@ def particle(self, particle: Union[str, Particle]):
...
+                     For scalar input, equidistant bunches are assumed.
+                     ``ring.harmonic_number`` has to be a multiple of
+                     ``bunches``.
+                     For array input the condition
+                     ``len(bunches)==ring.harmonic_number`` is required.
+                     (default=1, single bunch configuration).
+        """
+        if bunches == 1:
+            fp = numpy.ones(1)
+        elif numpy.isscalar(bunches):
```
lfarv, line 480: The special case `bunches == 1` does not create an array with the right size. You may take out this special case, `numpy.isscalar(1)` should work.

pyat/at/lattice/lattice_object.py
```diff
@@ -436,6 +455,71 @@ def particle(self, particle: Union[str, Particle]):
...
+    def bunch_spos(self):
+        """Bunch position around the ring [m]"""
+        bs = self.circumference/len(self._fillpattern)
+        allpos = bs*numpy.arange(len(self._fillpattern))
+        return allpos[self._fillpattern > 0]
+
+    @property
+    def nbunch(self):
+        """Number of bunches"""
+        return len(self.fillpattern)
```
lfarv, line 521:
```python
return numpy.count_nonzero(self._fillpattern)
```
is more efficient

lfarv:
>  I prefer it this way I think it is more readable / understandable

I can't believe that… The new version is so simpler !

Anyway, with the comments above, I think it can be merged. There are a few details that can be solved later:

- Due to the `nbunch` appearing in the `_disp_attributes`, saving and loading a lattice in the .repr format gives an unexpected result: the re-created lattice has a "normal" `nbunch` meaningless attribute, but a default filling pattern. Not crucial.
- The periodicity is not properly taken into account. Not critical, because a periodic lattice is not really adapted for multi-bunch tracking, unless the bunch pattern follows the same periodicity, which is a limitation. I guess you don't want to touch that, I'll take care later.

I'm a bit worried that "fillpattern' is not exactly the read/write property associated with `_fillpattern`: it would be more intuitive. Not critical, can be changed later.
I would also add a "bunch list" property, as another way to visualise the filling pattern.

So feel free to merge when you like.

lfarv (approved):
After minor modifications

pyat/at.c
swhite2401: Did you check this?

pyat/at/lattice/lattice_object.py
swhite2401: Ok but then I have to set a default value for `beam_current`, for me it is the same...anyway it is done

pyat/at/lattice/lattice_object.py
swhite2401: `numpy.ones(1)` is an accepted valid array, but I agree it is not useful, I have removed this functionality

pyat/at/lattice/lattice_object.py
swhite2401: Done

pyat/at/lattice/lattice_object.py
swhite2401: Ok but then I have to check that bunches is not a float, otherwise the error message is misleading (`len(bunches)` failing...)

swhite2401:
> > I prefer it this way I think it is more readable / understandable
> 
> I can't believe that… The new version is so simpler !
> 
> Anyway, with the comments above, I think it can be merged. There are a few details that can be solved later:
> 
>     * Due to the `nbunch` appearing in the `_disp_attributes`, saving and loading a lattice in the .repr format gives an unexpected result: the re-created lattice has a "normal" `nbunch` meaningless attribute, but a default filling pattern. Not crucial.
> 
>     * The periodicity is not properly taken into account. Not critical, because a periodic lattice is not really adapted for multi-bunch tracking, unless the bunch pattern follows the same periodicity, which is a limitation. I guess you don't want to touch that, I'll take care later.
> 
> 
> I'm a bit worried that "fillpattern' is not exactly the read/write property associated with `_fillpattern`: it would be more intuitive. Not critical, can be changed later. I would also add a "bunch list" property, as another way to visualise the filling pattern.
> 
> So feel free to merge when you like.

Ok final review I hope, I have implemated all your comments I believe.

Concerning the saving and loading of .repr format, I now force everything to default, at least value are consistent even though you lose the information on the `fillpattern`.

Concerning you comment on periodicity, I disagree. It is perfectly fine to simulate a non periodic fill pattern on a periodic lattice, both are independent, `periodicity` is already handled through the harmonic number. Please do not change this, adding more restrictions will be counter-productive. 
However, it is also true that it does not make any sense to do multi-particle tracking with `periodicity !=1` so I think this is a non issue.

I did more changes that expected, to I request for a last review if you have time...

swhite2401:
By the way, I very  often have to re-run the matlab test several times for them to complete properly, is this a known issue?

pyat/at/lattice/lattice_object.py
```diff
@@ -55,10 +55,13 @@ class Lattice(list):
     """
     # Attributes displayed:
     _disp_attributes = ('name', 'energy', 'particle', 'periodicity',
-                        'harmonic_number')
+                        'harmonic_number', 'beam_current', 'nbunch')
+    # excluded attributes
+    _excluded_attributes = ('nbunch', 'beam_current')
```
lfarv, line 60: I think that you should not exclude `beam_current`. It prevents initialising it with a keyword. It does not matter if `beam_current` is transmitted in save/load while `_fillpattern` is not, they are completely independent.

pyat/at/lattice/lattice_object.py
```diff
@@ -156,14 +181,17 @@ def __init__(self, *args,
...
         # set attributes
         self.update(kwargs)
 
         # Setting the harmonic number is delayed to have self.beta available
         if not (frequency is None or frequency == 0.0):
             rev = self.beta * clight / cell_length
             self._cell_harmnumber = int(round(frequency / rev))
+            self.set_fillpattern()
         elif not math.isnan(ring_h):
+            print('test')
```
lfarv, line 194: Remaining test output

pyat/at/lattice/lattice_object.py
```diff
@@ -436,6 +464,85 @@ def particle(self, particle: Union[str, Particle]):
...
+        """Filling pattern describing the bunch relative
+        amplitudes such that ``sum(fillpattern)=1``.
+        Calls the function ``Lattice.set_fillpattern``.
+        """
+        self.set_fillpattern(value)
+
+    @property
+    def bunch_list(self):
+        """Indices of filled bunches"""
+        return numpy.nonzero(self._fillpattern)
```
lfarv, line 526: It should be
```python
return numpy.flatnonzero(self._fillpattern)
```
otherwise you get a tuple enclosing the array

lfarv (approved):
I also hope it's the final one!

pyat/at/lattice/lattice_object.py
swhite2401: Right, I thought that setting `bunch_current=[beam_current] `behind the users back when reloading a file was dangerous...while setting the current at initialization was not so critical.
I'll remove it if you think it is better the other way

lfarv:
> By the way, I very often have to re-run the matlab test several times for them to complete properly, is this a known issue?

For a few days, I also noticed strange cancellations of Matlab test runs. Even before the test itself starts. No idea why…

Issue #454, opened 2022-08-02
pyAT: at.Mutlipole element creation issue


swhite2401:
For `at.Multipole` in case `K` is provided as input, `__init__()` fails because of a conflict with the property, this happens when saving a `.mat` lattice and then reloading it.
I suppose the property is defined here such that it is more convenient to create DQs, but this is a source of problems.
Adding the line `kwargs.pop('K', None)` in the `__init()__` of `at.Multipole` solves the problems but is not satisfactory...

Less critical: I do not understand why K is a required attribute for `at.Dipole`, can't it be made optional? I have also noticed that in the help k is defined as the 'field index', I am not sure what this means.

lfarv:
>  I do not understand why K is a required attribute for at.Dipole

`k` is an optional keyword argument. If you omit it, it is set to zero (no field gradient). But you are right concerning the help: it's the focusing strength, not the field index.

swhite2401:
> `k` is an optional keyword argument. If you omit it, it is set to zero (no field gradient). But you are right concerning the help: it's the focusing strength, not the field index.

Yes sorry it appears in the required attributes list...I got side tracked by this

lfarv:
@swhite2401 : Could you give the exact sequence that you tried? 

May be you tried to instantiate a `Multipole` element? `Multipole` in intended to be a base class, not to be instantiated. The rule is that you should use the class corresponding to the lower multipole present in the magnet:
- a DQ is a `Dipole` with a quadrupole component (crucial, otherwise you don't get the 'Bnd*' passmethod),
- a quadrupole-sextupole is a `Quadrupole` with a sextupole component
- ...

It's true that defining a pure dodecapole is ambiguous. I would use the Quadrupole class…

I'll look at what happens when instantiating a Multipole. In case of problems, there are 2 possibilities:
1. Correct the problem, if possible. At first glance, taking the `K` property out of `Multipole`, into both `Dipole` and `Quadrupole` looks simple,
2. turn `Multipole` into an abstract class

swhite2401:
@lfarv, you are right I instantiate a `Multipole`, that is in fact the problem.  This action has to be allowed, for example at ESRF lattices are all define with multipoles.
Multipole can be any arbitratry magnet and this is very convenient. For instance how would you define correctors doing H/V dipole + skew (ESRF), or corrector coils on octupoles (PSI)?

[...]

PR #455, opened 2022-08-04
Multipole can be instantiated

lfarv:
Issue #454 reveals a problem when saving/loading Multipole instances in .mat format. This is fixed here

### DIFF
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -421,7 +421,7 @@ class Multipole(LongElement, ThinMultipole):
 
 
 class _KMixin:
-
+    """Mixin class adding the K property to Dipole and Quadrupole"""
     # noinspection PyPep8Naming
     @property
     def K(self) -> float:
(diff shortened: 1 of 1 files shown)


swhite2401 (approved):
Ok for me

swhite2401:
@lfarv, sorry I take it back, save and loading still fails for again the same problem: `K` is defined as an attribute of multipole in the matlab lattice. You have to `pop()` it before you initialize the element otherwise `K` is set before `PolynomB` is defined (and I supposed the same should be implemented for `H`).

lfarv:
Hello @swhite2401

> save and loading still fails for again the same problem

I cannot reproduce your problem. `K` is no more saved for multipoles. Are you sure you don't try to load an old wrong file? Or use the wrong branch?

swhite2401:
Yes sorry, I looked in details and this was one specific case where I voluntarily instantiate a `Multipole` with`K` defined in `kwargs`.
I suppose we can live with this, I do not see why anyone would do this.

So ok for me to merge.

swhite2401 (approved):

Issue #456, opened 2022-08-10
pyAT get_tune and fit_tune using integer part of tune


simoneliuzzo:
In matlab AT the atfittune function gives the possibility to match the integer part of the tune using the flag 'UseIntegerPart'. 
This is apparently not possible in pyAT. Could it be added?

swhite2401:
Done in #458

PR #458, opened 2022-08-10
Fit integer tune

swhite2401:
This PR adds to possibility to fit the integer part of the tune either by set `fit_integer=True` as argument or providing a non-zero integer part for one of the target values.
Follow-up from #456

### DIFF
--- a/pyat/at/matching/globalfit.py
+++ b/pyat/at/matching/globalfit.py
@@ -4,69 +4,78 @@ such as the tune and the chromaticity
 """
 from typing import Optional
 import numpy
+from warnings import warn
 from ..lattice import Lattice, Refpts, set_value_refpts
+from ..lattice import AtWarning, AtError
 from ..physics import get_tune, get_chrom
 
 
 __all__ = ['fit_tune', 'fit_chrom']
 
 
-def _get_tune(ring: Lattice, dp: float):
-    return get_tune(ring, dp=dp)[0:2]
+def _get_tune(ring: Lattice, dp: float, **kwargs):
+    get_integer = kwargs.pop('fit_integer', False)
+    return get_tune(ring, dp=dp, get_integer=get_integer)[0:2]
 
 
-def _get_chrom(ring: Lattice, dp: float):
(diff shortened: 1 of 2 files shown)


simoneliuzzo:
Dear @swhite2401 , thank you for this addition. The function get_tune may be should have the same optional input? 
for example, to check that the tune correction is correctly done I would do:
```
initial_tune = get_tune(ring, ..., use_integer = True)
ring_tune = at.matching.fit_tune(ring, ...,fit_integer=True)
final_tune = get_tune(ring_tune, ..., use_integer = True)
```

simoneliuzzo:
Dear @swhite2401 

the help of at.matching.globalfit.fit_tune gives examples that are may be not up to date.
The help of get_tune could be complemented with the get_integer optional flag

simoneliuzzo:
works for me

simoneliuzzo (approved):

pyat/at/matching/globalfit.py
```diff
@@ -4,91 +4,105 @@
 """
 from typing import Optional
 import numpy
+from warnings import warn
 from ..lattice import Lattice, Refpts, set_value_refpts
+from ..lattice import AtWarning, AtError
```
lfarv, line 9: I have "Unused import statement" warnings. Apparently warn is not used… Are these lines left after tests?

pyat/at/matching/globalfit.py
```diff
@@ -4,91 +4,105 @@
...
         refpts1:    Selection of the 1st family
         refpts2:    Selection of the 2nd family
-        newval:     New tunes
+        newval:     New tunes, in case an non-zero integer part
+                    is provided, fit_integer is set to True
         tol:        Tolerance for the matching; Default: 1.0e-12
         dp:         Momentum deviation. Default: 0
         niter:      Maximum number of iterations. Default 3
-    
+        fit_integer: bool (default=False), use integer tune
```
lfarv, line 90: `fit_integer` is not in the argument list. It should be in a "Keyword Arguments:" section.

PR #459, opened 2022-08-10
patpass poolsize as global DConstant option

swhite2401:
This PR allows to set the pool_size of all functions using `patpass` using `DConstant.patpass_poolsize`


lfarv (approved):
Nice. No problem for me

Issue #433, opened 2022-07-04
atgeometry for pyAT


swhite2401:
If it does not exist, a function to get the elements 2D positions similar to `atgeometry()` would be a very useful addition to pyAT

simoneliuzzo:
Dear @swhite2401, I have written such functions recently. I doubt they are at pyAT level. The file is here: 
https://gitlab.esrf.fr/BeamDynamics/newbooster/DBA_long/-/blob/master/booster_lattice_tools/survey.py

swhite2401:
Very nice thanks!

lfarv:
Added in #460

PR #460, opened 2022-08-10
Compute and plot 2D ring geometry

swhite2401:
This is based on the code proposed by @simoneliuzzo in #433.
Two functions are provided:

- ``ring.get_geometry()`` to compute the ring geometry
- ``ring.plot_geometry()`` to plot and compute the ring geometry

### DIFF
--- a/pyat/at/lattice/utils.py
+++ b/pyat/at/lattice/utils.py
@@ -771,7 +771,7 @@ def get_geometry(ring: List[Element], start_coordinates=(0, 0, 0),
     yy = numpy.zeros((len(ring)+1, 1))
     angle = numpy.zeros((len(ring)+1, 1))
     x, y, t = start_coordinates
-    cost, sint = numpy.cos(t), numpy.sin(t) 
+    x0, y0, t0 = start_coordinates
 
     for ind, el in enumerate(ring+[ring[0]]):
         ang = 0.0
@@ -787,11 +787,11 @@ def get_geometry(ring: List[Element], start_coordinates=(0, 0, 0),
         yy[ind] = y
         angle[ind] = t
 
-    radius = get_s_pos(ring, len(ring))[0] / abs(t) \
+    radius = get_s_pos(ring, len(ring))[0] / abs(t-t0) \
         if t != 0.0 else 0.0
     if centered:
-        xx += -abs(radius)*sint - start_coordinates[0]
-        yy += abs(radius)*cost - start_coordinates[1]
+        xx += -abs(radius)*numpy.sin(t0) - x0
+        yy += abs(radius)*numpy.cos(t0) - y0
     geomdata['x'] = xx
     geomdata['y'] = yy
     geomdata['angle'] = angle
--- a/pyat/at/lattice/utils.py
+++ b/pyat/at/lattice/utils.py
@@ -20,7 +20,7 @@ import functools
 from typing import Callable, Optional, Sequence, Iterator, TypeVar, Union, Tuple
 from itertools import compress
 from fnmatch import fnmatch
-from .elements import Element
+from .elements import Element, Dipole
 
 Refpts = TypeVar("Refpts", int, Sequence[int], bool, Sequence[bool])
 ElementFilter = Callable[[Element], bool]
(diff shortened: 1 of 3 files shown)


pyat/at/lattice/utils.py
```diff
@@ -721,6 +722,12 @@ def set_tilt(ring: Sequence[Element], tilts, relative=False) -> None:
 
 def set_shift(ring: Sequence[Element], dxs, dzs, relative=False) -> None:
     """Sets the translations of a list of elements.
+    
+    Usage:
+    
+    .. code-block:: python
+    
+       ring.get_geometry()
```
lfarv, line 730: This seems to be in the wrong function (`set_shift` instead of `get_geometry`).

pyat/at/lattice/utils.py
```diff
@@ -734,3 +741,62 @@ def set_shift(ring: Sequence[Element], dxs, dzs, relative=False) -> None:
...
+        t -= ang
+        xx[ind] = x
+        yy[ind] = y
+        angle[ind] = t
+
+    try:
+        radius = ll / abs(t)
+    except ValueError as ex:
+        print(ex)
+        print('geometry: radius not computed, set to 0.0')
```
lfarv, line 793: Is this printout necessary? If so, it should be printed only if `centered` is True. Otherwise it doesn't matter. And should preferably converted in a warning, or printed to stderr.

pyat/at/lattice/utils.py
```diff
@@ -734,3 +741,62 @@ def set_shift(ring: Sequence[Element], dxs, dzs, relative=False) -> None:
...
+        t -= ang
+        x += ll * numpy.cos(t)
+        y += ll * numpy.sin(t)
+        t -= ang
+        xx[ind] = x
+        yy[ind] = y
+        angle[ind] = t
+
+    try:
+        radius = ll / abs(t)
```
lfarv, line 790: This implies that `ll` is the circumference. But is appears it's the length of the last element. Did I miss something?

pyat/at/lattice/utils.py
```diff
@@ -734,3 +741,62 @@ def set_shift(ring: Sequence[Element], dxs, dzs, relative=False) -> None:
...
+    except ValueError as ex:
+        print(ex)
+        print('geometry: radius not computed, set to 0.0')
+        radius = 0.0
+    rshift = radius if centered else 0.0
+
+    xx += offset[0]
+    yy += offset[1] + abs(rshift)
```
lfarv, line 798: I think that if you set a `t` offset (initial angle), the radius affects both `xx` and `yy`.

pyat/at/lattice/utils.py
```diff
@@ -734,3 +741,62 @@ def set_shift(ring: Sequence[Element], dxs, dzs, relative=False) -> None:
...
+    geom_dtype = [('x', numpy.float64, (1, )),
+                  ('y', numpy.float64, (1, )),
+                  ('angle', numpy.float64, (1, ))]
+    geomdata = numpy.recarray((len(ring)+1, ), dtype=geom_dtype)
+    xx = numpy.zeros((len(ring)+1, 1))
+    yy = numpy.zeros((len(ring)+1, 1))
+    angle = numpy.zeros((len(ring)+1, 1))
+    x, y, t = start_coordinates
+
+    for ind, el in enumerate(ring+[ring[0]]):
```
lfarv, line 775: The + operator cannot be used on a `Sequence[Element]` (line 746). Ideally, you would use `Lattice`, but `utils.py` must be imported before `lattice_object.py`, so `Lattice` is not available yet. So here I would use `List[Element]` as type hint for the `ring` argument.

simoneliuzzo:
Dear @swhite2401,

I think there is some problem with the radius computation.

When asking for centered=True for a ring lattice, the output radius is zero, and thus the coordinates are unchanged. The resulting figure is not centered on zero.

`
data, radius, ax = ring.plot_geometry(centered=True)
`





When computing the coordinates for a subset of the lattice, the radius seems also not correctly computed:

resulting figure is not centered on zero.

`
data, radius, ax = arc.plot_geometry(centered=True)
`

Screenshot 2022-08-16 at 14 41 01


I expected the radius to be independent of the portion of lattice requested (unless it is just a drift).

lfarv (approved):
It seems ok now !

simoneliuzzo (approved):

PR #483, opened 2022-09-09
Update packaging and installation

lfarv:
This updated the packaging and installation setup with two main changes:

#### Upgrade `setuptools`:

The main motivation is to ease the compilation of AT with MPI. To allow the modification of the build requirements depending on the context, setuptools now provides in-tree backends: `_custom_build/backend.py` here. The command for compilation with MPI is now:
```
MPI=1 pip install ".[mpi]".  # from the local clone
MPI=1 pip install --no-binary accelerator-toolbox "accelerator-toolbox[mpi]"  # from the PyPI repository
```
Note both the `[mpi]` option to add mpi4py to the **installation** requirements, and the `MPI=1` environment variable which will add mpi4py to the **build** requirements and select the `mpicc` compiler.

As a side effect, most of the configuration moves from `setup.cfg` to `pyproject.toml`. Completely removing `setup.cfg` works but is still considered as beta. So for the time being, we keep a minimal `setup.cfg`.

Also, `setuptools >= 62` cannot be installed on python 3.6. Removing support for python 3.6 also removes some complexity in `at/__init__.py`.

#### Move to project to the root of the repository:

1. This is a common practice to have the project definition at the top-level
2. This simplifies a lot the installation process: the source code of integrators (`atintegrators` directory) is now part of the project, so we do not need any more to make copy into `pyat/integrator-src`. `setup.py` is simpler and only deals with the definition of extensions.
3. The idea was also to have a single `machine_data` package for both Matlab and python by having the top-level `machine_data` also part of the project. This still does not work because of restrictions in the way `setuptools` looks for packages. It looks impossible to have package search at two levels, top-level for `./machine_data` and `pyat/at` for the `at` package. So we still have 2 `machine_data` repositories.

The difference for users is that `pip install` must now be run from the top-level instead of from the `pyat` directory. The documentation is updated.


lfarv:
In addition to the syntax for compiling with MPI involving an environment variable, there is now another syntax:
```
pip install --config-settings mpi=1 ".[mpi]".  # from the local clone
pip install --no-binary accelerator-toolbox --config-settings mpi=1 "accelerator-toolbox[mpi]"   # from the PyPI repository
```
This uses recent features of PEP 517. However it's only cosmetic, since the environment variable is still used internally to communicate with `setup.py`.

This is also valid for compilation with OpenMP:
```
pip install --config-settings openmp=1  .  # from the local clone
pip install --no-binary accelerator-toolbox --config-settings openmp=1 accelerator-toolbox   # from the PyPI repository
```

lfarv:
Rebased on `master`,

swhite2401:
Very nice, could you improve the doc page with these informations? I could not find any modifications related to MPI

lfarv:
@swhite2401 : right, I'll do that.

swhite2401 (approved):
Ok after help is updated

lfarv:
@swhite2401: I'd like to have your feedback on the updated documentation that you can preview [here](https://at.readthedocs.io/en/update_installation/p/index.html).

swhite2401:
> @swhite2401: I'd like to have your feedback on the updated documentation that you can preview [here](https://at.readthedocs.io/en/update_installation/p/index.html).

This looks fine to me, at some point we will provide examples but for now we are still working on the multibunch/beamloading devs

PR #493, opened 2022-09-21
Multibunch wakefieldpass

swhite2401:
This PR implements multibunch wakefield computation making use of the fill pattern information introduced in #452

The particles are distributed into bunches using `particle_number % number of bunches` and this on each MPI process. The advantage is that the bunch structure is entirely independent of the number of CPUs which is a strong limitation in some existing codes where `number of bunches = number of CPUs` was required.
The drawback is that each process has to fulfill  `number of particle >= number of bunches` which is the case for a vast majority of potential usage.

Changes associated to this branch:

- adaptation of `WakeFieldPass.c` and `atimplib.c` for multibunch
- simplification of `wake_elements.py` to use the fillpattern information and minor tweaks to related functions in `lattice_object.py`
- python function to access bunch information (mean and std) are provided in at.tracking.utils, these are very basic and should be extended based on user's experience

Mutlibunch wake calculation is activated by setting the fillpattern using:
```
ring.set_fillpattern(nbunch)
ring.beam_current  = current
```

Besides these the syntax is identical to single bunch calculations.


lfarv (changes requested):
`WakefieldPass.c` does not compile in Matlab:

WakeFieldPass.c:148:29: error: use of undeclared identifier 'PyExc_ValueError'

All files in `atintegrators` are compiled, even if they are not used. So you must provide the mex entry points (`mexFunction` and `trackFunction`), and provide a compatible way to throw exceptions: for instance see `AtGetLong` in the included `atelem.`c file.

If you want, I can modify `atelem.c` to provide a function to throw arbitrary errors.

swhite2401:
> `WakefieldPass.c` does not compile in Matlab:
> 
> WakeFieldPass.c:148:29: error: use of undeclared identifier 'PyExc_ValueError'
> 
> All files in `atintegrators` are compiled, even if they are not used. So you must provide the mex entry points (`mexFunction` and `trackFunction`, possibly just throwing an error if it cannot be used in Matlab), and provide a compatible way to throw exceptions: for instance see `AtGetLong` in the included `atelem.`c file.
> 
> If you want, I can modify `atelem.c` to provide a function to throw arbitrary errors.

Ok I will change this, I thought the mexfunction was optional.
A function to throw warning and errors would be useful yes.

lfarv:
Added the `atError(const char *format,...)` and `atWarning(const char *format,...)` functions. See the modified `WakefieldPass.c` for an example. However, it still blocks in Matlab:

```
Could not compile /Users/famille/dev/libraries/at/atintegrators/WakeFieldPass.c: Undefined symbols for architecture x86_64:
  "_mexFunction", referenced from:
     -exported_symbol[s_list] command line option
  "_passFunction", referenced from:
     -exported_symbol[s_list] command line option
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
Now, we need the 2 entry points !

lcarver:
Simulation of Transverse Resistive Wall for Q'=0. The formula (eq 31 of https://journals.iucr.org/s/issues/2014/05/00/xe5009/xe5009.pdf ) expects a mode -1 with a given growth rate. 

Simulating a case in uniform filling with similar un-converged parameters as above gives the following:

image

As before, the data was acquired by doing an SVD of the 992xNturn data set. As before, I will soon put an example file in the directory as this is a fundamental test.

swhite2401:
> Now, we need the 2 entry points !

Thanks! Very useful, I have added the mexfunction, it just prints an error message.

swhite2401:
> Simulation of Transverse Resistive Wall for Q'=0. The formula (eq 31 of https://journals.iucr.org/s/issues/2014/05/00/xe5009/xe5009.pdf ) expects a mode -1 with a given growth rate.
> 
> Ib = 200mA M=992 b=15mm sigma_c = 1e8
> 
> Simulating a case in uniform filling with similar un-converged parameters as above gives the following:
> 
> image
> 
> As before, the data was acquired by doing an SVD of the 992xNturn data set. As before, I will soon put an example file in the directory as this is a fundamental test.

Very nice! thanks @lcarver is it just an impression or the rise time seems systematically lower than theory?

lfarv:
> Thanks! Very useful, I have added the mexfunction, it just prints an error message.

OK, it now compiles correctly. But why not make it work ? At least the case `nrhs == 0` to get the list of required and optional fields…

lfarv (approved):
No compilation problem in Matlab any more, so it's ok !

swhite2401:
> OK, it now compiles correctly. But why not make it work ? At least the case `nrhs == 0` to get the list of required and optional fields…

Ok to print out the required fields, but the full implementation does not make much sense does it?

lfarv:
You're right, it does not look realistic to run this on a single core in Matlab. But `trackFunction`, the entry point used in tracking a lattice, works, so why `mexFunction` would not ? Anyway, keep it as it is, it's ok for me !

swhite2401:
> You're right, it does not look realistic to run this on a single core in Matlab. But `trackFunction`, the entry point used in tracking a lattice, works, so why `mexFunction` would not ? Anyway, keep it as it is, it's ok for me !

I have restored the mexfunction, still I think it is meaningless as the ring parameters cannot be passed...anyway it is done now, could be of some use for debugging...

lcarver (approved):
Added example files for LCBI and TRW which can be found in pyat/examples/CollectiveEffects

For me, it has passed the 2 fundamental tests for longitudinal and transverse, some more detailed benchmarking will of course take place but for me this implementation is working well and is correct.

Issue #463, opened 2022-08-18
confusing output of atx, ringpara, atsummary


simoneliuzzo:
![Screenshot 2022-08-17 at 11 31 02](https://user-images.githubusercontent.com/30342742/185085415-0898a5a9-5d36-44fe-b40b-5a0254a66f8e.png)

Computing the horizontal natural emittance of a lattice with `atx`, `ringpara` or `atsummary`, shifting the rf frequency with `atSetRingProperties` the values of emittance off-energy are largely different. 

For my experience, the values of `atx` are those reproduced in Control Room with beam.

Below the code to produce the image above (using Matlab_properties #443 branch)

```
 
drf = -300:50:300;

for i = 1:length(drf)

r = atSetRingProperties(ring,'rf_frequency',352372174+drf(i));

val = atsummary(r);

ex(i) = val.naturalEmittance;

[a,b] = atx(r);

ex2(i) = b.modemittance(1);

param = ringpara(r);

ex3(i) = param.emittx;


end

figure;
plot(drf,ex,'DisplayName','atsummary'); hold on;
plot(drf,ex2,'DisplayName','atx');
plot(drf,ex3,'DisplayName','ringpara');
xlabel('\Delta RF');
ylabel('\epsilon_h');
legend;
ax = gca;
ax.FontSize = 14;


```

lfarv:
@simoneliuzzo: you do not indicate if your lattice has radiation ON or OFF. I suspect ON, otherwise changing the frequency makes no sense.

 At the moment, `atsummary` and `ringpara` **cannot be used with an active cavity**: the need "radiation OFF". Changing the RF frequency has no effect (`ringpara`) or unpredictable results (`atsummary`). So:

1. the figure is irrelevant,
2. the problem is indeed a real one: both `atsummary` and `ringpara` have to be updated to accept:
   - 6d lattices by using `atlinopt6` instead of `atlinopt`
   - `dp` or `dct` specification on 6d lattice by acting of the RF frequency, using the `wrapper_6d` wrapper.

This is on the todo list.
However note that even when it's done, using a 6d lattice for `atsummary` or `ringpara` has no interest: they are based on linear optics, so you'll get identical results with more computation time.

Both functions are very similar, but each has its users. So we have to keep both, but ensure that they give identical results. The comparison with `atx` is different, because its method is completely different: Ohmi's envelope tracking based formalism versus analytical radiation integrals.

lfarv:
This issue was pointing out 2 different problems:
[...]

PR #500, opened 2022-09-28
"atsummary", "ringpara" and "atx" use "atlinopt6" for optics

lfarv:
This PR updates `atsummary`, `ringpara` and `atx` to use `atlinopt6` for optics. As a consequence:
- the input lattice may be either 4d or 6d ("radiation" ON or OFF),
- the off-momentum is properly taken into account. including for 6d lattices.

`atx` has deeper modifications:
- Since the input ring may be directly entered as a 6d lattice, optional syntaxes for configuring the "radiation ON" conversion are now useless. The are kept for compatibility but are deprecated,
- The optics method can be specified as a `'method'` keyword, with the following possibilities:
  - `@atlinopt6` is now the default value, applied to the 4d version of the lattice,
  - `@atlinopt4` may be used to apply the same method as the old version (it's equivalent to `atlinopt`),
  - `@atlinopt2` is available, but has very little interest.
  
  Some output fields are identical for the 3 methods (`alpha`, `beta`, `ClosedOrbit`, `Dispersion`…) but others are specific to each method,
- By default the linear optics is computed with the 4d version of the lattice, for speed and compatibility with the old version. However, with the `@atlinopt6` method, it can be computed with the 6d version by adding the `'6d'` flag.

This addresses some problems reported in #463. However there are still some incoherencies with the emittance. When off-momentum, the radiation integrals are computed with the corresponding modified values od betas and dispersion. However the result is still not in agreement with `atx`. A simpler example appears with the momentum compaction factor. It may be computed either directly by tracking (`mcf` function), or from the I<sub>1</sub> radiation integral. `atsummary` displays both results, which are in perfect agreement on-momentum, but differ off-momentum, suggesting that the I<sub>1</sub> integral is wrong. The only ingredients are dispersion and bending radius, so there is very little to do. Replacing the radius &rho; with &rho;*(1+&delta;) does not help…

Any idea, someone?


lnadolski:
Dear @lfarv 

It seems almost fine.
I do not understand why for esrf.m lattice

atx does not compute per default emittance, bunch length and so on.
I need to force atradon to get the expected result.

r = esrf;
atx(r) KO
atx(atradon(r)) OK
...
modemittance =
   NaN   NaN   NaN
modcoupling =
   NaN
projemittance =
   NaN   NaN   NaN
projcoupling =
   NaN
dampingtime =
   1.0e+07 *
  -0.002362334449781   0.408790442714825  -4.299921165235869
dampingJ =
   4.021027761813939  -0.023236874967468   0.002209113153529
fs =
     2.165810245594138e+03
espread =
     0
blength =
     0

lfarv:
Hello @lnadolski 

For some reason, the lattice generated by `esrf()` has active cavities but no radiation. That's why you get crazy damping times and no equilibrium emittance.

In the previous `atx` version, you had to provide a lattice **without longitudinal motion**, to get the optics with `atlinopt`, and it was internally switched to "radiation ON" with default options (RF, dipoles, quadrupoles) for `ohmienvelope`.

In this new version, the same thing happens if `'is_6d'` is false, but alternatively the user may provide a lattice **with longitudinal motion** (`'is_6d'` true). In this case, it is switched to "radiation OFF" for linear optics, and taken as it is for `ohmienvelope`. The advantage is that the user may tune the longitudinal motion as he likes. For instance here with the `esrf()` lattice: because of the cavity, `'is_6d'` is true but there is no damping.

I think this is how `atx` should behave:
- either you enter a lattice without longitudinal motion, and 6d is internally turned on with default options,
- or you enter a lattice with longitudinal motion, and you are responsible for the options.

lfarv:
However, you example shows one incoherency: the synchronous phase is computed with the energy loss per turn: `eloss=atgetU0(ron);`.
By default `atgetU0` uses the I<sub>2</sub> integral, which ignore the 6d state and takes all dipoles. So for the `esrf.m` lattice, `eloss` and `synchrophase` are not coherent with the other results.

`atgetU0` has another method using tracking which would be preferable here for coherence.

I'll look at that!

lfarv:
`atx`, `atsummary` and `ringpara` now get the energy loss with `atgetU0(ring,'method','tracking')` when the input has longitudinal motion ON. This gives losses according to the radiative elements declared in the lattice.

lnadolski:
Dear @lfarv 

Thanks for you reply.
I understand your statement.
At the same time, this is a change of habits for the users.
Up to now, atsummary, atx, etc were providing all parameters as soon as an active cavity where in the lattice.
The default lattice is in the control room and in most our simulations is a lattice with an active cavity for 6D tracking, 6D closed orbit, etc. but without any damping.

I would prefer to automatically activate radiation if not present in atsummary just for the computation of the radiation integrals, the bunch length etc.

Best regards,

Laurent.

simoneliuzzo (commented):
I would like to be replaced by someone else to review this pull request. I will be absent till January 2023.

lfarv:
> Up to now, atsummary, atx, etc were providing all parameters as soon as an active cavity where in the lattice.

**Problem**: it's just the opposite: up to now `atsummary`, `atx`, etc need as input a lattice **without any longitudinal motion**. They need a cavity in the lattice to get the RF voltage, but it has to be inactive (`IdentityPass`). The reason is that the optics computation uses `atlinopt`, which forbids any change of momentum. See the help in `atlinopt`:
```
%
%ATLINOPT only works for CONSTANT energy. So no PassMethod should:
%   1. change the longitudinal momentum dP (cavities, magnets with radiation)
%   2. have any time dependence (localized impedance, fast kickers etc)
%
```
The way it works until now is:
- `atsummary` and `ringpara` return analytical results: radiation integrals, longitudinal motion... So they don't need to turn anything on, neither cavities nor radiation. They work only with non-coupled lattices. And as I said earlier, it seems that radiation integrals are not correct off-momentum, even with the adequate optical functions
- `atx` computes optics with radiation and cavities OFF, and then turns them on to call `ohmienvelope`. It works in the general case.

This PR does not change anything in this case: if there is no radiation or cavity, it works exactly the same as before and gives the same results.

What this PR adds is the ability to call these functions **with radiation or cavities ON**. And in that case, the attempt is to give results corresponding to the properties of the input lattice: with or without damping.

**Context**: the problem of using `atlinopt` with radiation or cavities is long-standing issue. @simoneliuzzo may say a lot about that. The fundamental cause is the fact that the presence of radiation or cavities is not checked. And if you violate the rule, you get results, usually not so wrong and even possibly right in some conditions (on-momentum, without radiation and if the cavity is the 1st element in the lattice, the results are correct). The consequences of this bad practice are really problematic:
- Adding a check preventing the use of 6d motion in `atlinopt` and its derivatives would break many things, possibly even within AT. So it is postponed, probably for ever,
- trying to give correct results either by internally turning radiation and cavities OFF or using `atlinopt6` which accepts longitudinal motion changes the results, which also creates problems (like here),
- leaving things unchanged  allows users to use AT in a way which gives "almost correct", but in fact wrong, results. It is not acceptable in my opinion.

I may add that the wrong naming of `atradon`, `atradoff` `"radiation ON"` etc. added to the confusion by hiding their action on cavities. This should be better with the aliases recently introduced.

**So now**: what to do with this PR:
- forget it and leave things as they are: as I said before, I don't like the possibility of allowing users to get wrong results, even usually "slightly" wrong.
- turn radiation and cavities OFF as the 1st step of `atx`, `atsummary` etc. It may slightly change the results, and it cancels the purpose of the PR: allowing computation with 6D on,
- keep it, forcing users using bad practices to change their habits: in that case, the simplest is to initially define the lattice with 6d turned off: no radiation and inactive cavity. And turn 6D on for the very few computations involving longitudinal motion: 6D tracking, 6D closed orbit, 6D transfer matrix, `ohmienvelope`.

I understand your point, so we can take some time to find an acceptable solution…

simoneliuzzo:
Indeed, I have said a lot in the past, so I am not so keen to revive this discussion. 

In general I think the user should be free to : 
1) chose to run simulations with any combination of cavity and radiation states
2) set energy shifts in the lattice via dpp or cavity Frequency or both.

I personally do not know how to make this work.

May be accepting the (slightly) wrong results informing the user with a warning ('Warning: backward compatibility mode. The values of... and ... may be incorrect with RF .. and cavity .... . Please consider using <commands> instead.'), could be a solution. The motivation to accept wrong results would be to keep backward compatibility.

lfarv:
Hello @simoneliuzzo !

> chose to run simulations with any combination of cavity and radiation states

That's exactly what this PR tries. But it introduces changes pointed out by @lnadolski : a lattice with a cavity and no radiation gives 0 losses, 0 (or pi) synchronous phase, no equilibrium emittance and infinite damping times: it's correct, but it was no such before…


> set energy shifts in the lattice via dpp or cavity Frequency or both.

It's too bad you are leaving now: I hope you would appreciate the next PR to be proposed right after this one: specify everywhere the off-momentum through one of these 3 keywords: 
- dp=* (off-momentum)
- dct=* (path lengthening)
- df=* (RF frequency shift)

A single one, to avoid contradicting requirements.

lfarv:
> So now: what to do with this PR:
>
> - forget it and leave things as they are: as I said before, I don't like the possibility of allowing users to get wrong results, even usually "slightly" wrong.
> - turn radiation and cavities OFF as the 1st step of `atx`, `atsummary` etc. It may slightly change the results, and it cancels the purpose of the PR: allowing computation with 6D on,
> - keep it, forcing users using bad practices to change their habits: in that case, the simplest is to initially define the lattice with 6d turned off: no radiation and inactive cavity. And turn 6D on for the very few computations involving longitudinal motion: 6D tracking, 6D closed orbit, 6D transfer matrix, `ohmienvelope`.
> - Any other idea, someone?

@lnadolski : what do we do with the PR? Is it acceptable for you as it is, or do we postpone or modify?

lfarv:
No more comments, so I merge !

PR #501, opened 2022-10-12
Introduce an RF frequency shift input to specify off-momentum computation

lfarv:
Wherever there is an input for `dp` or `dct` to specify off-momentum computation, we can now alternatively use `df`, the RF frequency deviation from nominal. This off-momentum specification is closer to the way it is obtained in a real machine. So we can use one of these 3 keywords:

- dp=* (off-momentum)
- dct=* (path lengthening)
- df=* (RF frequency shift)

At this time, the following functions accept these inputs:

- `atsetcavity`
- `findorbit`(4D),  `findorbit4`, `findsyncorbit`
- `findm44`
- `atlinopt2`, `atlinopt4`, `atlinopt6`
- `tunechrom`

Missing functions will be added later (`atx`, `atsummary` and `ringpara` are pending in #500)


simoneliuzzo:
Dear Laurent, 
I find this is a really nice addition. 
I will not be able to review the pull request till January 2023. May be @carmignani could review the pull request instead of me?
all the best
Simone

Issue #502, opened 2022-10-18
Matlab: *QuantPass.c not working anymore


swhite2401:
@JonasKallestrup has found that the element-by-element photon emission does not work anymore.

After some testing it turns out that this problem is specific to matlab: particle are lost after they cross an element with `*QuantPass.c.`

The same tracking in pyAT works perfectly fine, I would therefore be tempted to blame the matlab `mex` command, and the way file are compiled in matlab. Any matlab expert have a good idea? @lfarv @lnadolski @carmignani @simoneliuzzo ?

A strange thing to note: adding a `printf()` statement in the passmethod magically fixes the issue.

There seem to be bugs in the passmethods:
-p_norm, NormL1, NormL2 probably need to be recomputed at each loop
-updating the r6[1] and r6[3] coordinates after the energy change does not seems right

These need to be confirmed and fixed as well.

swhite2401:
Below a python test script:

```
import at
import numpy
import matplotlib.pyplot as plt
from at.tracking import lattice_pass

path  = './'
filename = 'S28F.mat'
key = 'LOW_EMIT_RING_INJ'
latticef = path+filename
ring = at.load_lattice(latticef,key=key)
ring.radiation_on()

o6, _ = ring.find_orbit()
ncases = 5

xrad = []
for a in range(ncases):
    rout = numpy.squeeze(lattice_pass(ring, o6.copy(), refpts= range(60)))
    xrad.append(rout[0,:])

xquant = []
ring.radiation_off(cavity_pass='RFCavityPass')
rdip = ring.get_refpts(at.Dipole)
for d in ring[rdip]:
    d.PassMethod = 'BndMPoleSymplectic4QuantPass'
    d.Energy = ring.energy
    
for a in range(ncases):
    rout = numpy.squeeze(lattice_pass(ring, o6.copy(), refpts= range(60)))
    xquant.append(rout[0,:])


for i, xr, xq in zip(range(ncases),xrad,xquant):
    plt.subplot(211)    
    plt.plot(xr, label = 'Case #'+str(i))
    plt.subplot(212)    
    plt.plot(xq, label = 'Case #'+str(i))    
    
plt.subplot(211)    
plt.xlabel('Element index')
plt.ylabel('x [m]')
plt.legend()
plt.subplot(212)    
plt.xlabel('Element index')
plt.ylabel('x [m]')
plt.legend() 
plt.show()
```

swhite2401:
Below a matlab test script:

```
clear all;

[...]

PR #504, opened 2022-10-21
quantpass bugfixes

swhite2401:
This PR is a follow-up of #502. It was found that `*QuantPass.c` passmethods had erratic behavior and were producing wrong results.

Few bugs were identified and are fixed in the proposal. 
A more detailed code review and testing might be necessary for validation.

### DIFF
--- a/atintegrators/BndMPoleSymplectic4QuantPass.c
+++ b/atintegrators/BndMPoleSymplectic4QuantPass.c
@@ -80,9 +80,6 @@ void BndMPoleSymplectic4QuantPass(double *r, double le, double irho, double *A, 
         double *r6 = r+c*6;
         if (!atIsNaN(r6[0])) {
             int m;
-            double p_norm = 1/(1+r6[4]);
-            double NormL1 = L1*p_norm;
-            double NormL2 = L2*p_norm;
             /*  misalignment at entrance  */
             if (T1) ATaddvv(r6,T1);
             if (R1) ATmultmv(r6,R1);
@@ -104,9 +101,12 @@ void BndMPoleSymplectic4QuantPass(double *r, double le, double irho, double *A, 
                 double ng, ec, de, energy, gamma, cstec, cstng;
                 double ds, rho, dxp, dyp;
                 int nph;
+                double p_norm = 1/(1+r6[4]);
+                double NormL1 = L1*p_norm;
+                double NormL2 = L2*p_norm;
                 double dpp0 = r6[4];
-                double xp0 = r6[1]/(1+r6[4]);
-                double yp0 = r6[3]/(1+r6[4]);
+                double xp0 = r6[1]*p_norm;
+                double yp0 = r6[3]*p_norm;
                 double s0 = r6[5];
                 
                 fastdrift(r6, NormL1);
@@ -123,8 +123,8 @@ void BndMPoleSymplectic4QuantPass(double *r, double le, double irho, double *A, 
                 cstec = 3.0*gamma*gamma*gamma*clight/(2.0)*hbar/qe;
                 cstng = 5.0*sqrt(3.0)*alpha0*gamma/(6.0);
                 
-                dxp = r6[1]/(1+r6[4])-xp0 - irho*SL;
-                dyp = r6[3]/(1+r6[4])-yp0;
+                dxp = r6[1]*p_norm-xp0 - irho*SL;
+                dyp = r6[3]*p_norm-yp0;
                 ds = r6[5]-s0;
                 
                 rho = (SL+ds)/sqrt(dxp*dxp+dyp*dyp);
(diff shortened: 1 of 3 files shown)


swhite2401:
@JonasKallestrup I cannot add you as reviewer but it would be nice if you could take a look too

swhite2401:
First benchmarking test:

Comparison between (python syntax):
```
ohmi_envelop(ring.radiation_on())
lattice_pass(ring.radiation_on()+[quantdiff_elem])
lattice_pass(ring.radiation_on(dipole_pass='BndMPoleSymplectic4QuantPass', quadrupole_pass='StrMPoleSymplecticQuantPass')
```
This was done for the standard ESRF-EBS S28F lattice.

image

swhite2401:
Bencmarking continued:
This cases shows the energy loss after one quadrupole, in this case a very (unrealistic) strong quadrupole was used to increase the probability for photon emission

image

swhite2401:
Last verification:
Now I apply a frequency shift equivalent to a shift of dp/p of 1%, the following are expected:
-a reduction of horizontal emittance w.r.t. on-energy
-an increase of long. emittance w.r.t. on-energy
-> these are driven by the dispersive orbit in quadrupoles, this a good check that radiation in quadrupoles is properly implemented

I still track 1000 particles, the slicing of quadrupoles was increased and RFCavityPass has to be used.

image

swhite2401:
**Conclusions**: 
- There were several bugs in the present implementation
- Consistency checks of individual magnets and full ring simulation have been performed after the fixes
- All simulations show good agreement with expectations

**For the moment everything looks correct to me.** Comments and additional tests more than welcome

lfarv:
I tested the GetEnergy function by building the histogram of 10000 calls:


getenergy

The result is perfect: the generated values of photon energy follow the expected value

lfarv:
The threshold of 0.21 modified in the PR is the low limit of the tabulated value for photon energy. I looked at the behaviour around that point: below 0.21, the energy is computed analytically (21 % of the emitted photons), above it is interpolated in a table (79 % of the photons). The high end is used for only 1/100000 of the photons !



discontinuity

There is a small gap in energy at &epsilon;/&epsilon;<sub>0</sub> = 0.005, with a width of 0.00014. This is negligible and does not appear in the histogram (see previous comment).

So the generation of photon energy is correct !

atintegrators/BndMPoleSymplectic4QuantPass.c
```diff
@@ -123,27 +123,25 @@ void BndMPoleSymplectic4QuantPass(double *r, double le, double irho, double *A,
...
                 de = 0.0;
                 if(nph>0){
                     for(i=0;i<nph;i++){
                         de = de + getEnergy(ec);
                     };
-                };
-                
-                r6[1] = r6[1]/(1+r6[4])*(1+r6[4]-de/E0);
-                r6[3] = r6[3]/(1+r6[4])*(1+r6[4]-de/E0);
+                };
```
lfarv, line 141: Just a detail: the first test is redundant: the loop will not execute if the initial condition is not fulfilled…

lfarv (approved):

atintegrators/BndMPoleSymplectic4QuantPass.c
swhite2401: Right, corrected

swhite2401:
@lfarv, should we use this branch to provide on/off features?

lfarv:
@swhite2401 : I think we can merge this now: it solves a real bug, while quantum diffusion can still be turned on/off using `enable_6d`/`disable_6d`, even if it is not so convenient.

For the following, I even see 2 steps:
1. Correcting the fact that `check_6d` and `disable_6d` ignore the `QuantPass` suffix. It's a potentially dangerous bug,
2. Find the best way to  enable local quantum radiation, either with a dedicated function, of by adapting `enable_6d`. For instance one could imagine replacing the 'auto' code which selects the adequate pass method by either 'rad' for "classical" radiation or 'quantum' for quantum radiation. A dedicated function might be simple to use, I just would try to avoid duplicating the `enable_6d` code…

lfarv:
I have a doubt concerning OpenMP, or other parallel execution: will the `rand()` function generate the same sequence in the different threads (which would be bad), or still generate a new number on each call, possibly from different threads ? It's not clear in the POSIX documentation, and may depend on the implementation. GNU libc specifies that `rand` is thread safe, is it enough ?

swhite2401:
> I have a doubt concerning OpenMP, or other parallel execution: will the `rand()` function generate the same sequence in the different threads (which would be bad), or still generate a new number on each call, possibly from different threads ? It's not clear in the POSIX documentation, and may depend on the implementation. GNU libc specifies that `rand` is thread safe, is it enough ?

Very good point! I am afraid we have to initialize the random number generator with different seeds in each thread....that is quite a bit of a change w.r.t. the present implementation

swhite2401:
@lfarv so I definitely confirm this is a problem, and not only for this part but for functions using random number generation:
-particles distribution
-lumped diffusion element
-element by element photon emission
-others?

and this for all parallelized processes: MPI/OpenMP/python multiprocessing

The fix for each case is different and will have to take place in different parts of the code, but generally speaking we have have to produce a different seed for each thread.

The most difficult case at first sight is the python multiprocessing as I have no idea on how to pass the process number to the C engine in this case...

What I propose is to merge this branch as is because this is self contained and fixes obvious bugs and then make a separate PR to handle random number generation for parallelized applications

What do you think?

lfarv:
@swhite2401 : I agree to merge this branch right now.

It would be useful to make some tests to see what happens for different parallel methods and rand() implementations, but if looks not so simple…

JonasKallestrup:
> @JonasKallestrup I cannot add you as reviewer but it would be nice if you could take a look too

I've done the tracking simulation with the SLS 2 lattice (i.e. with anti-bends) and with a strong dispersive coupling source, and it results in the correct transverse emittances.

swhite2401:
Yes I fully agree testing is necessary but complicated...

Thanks both for your feedback, I merge this as is then

PR #505, opened 2022-10-23
Introduce an RF frequency shift input to specify off-momentum computation

lfarv:
Wherever there is an input for `dp` or `dct` to specify off-momentum computation, we can now alternatively use `df`, the RF frequency deviation from nominal. This off-momentum specification is closer to the way it is obtained in a real machine. So we can use one of these 3 keywords:

- dp=* (off-momentum)
- dct=* (path lengthening)
- df=* (RF frequency shift)

This is similar to what #501 introduced for Matlab

The following functions are modified to introduce the `df` keyword:

- `find_orbit4`, `find_syncorbit`, `find_orbit`
- `find_m44`
- `get_revolution_frequency`
- `linopt2`, `linopt4`, `linopt6`, `get_optics`, `avlinopt`
- `get_tune`, `get_chrom`

As a side effect, this PR also introduces new properties for the lattice object:
- `cell_length`
- `cell_revolution_frequency`
- `cell_harmnumber`

They are the 1-cell equivalent of the unchanged following properties related to the full ring (periodicity x cells):
- `circumference`
- `revolution_frequency`
- `harmonic_number`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -57,6 +57,10 @@ templates_path = ['_templates']
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ["README.rst", "**/*.so"]
+rst_prolog = """
+.. role:: pycode(code)
+   :language: python
+"""
 autodoc_default_options = {
         # Make sure that any autodoc declarations show the right members
         "members": True,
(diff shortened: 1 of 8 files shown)


swhite2401:
Very useful, would it make sense to issue an error if 2 contradicting argument are provided. I saw you mention in the help that `dct` and `df `overwrite `dp` but is that sufficient?

lfarv:
@swhite2401 : the test is added

swhite2401 (approved):

PR #507, opened 2022-10-24
Take "QuantPass" into account when checking 6D motion

lfarv:
The passmethods ending in `QuantPass` are forgotten when checking for 6D motion or disabling it. this PR fixes that. It upgrades:

- **Matlab**: `check_6d`, `check_radiation`, `atenable_6d`, `atdisable_6d`, `atradon`, `atradoff`
- **python**: `check_6d`, `check_radiation`, `Lattice.enable_6d`, `Lattice.disable_6d`, `Element.get_longt_motion`, `Element.set_longt_motion`

`enable_6d` is unchanged: by default it enables "classical" radiation by setting '*RadPass', possibly replacing '*QuantPass' if necessary.

Enabling *QuantPass can be explicitly required, but an easier way to do that is still to be discussed.

### DIFF
--- a/atmat/atphysics/Radiation/atdisable_6d.m
+++ b/atmat/atphysics/Radiation/atdisable_6d.m
@@ -103,7 +103,8 @@ end
         
         function elem=varelem(elem)
             % 'auto' multipole modification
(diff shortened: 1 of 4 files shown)


swhite2401 (approved):
OK!

PR #511, opened 2022-10-31
Secure the emittance computation in atsummary and ringpara

lfarv:
This is a follow-up of #463. `atsummary` and `ringpara` provide a computation of the equilibrium emittance based on the radiation integrals in dipoles. This value may be wrong for off-momentum lattices because it misses the contribution of quadrupoles in dispersive regions.

In this PR, the output of equilibrium emittance, energy spread and derived parameters is suppressed for off-momentum lattices, and a warning is issued.

### DIFF
--- a/atmat/atphysics/ParameterSummaryFunctions/atsummary.m
+++ b/atmat/atphysics/ParameterSummaryFunctions/atsummary.m
@@ -1,5 +1,22 @@
 function varargout = atsummary(varargin)
-%ATSUMMARY  Print out the parameters of the current AT lattice
+%ATSUMMARY  Print out various parameters of the current AT lattice
+%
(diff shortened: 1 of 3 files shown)


simoneliuzzo:
Dear @lfarv,

sorry, I could not find time to look at this pull-request. 

I think that a removed output will likely create (me) some backward compatibility issue. I will try to solve them in my code when/if they will appear. 

I am in the process to give up matlab. Hopefully by the end of 2023 I will be mostly converted to pyAT.

lfarv:
Hello @simoneliuzzo.

I don't think you'll miss anything: the output are removed if dp !=  0, but there was no `dp` input before #500. Unless you jumped immediately on that version (12 days ago), you should not be concerned. Though I know you are expert in discovering unexpected issues…

PR #515, opened 2022-11-14
Improve patpass

lfarv:
`patpass` provides parallel tracking using python multiprocessing. Up to now, `patpass` spawns one process per particle to be tracked. This is not optimal since it usually needs much more processes than available in the process pool. The processes are kept waiting until one slot in the process pool is available. So there is a large overhead in starting many processes and gathering the results.

Here we split the input of particle coordinates in as many slices as available processes in the pool, and then start one process per slice. The number of processes is equal to the number of possible parallel processes.

the output formatting is also simplified.

This will simplify the handling of random number generators.

### DIFF
--- a/pyat/at/tracking/patpass.py
+++ b/pyat/at/tracking/patpass.py
@@ -7,54 +7,53 @@ from at.lattice import uint32_refpts
 from warnings import warn
 # noinspection PyUnresolvedReferences
 from .atpass import atpass as _atpass
+from .track import fortran_align
 from at.lattice import AtWarning, DConstant
-import numpy
+import numpy as np
 
 
 __all__ = ['patpass']
 
+_atpassf = fortran_align(_atpass)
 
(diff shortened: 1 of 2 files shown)


swhite2401 (approved):
Looks ok to me, did you compute the speedup?

lfarv:
>> did you compute the speedup?

I probably depends on the number of particles, platform… One example:
MacOS, poolsize 4, 500 particles, 1000 turns of hmba: 5.5s instead of 6s, so ~10% improvement.

swhite2401:
> I probably depends on the number of particles, platform… One example:
> MacOS, poolsize 4, 500 particles, 1000 turns of hmba: 5.5s instead of 6s, so ~10% improvement.

Ah ok so not so huge, all good for for me!

MJGaughran (approved):

PR #516, opened 2022-11-15
Multibunch Beamloading

swhite2401:
The multi-bunch beam loading is introduced. Several additions are present in this PR:
- a passmethod `RFCavityBeamloadingPass.c` extenting the functionalities proposed in #191 with multibunch, MPI, a 2 energy loss calculation methods
- a python interface for element creation and adding/removing beamloading from cavities
- adaptation of examples and minor modifications of the `collective` subpackage and `lattice_object.py`
- Examples are updated

The proposed strategy is to modify the cavity elements in-place to integrate the beam loading, all the calculations are the done in the C tracking engine. Similarly beam loading can be removed from cavities.
The energy loss in the cavity can be calculated using the phasor method or the wake function, both are compatible with MPI and multibunch tracking.
A feedback loop is integrated and can be tuned using the `*Gain` arguments, this allows to maintain the effective voltage and cavity phase constant using the tuning angle and generator amplitude regardless of the beam induced voltage. This feedback is presently ideal but can be further extended to integrate more realistic circuits.

This implementation was benchmarked using `mbtrack` and theory.


pyat/at/physics/energy_loss.py
```diff
@@ -68,13 +68,7 @@ def dipole_i2(dipole: Dipole):
...
         """Losses from tracking
         """
-        ringtmp = ring.radiation_off(dipole_pass=None,
-                                     quadrupole_pass=None,
-                                     wiggler_pass=None,
-                                     sextupole_pass=None,
-                                     octupole_pass=None,
-                                     copy=True)
-
+        ringtmp = ring.disable_6d(copy=True)
```
lfarv, line 71: That's wrong: By default (no argument), disable_6d disables everything, and you get 0 losses! You can use:
```python
ringtmp = ring.disable_6d(RFCavity, copy=True)
```

lfarv:
3 comments:

1. Is it possible to change the passmethod name to `BeamLoadingCavityPass` instead of `RFCavityBeamLoadingPass`? In Matlab, disabling 6d cannot use classes (they don't exist), so it is based on names: cavities are identified by names ending in 'CavityPass'. Beamloading cavities would fall in that range.

2. Similarly, in python, disabling beamloading cavities would be easier if they inherit from `RFCavity`. So that `disable_6d(RFCavity)` will also disable beamloading cavities. At 1st glance, this looks possible:
   - By setting the order of base classes to `(RFCavity, Collective)`, you can inherit `_get_long_motion` and `set_long_motion` from `RFCavity`, while at the moment `set_long_motion` does not properly handle long cavities: they become `Identity` instead of `Drift` on a disable event. For this, you also need the name modification of point 1,
   - You could overload `is_compatible`, `_part` and `merge`. The simplest would be for `is_compatible` to always return False, there is no big interest in splitting and merging beamloading cavities. Or you can fully implement them. At the moment, they are inherited from `Element`, which is not correct: a beam loading element is now compatible with any other beam loading element, whatever their attributes are.
 
   If this works (to be checked, multiple inheritance is tricky), there is then another possible change. That would be to have a "standard" constructor for creating a beamloading cavity from scratch, by extending the `RFCavity` constructor with the necessary parameters: ring and resonator parameters. And add a factory function for creating a beamloading element based on an existing RF cavity, to be used in lattice conversion.
  This additional modification is for cleanliness, but is not strictly necessary.

3. Is it possible to have the `trackCavity` C function common with RFCavityPass.c for easier maintenance?

pyat/at/physics/energy_loss.py
swhite2401: You are right...but I don't understand how this could have worked and went unnoticed...

swhite2401: Done

swhite2401:
> 1. Is it possible to change the passmethod name to `BeamLoadingCavityPass` instead of `RFCavityBeamLoadingPass`? In Matlab, disabling 6d cannot use classes (they don't exist), so it is based on names: cavities are identified by names ending in 'CavityPass'. Beamloading cavities would fall in that range.

Ok. This is done. Comments 2 and 3 will take more time

swhite2401:
> 3\. Is it possible to have the `trackCavity` C function common with RFCavityPass.c for easier maintenance?

Done

lcarver:
1. I have added an example file that should work with both MPI on or off. It takes the old ESRF machine and runs a case with 16 symmetrically placed bunches. This makes the comparison with the analytical formula. We can observe that we are slightly lower than the analytical value. it is possible to see the coherent tune close to where it is expected, and the incoherent tune which is on the right hand side

image


2. I attach below a plot I made for 16 bunch filling, where I compare the WAKE and PHASOR BeamLoading methods with MBTRACK2 (Phasor method exclusively) and the analytical theory. This is for the EBS lattice. As the plot shows, we have very good agreement between WAKE and PHASOR, which is all slightly lower than the analytical. 

16b_benchmark

swhite2401:
I have implemented the suggested changes. @lcarver be careful the syntax has changed.
Other ideas / suggestions?

lcarver (approved):

pyat/at/lattice/lattice_object.py
```diff
@@ -540,6 +541,15 @@ def particle(self, particle: Union[str, Particle]):
...
             self._particle = Particle(particle)
 
+    def set_wake_turnhistory(self):
+        """Function to reset the shape of the turn history
+        in collective elements based on the number of slices,
+        turns and bunches
+        """
+        for e in self:
+            if e.is_collective:
+                e.clear_history(ring=self)
```
lfarv, line 551: This assumes that all `Collective` elements have a `clear_history()` method. This could be enforced by adding an `@abstact_method` to the `Collective` class. I propose an another PR for that, this one is large enough!

lfarv:
@swhite2401 and @carmignani : could you both confirm that this PR supersedes #191, and that that #191 could be closed?

pyat/at/collective/beam_loading.py
```diff
@@ -0,0 +1,282 @@
...
+                        PhaseGain=float, VoltGain=float, _mode=int,
+                        _beta=float, _wakefact=float, _nslice=int,
+                        ZCuts=lambda v: _array(v),
+                        _nturns=int, _phis=float,
+                        _turnhistory=lambda v: _array(v),
+                        _vbunch=lambda v: _array(v),
+                        _vbeam_phasor=lambda v: _array(v),
+                        _vbeam=lambda v: _array(v),
+                        _vcav=lambda v: _array(v),
+                        _vgen=lambda v: _array(v))
```
lfarv, line 121: Apparently, `_vgen` must have shape (2,) and can be set through `@Vgen.setter`. You could enforce the right shape with:
```python
_vgen=lambda v: _array(v, shape=(2,)))
```
It's also true for other attributes, but if they cannot be set by the user, it's not so important.

pyat/at/collective/beam_loading.py
```diff
@@ -0,0 +1,282 @@
+import numpy
+from enum import IntEnum
+from ..lattice import Lattice
+from at.lattice import Element, RFCavity, Collective
+from at.lattice.elements import _array
+from at.lattice.utils import Refpts, checktype
```
lfarv, line 6: `Element` and `check_type` are not used.

atmat/attests/pytests.m
```diff
@@ -90,7 +90,7 @@ function orbit6(testCase,lat2)
             porbit6=double(porbit6)';
             % Matlab
             [~,morbit6]=findorbit6(lattice.m);
-            testCase.verifyEqual(morbit6,porbit6,AbsTol=1.E-15 );
+            testCase.verifyEqual(morbit6,porbit6,AbsTol=5.E-15 );
```
lfarv, line 93: Did anything changed in the computation of the closed orbit? Do you get problems with the previous value?

swhite2401:
> @swhite2401 and @carmignani : could you both confirm that this PR supersedes #191, and that that #191 could be closed?

All the functionalities of #191 are integrated in this one, so yes it can be closed

swhite2401:
> About [remove mail adresses from C files](https://github.com/atcollab/at/pull/516/commits/a4398dacd197f711ceca9d78d9c581086b2d6835): it would have been preferable to do this in another PR: you are mixing different things and increasing a lot the number of modified files ! For now, keep it here, of course!

Yes sorry, I realized after that this was a bit bold...anyway now it is done

atmat/attests/pytests.m
swhite2401: I did not change anything but the matlab test was systematically failing so I just matched the tolerance of the matlab test to the one of the python test.

lfarv: Failing on GitHub ? It still succeeds on the other branches, so it looks strange…

swhite2401: So there is one thing. `energy_loss.py` was changed and is used to determine the initial condition for `find_orbit6`. This could have an impact on the final result to that level no?

pyat/at/lattice/lattice_object.py
swhite2401: Right! Maybe put an issue as a reminder? Or we just add it here it is a small addition

atmat/attests/pytests.m
lfarv: Ok, that could be an explanation. Do you confirm that it fails on GitHub (which platform?), because it still works for me (MacOS).

pyat/at/lattice/lattice_object.py
lfarv: As you like !

atmat/attests/pytests.m
lfarv: I found the last failure in the GitHub history: it failed on ubuntu-latest, with an error of 3.2e-15 instead of 1e-15. We don't have the results on the other platforms because all runs are cancelled as soon as one fails.

So let's assume `energy_loss.py` is the reason for the change.

swhite2401: The matlab tests fail on GitHub, I could not install the matlabengine in my pyenv, either it complains about the matlab version or the python version so I gave up.

lfarv:
@swhite2401 : another question: why do you implement you own turn counter in the integrator? This prevents you from stopping/restarting the tracking, and it may lead to different turn numbers depending on the integrator.

Also, by relying upon the "standard" turn counter, you could clear the turn history on turn zero in the integrator, avoiding the `Lattice.set_wake_turnhistory`. Well, this would either be a rewriting in C or a call to python from C, so it may not be a good idea…

swhite2401:
> @swhite2401 : another question: why do you implement you own turn counter in the integrator? This prevents you from stopping/restarting the tracking, and it may lead to different turn numbers depending on the integrator.
> 
> Also, by relying upon the "standard" turn counter, you could clear the turn history on turn zero in the integrator, avoiding the `Lattice.set_wake_turnhistory`. Well, this would either be a rewriting in C or a call to python from C, so it may not be a good idea…

Well, I was facing issues with memory allocation. The length of the position and size vector are fixing in lattice_pass bu the argument `nturn`, if you restart the tracking from turn `n` you have to modify this length and then the `keep_lattice` option cannot be used.
I made the assumption that the user going out of the C tracking engine would fetch the data, and then restarting the tracking would be equivalent to acquiring a new data set... this may not be optimal.
You have any suggestions?

Concerning the clear history, you may not necessarily want to clear it for a new tracking...maybe the oppsoite in fact to start from steady state. For me it is more relevant when you change beam parameters like the `fillpattern` and the history is clearly not valid anymore

lcarver:
So I have already approved these changes. but I did some further tests following comments from @carmignani and @lfarv. Below are some quick plots showing some expected effects.

incoherent
992 bunches with single particles. Ib=200mA. I kick one bunch and get the incoherent tune and I kick all bunches and get the coherent tune. This is as expected.

singlebunch_incoherent
I tried the same with a single bunch at 200mA. Now I try to kick one particle within the bunch for the incoherent and kick the full bunch for the coherent. This was trickier to simulate, I needed Nslice=50 and I had a big problem with not having enough turns to sufficiently sample the frequency I was trying to find. Strong kicks were needed, but eventually this could be simulated. However this is not a realistic case so it is good that it works as expected but it will not come up in regular use.

lfarv:
@lcarver : thanks for the checks: very nice, it's convincing!

PR #519, opened 2022-11-16
Elemrnt rotations

swhite2401:
This PR is a follow-up of #512. It was pointed out that there was no function to provide pitch and yaw in python. The proposed solution was using the wrong coordinate system and needed to be improved.

The change of coordinate from angle (input argument) to momentum is done using the `R1` and `R2` matrices.

`relative` option is still available but a bit tricky because some of the transformations are not commutative. 
This implementation is a at the stage of a proposal, ideas and suggestions welcome!

If we are happy with this, it can easily be ported to matlab.

### DIFF
--- a/pyat/at/lattice/utils.py
+++ b/pyat/at/lattice/utils.py
@@ -720,6 +720,14 @@ def rotate_elem(elem: Element, tilt: float = 0.0, pitch: float = 0.0,
     the ``y``-axis.
     A positive angle represent a clockwise rotation when looking in
     the direction of the roation axis.
+    
+    The transformations are not all commmutative, the tilt is always the
+    last transformation applied.
+    
+    In case ``relative=True``, the previous angle and shifts are derived
+    form the R and T matrix and incremented by the input arguments.
+    
+    The shift is always conserved regardless of the value of ``relative``
 
     Parameters:
         elem:           Element to be tilted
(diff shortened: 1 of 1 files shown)


pyat/at/lattice/utils.py
```diff
@@ -709,6 +712,72 @@ def get_s_pos(ring: Sequence[Element], refpts: Optional[Refpts] = None) \
...
+        ap, ay = -0.5*le*numpy.sin(pitch), -0.5*le*numpy.sin(yaw)
+        rr1 = numpy.asfortranarray(numpy.diag([ct, ct, ct, ct, 1.0, 1.0]))
+        rr1[0, 2] = st
+        rr1[1, 3] = st
+        rr1[2, 0] = -st
+        rr1[3, 1] = -st
+        rr2 = rr1.T
+        t1 = numpy.array([ay, -yaw, ap, -pitch, 0, 0])
+        t2 = numpy.array([ay, yaw, ap, pitch, 0, 0])
+        rt1 = numpy.asfortranarray(numpy.diag(numpy.ones(6)))
```
lfarv, line 745: Could be:
```python
rt1 = numpy.eye(6, order='F')
```
And the same in other locations.

swhite2401:
@elafmusa, I could not add you as reviewer but could you please check this is doing what you expected? Thanks!

lfarv:
@swhite2401 : Just for my understanding: why is it necessary to round the rotation values?

lfarv (approved):

swhite2401:
> @swhite2401 : Just for my understanding: why is it necessary to round the rotation values?

I had rounding errors of ~1e-19 when applying a transformation and then reverting if with `-angle` and was not getting exactly identity and zero translation.... I got annoyed by this so I am rounding the angles.
I suppose this comes from the calculation of the previous angles a shift.

It is not necessary, more of a cosmetic measure

simoneliuzzo:
I think this is a nice addition. I do not have time to review, so I removed myself from reviewers.

swhite2401:
> @swhite2401 I tested the implemented functions, and it works as expected thanks!

Ok I merge, you may switch back to the master branch

Issue #517, opened 2022-11-15
scipy requirements too low


swhite2401:
We are presently requiring `scipy =>0.16`
However, in `constants.py` we declare

`_hbar_c = _cst['reduced Planck constant times c in MeV fm'][0]`

It looks , the key `reduced Planck constant times c in MeV fm` was introduced in scipy version 1.4.0, the requirements have therefore to be updated to `scipy =>1.4.0`.

swhite2401:
Fixed in #520

PR #520, opened 2022-11-16
Upgrade scipy requirement to 1.4.0

lfarv:
#517 reported a problem with scipy constants. Requiring scipy >= 1.4.0 solves this. Scipy 1.4.0 was issued on 17/12/2019 and includes the upgrade of all physical constants to the [CODATA 2018 values](https://physics.nist.gov/cuu/Constants/index.html).


swhite2401:
You do not need to change this in the `pyproject.toml` as well?

PR #521, opened 2022-11-18
Beam monitor

swhite2401:
A pass method to compute the std and mean of the bunches is provided.
The results (arrays) in attributes of the elements.

This allows to get turn-by-turncenter of mass and sizes without coming out of the C tracking engine.
The passmethod is compatible with MPI.


lcarver:
Hello,
Works fine for me. 

bm_elem = at.BeamMoments('monitor')
ring.append(bm_elem)

then I have to run with
_ = at.lattice_pass(ring, part, nturns=100)

then bm_elem._positions has shape (6,Nbunches,Nturns) as expected.

I also tested with MPI, and I found that on each rank the bm_elem._positions was the same. So I would say that MPI was working.

Few comments:
Can we change the name from _sizes to something different. Maybe sigmas?
I am considering a possible extension to this PassMethod that may allow for extraction of slices as well. But it would be an extra as this provides the basic function of means and stds

lcarver (approved):

lcarver:
As an extra note, I see some signs of slow down with MPI on when the monitor is added. I think it is to be expected but will be need to be properly quantified.

swhite2401:
> I also tested with MPI, and I found that on each rank the bm_elem._positions was the same. So I would say that MPI was working.

Yes tis is expected, the moments are distributed on all ranks

swhite2401:
> Can we change the name from _sizes to something different. Maybe sigmas?

This can be done yes

swhite2401:
> As an extra note, I see some signs of slow down with MPI on when the monitor is added. I think it is to be expected but will be need to be properly quantified.

This is interesting, could you please evaluate the slow down? Depending on its significance we may worry or not. The monitor requires to exchange data between threads a small slow down is therefore expected w.r.t. the fully parallel case. This slow-down should be independent or the number of particles as only average data are passed.

If it is large or strongly depends on the particle number then we have a problem.

lfarv:
Don't we miss 2 properties to provide a public read-only access to the private attributes `_positions` and `_sizes`?

atintegrators/BeamMomentsPass.c
```diff
@@ -0,0 +1,130 @@
...
+
+void BeamMomentsPass(double *r_in, int nbunch, int num_particles, struct elem *Elem) {
+
+    int turn = Elem->turn;
+    double *sizes = Elem->sizes;
+    double *positions = Elem->positions;
+        
+    int i, ii, ib; 
+    size_t sz = 2*nbunch*6*sizeof(double) + nbunch*sizeof(double);
+    void *buffer = atMalloc(sz);
```
lfarv, line 31: Where is `buffer` freed ? Am I missing something?

atintegrators/BeamMomentsPass.c
```diff
@@ -0,0 +1,130 @@
...
+    double *sizep=dptr; dptr += nbunch*6;
+    double *nparts=dptr;
+    
+    for (i=0; i<nbunch; i++) {
+        nparts[i] = 0.0;
+        for(ii=0; ii<6; ii++) {
+            avep[i*6+ii] = 0.0;
+            sizep[i*6+ii] = 0.0;
+        }
+    }
```
lfarv, line 44: You can avoid initialisation by using `atCalloc` which initialises the allocated memory with 0 (as `calloc` does).

pyat/at/lattice/elements.py
```diff
@@ -435,6 +435,21 @@ class Monitor(Element):
...
 
+class BeamMoments(Element):
+    """Element to compute bunches mean and std"""
+
+    def __init__(self, family_name: str, **kwargs):
+        self.FamName = family_name
+        kwargs.setdefault('PassMethod', 'BeamMomentsPass')
+        self._sizes = numpy.zeros((6, 1, 0), order='F')
+        self._positions = numpy.zeros((6, 1, 0), order='F')
+        self.update(kwargs)
```
lfarv, line 446: We are missing `super().__init__`:
```python
super().__init__(family_name, **kwargs)
```
lines 441 and  446 can be removed

atintegrators/BeamMomentsPass.c
```diff
@@ -0,0 +1,130 @@
...
+    
+    for (i=0; i<num_particles; i++) {
+        double *r6 = r_in+i*6;
+        if(!atIsNaN(r6[0])){
+            ib = i%nbunch;
```
lfarv, line 49:
```
ib = i%nbunch;
```
This worries me: the bunch number should be extracted from the ct coordinate of the particle. Here you assign a bunch number from the index in the `r_in` array, but you keep the ct coordinate as if the particle was in bunch 0. So for the tracking, all particles are in bunch 0.
Of course, the same trick is applied in the `slice_bunch` function for impedance, but it would be simpler to work with the real ct coordinates.
For instance, with this beam monitor feature, I would expect the bunch average positions separated in ct by &beta;c/f<sub>RF</sub>, while they are all at 0 !

For me, the bunch distribution should be entirely defined by the `r_in` coordinate buffer. One needs a convenience function to generate a `r_in` buffer for a given filling pattern, but that looks easy.

Is it possible to convert the multibunch behaviour in this way? 
Would would happen now if we declare some filling pattern in the lattice, but set `r_in` so that particles are distributed differently?

swhite2401: Why are you worried?

This very clean and simple: the bunch id depends only on the particle index and is therefore fully deterministic. Extracting single bunch information is then trivial using the same modulo. It cannot be any simpler. Oscillating around 0 is perfectly fine: it is still a time lag w.r.t to the reference particle just in a different bucket, I see no issue with that. 

If anyone needs the absolute `ct` coordinate (which I strongly doubt) it can be easily reconstructed using the bunch id and filling pattern. I can provide such function and integrate if in the beam monitor if needed.
The only constraint is that the user should not try to generated the `ct` coordinates himself which is fine since no AT function can do it presently. We can document it for safety.

I have considered working with `ct` coordinates but then you need to forget about the `fillpattern` because we cannot guaranty that it is consistent with `r_in`. You then need an algorithm to split the beam in bunches using `ct` which involves bunch boundaries etc... and this will fail with off-energy tracking and drifting `ct` .  It does look very complicated to be... I am really not in favor of such change.

For me the real improvement would be for each particle to carry the information on bunch id, charge, mass, etc... but it would probably require to define a beam object which is a major change for AT...

lfarv: Ok, I understand it's a choice, and I don't intend to trigger large modifications. But this choice needs some precautions:
- the number of particles must be a multiple of the number of bunches. Is it checked somewhere?
- what happens if `r_in` does not agree with the `lattice.fill_pattern` ?  I'm using such such input (very easy to create) and I fear that this will break the multibunch processing. Checking that all particles are in bunch 0 is not so easy. I agree that a *strong* warning is necessary.
- off-momentum treatment is more difficult in the present setup: while ct easily gives you the absolute arrival time, and consequently phase with respect to any resonator, here you have to "rebuild"  the absolute time in some way. This is for me the main advantage of using ct. I don't have at the moment a way to check the current behaviour. Are you sure it's correct?

So no point in modifying anything at the moment, I was just very surprised looking at the results of "beam monitoring"!

lfarv: Hello @swhite2401: this night, my worries concretised into a real problem: you approach (shifting bunches by a constant time lag) works only for a single RF system (or a pure harmonic system).

Let's take a the simplest example of multiple RF: a main cavity, on harmonic h, and a voltage modulation cavity, on harmonic h-1. Then all the bunches see a different voltage, have  different synchronous phase and synchrotron frequency. They are not equidistant. You cannot simulate that by putting all particles in bunch 0.

The "time lag" trick can only be applied for an exact multiple of RF periods. It is obvious for a single RF system (possibly with pure harmonics), it does not work if the bunches are not equidistant. Cheating with time has its limitations…

I understand you concern on identifying the bunch number of each particle. It's only a visualisation problem, for tracking only ct matters, you don't care. But there is an intermediate solution: keep you convention of distributing the bunches according to the column index in `r_in`, but put the real ct in the 6th coordinate.

I came to the problem be remembering the 1st reason I had to generate multibunch `r_in`. And in this case, it's not so easy: you should compute the 6d closed orbit for each bunch. I did not do that, and got synchrotron oscillations in non-zero bunches.

To summarise: this discussion does not concern specifically the PR. Let's take it as it is and look at the problem in another PR. For the time being, we can say the multibunch treatment is limited to single RF systems (I'm still not sure for off-momentum, but you may be right).

swhite2401: What you describe is really a very specific case where you voluntarily introduce different voltages on the different bunches, I don't think there are many machines running this way nowadays.
For most light source multiple RF systems run on exact harmonics of the main: all bunches see the same voltage in the absence of beam loading, with transient beam loading this is different but the present implementation handles this correctly I believe (@lcarver will start looking into these issues soon).

I therefore do not agree with the statement: "the multibunch treatment is limited to single RF systems".  For what you call "pure harmonics" it should be valid and this will cover most of the cases you would like to model.

For more exotic cases, like the one you describe, you may set fillpattern=[1] and generate the multibunch r_in that you were trying to model, however you will have to slice the full beam -> many slices, and many of them empty (which is probably ok)

swhite2401: Now concerning your bullets:

1. This is not true, the only constraint is that the number of particles per process has to be greater or equal than the number of bunches: this is checked in the passmethod, an error is issued if this constraint is not fulfilled, in case the number of particles is not a multiple of `nbunch` a warning is issued (also handled in the passmethod)
2. In that case the bunch distribution will be wrong, I can check that all paticles are within +/-lambdaRF/2, this is relatively easy. Just as a side comment: it will be similarly complicated to associate a particle with a bunch in your proposal and this is not just visualization. The wake field kick calculation requires this information unless you slice the entire beam in one go which will be very inefficient and numerically unstable (to be checked). Post processing analysis also requires tbt bunch-by-bunch data, so you will need to be able to split the beam in bunches
3. Presently we have:
- Single bunch -> at each turn the `ct` coordinate is incremented by `cT0` in the turn history, I think this is OK and valid in all cases
- Multibunch -> `ring.bunch_spos` is used. These positions are derived using the nominal circumference. I think this is wrong for off-momentum and one should use `beta*c/frev` instead. This can be changed

carmignani: Hello,
there is also the case of third harmonic and 3.5 harmonic, as proposed for Bessy VSR. I did some simulations using 3.5 harmonic for the esrf a few years ago and I was checking the bunch length on different buckets at different ct values. 
That is another case where you need some bunches to be not at ct=0.

atintegrators/BeamMomentsPass.c
```diff
@@ -15,79 +15,71 @@
...
      
-    positions += 6*nbunch*turn;
-    sizes += 6*nbunch*turn;
-    memcpy(positions, avep, 6*nbunch*sizeof(double)); 
-    memcpy(sizes, sizep, 6*nbunch*sizeof(double));     
+    means += 6*nbunch*turn;
+    stds += 6*nbunch*turn;
+    memcpy(means, meanp, 6*nbunch*sizeof(double)); 
+    memcpy(stds, stdp, 6*nbunch*sizeof(double));
+    free(buffer);
```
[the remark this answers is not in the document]
swhite2401: done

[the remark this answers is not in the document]
swhite2401: > You must use `atFree`. In Matlab, we are using the Matlab memory allocator, so it must be consistent with `atCalloc`. In python, we are using the standard C allocator, so it would work anyway.

@lfarv, should I always use `atMalloc` /`atFree`? In several places `malloc`/`free` was used instead...

atintegrators/BeamMomentsPass.c
lfarv: > with transient beam loading this is different but the present implementation handles this correctly I believe (@lcarver will start looking into these issues soon).

I also think this is correct: the (shifted) ct should carry the necessary drift information. However, a beam loading example case simulating a "1/3" filling would be interesting: I intended to propose that in #516, it's more interesting than 16 bunches, may be @lcarver could add that?

> Multibunch -> ring.bunch_spos is used. These positions are derived using the nominal circumference. I think this is wrong for off-momentum and one should use beta*c/frev instead. This can be changed

I agree, it should be beta*c/frev

Now for your proposals:

> full slicing

Agreed, it's inefficient. But as I proposed before, using the column index for the bunch number as it is done now is OK. But use ct rather than a computed bucket position for timing.

> define a multi-bunch RF element for which the voltage depends on the bunch id

As you can imagine, I find this solution really non-intuitive: try to explain a user that for bunch n, he should put the particles in bunch 0 and have a special cavity which would act as if they were in bunch n !!
**It's so simple to work with absolute timing**: no ambiguity, you are sure it always works.

swhite2401: Ok so after discussion with @carmignani who is very keen on the VSR machine, I can propose to set the beam unfolding in the slicer as optional, you then have 2 cases:
-**Activated:** all bunches will be shifted according the `ring.bunch_spos`, this is the present implementation. A function to verify the `r_in` is in bucket zero will be provided
-**Deactivated:** bunches will not be shifted, bunch ids will however be used to separate bunches, a function will be provided to generate the correct distribution based on RF settings and `fillpattern`. If the user decides to not use this function there is not much we can do about it... results will be wrong

Is this fine for everyone? Any other suggestions?

swhite2401: I would like to implement this in a future PR, this one and #516 can be potentially be merged independently of this proposed enhancement

atintegrators/BeamMomentsPass.c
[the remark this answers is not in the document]
lfarv: `malloc`/`free` always works: it's standard C. But Mathworks claims that its allocator is better, that's why there are the macros `atMalloc`/`AtFree` which do the conversion to the proper low-level allocator. It's preferable to use them, but the critical point is that you must not mix them: allocating with one and freeing with another one will crash or corrupt the memory.

atintegrators/BeamMomentsPass.c
lfarv: @carmignani is right: there are countless possible bunch manipulations with RF systems, we should not restrict to the simplest ones. The proposal is fine for me. And of course it is for another PR.

> Activated: all bunches will be shifted according the ring.bunch_spos, this is the present implementation. A function to verify the r_in is in bucket zero will be provided

For off-momentum, I think you must use beta*c/frev: `ring.bunch_spos` is the on-momentum bunch position, the circumference is different off-momentum. I'm doubtful about the check function: it will add an overhead on each tracking, it must be run only if there is an impedance element in the ring, otherwise the user is free to do whatever he likes… Not obvious. I would rather put the effort in the documentation of multibunch tracking!

We'll have the opportunity to continue the discussion in the upcoming PR!

swhite2401: > For off-momentum, I think you must use beta*c/frev: `ring.bunch_spos`

Agreed! I am on it but facing recursive calls issues...I was thinking of using ring.find_orbit() but then `ring.bunch_spos()` is called in `lattice_pass()`, that is called in `find_orbit()`, not as simple as I thought

atintegrators/BeamMomentsPass.c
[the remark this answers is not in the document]
swhite2401: Understood, thanks

atintegrators/BeamMomentsPass.c
swhite2401: > otherwise the user is free to do whatever he likes…

Of course, what I had in mind was a warning only in Collective passmethods

lfarv (approved):

PR #524, opened 2022-11-24
Unfold beam

swhite2401:
This PR is a follow-up of the discussion in #521 and #516: for multi-RF systems with non integer harmonic different bucket will see different RF voltages. It is not possible to model such scenario with the present implementation since are artificially located in bucket zero.

In this proposal a function to unfold the beam is propose, the particle are shifted on the 6D orbit in their respective bucket. A optional argument is provided in `lattice_pass` to disable the internal unfolding that is based on equidistant bunches (`ring.bunch_spos` filled with zeros)


swhite2401:
Hello @lfarv , @lcarver , @carmignani . This is a minimal proposal that does not require any change in the C code which I find quite convenient.

We may want to propose to optionally call the `unfold` function in `lattice_pass` and `at.beam`, what do you think?

swhite2401:
While developing this I found a problem. Presently the collective passmethod require the number of particles >= number of bunches.
This is a source of conflict in functions using lattice_pass with a defined number of particles (orbit, optics, etc...) in case nbunch!=1 and a collective effect passmethod is present....

I propose to decorate all these function with `checkcollective(False)` since it make no sense to run these single particle calculations with a collective passmethod activated. This will be the topic of a new PR.

lfarv:
OK for the "unfolding" function. It makes sense to call it from `at.beam`. But I would not call it from `lattice_pass`: this should stay as a basic function tracking **exactly** what is given as input. Keep in ming that it does not need a Lattice as input, but an `Iterable[Element]`: it could even be an iterator rather than a sequence.

But does this answer the question of slicing the whole circumference?

swhite2401:
> But does this answer the question of slicing the whole circumference?

Yes I think it does: particles are distributed in bunches using the particles index and the same logic is used in the slicer.
The slicer is using the max. and min. `ct` of each bunch to determine the slicing range and this should not exceed one bucket if particles were correctly distributed.

For this to work there are several constraints:
-the initial distribution has to be generated in bucket zero followed by a call of the unfold function to distribute the particles in their respective bucket (can be optionally integrated in `at.beam`)
-the option `unfold_beam` has to be deactivated in `atpass` (not internal shifting)
-the `ZCuts` argument cannot be used in collective effects passmethods (presently applies to all bunches)

Did you see an issue? Or have doubts on this method?

lfarv:
> Did you see an issue? Or have doubts on this method?

No, I don't see. Still have doubts with off-momentum and ct coordinate increase, but that's another topic ! I did not go deep inside in the processing (I should!), so I cannot say.

So for me it's ok, at least as a step forward.

swhite2401:
> No, I don't see. Still have doubts with off-momentum and ct coordinate increase, but that's another topic !

Yes this probably has to be looked at in details, for sure the fixed `ZCuts` argument would fail, the rest is ok because `ds` is a difference between slices/bunches so the absolute `ct` does not matter, but I must admit that I have done extensive checks.

pyat/at/tracking/track.py
```diff
@@ -137,6 +146,11 @@ def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
     # define properties if lattice is not a Lattice object
     nbunch = getattr(lattice, 'nbunch', 1)
     bunch_currents = getattr(lattice, 'bunch_currents', numpy.zeros(1))
+    unfold_beam = kwargs.pop('unfold_beam', True)
+    if unfold_beam:
+        bunch_spos = getattr(lattice, 'bunch_spos', numpy.zeros(1))
+    else:
+        bunch_spos = numpy.zeros(len(bunch_currents))
     bunch_spos = getattr(lattice, 'bunch_spos', numpy.zeros(1))
```
lcarver, line 154: line 154 is a bug and should be removed

lcarver:
As mentioned above, a bug in line 154 of track.py which needs to be removed.
Also, in the __init__.py of tracking should it not also have 

`from .utils import *`

maybe the merge removed it by mistake. 

With the line 154 removed, it works as expected, although I have one concern...

If, in a multibunch simulation, I use call lattice pass with unfold_beam = False but I provide a beam that has NOT been unfolded, I effectively have provided a case where I have all of my particles in bucket 0 and no particles anywhere else. 

I ran this case, and it does not give an error, as the analysis still thinks that the particles are in other buckets, but they are not. Is it possible to catch this case and throw an error?

swhite2401:
> `from .utils import *`
> 
> maybe the merge removed it by mistake.

Right this has been corrected

swhite2401:
> If, in a multibunch simulation, I use call lattice pass with unfold_beam = False but I provide a beam that has NOT been unfolded, I effectively have provided a case where I have all of my particles in bucket 0 and no particles anywhere else.
> 
> I ran this case, and it does not give an error, as the analysis still thinks that the particles are in other buckets, but they are not. Is it possible to catch this case and throw an error?

Yes this is correct, and is also true for the opposite where `unfold_beam(at.beam)` was applied and `unfold_beam=True` in `lattice_pass` is used.
We have discussed this with @lfarv and for now we think it is more important to properly document all this than adding extensive and complicated tests.

For the future, we have discussed in #525 the possibility of introducing a beam object for collective effects simulations. This development would certainly make things much cleaner and allow to pass more detailed information on the beam characteristics including whether it has been unfolded or not.

lfarv (approved):

PR #525, opened 2022-11-24
check_collective

swhite2401:
This PR prevent running single particle calculation with collective passmethods activated in the lattice.
This was causing some crashes in case a multi-bunch fillpattern was initialized as the passmethods were expecting nparticles >=nbunch.

This is a bit cumbersome, I may have overlooked some function, could you please check carefully that this is ok and that I not forgotten anything?

### DIFF
--- a/pyat/at/acceptance/acceptance.py
+++ b/pyat/at/acceptance/acceptance.py
@@ -5,7 +5,7 @@ from .boundary import GridMode
 from .boundary import boundary_search
 from typing import Optional, Sequence
 import multiprocessing
-from ..lattice import Lattice, Refpts
+from ..lattice import Lattice, Refpts, check_collective
 from ..physics import frequency_control
 
 
@@ -14,6 +14,7 @@ __all__ = ['get_acceptance', 'get_1d_acceptance', 'get_horizontal_acceptance',
 
 
 @frequency_control
+@check_collective(False)
 def get_acceptance(
         ring: Lattice, planes, npoints, amplitudes,
         nturns: Optional[int] = 1024,
--- a/pyat/at/lattice/utils.py
+++ b/pyat/at/lattice/utils.py
@@ -31,7 +31,7 @@ Key = Union[type, Element, str]
 
 
 __all__ = ['AtError', 'AtWarning', 'check_radiation', 'check_6d',
-           'set_radiation', 'set_6d',
+           'set_radiation', 'set_6d', 'check_collective',
            'make_copy', 'uint32_refpts', 'bool_refpts',
            'checkattr', 'checktype', 'checkname',
            'get_cells', 'get_elements', 'get_refpts', 'get_s_pos',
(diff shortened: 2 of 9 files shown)


lfarv:
This seems to work, but it looks like these checks are not at the right place: with this, we are penalising all the basic functions to protect the very special case where we would have

1. nbunch > 1
2. a Wakefield element in the lattice

This is not the case for 95% of the design work for a storage ring (matching, parameter scans, linear and non-linear optimisations). These all involve the so-called "zero-current optics". But the check for collective elements requires each time a full scan of the lattice.

For me, this results from a design problem: the beam structure should not part of the lattice object. One solution to deal with that is the following:

- we take the "beam description" out of the lattice object and remove related parts from `lattice_pass`. `lattice_pass` gives `atpass` something saying "zero-current", for instance  nbunch=0 or 1, beam_current=0 or so,
- We protect the Wakefield element so that for "zero-current", it is fully transparent. This is anyway how it should behave, and it's may be already the case. 

This way, whatever happens if the future, all the present optics functions will never be disturbed, without needing any further check.

Then for impedance studies, we implement a wrapper around `lattice_pass`: kind of `impedance_pass`, or simply `beam_pass`. It must be provided with the filling pattern information and beam coordinates. It will provide `lattice_pass` the adequate `bunch_current`, `bunch_spos`,… data or call directly `atpass`, to be seen. The upcoming `unfold_beam` function can be used to provide the coordinates. So the filling pattern info must be provided to both `unfold_beam` to compute the coordinates and to `impedance_pass` to generate the bunch positions and currents. In a second step, one could make this even simpler with a "beam object", combining particle coordinates and fill_pattern info. But this object will not be involved in `lattice_pass`, so that all existing code is not affected.

**Advantages**:
- The main objects are well defined: the `Lattice` object contains lattice info, a possible `Beam` object carries beam info. This is not only cosmetic: I'm sure that it helps producing cleaner code.
- The basic `lattice_pass` stays basic: it does not make any assumptions nor modifies its input coordinates and ignores all collective effects data (but can pass them through to `atpass`).
- all existing "pre-impedance" code is unchanged and cannot break in future developments about collective effects because it does not carry any bunch current info. It correctly handles multi-bunch beams for exotic RF systems or non-integer harmonics, without any constraint of the indexing scheme: one can put the full bunch 0, followed by bunch 1, or interleave the bunches as the `unfold_beam` function does.

**Disadvantage**: for collective effect studies, one must call a new "`beam_pass`" function instead of `lattice_pass`. But it's probably the only needed function, everything else being the preparation of the beam and the processing of the tracking results.

**For now**: This proposal is far from being concrete. It must be discussed and better evaluated. I don't think it needs huge modifications, so it's not urgent and should not stop developments on collective effects. In particular #524 should go on!

Concerning this PR, I propose to postpone it since the danger is relatively small and should not concern the present users. In addition, it should result in an error ("Number of particles has to be greater or equal to the number of bunches.") rather than a crash. Or is it worse? It should not!

swhite2401:
I fully support the idea of a beam object as I have been thinking about it for some time now! A wrapper sound like a good idea to keep backwards compatibility.

Yes, the code throws the error: "Number of particles has to be greater or equal to the number of bunches." that's it so not too bad. It is a bit cryptic but may steer the user towards an issue with the fill pattern...

Ok to put this on hold for now and instead neutralize the passmethods and errors when the current equals zero as an additional safety. I close it then

PR #531, opened 2022-11-30
Update at.beam

lfarv:
`at.beam` is updated to use the recently introduced random number generators (RNGs).

At the same time, the `at.sigma_matrix` function is rewritten to remove some holes in the previous logics.


swhite2401 (approved):

lcarver:
All works for me. Also I like the fact that you can do the reverse now, computing the sigma_matrix from a given set of coordinates, which is a very useful addition.

lcarver (approved):

PR #534, opened 2022-12-01
Update quantum diffusion pass methods

lfarv:
`QuantDiffPass`, `BndMPoleSymplectic4QuantPass` and `StrMPoleSymplectic4QuantPass` are now using the RNGs introduced in #513.

Also `atrandom.c` is modified to add mean and stdev input arguments to the Gaussian generator, so `VariableThinMPolePass` had also to be updated.

The `atrandom.c` interface should now be stable.


lfarv:
Waiting for #532 to be merged

swhite2401:
Looks ok to be but has this been tested? Is tracking giving the correct equilibrium emittance?

lcarver:
> Looks ok to be but has this been tested? Is tracking giving the correct equilibrium emittance?

Yes, this was the only test that I did.

swhite2401 (approved):

PR #535, opened 2022-12-01
Failure of GitHub action for Matlab compilation error

lfarv:
As discussed [here](https://github.com/atcollab/at/pull/527#issuecomment-1333434260), the "Run Matlab tests" GitHub action fails if Matlab fails to compile AT

There is a new `-fail` flag in atmexall to stop at the 1st error:
```Matlab
atmexall -fail
```

The failure in this branch demonstrate that, but should not prevent merging !


carmignani:
It's ok for me.

carmignani (approved):

PR #538, opened 2022-12-06
Bpm errors

swhite2401:
This proposal introduces a possible solution to handle BPM, field and alignment errors.

Errors are added to an element with additional arguments `BPM*` for BPM errors and `*Err` for `PolynomA/B` and alignment errors. For BPM errors, it is only possible to modify the output of `find_orbit`. For field and alignment errors one can either use a wrapper for selected functions (`lattice_pass`, `find_orbit` and `get_optics` for now) or generate a new ring. In both cases the `*Err` attribute is added to the standard existing attribute (For example `PolynomB = PolynomB + PolynomBErr`).

Function provided:
-wrapper are provided for the functions `lattices_pass`, `find_orbit` and `get_optics`. This does not modify the input ring
-associated new functions `*_err` are defined with the modified outputs
-a function `enable_errors` is provided to return a copy of the lattice with errors included in existing attributes: this lattice can be used with any existing `at` function, this new ring is protected by a private flag to avoid that the errors are applied twice
-a function is provided to fetch the mean and std error of selected keys/attributes

Since this is a proposal I leave as draft for now, any suggestions?


lfarv:
Hello @swhite2401 

This looks very nice! I have no comment on the global design:
- the provided functions seem to fulfil most of the needs
- they are simple enough and easy to use

May be one could look at a way to look at the effect of specific kinds of errors: I take the case where all kinds of errors are assigned once for ever (but not enabled). One could want to look at:

- magnet errors without BPM errors: for this you can explicitly call `enable_errors`, and then use the "normal" optics functions which ignore BPM errors,
- BPM errors only: for this you could explicitly set the `_has_errors` attribute on the initial lattice and call the `*_err` functions. Not very elegant, but possible,
- Look independently at alignement, tilt or field errors. That would need a version of `enable_errors` with a selector, with a default of enabling everything.

Otherwise I have a few comments concerning implementation details, that can wait…

swhite2401:
Hello @lfarv, thanks for the feedback. If you agree with the design I will extend this as you suggest to provide more flexibility. I agree the implementation needs to be improved, we can iterate on this once the desired functionalities are all implemented.

swhite2401:
@lfarv I have pushed my latest modifications (not tested), it is all yours!

lfarv:
I have some proposals to add to the error assignment function:

1. **For all errors**: It should be possible to specify both systematic errors (identical on all elements) and random errors (randomly distributed over elements). For this, we could decide that the "value" in keyword arguments could be either a tuple "(systematic, random)" or a single item, which would be attributed to random, as it is done now. Also, by making use of the broadcasting feature, the "systematic" value could set a completely predefined set of errors to all elements.
2. **For magnet field errors**: it would be convenient to specify `PolynomA/BErr` either in absolute (added to `PolynomB`), or in relative (multiplied be the "strength" of the magnet before addition). That's the easiest way to introduce gradient integral random errors or systematic multipolar errors. The "strength" would be `PolynomB[1]` for quads, `PolynomB[2]` for sexts and so on. What about "Multipole" objects ? How to indicate the choice ? A `"relative"`bool keyword argument ?
3. **For gain, offset, shift errors**: they are (2,) arrays, for H and V components. Do we allow a single value to indicate the same value in both planes ?

Given the specificity of magnet errors, it could be convenient to separate `assign_bpm_errors` and `assign_magnet_errors`. This would also have the advantage of making the documentation cleaner: it will be quite long. What do you think of that?

Finally, I was also thinking of a "noise" BPM error, varying on each BPM crossing. This would make sense for `lattice_pass`, possibly affecting the determination of tunes. But problematic for closed orbit: the orbit would change on each call… But why not? Let's keep this for later…

swhite2401:
I do have some concern about the relative errors though...I think the simplest would be to add another field `Polynom[AB]Scale` for instance that would just multiply the result after the offset was applied (default being one for all multipoles). I think this is more flexible and in case of calibration errors I believe all multipoles would scale no?
Similarly to other fields it is activated when it is present.

I do not understand what you mean by separating magnet and BPM errors, isn't this already the case? Or you would like to produce to separate decorators?

Besides that all the rest is fine for me!

BPM noise could be interesting in fact! We may simply generate/apply them in the function `_apply_bpm_track_error`. It would just require an additional sigma argument. and also separated decorator for each function producing an orbit, but you had in mind to separate them anyway right?

lfarv:
> I do not understand what you mean by separating magnet and BPM errors, isn't this already the case?

There is at the moment a single `assign_errors` function for all kinds of errors. If we need keyword flags specific for BPMs or magnets, it could be cleaner to have rather `assign_bpm_errors` and `assign_magnet_errors`. The description of all errors will also be rather long, and could be shared between both functions. But if there is no specific feature, we keep is as it is!

lfarv:
Some of the modifications discussed above are implemented:

- a tuple (systematic, random),  or a single value for random only is implemented for all errors,
- for offsets/shifts and gains, a [x, y] array defines H and V values, but a single number is accepted to set both planes to the same value.
- for rotations, a single value sets tilt only, and a triplet sets [tilt, pitch, yaw].

On the other hand, a single `assign_errors` function is kept for all errors (no need to split), and the question on PolynomA/B is postponed until we find a satisfactory solution.

The documentation is ready.

swhite2401:
> the question on PolynomA/B is postponed until we find a satisfactory solution.

@lfarv, have you thought about this? An alternative could be to provide a relative option for `PolynomA/B` in the `assign_errors` function?

lfarv:
@swhite2401 : Still there ?

> An alternative could be to provide a relative option for `PolynomA/B` in the `assign_errors` function?

This was exactly my initial idea, but you convinced me that another attribute would be simpler than an option (we can add both attributes). My idea is to provide a multipole description directly coming out of the magnetic measurements, normalised to the main component. This way, you have a single description for mechanically identical magnets, the multipoles just scale with the main strength. This is perfectly valid in the linear range. At saturation (which is unfortunately the usual case nowadays), it is approximate, but anyway better than a fixed value if you intend to vary the strength around to nominal working point. This seems to me the simplest way to describe systematic multipole errors, and is also probably useful for random errors: errors related to mechanical random errors also roughly scale with the current.

By using a different attribute ("ScalingPolynomAErr", or better if you have ideas), I just comes as an addition to the present code, and can be implemented now or later: no fundamental change is needed.

swhite2401:
Hello @lfarv, yes still around for a couple days.
Well, I was thinking of some simpler: instead of providing absolute value to `assign_errors` you provide a percentage instead.
Then the absolute value is derived internally using `PolynomA/B*error` and assigned to `PolynomBErr`.

Or course in this case if you change the main strength the error remains fixed, but my feeling is that for error studies we look at static lattices in general.

I agree this can be added at any time...up to you!

swhite2401:
> By using a different attribute ("ScalingPolynomAErr", or better if you have ideas), I just comes as an addition to the present code, and can be implemented now or later: no fundamental change is needed.

I case we would like a 'dynamic' behavior it think we would need to go for this one right?

lfarv:
Still not clear for me… I'll detail the way I see it:

1. The perfect lattice is defined with "standard" quadrupoles: PolynomB[1]=K, nothing else, neither in B nor in A,
2. we take the normalised multipole analysis from the magnetic measurement of let's say type "A" quadrupoles: it contains octupole, dodecapole etc. We introduce this with `assign_errors` in the "ScalingPolynomBErr" attribute of all type "A" quadrupoles,
3. we can play with this (still perfect) lattice: matching, tune optimisation...
4. to look at the effect of errors, in `enable_errors`, we sum PolynomB, K*ScalingPolynomBErr and PolynomBErr (if any), store the sum in PolynomB to run say dynamic acceptance,
5. We can make any new tuning on the initial lattice, no need to re-`assign_errors`, the magnetic contents of each magnet is frozen in the ScalingPolynomBErr attribute.

The point is that this make very easy to convert the results of magnetic measurements into an error attribute. Even errors on gradient integrals are easily described: you put 0.001 in the random ScalingPolynomBErr[1], and that's it!

This leaves the "static" PolynomBErr acting as you defined it, it's just a new feature. And we may add more…

By the way, we could add a `copy` option in `enable_errors` ?

swhite2401:
Ok it is now all clear and I agree with what your propose, this is indeed very simple!

Why should `copy` be optional in `enable_errors`, you never want to modify `ring` in-place no?

swhite2401:
Then the question on multipoles remain... the scaling applies to all non-zero orders in `PolynomB/A`?

lfarv:
> Why should copy be optional in enable_errors, you never want to modify ring in-place no?

You are right, no need!

> Then the question on multipoles remain... the scaling applies to all non-zero orders in PolynomB/A?

The scaling would apply to all orders: you can include a spurious dipole component (systematic or random) in a quadrupole. For systematic errors, you should set 0 in the "main" component (it is already the PolynomB value), but you may set it for random errors (see above). The errors would scale with
- 1/(bending radius) for `Dipole` [m<sup>-1</sup>]
- PolynomB[1] for `Quadrupole` [m<sup>-2</sup>]
- PolynomB[2] for `Sextupole` [m<sup>-3</sup>]
- PolynomB[3] for `Octupole` [m<sup>-4</sup>]

This can be done easily  by adding a "Strength" or "IntegratedStrength" property to the various elements ("DefaultOrder" is already defined).

Assigning a "ScalingPolynomErr" to a `Multipole` should fail: there you must use fixed errors

One could normalise to the **integrated** strength (bending angle, K.l, H.l, etc.). Do you have a preference ? Integrated could apply to "thin" magnets, but there are no `ThinQuadrupole` or so.

For "DQ"s (described as `Dipole`), it will be static whether you use PolynomBErr or ScalingPolynomBErr (unless the radius varies): it's supposed to be a fixed magnet.

swhite2401:
> One could normalise to the **integrated** strength (bending angle, K.l, H.l, etc.). Do you have a preference ?

No real preference, integrated strength potentially offers some advantages as you mentioned and is generally what you get from magnetic measurements.

lfarv:
So let's go for integrated strengths!

lfarv (approved):
ready for me

lcarver:
In my opinion, if you have a bpm offset of +x, when you run find_orbit_err you should find an orbit at this location of -x. However with this implementation, the orbit at this point is equal to the BPM offset.

lfarv:
@lcarver: good point.

In the present implementation, the conversion (forgetting rotations) is: x<sub>bpm</sub> = gain<sub>bpm</sub> * (x<sub> orbit </sub> + offset<sub>bpm</sub>). The full formula in the doc is [here](https://at.readthedocs.io/en/bpm_errors/p/api/at.errors.error_def.html#at.errors.error_def.assign_errors), clearly mentioning that it results in a BPM shift of -offset, as you noticed. That also looks odd to me, and I'm ready to change the offset sign if everybody agrees.

If we completely invert the formula, one would also switch to 1/gain… Not so obvious

lcarver:
> If we completely invert the formula, one would also switch to 1/gain… Not so obvious

Why is the switch to 1/gain needed? Why would the formula not be exactly the same except for a minus sign on the offset?

lfarv:
@lcarver : one clean way to change the sign of the offset is to invert the formula and write (again forgetting the rotation):

 x<sub>orbit</sub> = gain<sub>bpm</sub> * x<sub>bpm</sub> + offset<sub>bpm</sub>

Now "offset" would be the opposite of the present one (as you like it), and "gain" the inverse of the present one.

Of course, one could also put minus signs in the formulae and choose whatever we like, but that also looks strange…

Anything is good for me, provided the formula is clearly appearing in the documentation. Any advice, @swhite2401 ?

pyat/at/errors/error_def.py
```diff
@@ -0,0 +1,449 @@
...
+
+def _sort_flags(kwargs):
+    errargs = {}
+    for key in _SEL_ARGS:
+        if key in kwargs:
+            errargs[key] = kwargs.pop(key)
+    return kwargs, errargs
+
+
+def find_orbit_err(ring: Lattice, refpts: Refpts = None, **kwargs):
```
simoneliuzzo, line 349: this function may be used with a lattice without errors or it gives errors? why a new function is needed? I think that it would be a much larger step forward if we could avoid using `*_err`. This startegy was used in the past for private codes. a `find_orbit` sufficiently clever to understand if a lattice has errors or not would be highly user friendly.

pyat/at/errors/error_def.py
```diff
@@ -0,0 +1,449 @@
...
+    All other keywords are forwarded to :py:func:`.find_orbit`
+
+    See also:
+        :py:func:`assign_errors`, :py:func:`.find_orbit`,
+        :py:func:`get_optics_err`, :py:func:`lattice_pass_err`
+    """
+    kwargs, errargs = _sort_flags(kwargs)
+    ring = enable_errors(ring, **errargs)
+    orbit0, orbit = find_orbit(ring, refpts=refpts, **kwargs)
+    _apply_bpm_orbit_error(ring, refpts, orbit)
```
simoneliuzzo, line 373: `orbit0` does not need to be modified?

pyat/at/errors/error_def.py
```diff
@@ -0,0 +1,449 @@
...
+    return orbit0, orbit
+
+
+def get_optics_err(ring: Lattice, refpts: Refpts = None, **kwargs):
```
simoneliuzzo, line 377: same comment as for `find_orbit`

pyat/at/errors/error_def.py
```diff
@@ -0,0 +1,449 @@
...
+        :py:func:`find_orbit_err`, :py:func:`lattice_pass_err`
+    """
+    kwargs, errargs = _sort_flags(kwargs)
+    ring = enable_errors(ring, **errargs)
+    ld0, bd, ld = get_optics(ring, refpts=refpts, **kwargs)
+    _apply_bpm_orbit_error(ring, refpts, ld.closed_orbit)
+    return ld0, bd, ld
+
+
+def lattice_pass_err(ring: Lattice, r_in, nturns: int = 1,
```
simoneliuzzo, line 406: same comment as for `find_orbit`

pyat/at/errors/error_def.py
simoneliuzzo, line 349: an additional issue to be dealt by `find_orbit` are BPM misalignments (T1/T2 R1/R2) for my personal usage I have always considered this to be the position of the BPM in the machine. A different information compared to the BPM offset/rotation. In a lattice with very large misalignment, (potentially following a smooth curve with amplitudes reaching several millimeters), T1/T2 could be order of mm, offsets of order of micrometers and the final corrected orbit also micrometers. To read the "absolute" orbit in space, one would have to sum T2 to the BPM readings. However this information is almost never relevant, it may be needed just for a single plot to show what find_orbit returns. It may be provided via an absolute=False flag. If True T2 are ignored.
`find_orbit` make indirectly use of the T1-2 in all elements to give eventually a very large orbit. 
T1 and T2 errors are ignored by `_apply_bpm_orbit_errors` but it may be needed to include them in order to give the user the real expected measured orbit with errors.

pyat/at/errors/error_def.py
```diff
@@ -0,0 +1,449 @@
...
+
+    All other keywords are forwarded to :py:func:`.get_optics`
+
+    See also:
+        :py:func:`assign_errors`, :py:func:`.get_optics`,
+        :py:func:`find_orbit_err`, :py:func:`lattice_pass_err`
+    """
+    kwargs, errargs = _sort_flags(kwargs)
+    ring = enable_errors(ring, **errargs)
+    ld0, bd, ld = get_optics(ring, refpts=refpts, **kwargs)
```
simoneliuzzo, line 401: ld0 closed orbit does not seem to be updated by these lines

pyat/at/errors/error_def.py
```diff
@@ -0,0 +1,449 @@
...
+        ring = _apply_field_errors(ring, **kwargs)
+        ring = _apply_alignment_errors(ring, **kwargs)
+        ring._has_errors = True
+    return ring
+
+
+def has_errors(ring):
+    """Get the error status of a lattice"""
+    return getattr(ring, '_has_errors', False)
+
```
simoneliuzzo, line 333: an additional function to print a text table of errors is often really useful to read a lattice

pyat/at/lattice/elements.py
```diff
@@ -243,12 +260,19 @@ class Element(object):
                         T1=lambda v: _array(v, (6,)),
                         T2=lambda v: _array(v, (6,)),
                         RApertures=lambda v: _array(v, (4,)),
-                        EApertures=lambda v: _array(v, (2,)),
-                        KickAngle=lambda v: _array(v, (2,)),
+                        EApertures=_array2,
+                        KickAngle=_array2,
                         PolynomB=_array, PolynomA=_array,
+                        ScalingPolynomBErr=_not_allowed,
+                        scalingPolynomAErr=_not_allowed,
```
simoneliuzzo, line 267: small S in `scalingPolynomAErr`

pyat/at/lattice/elements.py
```diff
@@ -618,6 +645,11 @@ def __setattr__(self, key, value):
 
         super(ThinMultipole, self).__setattr__(key, value)
 
+    @property
+    def integrated_strength(self):
```
simoneliuzzo, line 649: there is an identical function to this one in line 718 that does the multiplication by length. 
may be this is only a template function that is overloaded within the single elements?

simoneliuzzo (commented):
I find this pull request a really needed addition to pyAT. In my opinion the approach is still a bit conservative and could be made more user friendly by removing the *_err functions. The equivalent for matlab AT would be also most welcome.

simoneliuzzo:
It is not clear to me why: 

> Assigning a "ScalingPolynomErr" to a `Multipole` should fail: there you must use fixed errors

For user friendlyness I would prefer that ScalingPolynomA/BErr applies to any element. For Multipole the main multipole is not known then it has to be somehow defaulted (largest absolute PolynomB/A) or specified by the user when assigning these errors.

pyat/at/errors/error_def.py
lfarv: Hi @simoneliuzzo : the idea is the following:

You may assign errors to a lattice once for ever: they are stored in dedicated attributes and are ignored in all optics computations. So `find_orbit`, `get_optics` and so on are not affected by the "assigned" errors.

The errors are processed only in the `*_err` functions. This avoids confusion: you don't have to care whether assigned errors are present in the lattice or not, they are ignored unless you explicitly call a` *_err` function.

pyat/at/errors/error_def.py
lfarv: And same answer…

pyat/at/errors/error_def.py
lfarv: At the moment, T1-2 and R1-2 in `Monitors` don't have any effect on anything: they don't affect the orbit, neither with nor without errors. So you can indeed use them for your personal use to keep track of any information you want. On the other hand, if you have a mechanical offset of several millimetres, you can put it in the `BpmShift` attribute (beware: the sign is currently debated with @lcarver), you may get a closed orbit of micrometers, or even zero if there are no magnet errors, but the BPM reading as returned by `find_orbit_err` will be millimetres: that's correct !

pyat/at/errors/error_def.py
lfarv: No, it just shows the orbit resulting from magnet errors enables in the previous line

pyat/at/errors/error_def.py
lfarv: Good point. I take note

pyat/at/lattice/elements.py
lfarv: Well done ! Thanks !

lfarv:
> For user friendlyness I would prefer that ScalingPolynomA/BErr applies to any element

The problem is that there is no obvious default order: we are talking here of magnets which have not been declared neither as Dipole, Quadrupole, Sextupole nor Octupole. For combined function magnets, you are free to decide which class you choose, and the scaling will be made with the corresponding order. So we are mostly talking here of higher order multipoles, or skew magnets. 

In that case, I think the best option is to set explicitly the `DefaultOrder` attribute to the `Multipole` element: you decide what it is, once for ever, and then everything behaves as expected. Be careful, the `strength` or `integrated_strength` of a magnet is the "normal" one (not skew, it's based on PolynomB !) . Any "automatic" order attribution might be confusing.

In any case, you can still use the "static" error: for a magnet you don't intend to vary, that's good enough.

pyat/at/errors/error_def.py
swhite2401: I tend to agree this is a bit tricky, how do you handle the case where T1/T2 are non zero in the elements upstream and downstream the BPM?

pyat/at/errors/error_def.py
swhite2401: @simoneliuzzo we had been discussing the use of ``pandas`` for this but I did not implement it since you were interested to look at it yourself, is it still something you intend to do?

swhite2401:
> The problem is that there is no obvious default order: we are talking here of magnets which have not been declared neither as Dipole, Quadrupole, Sextupole nor Octupole.

Presently the proposal is to have (quadrupole as example):
``PolynomB = K*ScalingPolynomBErr + PolynomBErr``

This is fine and will work for classes with predefined order.
For arbitrary multipole it is a bit tricky since we cannot determine the order, however is it really a problem to use the following?
``PolynomB = PolynomB*ScalingPolynomBErr + PolynomBErr``

My feeling is that it would result in the same from predefined orders (`PolynomB` is filled only where needed) but would cover also the case of arbitrary multipoles (all orders scaled).
There might be some complication with orbit in case `PolynomB[0]` is changed but we may well apply the scaling only starting from `order=1` in case the class is not a dipole.

lfarv:
@lcarver and @swhite2401 

So I propose to define the BPM errors according to this:

 xy<sub>bpm</sub> = gain<sub>bpm</sub> * rot<sub>bpm</sub> * (xy<sub> orbit </sub> - offset<sub>bpm</sub>)

The only difference from the initial formula is the sign of the offset. Offset and rotations are equivalent to mechanical displacements of the monitor, gain is the magnification applied by the monitor to the beam position.

I change the documentation accordingly.

pyat/at/errors/error_def.py
lfarv: @simoneliuzzo and @swhite2401 : I hope you don't intend to have PyAT depend on `pandas`…

lfarv:
> For arbitrary multipole it is a bit tricky since we cannot determine the order, however is it really a problem to use the following?
> `PolynomB = PolynomB*ScalingPolynomBErr + PolynomBErr`

This does not work ! The scaling corresponds to **current** you put in the magnet, so it has to be a single **scalar** applied to all components. Otherwise you'll change the shape of the field.

When you get the field measurement, you can arbitrarily normalise it with any of its non-zero components. For a non-standard magnet, just take any component, and specify the corresponding order in the `DefaultOrder` attribute, that's it.

pyat/at/errors/error_def.py
swhite2401: Well I do not see a real need for now, but `pandas` offers a lots of functionalities that could become handy and is sufficiently well documented and widely used that it could be considered.
Why do you see this as a problem? How is it more of a problem than `mpi4py` for instance?

swhite2401:
> The scaling corresponds to **current** you put in the magnet, so it has to be a single **scalar**

Ok then what about:
`PolynomB = (PolynomB + PolynomBErr)*ScalingPolynomBErr`

ThorstenHellert:
> > The equivalent for matlab AT would be also most welcome.
> 
> I fully agree and started looking at that. @simoneliuzzo, I think you already did a lot about errors and corrections. Is there something we can take out of that ? I'd like to have something similar to this branch: as simple, using the same attributes, same concepts, etc. Is there some place I should look at?

Hi Laurent, you might want to take a look at https://sc.lbl.gov, I implemented a sophisticated error model for Matlab AT. I always wanted to get back to you guys and see if and how that could be integrated into AT

lfarv:
Does not work either… When developing, we again get a `PolynomB*ScalingPolynomBErr` which is not a proper scaling.

Look at the way to generate these error attributes. As an example, I take a quadrupole

**Systematic errors:**
The magnetic measurements at the "nominal" current give systematic octupole and a 12-pole components in addition to the quad:
`[0, B2, 0, B4, 0, B6]`
The 1st step is to normalise with respect to the quad component:
`[0, 1, 0, B4/B2, 0, B6/B2]`
The 1 goes to the main field, the rest goes to errors:
`PolynomB = [0, B2]`
`ScalingPolynomBErr = [0, 0, 0, B4/B2, 0, B6/B2]` (systematic)

Now assume that after tuning we set the quadrupole to C2 instead B2:
`PolynomB = [0, C2]`
`enable_errors` will compute:
`[0, C2] + C2*[0, 0, 0, B4/B2, 0, B6/B2] = [0, C2, 0, C2/B2*B4, 0, C2/B2*B6]`
which is the exact scaling of the measured field by a ratio C2/B2: it's correct

**Random errors:**
In addition , we know that we have random relative field integral error of 10<sup>-3</sup>,and a random sextupole B3.  All this also scales with the magnet current, so we define:
`ScalingPolynomBErr = [0,  1.E-3, B3/B2,  1.E-3/B2, 0, 1.E-3/B2]` (random)
After generating a random number and multiplying by B2, we get the correct random contribution.

All this is based on the fact that we have a complex magnet but a single knob to scale it: the current, that we arbitrarily assign to B2 (for a quad).

For this simple case, we don't use the "static" error attribute, and we can scale the magnet without needing to reassign different errors.

**Arbitrary multipole:**
Now let's take a combined magnet , quadrupole-sextupole, let's say:
`PolynomB = [0, B2, B3]`
For errors, we can arbitrarily decide it's a quad and scale with B2: `ScalingPolynomBErr = [0, 0, 0, B4/B2, 0, B6/B2]`, which will then be multiplied by C2
or decide it's a sextupole: `ScalingPolynomBErr = [0, 0, 0, B4/B3, 0, B6/B3]` which will be multiplied by C3.

The result will be **exactly the same** (provided the users correctly scales the main field [0, C2, C3]). But we had to decide what is the "main order" of the magnet.

`ScalingPolynomBErr` is meant to be used for "single knob" magnets (the field scales with a single : it ensures the correct scaling of errors without intervention, but it requires the "knob" to be associated with a single order. This works for the general Multipole object provided you assign it an arbitrary DefaultOrder attribute and normalise the errors accordingly. I don't see any difficulty in that.

swhite2401:
> `ScalingPolynomBErr` is meant to be used for "single knob" magnets (the field scales with a single : it ensures the correct scaling of errors without intervention, but it requires the "knob" to be associated with a single order. This works for the general Multipole object provided you assign it an arbitrary DefaultOrder attribute and normalise the errors accordingly. I don't see any difficulty in that.

Sorry I had misunderstood what you were doing I should have looked in more details in the code. I agree this is fine and the additional `DefaultOrder` should allow to cover all relevant cases.

lfarv:
>  I should have looked in more details in the code

To avoid this step (quite uncomfortable for users), I could add something similar to my previous post in the prolog of the `error_def.py` file. It can be improved later according to comments I get.

pyat/at/errors/error_def.py
simoneliuzzo: the first element of the ring may be a BPM

pyat/at/errors/error_def.py
simoneliuzzo: I find the choice to use *_err not really user friendly.  I have done like you propose in my code for many years, simply because it would have been a too big change to do it "in the correct way". I expect that find_orbit gives me an orbit with errors, as it does for classic magnet errors T1/T2 R1/R2. Why for BPM errors should it be different? As a user, I would think that find_orbit does not work if I set bpm errors and I see no change in the orbit provided by find_orbit. I would not think (at least immediately) that I am supposed to use a specific functions for the task (that I would then rather call find_orbit_bpm_errors)
I think this point is rather important to be discussed as it changes the present architecture of this pull request. What is the opinion of @swhite2401 , @lcarver, @carmignani, @ThorstenHellert and other users?

simoneliuzzo:
> 

I think the code from @ThorstenHellert is more advanced and documented than what I have used for EBS studies.

ZeusMarti:
> > > The equivalent for matlab AT would be also most welcome.
> > 
> > 
> > I fully agree and started looking at that. @simoneliuzzo, I think you already did a lot about errors and corrections. Is there something we can take out of that ? I'd like to have something similar to this branch: as simple, using the same attributes, same concepts, etc. Is there some place I should look at?
> 
> Hi Laurent, you might want to take a look at https://sc.lbl.gov, I implemented a sophisticated error model for Matlab AT. I always wanted to get back to you guys and see if and how that could be integrated into AT

As both AT and SC user I would highly apreciate it!

pyat/at/errors/error_def.py
swhite2401: @simoneliuzzo , in fact my initial intention was to decorate `find_orbit` which would results in the behavior you are seeking for.

There are nevertheless 2 issues with that in the present implementation:
1- it is less efficient since you will always scan through the whole lattice to check for the presence of errors
2- in case you assign errors to an element, calculations will always include then

A possible workaround is to define a lattice attribute `ring.errors_enabled` that one would check before scanning the lattice, it can be done without much effort and thinking about it I do agree with you that such implementation resembling the `enable/disable_6d` functionality would be much nicer for the end user.... now let's see what other users think.

Any arguments against this @lfarv?

pyat/at/errors/error_def.py
swhite2401: This is true, we could add the BPM errors to `orbit0` in case BPM errors are set for the first ring element, and in a similar way affect the closed orbit output of `get_optics`.

lcarver:
I agree with the comment from @simoneliuzzo (in code review) that just using find_orbit is better than having a separate function. The proposals from @swhite2401 would work for me.

lfarv:
> The proposals from @swhite2401 would work for me.

Which ones ?

lcarver:
> > The proposals from @swhite2401 would work for me.
> 
> Which ones ?

Hello, sorry I was replying in the main PR window to a comment on the code review. The proposal to have something called `ring.enable_errors` which means you only need to call `find_orbit `instead of having a different function called `find_orbit_err`.

pyat/at/errors/error_def.py
lfarv: I think you are mixing 2 things: orbit and BPM readings. `orbit0` is the orbit, it is independent of BPM errors and must stay so because it's the input for subsequent computations (matrix, optics, envelope). Only `orbit` is affected.

The 1st element may be a BPM, and if you select it in `refpts`, you get its reading as for any other element in orbit.

Ideally I would have added another output `orbit_reading` to keep it separate from `orbit`…

pyat/at/errors/error_def.py
lfarv: @simoneliuzzo, @lcarver, @carmignani @ThorstenHellert:

Keep in mind that there are 2 different things: the orbit and the BPM readings. Here there are 2 functions: `find_orbit` gives you the true orbit (affected by T1/T2, R1/R2 and magnet errors, as Simone says, but independent of BPM errors), and `find_orbit_err` which returns the BPM readings.

Maybe the function names could be improved: `find_orbit_view` , `find_optics_view` ?

If there is a strong argument for having a single function, I see 3 possibilities:
- adding another output argument (or field name for `get_optics`) to keep orbit and orbit_view separated. That's the cleanest solution,
- adding a flag `errors_enabled` which will trigger the sequence "enable errors, compute, process BPM errors", as the *_err functions do,
- putting the flag at the lattice level `ring.errors_enabled ` (Simon's proposal), with the same effect. In this case, when should the random errors be generated ? When setting the flag or when starting a computation ?

lfarv:
There is a question about random errors. At the moment they are generated in assign_errors and random and systematic errors are combined into a single value. I think it would be better to make the combination in enable_errors to avoid restarting from the bare lattice and reassigning the errors when needing a new test case.

It means that the error attributes should be tuples rather than numpy arrays. To be checked…

swhite2401:
> If there is a strong argument for having a single function, I see 3 possibilities:
>     * adding another output argument (or field name for `get_optics`) to keep orbit and orbit_view separated. That's the cleanest solution,
>     * adding a flag `errors_enabled` which will trigger the sequence "enable errors, compute, process BPM errors", as the *_err functions do,
>     * putting the flag at the lattice level `ring.errors_enabled ` (Simon's proposal), with the same effect. In this case, when should the random errors be generated ? When setting the flag or when starting a computation ?

I am not very keen on variable output functions, we did it in the past for other functions and I regret it very much. It is very convenient for the user the run the same code for a lattice with or w.o. errors and this approach discards this possibility.

Now for the other 2 options I am a bit confused by your comment, my understanding in the sequence/logic of applying errors and computation on lattice with errors is the following:

1. You call the initialization function `assign_errors` that fills the additional attributes with ALL errors, random, systematic, etc..., this is done once and for all and fully characterizes the errors associated with your lattice. Errors are never generated again unless `assign_errors` is called in which case they are overwritten. This lattice file with errors can be used to run many calculation using INDENTICAL errors set.
2. For now, 2 function are provided `*_err` to compute optics, orbit and trajectory in the presence of errors. In this case, errors are summed to design values internally in the decorator. The lattice is not modified. These are the functions that were questioned by @simoneliuzzo
3. There is a possibility to generate a lattice with errors: this generates a copy of the lattice with errors summed to the design values, in this case all errors are included except BPM errors. This lattice can be used to computed DA, lifetime, or as input to any function that is not decorated. This lattice with errors is "flagged" with an `has_errors` property to prevents that errors are added twice

Do you confirm this? If that is the case I do not understand the comment on random errors in 3... since ALL errors are generated in the initialization function `assign_errors`.

Now as far as I understand you are proposing 2 things (except for the first option that would not be my preference):
1-add a flag argument to `find_orbit / get_optics / lattice_pass` to trigger the calculation in the decorator `enable_errors`
2-add a lattice property to trigger the calculation in the decorator `enable_errors`

My personal preference goes to 2 in this case: no need to modify existing functions, just decorate them. Any code can be run with exactly the same syntax to compare results with and w.o. errors just by placing the lattice property `ring.errors_enabled` to `True` or `False`. This is by far the most convenient interface for users.

Did I miss something?

swhite2401:
There is only one case that I would find useful to "randomize" -> we may want at some point to introduce tbt BPM fluctuations for `lattice_pass` but this can be done in a later PR.

lfarv:
Concerning the choice of a single function for all case:

> My personal preference goes to 2 in this case: no need to modify existing functions, just decorate them. Any code can be run with exactly the same syntax to compare results with and w.o. errors just by placing the lattice property ring.errors_enabled to True or False. This is by far the most convenient interface for users.

So we are left with optics 1 and 2 !

I think 1 is cleaner because there is no possible confusion between orbit and BPM readings. Of course, I'm not thinking of variable outputs ! The BPM reading is always there, it's equal to orbit if there is no BPM error. For `get_optics`, it's easy: just adding a field in a structure. For `lattice_pass` and `find_orbit`, we need new functions: a single one for everything, but with a new name… Not the ideal solution.

But option 2 is good enough for me. my only concern is that you need to remember the flag to know whether it's an orbit or bpm reading. Positive point is that it's the easiest to implement.

I don't like optics 3, but apparently it's already out.

swhite2401:
> my only concern is that you need to remember the flag

Sorry I am still confused...the limits of github chat... I personally would discard option 1, we should not change the output structure of standard AT functions, we have seen all the damage it caused to users with the changes in `linopt.`

With this in mind we need a flag to enable or not the calculations with errors.

My preference clearly goes to a global lattice property that would activate errors calculation for all subsequent calculations. This is in my opinion much simpler: no changes of syntax in existing function, single place to activate/deactivate errors, no modification needed to existing users scripts to be run with errors.

You seem to prefer option 2, but my understanding is that it implies adding function arguments and that errors would be activate at the function level: this is not convenient for users and is exactly the opposite of what I would expect from such functionality in an ideal world, i.e. the same behavior as `ring.enable_6d()`
But maybe I am missing something: could you describe exactly what this flag is and how you intend to implement it? and also what you do not like about option 3? Because I don't seem to get it...

lfarv:
Concerning the random generation:

The assignment of errors is a rather long operation: you have to take care of different elements, magnets BPMs, and for each set a number of different parameters. So I had in mind that it could be done once for ever. The alignment tolerances and others are part of the lattice design… The lattice with assigned errors could be the one which is stored on file.

Instead, at the moment, if we want 10 error sets, we have to run 10 times the error assignment, with exactly the same parameters. That looks like a waste or resources. And since we have to start again from the bare lattice, a "copy" option in `assign_errors` is necessary.

What I propose is to keep `assign_errors` as it is, but postpone the random generation to `enable_errors`. This way, `assign_errors` is done once for ever, and if you run `enable_errors` 10 times, you get 10 copies of the ring with different error sets (BPM errors included, I come to that later). For random errors, statistics are crucial!

Of source you may want to keep a single error set for deeper studies:
- you can keep the output of `enable_errors` as long as you want, and use it everywhere you want, the `has_errors` flag freezes any further modification,
- you can use the `seed` argument to generate many times the same error set.

The problem is the same in the present situation: unless you give it a seed, `assign_errors` will generate different sets each time it is called, you cannot rely on it.

I do not see any drawback in this modification, only improvement. In addition, with this modification, the `has_errors` flag of the lattice could replace the proposed flag argument of `find_orbit`:
- if `ring.has_errors` is true, the flag argument is ignored (or gives a warning is it is explicitly set to false)
- if `ring_has_errors` is false or absent, the flag argument generates a temporary ring with errors, as `find_orbit_err` does at the moment.

Simple and clean !

swhite2401:
I am afraid we do not understand each other and that the github chat is somewhat inefficient for this kind of discussion.
I have a very different opinion on how this should be implemented, so what I propose is to just implement what I have in mind in a separate branch and we can continue this discussion once we can test and look at both versions.

lfarv:
Posts are crossing…

What I have in mind:

the equivalent of enable/disable_6d is not possible because with errors there is no way back with errors. So

- `assign_errors` stays as it is except that random generation is postponed
- `enable_errors` effectively generates random values (**all** the assigned errors including BPMS) and generates a new lattice with `has_errors` positioned. Practically, we just need 3 new hidden attributes on BPMS to store the generated cumulated errors. If needed, we may also add a kind of mask in the future to select some errors, I don't see how at the moment and it's not critical for now.
- `find_orbit` and `lattice_pass` detect the `has_errors` flag in the lattice to post process the results with BPM errors or not.
  I would also add an optional "with_errors" flag:
  - if `ring.has_errors `is already set: nothing more to do
  - if not, do the same as `find_orbit_err`: enable the errors in a temporary ring before computing, It avoids generating the lattice with errors in advance. For statistics it's simpler: a single function to call.
- the `*_err` functions disappear

If we don't go for the function flag, I have a question for `get_optics`. It could stay as it is: the magnet errors are in the lattice and will be taken into account. "closed_orbit" is the real closed orbit, no problem. Do we need BPM processing, knowing that this is already available in `find_orbit`? And if yes, I would store the result in a new field, so that there is no confusion with the real closed orbit.

lfarv:
@swhite2401 : I summarise what I understand: we are discussing about 2 points:

1. Suppression of the `*_err` functions (@simoneliuzzo end @lcarver comments). This now looks trivial for me:
   - `assign_errors` and `enable_errors` unchanged,
   - a simple test in `find_orbit` and others: if `has_errors` is set, process the BPMs after computation.
   You may think of more complicated solutions, but that one is the simplest and I think we can easily converge on that point.

2. Generation of random errors: there we disagree on time to do that, we have to look at number of use cases to make a decision, but may be a full implementation is not necessary. Let's think…

On my side, I'll not be too much available in the next 8 days, so we have time.

swhite2401:
Sure, please go ahead with your proposal and finalize it, I have something more ambitious in mind but it will take me a bit of time to come up with a proposal (also because I have many other things ongoing).
I just think it is much simpler to comment on written code in this github context...so give me a bit of time and I can propose something as a basis for discussion.

swhite2401:
One comment: in your proposal I do not see much interest in keeping the `*err` attribute once they have been added to the main ones. You may consider deleting them once they have been added and the lattice with errors generated.
This would make the implementation safer as it remove any possibility of counting errors twice and allows to change the `_has_errors` flag into `_bpm_errors_enabled` which makes more sense. It could be read/write to be used as a switch by the user.

lfarv:
> One comment

Agreed !

lfarv:
What's new in the last commit:

- the generation of random values is moved to `enable_errors`, with the following benefits:
  - no need to re-run `assign_errors` to generate a new error set: the errors are assigned once for ever,
  - easier to generate a **reproducible** set of errors: a single seed in `enable_errors` is needed rather than a seed in each of multiple `assign_errors` commands.
  
  For a given seed, the errors generated for an error category do not depend on the other enabled errors: for instance random multipole errors do not depend upon whether you enable alignment errors or not.
- there are no more special `*_err` functions. But a new boolean `monitor_errors` keyword is added to `lattice_pass`, `find_orbit` and `get_optics` commands. It is necessary because in the absence of special "bpm_reading" output argument, these function must still give the real orbit for internal use (optics or other parameter computation), even when errors are enabled in a lattice. The BPM error contribution is added to the "closed orbit" output in the last step when a user asks for monitor errors. Otherwise, the `apply_bpm_orbit_errors` function is available to convert any closed orbit output into monitor readings.

swhite2401:
@lfarv, I think we can close this one if you agree

PR #550, opened 2023-01-11
Fix bug in convolve_wakefun

lcarver:
convolve wakefun was incredibly wrong. Now it is working. 

image

### DIFF
--- a/pyat/at/collective/wake_functions.py
+++ b/pyat/at/collective/wake_functions.py
@@ -6,28 +6,32 @@ from at.constants import clight
 from scipy.interpolate import interp1d
 
 
-def convolve_wakefun(srange, w, sigs):
+def convolve_wakefun_fft(srange, w, sigs, xr=2*0.09, npoints=500000):
     """Convolution of a wake function with a pulse of rms
     length sigs, this is use to generate a wake potential
     that can be added to the output of EM code like GDFIDL"""
+
     sigt = sigs / clight  # Convert sigmaz to sigmat
-
+    nt = npoints + npoints - 1
+    
     # First we have to uniformly sample what is provided
-    min_step = numpy.diff(srange)[0]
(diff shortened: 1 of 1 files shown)


swhite2401:
Why not try to use numpy.convolve? This would save all the headache in trying to implement it yourself

lfarv:
@lcarver and @swhite2401:

> Why not try to use numpy.convolve? This would save all the headache in trying to implement it yourself

Any conclusion on that ?

lcarver:
```
import at
import numpy
import time
    
sigmaz = 1e-3
srange = numpy.linspace(-100*sigmaz, 0.2+1e-15, 2000)
long_wf = at.collective.wake_functions.long_resonator_wf(srange, 300e6, 1000, 1e6, 1)
#long_wf = numpy.heaviside(srange, 1)
now = time.time()
conv_fft = at.collective.wake_functions.convolve_wakefun_fft(srange, long_wf, sigmaz, xr=0.3, npoints=2000)
print(time.time() - now)

now = time.time()
conv_np = at.collective.wake_functions.convolve_wakefun_numpy(srange, long_wf, sigmaz)
print(time.time() - now)

fig, ax = plt.subplots()
ax.plot(srange, conv_fft, 'r')
ax.plot(srange, conv_np, color='b', linestyle='dashed')

#ax.set_xlim(-0.005, 0.005)
plt.show()
```
20-Feb-23-144906

I rewrote the convolve function and now there are 2 convolve functions in `wake_functions`. The numpy convolve function is much faster (0.7ms vs 3.5ms), is a much cleaner script and gives the same result. I have left both functions in the file for now, so that if you wish to test both the old and new functions you can. But I will delete the old method and keep only convolve_wakefun_numpy which I will rename to convolve_wakefun to be consistent with the old function

lcarver:
ready for merge (I will do it in about 7 hours if no response)

Issue #553, opened 2023-01-24
Frequency map with PyNAFF and patpass


oscarxblanco:
Dear all,

I have created a python function to calculate the frequency and diffusion maps for a given ring, using patpass parallel tracking and the PyNAFF python library.

I attach a copy of the function to this message in case it is of any use.
(Please remove the .txt name extension at the end of the filename)

[fmap_parallel_track.py.txt](https://github.com/atcollab/at/files/10489591/fmap_parallel_track.py.txt)

Best regards,
o

swhite2401:
Handled in PR #556

Issue #552, opened 2023-01-23
patpass number of cpu in a pool


oscarxblanco:
Dear all,

is there a way to know the number of cpus used in a patpass pool ?
it seems you that the number of cpus asked and given from a pool could be different.

Best regards,
o

swhite2401:
Hello @oscarxblanco,

If no users input is given, the poolsize is defined as below:
`pool_size = min(len(r_in[0]), multiprocessing.cpu_count(), DConstant.patpass_poolsize)`
which is the minimum between, number of particles tracked, available CPUs or global option `DConstant.patpass_poolsize`.

If you set the `pool_size` argument, then this number is used, however if the number of particles is smaller than the `pool_size` `patpass` will spawn only the needed number of threads which is in this case equal to the number of particles.

oscarxblanco:
Great, thank you !

Is there anyway to confirm the value of the poolsize once created ?

Best regards,
o

swhite2401:
Sorry having a closer look at the code to answer your question, I realized I gave you wrong information: if you set the argument `pool_size ` of `patpass` this number is used for the pool generation regardless of the number of particles.
If you want to control the number of processes this is probably what you want to use.

However if the number of particles is smaller than `pool_size`  you will have empty processes, I don't know whether this is a problem for you.

Issue #522, opened 2022-11-18
atidtable_dat (IdTablePass) in pyat


oscarxblanco:
Dear all,

is there a pyat function to create an element from a Radia magnetic field map (pass method 'IdTablePass') ?

In matlab AT it is called atidetable_dat, however, I haven't been able to find an equivalent in pyat.
The 'IdTablePass' is in the integrators folder but I see no usage of it.

Best regards,
o

lcarver:
Hello,
Have you tried generating a general element?
```

from at.elements import Element
elem = Element('name', PassMethod='IdTablePass', kwargs*)
```

where kwargs needs to provide all of the correct arguments for the passmethod.

oscarxblanco:
Dear @lcarver ,
thank you for pointing me out to the general element function. I had a quick view and it seems that it does not have all the arguments required by the field map tracking. 

Here is a list of the element arguments definitions in `elements.py` :

at/lattice/elements.py 
231     _BUILD_ATTRIBUTES = ['FamName']                                             
232     _conversions = dict(FamName=str, PassMethod=str, Length=float,              
233                         R1=_array66, R2=_array66,                               
234                         T1=lambda v: _array(v, (6,)),                           
235                         T2=lambda v: _array(v, (6,)),                           
236                         RApertures=lambda v: _array(v, (4,)),                   
237                         EApertures=lambda v: _array(v, (2,)),                   
238                         KickAngle=lambda v: _array(v, (2,)),                    
239                         PolynomB=_array, PolynomA=_array,                       
240                         BendingAngle=float,                                     
241                         MaxOrder=int, NumIntSteps=int,                          
242                         Energy=float,                                           
243                         )

While in matlab I get :
![IDmatlab](https://user-images.githubusercontent.com/108396048/204849481-000007c6-910d-4a67-896e-06d9fd013124.png)


Do these differences have an impact on the tracking ?

lcarver:
From looking at IDtablepass, the required arguments are: Length, xkick, ykick, xtable, ytable, Nslice

So if you use 
`elem = Element('name', PassMethod='IdTablePass', xkick=x0, ykick=x1, xtable=x2, ytable=x3, Nslice=x4)
`
then they will be passed to the C passmethod and it should work (assuming your input is ok).

swhite2401:
Hello @oscarxblanco, you are also more than welcome to propose a python implementation of this `at` element, we can help you during the review process.

oscarxblanco:
[...]

PR #556, opened 2023-02-05
Frequency Analysis

oscarxblanco:
Dear all,
deall @lfarv , @lcarver , @carmignani , @simoneliuzzo 

I have a first version of the frequency analysis function for your review.
It does the frequency analysis of a ring using PyNAFF lib and pyat parallel tracking (patpass).
The only fix requirement is a ring, and other default values could be set as needed.

The function returns two arrays : first, an array containing [xoffset, yoffset, nux, nuy, log10(|dnu|)/turns] for every particle that survives a given number of turns with a defined frequency; second, it returns particle losses if the lossmap flag was set or an empty array otherwise.

During the integration with pyat I came across with a deprecated use of numpy.Complex inside the PyNAFF library that limits the compatibility with numpy up to version 1.23.5.

This pulls request is related to issues  #522 , #552 and #553 .

Best regards,
o

## function usage example
```python
print(f'FMAP in 4D');
ring.disable_6d();
[xy_tune_nudiff_array, plosses_array] = fmap_parallel_track(ring,  coords=[-8,8,-3,3], steps=[100,100], scale='linear',  lossmap=False, verbose=False)
```


lcarver:
Thanks a lot!

Before I dive into it and provide a review. My first comment is about `PyNAFF`. 

We have included within PyAT a python implementation of the SUSSIX algorithm which is called harmonic analysis (written originally by J.Coello de Portugal of CERN around 2016).

You can find the functions in `at/pyat/at/physics/harmonic_analysis.py` and details are within.

We stripped down some of the functionalities from his original implementation to just provide some basic tune analysis from a provided centroid. Are there any functionalities missing from harmonic analysis that prevent you from replacing PyNAFF with the harmonic analysis module? I think this would be preferable to having a dependency on an external module.

lfarv:
@oscarxblanco : the reasons you give for the choice of PyNAFF make sense, and having a good NAFF implementation available in PyAT may be useful. I have however a question concerning the compatibility with `numpy`: for now, I think that staying one version behind the last one is not a problem, but that must be a temporary situation. Is the PyNAFF project still active ? it has been sleeping for about 2 years… Can we expect an update solving the "Complex" problem ? I see you raised the question few days ago… The correction looks simple.

Otherwise, we may need in the future to fork another branch, not a very elegant solution…

In the end, whatever the choice between the present "harmonic_analysys" implementation and PyNAFF, we should keep only one.

lfarv:
@oscarxblanco : I get 88 syntax warnings (trailing backslashes or semicolons) and 317 non-conformities with the PEP8 style guide. Could you do some cleaning ?

oscarxblanco:
@lfarv Those are a lot of warning ! Yes, I will do the cleaning. 
Is there any tool that you would suggest to check the code ?

@lfarv With respect to PyNAFF, none of the two developers is active on github, so I don't expect updates on the project. The installation method, using setup.py, might be also deprecated very soon. 
It seems to me that either PyNAFF sources are adopted by pyat, or I use another library.

If you are already convinced that harmonic_analysis do the task, it would be easier to just change the library. I'll have a look on harmonic_analysis library and get back to you.

lfarv:
@oscarxblanco:
> If you are already convinced that harmonic_analysis do the task, it would be easier to just change the library. I'll have a look on harmonic_analysis library and get back to you.

No, I'm not convinced! Looking at the code, I very much prefer PyNAFF. It's a single module, so if the license allows copying and modifying, I would indeed adopt it. Do you know how to do that in a clean way (PyAT itself is not licensed !)

About warnings and code style: I'm using the `PyCharm` IDE which gives all information while writing code,  but there are many IDE's doing the same (VSCode,...). For a dedicated verification tool, there is `flake8` which is used in the GitHub test sequence (you can look at the GitHub test results).

lfarv:
Another comment: having a module name identical to a function/class name creates problems. Once the function name is imported in the containing package name (`at.physics` here), the module becomes inaccessible: `at.physics.fmap_parallel_track` now means the function and hides the `at.physics.fmap_parallel_track.py` module. This has no consequence anywhere in code, but it has for the Sphinx documentation generator: it cannot document the module, unless we do it manually. So I would prefer you rename one of the two.

oscarxblanco:
I did a simplified comparison of the speed and precision of PyNAFF and J.Coello de Portugal (JCdP) library. Here is the difference between the returned fundamental frequency and the input frequency, for a variable sample length and two orthogonal signals (sin and cos).

PyNAFFvsJCdP
With respect to speed, I get that JCdP is 20 times faster : 
```python
  PyNAFF_time: 86.517541764013
  JCdP_time: 3.732474807999097
```
The behavior with random samples is also different. PyNAFF returns a void array with a warning message to delete de DC component, while JCdP returns zero.
If I pass a sin function and I ask for higher order components (should be zero, but, I wanted to see the result), PyNAFF returns a small value, while JCdP returns the fundamental.
In JCdP's inplementation, only the 'laskar' method gave a result. The 'fft' gave me only the integer part of the tune and a residual of order O(-1)

For the plot I used two orthogonal signals (sin and cos functions) for a fixed frequency and variable number of samples.
```python
fsin = 54.321123456789987654
fcos  = 12.345678998765432112
signalx = 10e-6*numpy.sin(2*numpy.pi*fsin*nss)
signaly = 20e-6*numpy.cos(2*numpy.pi*fcos*nss)
#and I pass the signals to both libraries, for example, the sin function :
FAsinJCdP[iter_index] = nts*get_tunes_harmonic(signalx,method='laskar', num_harmonics=1, hann=False)
FAsinPyNAFF[iter_index] = nts*PyNAFF.naff(signalx,nts,1,0,False)[0][1]
```

lfarv:
Looking at the code of the 2 versions, I think that what is called the 'laskar' method in harmonic_analysis.py is not at all a NAFF method: it simply recursively subtracts the main frequency obtained with fft from the signal and starts again up to the number of desired number of harmonics. It' simple and fast (your test shows it), but it's still basically an fft. It possibly does not give the accuracy and resistance to noisy data that NAFF is supposed to give…

Your test shows that PyNAFF gives a results 1 to 2 orders of magnitude closer to the true value than JCdP, but does not include noise.

oscarxblanco:
@lfarv 
I decided to do again the basic comparison w/o noise for a sin function of 100e-6 amplitude. This time I include also NAFFlib and normalize to the input frequency.
PyNAFFvsJCdPvsNAFFlib
Adding -56db of uniform noise, equivalent to noise on the 5th bit (.15e-6 amplitude), the response of all modules in similar.
PyNAFFvsJCdPvsNAFFlib_m28dB
Adding a maximum of -3dB of uniform noise, equivalent to noise in the 2nd bit, (70e-6 amplitude), the response is as follows :
PyNAFFvsJCdPvsNAFFlib_m3dB

With respect to speed, I get :
```
$ python timeit_PyNAFF_JCdP_NAFFlib.py 
  PyNAFF_time: 40.43559245800134
  JCdP_time: 2.5347508350387216
  NAFFlib_time: 4.345918281993363
```

I attach to this message the speed test script in python that I used.
[timeit_PyNAFF_JCdP_NAFFlib.py.txt](https://github.com/atcollab/at/files/10684302/timeit_PyNAFF_JCdP_NAFFlib.py.txt)


Best regards,
o

swhite2401:
Dear all,

A comprehensive comparison between algorithms can be found in:
https://journals.aps.org/prab/abstract/10.1103/PhysRevAccelBeams.19.054001

In this reference SUSSIX (on what `harmonic_analysis`) is based and NAFF gives similar results in the absence of noise while with noise all methods are more or less equivalent. It would be good to understand what was stripped down from SUSSIX that explains the loss in precision in the absence of noise.

However, it seems that in the presence of noise (that is the case we are interest in) all the benefit is lost. I would therefore strongly advise to keep the fastest method, a factor 20 is a big deal when you have to analyse large data sets and cannot afford a cluster.

lfarv:
Very interesting article. 

> It would be good to understand what was stripped down from SUSSIX that explains the loss in precision in the absence of noise

Looking at the `harmonic_analysis` code, as long as only the main harmonic is considered, it's a **simple fft** with an interpolation over the 3 highest peaks  (lines 42 - 42). All the rest deals with the following harmonics, which are not used in the two cases we are considering for the moment.

So I agree with @swhite2401, the fastest method is the best choice. 

In the future, if we have to extract additional frequencies with small amplitudes compared to the main one, one may raise the question again (and this is not checked in the article).

oscarxblanco:
@swhite2401 
Thank you for the article, it is very ilustrative on the effect of noise in the tune calculation.
I, however, doubt this is a good reference to compare with as the results refer to a particular application of the algorithms, and not to numerical precision of the implementation.

In any case, I'll make the changes to implement the frequency map with JCdP's algorithm and have a look on the result.

Best regards,
o

swhite2401:
Fig. 5 and Fig. 6 are considering simple cosine functions (see Eq. 18 and 19), I think it is very similar to what you have done. 

@lcarver is the hanning window applied in `harmonic_analysis`? If not, this may explain some of the discrepancies and we may consider to introduce it, it is quite simple.

oscarxblanco:
@swhite2401 
Yes, it is similar. And it is the result when considering noise.

swhite2401:
@oscarxblanco the option `hann` is there in `harmonic_analysis` but set to `False` by default. It could be interesting to evaluate its impact on the resolution.

swhite2401:
> Yes, it is similar. And it is the result when considering noise.

Ah ok, got it! In fact there is no data for the case without noise, sorry I scrolled too quickly -> this just confirms that in the presence of noise both algorithms are equivalent.

oscarxblanco:
@swhite2401 
I made a test with the following call setting hann=True and hann=False, but, the result is the same.
```python
from harmonic_analysis import get_tunes_harmonic
get_tunes_harmonic(xReadings, method='laskar', num_harmonics=1, hann=True)
```

lfarv:
@oscarxblanco , @swhite2401 

> I made a test with the following call setting hann=True and hann=False, but, the result is the same.

Of course, because whatever you put in `hann`, no window is applied with the so-called 'laskar' method. Which is by the way not at all related to Laskar, it's a simple fft !

But indeed, windowing should improve the results. This code should be rewritten, it's poorly written…

lfarv:
> properly implementing the hanning window in all modes seems like a relatively simple upgrade.

When restricted to the main harmonic, it's true. When more harmonics are requested (default is 20), it's not: this is done by subtracting the previously found harmonic from the signal. If a window is applied, it's complicated.

For now, we could take `harmonic_analysis` as it is (no window): it's fast and reasonably accurate. That means modify `fmap_parallel_track.py` to use `get_tunes_harmonic` and remove the dependency to PyNAFF.

Later, one can still upgrade `get_tunes_harmonic` to implement windowing or other methods, it will be a single function modify.

Splitting in 2 steps allows this PR to be merged so that frequency maps are operational, and leave the possibility to improve the tune detection later.

swhite2401:
> When more harmonics are requested (default is 20), it's not

Well you may use the raw signal for the subtractions and apply the windowing only to get the harmonic frequency/phase/amplitude. Anyway I agree that this does not belong to this PR.

oscarxblanco:
Dear all,
as a last qualitative example I leave here an image of two diffusion maps over a lattice.
The lost in numerical precision reduces the clarity in some traces.
Screenshot from 2023-02-15 16-44-19

This time I won't be able to provide the lattice I used, but here is the call to the function :
```python
[xy_tune_nudiff_array, plosses_array] = \
    fmap_parallel_track(ring, \
    coords=[-8,8,-3,3], \
    steps=[200,200], \
    scale='linear', \
    lossmap=True,
    verbose=False)
```
Best regards,
o

oscarxblanco:
> Another comment: having a module name identical to a function/class name creates problems. Once the function name is imported in the containing package name (`at.physics` here), the module becomes inaccessible: `at.physics.fmap_parallel_track` now means the function and hides the `at.physics.fmap_parallel_track.py` module. This has no consequence anywhere in code, but it has for the Sphinx documentation generator: it cannot document the module, unless we do it manually. So I would prefer you rename one of the two.


Dear @lfarv ,
are there any suggested names for the python file and function ?

Best regards,
o

lfarv:
@swhite2401:

> Well you may use the raw signal for the subtractions and apply the windowing only to get the harmonic frequency/phase/amplitude. Anyway I agree that this does not belong to this PR.

I suspect that that amplitude you get is affected by the windowing so that you cannot use it on raw data, but I may be wrong…

@oscarxblanco: Very nice plots !
why not using `frequency_maps.py` for the module and keep you function name? Unless someone has a better idea…
Otherwise, this branch is ok for me.

oscarxblanco:
Dear @lfarv ,

the filename has been changed to frequency_maps.py in the last commits.

Best regards,
o

swhite2401:
Hello @oscarxblanco, following some remarks by @simoneliuzzo I am fixing a few things in the tune calculation from tracking. In particular when the particle used for the fft was lost (i.e. nan coordinates) the function `get_tunes_harmonic` was crashing with a `ValueError`.
Instead the function `get_tunes_harmonic` will now return `nan` when the tracking fails. Is this ok with your implementation? Any suggestion on your side?

oscarxblanco:
@swhite2401 
I checked the modifications and they have no impact on this branch. 

I took the oportunity to add to more parameters to the output : dnux, dnuy. So, to recap, 

The function returns two arrays : 
- first, an array containing 
[xoffset, yoffset, nux, nuy, dnux, dnuy, log10(|dnu|/turns)] 
for every particle that survives a given number of turns with a defined frequency; 
- second, it returns particle losses if the lossmap flag was set or an empty array otherwise.

Best regards,
o

pyat/at/physics/frequency_maps.py
```diff
@@ -0,0 +1,283 @@
...
+              however, it is not returned by this function.
+    WARNING : loss map format is experimental
+    """
+
+    if co:
+        # closed orbit values are not returned. It seems not necessary here
+        addco = find_orbit(ring)
+        print(f'Closed orbit:\t{addco}')
+    else:
+        addco = [numpy.zeros((6, 1))]
```
lfarv, line 97: The usual way to handle closed orbit in PyAT is to add a keyword argument `orbit=None`. Then, here you would have:
```python
if orbit is None:
    # closed orbit values are not returned. It seems not necessary here
    addco = find_orbit(ring)
    print(f'Closed orbit:\t{addco}')
else:
    addco = orbit
```
You can then remove the `co` flag. And even simpler, use directly `orbit` instead of `addco` later.

pyat/at/physics/frequency_maps.py
```diff
@@ -0,0 +1,283 @@
...
+        z0 = numpy.zeros((6, lenixarray))
+        # z0 = z0 + add_offset6D + addco[0,:];
+        # z0[0,:] = z0[0,:] + xscale*ixarray + 1e-9;
+        # z0[2,:] = z0[2,:] + yscale*iy      + 1e-9;
+        z0[0, :] = xscale * ixarray + add_offset6D[0] + addco[0][0] + 1e-9
+        z0[1, :] = add_offset6D[1] + addco[0][1]
+        z0[2, :] = yscale * iy + add_offset6D[2] + addco[0][2] + 1e-9
+        z0[3, :] = add_offset6D[3] + addco[0][3]
+        z0[4, :] = add_offset6D[4] + addco[0][4]
+        z0[5, :] = add_offset6D[5] + addco[0][5]
```
lfarv, line 179: The commented version is more efficient (with `z0 = z0 + add_offset6D + orbit`)
Is there any problem there ?

pyat/at/physics/frequency_maps.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def fmap_parallel_track(ring,
+                        coords=[-10, 10, -10, 10],
+                        steps=[100, 100],
+                        scale='linear',
+                        turns=512,
+                        ncpu=30,
+                        co=True,
+                        add_offset6D=numpy.zeros((6, 1)),
```
lfarv, line 28: Why not
```python
add_offset6D=numpy.zeros(6),
```
Later on, you handle it as a 1D array

pyat/at/physics/frequency_maps.py
oscarxblanco: @lfarv 
This change has been implemented in commit @oscarxblanco
[removes co flag, adds orbit](https://github.com/atcollab/at/pull/556/commits/5998db4ec58873ab9b3226c61a66522ccbca9a17)

pyat/at/physics/frequency_maps.py
oscarxblanco: @lfarv 
These changes have been implemented in commit https://github.com/atcollab/at/pull/556/commits/6503cf4a5baefdf74176a3feb77b4c5921160185

@lfarv thank you for the notes on the transpose and memory organization.
I implemented the changes and they work, but, I don't know the details of C-aligned and Fortran-aligned memory handling. Please, let me know if you have any documentation I could read on this topic.

pyat/at/physics/frequency_maps.py
oscarxblanco: Implemented in commit https://github.com/atcollab/at/pull/556/commits/6503cf4a5baefdf74176a3feb77b4c5921160185

pyat/at/physics/frequency_maps.py
lfarv: Ordering is specified in numpy by the "order" parameter in many function calls, for instance for[ array creation](https://numpy.org/doc/1.23/reference/generated/numpy.array.html#numpy-array). You can retrieve it in the ["flags" attribute](https://numpy.org/doc/1.23/reference/generated/numpy.ndarray.flags.html?highlight=ndarray%20flags#numpy-ndarray-flags) of any array.

More generally, [this article](https://en.wikipedia.org/wiki/Row-_and_column-major_order) describes the memory layout of multidimensional arrays.

Usually, you don't have to care about memory layout because numpy handles this for you. The problem appears when the data is processed in another language (C for AT): the C code must know how the data is organised in memory. So `lattice_pass` and the element constructors reorganise the arrays if necessary, but it' more efficient if you can avoid it.

For info, Fortran-style ordering was chosen because it's the way Matlab organises its data. numpy lets you chose, and the default is C-style.

oscarxblanco: Thank you @lfarv

oscarxblanco (commented):
This code has been tested with a lattice and results are similar to MATLAB.
Speed tests with different frequency analysis libraries showed better results with this implementation and a relatively small loss on precision.
Numerical precision could be improved later on by changing/improving the frequency analysis library.

o

lfarv (approved):
Ok for me, you may merge

PR #559, opened 2023-02-17
improve help of non-linear module

swhite2401:
The PR was initially meant to improve the help of the non-linear module. 
In addition it fixes a nasty behavior of `get_tune_harmonics` that was crashing with an obscure `ValueError` when the tracking was failing.
If the tracking fails a nan is returned in this proposal (I hope this does not crash with #556).

More generally, my feeling is that the module 'nonlinear.py' would profit a lot from a good clean-up. 
Since this is closely linked to `harmonic_analysis`, that also needs some re-writing as mentioned in #556 , I would like to suggest to plan a complete re-write in a separate PR and keep this one as a temporary fix.

### DIFF
--- a/pyat/at/physics/harmonic_analysis.py
+++ b/pyat/at/physics/harmonic_analysis.py
@@ -261,6 +261,9 @@ def get_tunes_harmonic(cents, method: str = 'laskar',
                                           num_harmonics=num_harmonics,
                                           method=method,
                                           hann=hann)
-        tunes[i] = get_max_spectrum(freq, amp, fmin, fmax)
+        try:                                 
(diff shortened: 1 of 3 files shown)


lfarv:
@swhite2401: I fully agree with your plan. So if this PR does not conflict #556 (it should rather help!), it's good for me.

oscarxblanco (approved):
I saw the changes and they seemed not to interfere with the PR branch development.
Just to be sure, I did checkout the three modified files and launch an fmap calculation without any problem; I did not expect any problem because I check the tracking array for nans before analyzing it with the harmonic frequency library.

This modification does not crash with the frequency analysys on #556 .

Best regards,
o

PR #566, opened 2023-02-23
Axis definitions

lfarv:
Many functions require the user to specify a coordinate axis (x, p<sub>x</sub>, y, p<sub>x</sub>, &delta;, c&tau;)  or a plane (horizontal, vertical or longitudinal).
A new module `axisdef.py` introduces helper functions to make the specification of coordinate axes or plane selection easier.

These functions supersede the `axis_descr` function of the `utils.py` module.

### DIFF
--- a/pyat/at/lattice/axisdef.py
+++ b/pyat/at/lattice/axisdef.py
@@ -1,9 +1,11 @@
 """Helper functions for axis and plane descriptions"""
 from __future__ import annotations
 from typing import Optional, Union
+# For sys.version_info.minor < 9:
+from typing import Tuple
 
 AxisCode = Union[str, int, slice, None, type(Ellipsis)]
-AxisDef = Union[AxisCode, tuple[AxisCode, AxisCode]]
+AxisDef = Union[AxisCode, Tuple[AxisCode, AxisCode]]
 
 _axis_def = dict(
     x=dict(index=0, label="x", unit=" [m]"),
@@ -36,11 +38,6 @@ _plane_def['x'] = _plane_def['h']
 _plane_def['y'] = _plane_def['v']
 _plane_def[None] = dict(index=slice(None), label="", unit="", code=":")
 _plane_def[Ellipsis] = dict(index=Ellipsis, label="", unit="", code="...")
-
-_no_def = {
-    None: dict(index=slice(None), label="", unit="", code=":"),
-    Ellipsis: dict(index=Ellipsis, label="", unit="", code="...")
-}
 
 
 def _descr(dd: dict, arg: AxisDef, key: Optional[str] = None):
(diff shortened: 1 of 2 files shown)


pyat/at/lattice/axisdef.py
```diff
@@ -0,0 +1,163 @@
...
+
+          1, 'px', PX', 'xp'
+
+          2, 'z', 'Z'
+
+          3, 'py', 'PY', 'yp'
+
+          4, 'dp', 'delta'
+
+          5, 'ct'
```
swhite2401, line 74: What about `s` , this is also a notation that is used very often

pyat/at/lattice/axisdef.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Parameters:
+        plane:          plane code or tuple[plane code] selecting planes.
+          planes for the 3 planes are:
+
+          0, 'h', 'H', 'x' for horizontal plane,
+
+          1, 'v', 'V', 'y' for vertical plane,
+
+          2, 'l', 'L' for the longitudinal plane
```
swhite2401, line 129: Again why not `z` or `s` ? This is much more natural that `L` that is never used as it generally defines a length

swhite2401:
Very useful!

pyat/at/lattice/axisdef.py
lfarv: Right! Added 's'

pyat/at/lattice/axisdef.py
lfarv: I initially wanted to distinguish axes  (x, p<sub>x</sub>,...) and planes (h <=> (x, p<sub>x</sub>)...). But you are right, it seems unavoidable to also use x, y, z for planes.

pyat/at/lattice/axisdef.py
```diff
@@ -0,0 +1,164 @@
...
+
+    Parameters:
+        axis:           axis code or tuple of axis codes selecting axes in the
+          standard AT coordinate system. codes for the 6 axes are:
+
+          0, 'x', 'X'
+
+          1, 'px', PX', 'xp'
+
+          2, 'z', 'Z'
```
swhite2401, line 69: I did not see this.... `z` notation for the vertical axis is really not used by many anymore and for very good reasons.... in case you insist in keeping it could you at least add `y` and `Y` to the list?

pyat/at/lattice/axisdef.py
```diff
@@ -0,0 +1,164 @@
...
+
+          3, 'py', 'PY', 'yp'
+
+          4, 'dp', 'DP', 'delta'
+
+          5, 'ct', 'CT', 's'
```
swhite2401, line 75: What about `S` ?

pyat/at/lattice/axisdef.py
```diff
@@ -0,0 +1,164 @@
...
+    h=dict(index=0, label="x", unit=" [m]"),
+    v=dict(index=1, label="y", unit=" [m]"),
+    l=dict(index=2, label="z", unit="")
+)
+for xk, xv in [it for it in _plane_def.items()]:
+    xv['code'] = xk
+    _plane_def[xv['index']] = xv
+    _plane_def[xk.upper()] = xv
+_plane_def['h'] = _plane_def['x']
+_plane_def['v'] = _plane_def['y']
```
swhite2401, line 39: Either this or the _plane_def initialization has to be modified

pyat/at/lattice/axisdef.py
```diff
@@ -0,0 +1,164 @@
...
+def axis_(axis: AxisDef, key: Optional[str] = None):
+    r"""Return axis descriptions
+
+    Parameters:
+        axis:           axis code or tuple of axis codes selecting axes in the
+          standard AT coordinate system. codes for the 6 axes are:
+
+          0, 'x', 'X'
+
+          1, 'px', PX', 'xp'
```
swhite2401, line 67: Is `xp` (which for me corresponds to `x'`) really something we want to allow in AT, I think this brings confusion w.r.t. coordinates definitions

pyat/at/lattice/axisdef.py
lfarv: Typo: should be y and Y

pyat/at/lattice/axisdef.py
lfarv: Ok, I remove xp and yp

swhite2401 (approved):

PR #569, opened 2023-02-28
Bug fix on atrotatelattice

lfarv:
the Matlab function `atrotatelattice` was not updating the `cavpts` ring property which is stored in the RingParam element (bug reported be @simoneliuzzo). This is corrected.

In addition, as this function appears to be used, it is moved from the `pubtools` directory (non-validated functions) to the standard `atmat/lattice` directory.


simoneliuzzo (approved):
I have tested the method with and without RingParam element in the lattice. It appears working as expected.

PR #571, opened 2023-03-02
Undulator dipoles

simoneliuzzo:
I try to add a method to create an undulator made of dipoles. 

an example demo is provided in the pyat/examples/undulators

How to merge such list of elements (that gives the correct emittance and energy loss) with an IdTablePass element?

### DIFF
--- a/pyat/at/lattice/__init__.py
+++ b/pyat/at/lattice/__init__.py
@@ -9,6 +9,7 @@ from .options import DConstant, random
 from .particle_object import Particle
 from .elements import *
 from .idtable_element import InsertionDeviceKickMap
+from .undulator_dipoles import undulator_dipoles, insert_undulator
 from .utils import *
 from .lattice_object import *
 from .cavity_access import *
--- a/pyat/at/lattice/undulator_dipoles.py
+++ b/pyat/at/lattice/undulator_dipoles.py
@@ -82,14 +82,14 @@ def undulator_dipoles(L, nperiod,
 
         return undper
 
-    if magnetmodel is 'multipoles':
+    if magnetmodel == 'multipoles':
 
         undperiod = makeundperiod(
             at.Multipole('NegPole', LPole, -AngPole / LPole),
             at.Multipole('PosPole', LPole,  AngPole / LPole),
             at.Drift('PoleGap', DistPole))
 
-    elif magnetmodel is 'rectangularbend':
+    elif magnetmodel == 'rectangularbend':
 
         undperiod = makeundperiod(
             at.Dipole('NegPole', LPole, bending_angle=-AngPole,
@@ -147,10 +147,10 @@ def insert_undulator(ring, und, index, matchoptics=False):
     
         # add some PolynomB in case changes relative to 0.
         for ind, el in enumerate(ring_und):
-            if el.FamName is 'PosPole':
+            if el.FamName == 'PosPole':
                 el.MaxOrder = 1
                 el.PolynomB[1] = 0.0
-            if el.FamName is 'NegPole':
+            if el.FamName == 'NegPole':
                 el.MaxOrder = 1
                 el.PolynomB[1] = 0.0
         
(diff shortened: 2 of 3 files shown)


oscarxblanco:
Good question @simoneliuzzo ,

I was thinking to use a wiggler element, and did try to use pyat to calculate the radiation integrals with no sucess.
I'll have a look on your example asap.

o

simoneliuzzo:
Thank you @oscarxblanco 
there is no urgency, as you see I do not even pass the tests yet!

simoneliuzzo:
@lcarver and @swhite2401 suggested to give the possibility to define an undulator that has both the radiation effects and the kick map effects. However both elements are thick. How to put them together? @oscarxblanco and @lfarv may be have some ideas?

swhite2401:
Hello @simoneliuzzo do you want me to review this PR?

simoneliuzzo:
Dear @oscarxblanco ,

do you know if we can put 0 length in InsertionDeviceKickMap element? 
that may be an easy solution to make the two work together.

thank you
best regards

oscarxblanco:
@simoneliuzzo It is not possible because Length is divided by 2Nslice in IdTablePass.c (line 56).

I would suggest to create a new element with Length and IntegratedLength.
Length coud be varied as needed for plots, slices and so on.
IntegratedLength is always kept unchanged and passed to the InsertionDeviceKickMap.

That new element should inherit (all or most) properties from InsertionDeviceKickMap.

Best regards,
o

simoneliuzzo:
Dear @lfarv 

I understand and completely agree with your description. 

The problem for me is that there is no wiggler or undulator method doing really all that we would like to see from an ID or wiggler pass. IdTablePass and WigglerLinearPass do not give the correct emittance and energy loss. Dipoles model do not give the correct tracking. 

This pull request is open simply to point out this rather important missing feature for a code that is used mostly by synchrotron light sources.

I did not intend to merge the methods I have submitted, but only to trigger this discussion (and share the files) to think about what should be done in order to have what we we really need: a method to add to an AT lattice an ID that gives the correct tracking and global parameters. 

Concerning the dipole model we recently compared the radiated power computed by the dipole model and the values obtained from formulas and from the control system:

Screenshot 2023-03-10 at 08 47 10

93KeV from the dipoles model, 100KeV from the cavities.

simoneliuzzo:
I am preparing the test with wiggler pass and will share the result as soon as they are ready

simoneliuzzo:
I am unable to continue the tests with WigglerLinearPass. 
Even if the optics are visible with plot_beta(), at.envelope_parameters gives an error: Unstable Ring.
Screenshot 2023-03-10 at 09 50 21

and the lattice file if you want to try your self.
[ring_w150_wiggler_25mm.mat.zip](https://github.com/atcollab/at/files/10940276/ring_w150_wiggler_25mm.mat.zip)

simoneliuzzo:
I switched pass Method for the wiggler element to WiggLinearPass.

now the optics return nan 
and the tracking fails with error  'Wiggler' object has no attribute 'InvRho'

So I think that the either the Wiggler element has some problem or I am not using the correct element. Likely the second. 
Please advise.

simoneliuzzo:
Lattice with WiggLinearPass element:
[ring_WiggLinearPass.mat.zip](https://github.com/atcollab/at/files/10940377/ring_WiggLinearPass.mat.zip)

simoneliuzzo:
Dear @lfarv, @oscarxblanco, @swhite2401 and @lcarver,

@lcarver noticed that the new passmethod https://github.com/atcollab/at/pull/632 could solve the problem we are discussing here. By adding to the undulator pass also the energy loss field, we may have an undulator model for AT that reproduces both the kickmaps effects and the energy loss per turn. 
The energy loss would need to be computed before hand with this solution. 

regards
simone

swhite2401:
@simoneliuzzo can we close this PR? Now that the wiggler passmethod is fixed this seems to be obsolete to me right?

Issue #572, opened 2023-03-05
lattice operators: function and documentation


swhite2401:
It would be useful to add functions equivalent to the lattice operators and document them such that they appear on the sphinx page

PR #573, opened 2023-03-05
Lattice operators

swhite2401:
This PR proposes to introduce function as alternatives to the operators `+`, `+=`, `*`. This allows users to use their favorite syntax (function or operator) but also to document these functionalities in the sphinx webpage.

The default behavior remains unchanged except for the operator `*` that is now using deep copies of the elements of the original ring for the repetition. 
All functions are provided with an optional argument `copy` that allow to use deep copies of the elements to be appended, inserted or concatenated.

This issue was discussed in #565 and reported in #572. There might be other options to achieve this, please advise.

### DIFF
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -261,27 +261,15 @@ class Lattice(list):
 
     def __add__(self, elems):
         """Add elems, an iterable of AT elements, to the lattice"""
+        return merge_lattices(self, elems)
+
+    def __iadd__(self, elems):
         newring = Lattice(self)
         newring.extend(elems)
         return newring
 
-    def __iadd__(self, elems):
-        return super(Lattice, self).__iadd__(self._addition_filter(elems))
-
     def __mul__(self, n):
-        """Repeats n times the lattice"""
-        periodicity = self.periodicity
-        if n != 0 and periodicity > 1:
-            nbp = periodicity / n
-            periodicity = int(round(nbp))
-            if abs(periodicity - nbp) > _TWO_PI_ERROR:
-                warn(AtWarning('Non-integer number of cells: {}/{}. Periodi'
-                               'city set to 1'.format(self.periodicity, n)))
-                periodicity = 1
-        # noinspection PyTypeChecker
-        return Lattice(elem_generator,
-                       itertools.chain(*itertools.repeat(self, n)),
-                       iterator=self.attrs_filter, periodicity=periodicity)
+        return self.repeat(n)
 
     def _addition_filter(self, elems: Iterable[Element]):
         cavities = []
@@ -319,6 +307,13 @@ class Lattice(list):
         super().insert(idx, elem)
 
     def extend(self, elems: Iterable[Element]):
+        r"""This method adds all the elements of `elems` to the end of the
+            lattice. The behavior is the same as for a :py:obj:`list`
+            
+            Equivalents syntaxes:        
+            >>>ring.extend(elems)
+            >>>ring += elems          
+        """
         if hasattr(self, '_energy'):
             # When unpickling a Lattice, extend is called before the lattice
             # is initialized. So skip this.
@@ -326,7 +321,55 @@ class Lattice(list):
         super().extend(elems)
 
     def append(self, elem: Element):
+        r"""This method overwrites the inherited method :py:meth:`list.append()`,
+            it behavior is changed, it accepts only AT lattice elements 
+            :py:obj:`Element` as input argument.  
+
+            Equivalents syntaxes:        
+            >>>ring.append(elem)
+            >>>ring += [elem]                     
+        """
         self.extend([elem])
+        
+    def repeat(self, n: int, copy=True):
+        r"""This method allows to repeat the lattice `n` times. 
+            If `n` does not divide `ring.periodicity`, the new ring 
+            periodicity is set to 1, otherwise  it is et to
+            `ring.periodicity /= n`.
+            
+            Parameters:
+                n (int): number of repetition
+            
+            Keyword Arguments:
+                copy(bool): Default :py:obj:`True`. If :py:obj:`True`
+                            deepcopies of the lattice are used for the
+                            repetition
+                
+            Returns:
+                newring (Lattice): the new repeated lattice
+
+            Equivalents syntaxes:        
+            >>>newring = ring.repeat(n)
+            >>>newring = ring * n                     
+        """
+        """Repeats n times the lattice"""
+        periodicity = self.periodicity
+        if n != 0 and periodicity > 1:
+            nbp = periodicity / n
+            periodicity = int(round(nbp))
+            if abs(periodicity - nbp) > _TWO_PI_ERROR:
+                warn(AtWarning('Non-integer number of cells: {}/{}. Periodi'
+                               'city set to 1'.format(self.periodicity, n)))
+                periodicity = 1
+        # noinspection PyTypeChecker
+        elems = []
+        for i in range(n):
+            if copy:
+                elems += self.deepcopy() 
+            else:
+                elems += self            
+        return Lattice(elem_generator, elems, 
+                       iterator=self.attrs_filter, periodicity=periodicity)
 
     @property
     def attrs(self) -> Dict:
@@ -1212,6 +1255,33 @@ class Lattice(list):
             # noinspection PyAttributeOutsideInit
             self._radiation = radiate
             return radiate
+            
+            
+def merge_lattices(*lattices: Tuple[Lattice, ...], copy=False):
+    """Merge several lattices together
+    
+    Equivalents syntaxes:  
+    >>>newring = merge_lattices(ring1, ring2)           
+    >>>newring = ring1 + ring2 
+
+    Parameters:
+        lattices:  lattices to be merged
+        
+    Keyword Arguments:
+        copy(bool): Default :py:obj:`False`. If :py:obj:`True`
+                            deepcopies are used to produce the merged
+                            newring
+
+    Returns:
+        newring(Lattice): merged Lattice
+    """    
+    newring = []
+    for l in lattices:
+        if copy:
+            newring += l.deepcopy()
+        else:
+            newring += l
+    return newring 
 
 
 def lattice_filter(params, lattice):
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -1258,7 +1258,8 @@ class Lattice(list):
             return radiate
 
 
-def merge_lattices(*lattices: Tuple[Union[Lattice, List], ...], copy=False):
+def merge_lattices(*lattices: Tuple[Union[Lattice, List], ...],
+                   copy=False) -> Lattice:
     """Merge several lattices together
 
     Equivalents syntaxes:
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -261,10 +261,10 @@ class Lattice(list):
 
     def __add__(self, elems):
         """Add elems, an iterable of AT elements, to the lattice"""
-        return merge_lattices(self, elems, copy=True)
+        return merge_lattices(self, elems)
 
     def __iadd__(self, elems):
-        return merge_lattices(self, elems)
+        return merge_lattices(self, elems, inplace=True)
 
     def __mul__(self, n):
         return self.repeat(n)
@@ -1256,8 +1256,8 @@ class Lattice(list):
             return radiate
 
 
-def merge_lattices(*lattices: Tuple[Iterable[Element], ...],
-                   copy=False) -> Lattice:
+def merge_lattices(lattice0: Lattice, *lattices: Tuple[Iterable[Element], ...],
+                   copy=False, inplace=False) -> Lattice:
     """Merge several lattices together
 
     Equivalents syntaxes:
@@ -1265,20 +1265,29 @@ def merge_lattices(*lattices: Tuple[Iterable[Element], ...],
     >>>newring = ring1 + ring2
 
     Parameters:
-        lattices:  lattices to be merged
+        lattice0:  initial lattice
+        lattices:  lattices to be merged to lattice0
 
     Keyword Arguments:
         copy(bool): Default :py:obj:`False`. If :py:obj:`True`
-                            deepcopies are used to produce the
-                            merged newring
+                            deepcopies of the elements of lattices
+        inplace(bool): Default :py:obj:`False`. If :py:obj:`True`
+                       the lattice `lattice0 is modified in place.
+                       Otherwise a new Lattice object is created.
 
     Returns:
         newring(Lattice): merged Lattice
     """
-    newring = Lattice(lattices[0])
-    for lat in lattices[1:]:
-        newring.extend(lat, copy=copy)
-    return newring
+    if not isinstance(lattice0, Lattice):
+        raise AtError('Input argument lattice0 is not a Lattice object')
+
+    if inplace:
+        lattice = lattice0
+    else:
+        lattice = Lattice(lattice0)
+    for lat in numpy.atleast_2d(lattices):
+        lattice.extend(lat, copy=copy)
+    return lattice
 
 
 def lattice_filter(params, lattice):
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -272,7 +272,7 @@ class Lattice(list):
         return self.repeat(n)
 
     def __reversed__(self):
-        return self.reverse(copy=True)
+        return (el.swap_faces(copy=True) for el in self[::-1])
 
     def _addition_filter(self, elems: Iterable[Element], copy_elements=False):
         cavities = []
@@ -453,9 +453,8 @@ class Lattice(list):
         of elements. Alignment errors are not swapped
 
 
-        Equivalents syntaxes:
+        Usage:
         >>> newring = ring.reverse(copy=True)
-        >>> newring = reversed(ring)
 
         Keyword Arguments:
             copy(bool): Default :py:obj:`False`. If :py:obj:`True`
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -309,10 +309,10 @@ class Lattice(list):
     def extend(self, elems: Iterable[Element]):
         r"""This method adds all the elements of `elems` to the end of the
             lattice. The behavior is the same as for a :py:obj:`list`
-            
-            Equivalents syntaxes:        
+
+            Equivalents syntaxes:
             >>>ring.extend(elems)
-            >>>ring += elems          
+            >>>ring += elems
         """
         if hasattr(self, '_energy'):
             # When unpickling a Lattice, extend is called before the lattice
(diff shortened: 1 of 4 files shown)


lfarv:
@swhite2401 :this looks very nice, but I have a naming problem:

The `copy` keyword argument is already used in may lattice methods like `enable_6d`, `set_cavity`, `set_value_refpts`, `add_beamloading` and more, with a meaning identical to the `inplace` keyword of you new methods. On the other hand, your `copy` keyword has a different meaning (deep copy of elements). 

I agree that you convention makes sense, but I think it's too confusing. I suggest `element_copy` (or anything better) for the new keyword.

swhite2401:
> I agree that you convention makes sense, but I think it's too confusing. I suggest `element_copy` (or anything better) for the new keyword.

Ok no problem.

By the way, I have an issue with sphinx... I have not sure this can be solved easily. In the ` lattice_object` documentation I have that ALL the method added to `lattice_object` (find_optics, get_optics, etc....) are document again for `lattice_object` which makes it very difficult to read (the page is way too long...).
Do you think it would be possible to document only the methods present in the module in this part?

pyat/at/lattice/lattice_object.py
```diff
@@ -312,21 +301,125 @@ def _addition_filter(self, elems: Iterable[Element]):
...
+        Returns:
+            newring(Lattice): concatenated Lattice
+        """
+        if inplace:
+            lattice = self
+        else:
+            lattice = Lattice(self)
+        for lat in numpy.atleast_2d(lattices):
+            lattice.extend(lat, copy=copy)
+        return lattice
```
lfarv, line 422: For safety, it should return lattice only if `inplace=False`:
```
return None if inplace else lattice
```

pyat/at/lattice/lattice_object.py
```diff
@@ -312,21 +301,125 @@ def _addition_filter(self, elems: Iterable[Element]):
...
-        self.extend([elem])
+    def append(self, elem: Element, copy=False):
+        r"""This method overwrites the inherited method
+            :py:meth:`list.append()`,
+            it behavior is changed, it accepts only AT lattice elements
+            :py:obj:`Element` as input argument.
+
+            Equivalents syntaxes:
+            >>>ring.append(elem)
+            >>>ring += [elem]
```
lfarv, line 350: Missing space: `>>> ring.append(elem)`. (in many places)

lfarv:
>  I have an issue with sphinx

I don't think it's possible to filter the methods, apart from writing everything manually. Sphinx imports the Lattice object and documents everything it finds. 
On the other hand, we also need list of all available methods somewhere…

pyat/at/lattice/lattice_object.py
```diff
@@ -312,21 +301,125 @@ def _addition_filter(self, elems: Iterable[Element]):
...
+        if n != 0 and periodicity > 1:
+            nbp = periodicity / n
+            periodicity = int(round(nbp))
+            if abs(periodicity - nbp) > _TWO_PI_ERROR:
+                warn(AtWarning('Non-integer number of cells: {}/{}. Periodi'
+                               'city set to 1'.format(self.periodicity, n)))
+                periodicity = 1
+        elems = itertools.repeat(self, max(n-1, 0))
+        lattice = self.concatenate(*elems, copy=copy)
+        return Lattice(lattice, periodicity=periodicity)
```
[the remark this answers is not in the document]
lfarv: Remove `harmonic_number=self.harmonic_number`, sorry, it's specific to `develop`

swhite2401:
> > I have an issue with sphinx
> 
> I don't think it's possible to filter the methods, apart from writing everything manually. Sphinx imports the Lattice object and documents everything it finds. On the other hand, we also need list of all available methods somewhere…

Ok no surprise then. I was asking just in case, I thought maybe there could be some kind of filter to not should the docstring and instead but a link to the source. Not a big deal

pyat/at/lattice/lattice_object.py
swhite2401: I have to return `lattice` otherwise the operator `+=` does not work

pyat/at/lattice/lattice_object.py
swhite2401: done

pyat/at/lattice/lattice_object.py
swhite2401: I take it back I have modified `__iadd__` and it is fine (I also mixed up `inplace` and `copy` that have inverse logic...)

swhite2401:
@simoneliuzzo you may be interested in this. You can now repeat a cell using `cell * n` or `cell.repeat(n)` and elements are automatically copied.
Other functions to perform standard operations (append, concatenate,...) are provided that have a copy option too.

simoneliuzzo:
This looks nice! Do you have also the "-" operator to inverse the order of the elements?

swhite2401:
> This looks nice! Do you have also the "-" operator to inverse the order of the elements?

Ok, after discussion on the meaning of the `-` operator, this can be added using `ring.repeat(n)` where `n` is a negative number. In this case the mirror of the lattice is repeated, taking care of swapping exit and entrance faces of dipoles.
Similarly the `-` operator can be introduced with:
`ring1 - ring2 = ring1 + ring2[::1]`
of course properly swapping faces as well.

What do you think? @lfarv ? Any better ideas?

lfarv:
@simoneliuzzo and @swhite2401 : The "-" operator is not defined for lists, and it would be confusing to define it as "reverse": it is not the reciprocal of addition, `ring - ring` will not give `[]`. I don't know what could be the reciprocal of concatenation! So for reversing a list, the standard ways are:

- `newring = ring[::-1]` returns a new object (standard indexing syntax)
- `ring.reverse()` reverse the order in-place, (standard list method)
- `reversed(ring)`  returns an iterator over the elements is reverse order (built-in python function)

All work for `Lattice` objects, but don't swap enter and exit faces.
- The `ring[::-1]` form should be modified, it would break the meaning of indexing.
- The `reverse() `method can be overloaded and swap the faces. It could also introduce a copy keyword, but the default should be False, to be compatible with the normal behaviour.
- `reversed(ring)` can be customised by writing a `__reversed__()` method swapping the faces, but is it useful?

Concerning the use of `repeat`, I don't like it too much: python explicitly states that `ring * n` if n is less that 0 returns an empty ring. Can repeat(n) do differently ?

I think we must stick to python standards, so my suggestion is:
- implement the reverse() method reversing in-place with faces swapped,
- optionally implement `__reversed__`
- set the python behaviour to '*' and `repeat`: if n is less than 0, take it as zero.

swhite2401:
@lfarv, I have no strong feeling about this, I think the idea was just to reproduce some of the madx features for building sequences but they are not essential.
On the other hand I do not see why we cannot deviate from the `List` standards. If some features/behaviors are useful for AT they should be adopted regardless of what is done in `List`. In fact this is already the case with the `*` operator.

For this PR I will then adpat the `.reverse()` and `reversed()` methods to switch dipole faces, this is useful and I do not see why they should behave differently. I will also modify `repeat()` according to your comments.

pyat/at/lattice/lattice_object.py
```diff
@@ -312,21 +302,136 @@ def _addition_filter(self, elems: Iterable[Element]):
...
+        if n != 0 and periodicity > 1:
+            nbp = periodicity / n
+            periodicity = int(round(nbp))
+            if abs(periodicity - nbp) > _TWO_PI_ERROR:
+                warn(AtWarning('Non-integer number of cells: {}/{}. Periodi'
+                               'city set to 1'.format(self.periodicity, n)))
+                periodicity = 1
+        elems = (copy_fun(el, copy_elements) for _ in range(n) for el in self)
+        return Lattice(elem_generator, elems, iterator=self.attrs_filter,
+                       periodicity=periodicity)
```
lfarv, line 403: One could take care of the harmonic number:
```python
        try:
            cell_h = self._cell_harmnumber
        except AttributeError:
            hdict = {}
        else:
            hdict = dict(_cell_harmnumber = n * cell_h)
        elems = (copy_fun(el, copy_elements) for _ in range(n) for el in self)
        return Lattice(elem_generator, elems, iterator=self.attrs_filter,
                       periodicity=periodicity, **hdict)
```
Note: the `try/except` is necessary because  `_cell_harmnumber` may be missing.

This make the harmonic number be multiplied by the same amount as the circumference, ensuring a constant RF frequency.

pyat/at/lattice/lattice_object.py
```diff
@@ -312,21 +302,136 @@ def _addition_filter(self, elems: Iterable[Element]):
...
+
+    def concatenate(self, *lattices: Tuple[Iterable[Element], ...],
+                    copy_elements=False, copy=True):
```
lfarv, line 406: In all the examples where the `copy` keyword is used (`enable_6d`, `set_cavity`, `set_value_refpts`, `add_beamloading`), its default if `False`. It's not absolutely necessary to have the same default everywhere, but for me it would be easier to remember… 

Anyway, whatever the default value is, I'd push to explicitly specify `copy` in all internal calls to concatenate (within `lattice_object.py`). It's clearer!

pyat/at/lattice/lattice_object.py
```diff
@@ -312,21 +302,136 @@ def _addition_filter(self, elems: Iterable[Element]):
...
+        """Concatenate several `Iterable[Element]` with the lattice
+
+        Equivalents syntaxes:
+        >>> newring = ring.concatenate(r1, r2, r3)
+        >>> newring = ring + r1 + r2 + r3
```
lfarv, line 411: May be one could give examples of in-place concatenation:
```python
>>> ring.concatenate(r1, copy=False)
>>> ring += r1
```

lfarv:
@swhite2401 and @simoneliuzzo : I think there could be a way of modifying the `Lattice` slice indexing (`ring[start:stop:step]`) to handle dipole face swapping:

One could decide that in the **only case of slice indexing**, a negative `step` value would trigger the face swap. Integer indexing (`ring[i]`) and array indexing (`ring[[a, b, c]]`) are left unchanged.

This still looks reasonable for "getitem", though a question appears for the "setitem": what should a line like ` ring[0:4:-1] = [drift3, bend2, drift1]` do? Swap `bend2` faces or not?

swhite2401:
@lfarv why not yes. For me `__setitem__` should never swap, the user is responsible to properly set the RHS.
Let me first finish with the `reverse`  and `reversed` devs

swhite2401:
> Answer: **it must swap**, so that

Then we should give up on this it is too complicated I don't like that AT can change what I am trying to set, it should take my input as is

swhite2401:
Hello @lfarv @simoneliuzzo , I started looking into this and reversing a lattice is more complicated that it seems if you want to take into account the R1, R2 and T1, T2 attributes and not just face angles and fringe fields.
We may decide to ignore them but then the backtracking would not be valid in the presence of alignment errors.... any suggestion? Anything I did not think about?

lfarv:
@swhite2401 : In any case, the "reverting" operation must be handled by the `Element` class. The `Lattice` method should just call `elem.revert()` or similar. 

For T1, T2, R1, R2, I thought about it, and it's probably better to keep them as they are: a magnet shifted "inside" in the direct lattice should better be shifted inside also in the reverted lattice. Same for rotation: I would keep the rotation angle identical. So R1, R2, T1, T2 stay unchanged. This makes sense for "deliberately" displaced magnets. For random errors, anyway the displacements have to be regenerated…

Maybe this intervention on the `Element` class justified another PR? As you like!

lfarv:
T1, T2, R1, R2 continuation: one main interest of reversing a lattice is to build symmetrical cells:
`cell = line + line.reversed()`. 
For this purpose, one must keep T1, T2, R1, R2 unchanged (not swapped) at the element level. If a whole section is displaced, then it's more complicated…

lfarv:
> I was indeed adding a method to Element just swapping faces and finge field before I started thinking about this. If we stick to this it is simple enough to be kept in this PR. We will for now ignore misalignment, I will document accordingly.

Perfect.

Once you have done `reverse()` and `__reversed__()` I can take care of indexing in another PR`

swhite2401:
> I can take care of indexing in another PR`

As I said, I am against this proposal, `__setitem__` should not modify the users input in any condition (doing a swap or else)

lfarv:
@swhite2401 : OK, so we keep the slice indexing as it is: `ring[::-1]` returns the elements in reverse order, without any swapping. No problem for me !

lfarv:
**T1, T2 again**: what I wrote yesterday was wrong. It's more complicated ! When the electron turns back (which is the same as reverting the line), x changes sign (left become right), but y stays the same (top remains top). So only the signs of x and p<sub>x</sub> must be changed.

So the procedure is: take T2, change the sign of x and p<sub>x</sub>, put it in T1, and take T1 (the initial one), change the sign of x and p<sub>x</sub>, put it in T2…

For a simple z rotation (z changes sign), put -R2 in R1 and -R1 in R2.

For the combination of all rotations, it's a nightmare…

simoneliuzzo:
@lfarv and @swhite2401 

I think that the T1, T2, R1 and R2 may be ignored. The '-' feature is mostly used to create a lattice from scratch. Errors will be added after on the lattice already rotated. I doubt that other codes that have the "-" operator (MAD, SAD, elegant, OPA, ...) did deal with errors as well. 
If we want to protect the user , we may give an error "lattice reverse not allowed with errors" if T1, T2, R1 and R2 are set.

swhite2401:
Hello @lfarv @simoneliuzzo , I am facong another issue: when an element has been divide in n sub-elements the entrance face is applied to the first sub-element and the exit face to the last sub-element.
Reversing the the individual elements does not work in this case.... any ideas on how to handle this?

lfarv:
I do not see why it does not work element by element: taking elements in reverse order, you find the last part of a sliced magnet. Its exit face becomes the entry face of the new "inverted" magnet, its missing entry face becomes a missing exit face, and so on up to the 1st slice: missing exit face becomes missing entry face, and the original entry face becomes the final exit face. Do I miss something?

swhite2401:
That is correct, no problem then.

swhite2401:
Added `reverse()`, `reversed()` and some basic tests, I think is ready for merge

pyat/at/lattice/lattice_object.py
```diff
@@ -314,21 +306,174 @@ def _addition_filter(self, elems: Iterable[Element]):
...
+        Returns:
+            lattice(Lattice): concatenated Lattice, if `copy==True` the
+                              new lattice object is returned
+                              otherwise None
+        """
+        if copy:
+            lattice = Lattice(self)
+        else:
+            lattice = self
+        for lat in numpy.atleast_2d(lattices):
```
lfarv, line 446: I don't understand why you need to create a numpy array:
```python
for lat in lattices:
    lattice.extend(lat, copy_elements=copy_elements)
```
Works, at least in my tests

pyat/at/lattice/lattice_object.py
```diff
@@ -314,21 +306,174 @@ def _addition_filter(self, elems: Iterable[Element]):
...
+        except AttributeError:
+            hdict = {}
+        else:
+            hdict = dict(_cell_harmnumber=n*cell_h)
+        elems = (copy_fun(el, copy_elements) for _ in range(n) for el in self)
+        return Lattice(elem_generator, elems, iterator=self.attrs_filter,
+                       periodicity=periodicity, **hdict)
+
+    def concatenate(self, *lattices: Tuple[Iterable[Element], ...],
+                    copy_elements=False, copy=False):
```
lfarv, line 416: Apparently, is should be `*lattices: Iterable[Element],`

pyat/at/lattice/lattice_object.py
swhite2401: I think that for the case `ring.concatenate(ring2)` if `ring2` has only one dimension you will have as many calls to extend as the number of elements in `ring2`, while with the present implementation you have only one call. Isn't this correct?

pyat/at/lattice/lattice_object.py
swhite2401: `*lattices` is a variable length argument (similar to `*args` in generic function notation), it is therefore a `tuple` no? 
Why do you say apparently? You have syntax checks complaining?

pyat/at/lattice/lattice_object.py
lfarv: No, the `lattices` argument is then `(ring2,)`, a tuple with 1 element, and it works normally.

pyat/at/lattice/lattice_object.py
lfarv: "Apparently", because with `Tuple` I get a warning which disappears when I remove it. It seems that for variable argument lists, what matters is the single argument type itself.

By the way, the `Tuple` type is deprecated since python 3.9, better using the standard `tuple` type instead. The `from __future__ import annotations` line on top make this work also in 3.7 and 3.8.

swhite2401: Ok thanks, done

pyat/at/lattice/lattice_object.py
swhite2401: Done

pyat/at/lattice/lattice_object.py
```diff
@@ -314,21 +306,175 @@ def _addition_filter(self, elems: Iterable[Element]):
...
+                              new lattice object is returned
+                              otherwise None
+        """
+        if copy:
+            lattice = Lattice(self)
+        else:
+            lattice = self
+        lattice[:] = lattice[::-1]
+        for e in lattice:
+            e.swap_faces()
```
lfarv, line 476: There is a problem: even if `copy` is True, the faces of the original lattice are swapped !

You must make a copy of the element before swapping the faces.

pyat/at/lattice/lattice_object.py
```diff
@@ -263,29 +262,19 @@ def __str__(self):
...
-                periodicity = 1
-        # noinspection PyTypeChecker
-        return Lattice(elem_generator,
-                       itertools.chain(*itertools.repeat(self, n)),
-                       iterator=self.attrs_filter, periodicity=periodicity)
+        return self.repeat(n)
 
-    def _addition_filter(self, elems: Iterable[Element]):
+    def __reversed__(self):
+        return self.reverse(copy=True)
```
lfarv, line 275: This returns  a `Lattice` object, while it should return a "reverse iterator": iterator over the elements in reversed order:
```python
for elem in self[::-1]:
    yield elem.copy.swapfaces()
```

Or should the copy be done in the `swapfaces()` method?

pyat/at/lattice/lattice_object.py
lfarv: Or maybe the copy should be done if the swap_faces() method ? And only if necessary (if there is anything to swap?)

pyat/at/lattice/lattice_object.py
```diff
@@ -314,21 +306,176 @@ def _addition_filter(self, elems: Iterable[Element]):
...
+
+        Returns:
+            lattice(Lattice): reversed Lattice, if `copy==True` the
+                              new lattice object is returned
+                              otherwise None
+        """
+        if copy:
+            elems = (el.swap_faces(copy=copy) for el in self[::-1])
+            return Lattice(elem_generator, elems, iterator=self.attrs_filter,
+                           periodicity=self.periodicity)
```
lfarv, line 473: `__reversed__` still needs to be modified to return an iterator instead of a Lattice. It could be:
```python
def __reversed__(self):
    return (el.swap_faces(copy=True) for el in self[::-1])
```
Which could make this simpler:
```python
    if copy:
        return Lattice(elem_generator, reversed(self),
                       iterator=self.attrs_filter)
    else:
        reversed_list = list(reversed(self))
        self[:] = reversed_list
```
Specifying `periodicity` in the constructor is not necessary, the periodicity is transferred by `attrs_filter`

lfarv:
Still waiting for the fix of `__reversed__`: it must [return an iterator](https://docs.python.org/3.7/library/functions.html#reversed)

swhite2401:
> Still waiting for the fix of `__reversed__`: it must [return an iterator](https://docs.python.org/3.7/library/functions.html#reversed)

@lfarv I replied to this point several times! And I would like to discuss it. I know this is what the `list` does why should this absolutely return and iterator for the `Lattice` object? Could you please give your reasons

swhite2401:
> > Still waiting for the fix of `__reversed__`: it must [return an iterator](https://docs.python.org/3.7/library/functions.html#reversed)
> 
> @lfarv I replied to this point several times! And I would like to discuss it. I know this is what the `list` does why should this absolutely return and iterator for the `Lattice` object? Could you please give your reasons

I just realized all my comments were not submitted... this is why you did not see them...

lfarv:
@swhite2401 : You have no choice: the definition of the [`__reversed__`](https://docs.python.org/3.7/reference/datamodel.html?highlight=__reversed__#object.__reversed__) special method specifies that "**It should return a new iterator object that iterates over all the objects in the container in reverse order**". The built-in function `reverse()` is the symmetrical of `iter()`: both provide a way to get in iterator object (forward or reverse) from an iterable (a list, Lattice or any Sequence). Look at those two special methods in the [Python Language Reference](https://docs.python.org/3.7/reference/datamodel.html#special-method-names)

If you want a function returning a Lattice object, you need another one (but it's already there !)

swhite2401:
> @swhite2401 : You have no choice: the definition of the [`__reversed__`](https://docs.python.org/3.7/reference/datamodel.html?highlight=__reversed__#object.__reversed__) special method specifies that "**It should return a new iterator object that iterates over all the objects in the container in reverse order**". The built-in function `reversed()` is the symmetrical of `iter()`: both provide a way to get an [`iterator`](https://docs.python.org/3.7/glossary.html#term-iterator) object (forward or reverse) from an [`iterable`](https://docs.python.org/3.7/glossary.html#term-iterable) (a list, Lattice or any Sequence). Look at those two special methods in the [Python Language Reference](https://docs.python.org/3.7/reference/datamodel.html#special-method-names)
> 
> If you want a function returning a Lattice object, you need another one (but it's already there !)

It is done, ok to merge?

lfarv (approved):
OK !

PR #576, opened 2023-03-08
Improve the import order

lfarv:
The main point is to have the `frequency_control` decorator available for use in the lattice package.

- `frequency_control` has been moved to the `lattice/cavity_access.py` module, without any code change. Now it can be used everywhere in the `lattice` package. It's also still available in the `physics` package for compatibility,
- `get_revolution_frequency` is back in the `revolution.py` module, without any code change,
- all other changes are displacements of import statements to linearise the import order. So no real code change !


swhite2401:
Ok for me

Issue #575, opened 2023-03-07
pyat get_lifetime modifies the given refpts


oscarxblanco:
Dear all,

the function get_lifetime has a parameter to pass the reference points, refpts, where the momentum aperture momap was calculated.
However, in line 146 of pyat/at/acceptance/touschek.py this array is modified leading to a mismatch between momap and refpts.

https://github.com/atcollab/at/blob/fee2a6a462ad263aa722b2c8b0ffa3dafdce8332/pyat/at/acceptance/touschek.py#L142-L147

I think that refpts should not be modified when given by the user. One could think to use a marker as a reference point which should not be removed from the list even if it has zero length.

Best regards,
o

oscarxblanco:
One possible solution is to execute those lines only when refpts is not passed, i.e. move them inside the if refpts is None case.

oscarxblanco:
One other posibility is that due to the implementation of get_lifetime, one needs to pass only no-zero length elements. 

If that is the case the function should check the elements length and send a warning/error if any is of length zero. This could be done by comparing the length of the filtered list to the original one, or by checking that the list of elements with zero length is void.

swhite2401:
By the way in case you are interested strictly in momentum aperture you can run get_momentum_acceptance. This will not interfere with your refpts input.
Then you can give the result of this calculation together with refpts to get_lifetime.

Removing the refpts in get_lifetime that will not be used (zero length elements) is also ment to optimize the computer time.

oscarxblanco:
> Hello @oscarxblanco, Thanks for the feedback. In fact this is not fully satisfactory should be improved. The zero-length elements are indeed causing problems. This is why the internal refpts is returned.
> 
> What I can propose in the short term based on your remarks is: -if refpts is None no change -if refpts is provided check that all corresponding elements have non-zero length and throw a warning if this condition is not respected
> 
> Then we may think of trying to improve the algorithm to account for zero length elements (length could be in this case the distance to the next refpts but this could be dangerous if this distance is very large...)
> 
> Let me know in case you have better proposals. This was basically translated from the matlab and there is certainly room for improvement.

Thank you, it sounds alright, except for the warning. In my opinion it should throw an error as the algorithm does not yet support zero-length elements.



> By the way in case you are interested strictly in momentum aperture you can run get_momentum_acceptance. This will not interfere with your refpts input. Then you can give the result of this calculation together with refpts to get_lifetime.
> 
[...]

PR #577, opened 2023-03-09
zero length elements in lifetime calculation

swhite2401:
This PR is a bugfix in response to issue #575. The following corrective action are impleted:
-if zero-length elements are provided, they are automatically remove with a warning
-if the user provides the momentum aperture the zero-lenght elements are also remove from the momentum aperture such that refpts and momap always remain consistent
-the momentum aperture and refpts used for the lifetime calculation (with elements removed) are returned by the function

### DIFF
--- a/pyat/at/acceptance/touschek.py
+++ b/pyat/at/acceptance/touschek.py
@@ -150,8 +150,9 @@ def get_lifetime(ring, emity, bunch_curr, emitx=None, sigs=None, sigp=None,
 
     mask = [ring[r].Length > 0.0 for r in refpts]
     if not numpy.all(mask):
-        warnings.warn(AtWarning('zero-length elements removed '
-                                'from lifetime calculation'))
+        zerolength_warning = ('zero-length elements removed '
+                              'from lifetime calculation')
+        warnings.warn(AtWarning(zerolength_warning))
     refpts = refpts[mask]
 
     if momap is None:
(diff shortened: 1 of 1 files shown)


oscarxblanco:
@swhite2401 the mask you implemented is working.
The warning output, however, seems to be shredded 
```bash
touschek.py:153: AtWarning: zero-length elements removed from lifetime calculation
  warnings.warn(AtWarning('zero-length elements removed '
```

swhite2401:
Yes I could never find a way out of this problem: for multi-line string (required by PEP8) the message is first printed out and then the line at which it is issued is shown, this for multi-line messages results in printing only the first line....

If you have a solution for this I'dd be very happy to hear it! Or maybe @lfarv ?

oscarxblanco (changes requested):
I did a test with a momentum aperture calculated at indexes of elements with zero-length.
The mismatch between indexes and momap is not longer present.

The warning message seems to be descriptive enough, but, the standard output is shredded.
I have put below a suggestion on how to correct this issue.

swhite2401:
@oscarxblanco I have implemented the requested changes

oscarxblanco (approved):

oscarxblanco:
@swhite2401 
Unfortunately this modification introduced a bug when calculating the lifetime without the momap input.
```
 ... acceptance/touschek.py", line 156, in get_lifetime
    refpts = refpts[mask]
TypeError: range indices must be integers or slices, not list
```

It could be fixed by making refpts a numpy array.
```
refpts = numpy.array(refpts[mask])
```

pyat/at/acceptance/touschek.py
```diff
@@ -143,8 +144,16 @@ def get_lifetime(ring, emity, bunch_curr, emitx=None, sigs=None, sigp=None,
         refpts = range(len(ring))
```
oscarxblanco, line 144:
```python
refpts = numpy.array(range(len(ring)))
```

oscarxblanco (commented):
One modification required to be compatible with no momap input file option.

PR #579, opened 2023-03-13
Fix bug in memory allocation of wake method

lcarver:
double memory allocation of wake (2 numbers) led to runaway memory. Surprised it wasn't spotted before. Bug is now fixed by using pointer instead of extra allocation

### DIFF
--- a/atintegrators/atimplib.c
+++ b/atintegrators/atimplib.c
@@ -260,10 +260,9 @@ void compute_kicks(int nslice,int nturns,int nelem,
 };
 
 
-double *wakefunc_long_resonator(double ds, double freqres, double qfactor, double rshunt, double beta) {
+double *wakefunc_long_resonator(double ds, double freqres, double qfactor, double rshunt, double beta, double *wake) {
 
     double omega, alpha, omegabar;
-    double *wake = malloc(2*sizeof(double));
     wake[0] = 0.0;
     wake[1] = 0.0;
     double dt;
@@ -296,7 +295,6 @@ double *wakefunc_long_resonator(double ds, double freqres, double qfactor, doubl
         wake[0] = 0.0;
         wake[1] = 0.0;
     }            
-    return wake;
 }
 
 
@@ -310,7 +308,7 @@ void compute_kicks_longres(int nslice,int nbunch,int nturns, double *turnhistory
     double ds,wi;
     double *turnhistoryZ = turnhistory+nslice*nbunch*nturns*2;
     double *turnhistoryW = turnhistory+nslice*nbunch*nturns*3;
-    double *wake = malloc(2*sizeof(double));
+    double *wake = atMalloc(2*sizeof(double));
     double vba, vbp;
     double *vbr = vbunch;
     double *vbi = vbunch+nbunch;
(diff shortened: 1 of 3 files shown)


swhite2401:
Ok for me. There are most likely cleaner solutions but at least this one fixes the bug

swhite2401 (approved):

atintegrators/atimplib.c
```diff
@@ -339,7 +337,7 @@ void compute_kicks_longres(int nslice,int nbunch,int nturns, double *turnhistory
                 ds = turnhistoryZ[i]-turnhistoryZ[ii];
                 if(turnhistoryW[ii]>0.0 && ds>=0){
                     wi = turnhistoryW[ii];
-                    wake = wakefunc_long_resonator(ds,freq,qfactor,rshunt,beta);       
+                    wakefunc_long_resonator(ds,freq,qfactor,rshunt,beta,wake);       
                     kz[i-nslice*nbunch*(nturns-1)] += normfact*wi*wake[0];
                     vbeamk[0] += normfact*wi*wake[0]*energy/nbunch;
                     vbeamk[1] -= normfact*wi*wake[1]*energy/nbunch;
```
swhite2401, line 343: Here you probably have to divide by `nslice`

lcarver:
So in the latest version of this PR.
Memory leak problem is solved.
More accurate method of computing the Vbeam is now put in place. Now it computes (correctly!) the weighted mean of the voltage kick experienced by the beam. Previously I was seeing a few strange things, such as noisy Vbeam value turn by turn, which didn't converge with increasing nslice and nparts. This should now be resolved.

Before I merge I would like to redo the benchmarking for the usual things (Qs, transient and checking Vbeam with analytical values)

lcarver:
Full test of all different cases, comparing the Vbeam for a range of currents (with and without zcuts, for phasor and for wake, for Nslice=1 or Nslice>1) have all passed. Agreement is excellent.

image

image

lcarver:
For me this is now ready to merge, I am running the benchmark for Nslice=50 for Qs vs current but to be honest, the Vbeam agrees perfectly so there is no reason it should not pass. 

I leave it a few days anyway before I merge in case someone wants to comment. I will probably merge on friday morning (after Nslice=50 test has passed)

pyat/at/lattice/lattice_object.py
```diff
@@ -629,12 +629,10 @@ def bunch_list(self) -> numpy.ndarray:
...
-    def beam_current(self):
+    def get_beam_current(self):
         return self._beam_current
 
-    @beam_current.setter
-    def beam_current(self, value, clear_history=True):
+    def set_beam_current(self, value, clear_history=True):
         self._beam_current = value
         if clear_history:
             self.set_wake_turnhistory()
```
lfarv, line 638: Better defining the property here:
```python
def get_beam_current(self):
         return self._beam_current

def set_beam_current(self, value, clear_history=True):
    self._beam_current = value
    if clear_history:
        self.set_wake_turnhistory()

beam_current=property(get_beam_current, set_beam_current)
```
rather the defining it out of the class.

atintegrators/atimplib.c
```diff
@@ -306,15 +304,16 @@ void compute_kicks_longres(int nslice,int nbunch,int nturns, double *turnhistory
...
     int rank=0;
     int size=1;
-    int i,ii,ib;
-    double ds,wi;
+    int i,ii,ib,loopstart,loopend;
+    double ds,wi,wii;
     double *turnhistoryZ = turnhistory+nslice*nbunch*nturns*2;
     double *turnhistoryW = turnhistory+nslice*nbunch*nturns*3;
-    double *wake = malloc(2*sizeof(double));
+    double *wake = atMalloc(2*sizeof(double));
```
lfarv, line 311: Why are you allocating dynamically an array of fixed length ?
```C
double wake[2];
```
is more efficient and does not require freeing.

lfarv:
Hi @lcarver. Could you declare all functions in `atimplib.c` as `"static"`. It avoids cluttering the namespace (and prevents name collisions).

atintegrators/atimplib.c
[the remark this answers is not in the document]
lcarver: ok

lcarver:
image

Above plot for slightly higher Rs (13 cavs instead of 11 - explains different shift rate and threshold) and done for Nslice per bunch = 50.

All good for me.

swhite2401 (approved):

PR #580, opened 2023-03-13
BugFix: refpts=None in get_lifetime

swhite2401:
PR #577 introduced a bug when `refpts=None`, this PR fixes this issue

### DIFF
--- a/pyat/at/acceptance/touschek.py
+++ b/pyat/at/acceptance/touschek.py
@@ -150,8 +150,9 @@ def get_lifetime(ring, emity, bunch_curr, emitx=None, sigs=None, sigp=None,
 
     mask = [ring[r].Length > 0.0 for r in refpts]
     if not numpy.all(mask):
-        warnings.warn(AtWarning('zero-length elements removed '
-                                'from lifetime calculation'))
+        zerolength_warning = ('zero-length elements removed '
+                              'from lifetime calculation')
+        warnings.warn(AtWarning(zerolength_warning))
     refpts = refpts[mask]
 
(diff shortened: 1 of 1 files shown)


swhite2401:
@oscarxblanco sorry I overlooked the case where `refpts=None` in the previous PR, this one should fix the issue.

oscarxblanco (approved):
I have checked the call to get_uint32_index(at.All, endpoint=False) and it seems to work as expected for a given ring with 4043 elements
```
>>> len(ring)
4043
>>> ring.get_uint32_index(at.All, endpoint=False)
array([   0,    1,    2, ..., 4040, 4041, 4042], dtype=uint32)
>>> ring.get_uint32_index(at.All)
array([   0,    1,    2, ..., 4041, 4042, 4043], dtype=uint32)
```

When executing a test example, both options, with or w/o refpts work emit the expected warning message refering to zero-length elements in the index list. This is normal behavior.

I was not able to finish the whole momentum aperture calculation due to some segmentation fault. But this seems to be not related to the indexing problem.

swhite2401:
Ok so I propose to merge this as it fixes an obvious bug, on the other hand it would be good to understand this segmentation fault, I you have more details could you report it as a separate issue (it is most likely originating from the c part of the code).

I usually never run LT calculation for the full lattice but rather on a subset of elements (typically magnets) so it is well possible that for very heavy computations some problems have passed through the various validations.

oscarxblanco:
@swhite2401 ok, I will report the segmentation fault in another issue.

From my side, this fix could be merged.

PR #581, opened 2023-03-14
New integrators: ExactDriftPass, ExactMultipolePass

lfarv:
New integrators for drifts and straight multipoles.


lfarv:
Removed a sanity check on loading Matlab files blocking `ExactDriftPass`

lfarv:
Added `ExactRectangularBendPass` and `ExactRectangularBendRadPass`.

Note that `ExactRectangularBendPass` now accepts pole face angles. So it can be used to replace any dipole without higher order terms. Using it for focusing magnets is delicate since the reference trajectory is no more an arc of a circle. So it's likely to generate a non-zero closed orbit… An  `ExactSectorBendPass` is in preparation and will avoid those problems.

simoneliuzzo:
Dear @lfarv and @swhite2401 ,
I start again today to test this branch vs MADX-PTC. I will keep you posted with the results.

simoneliuzzo:
Dear @lfarv  and @swhite2401,

below my actual status of comparison AT vs MADX-PTC

I take the same identical initial coordinates (20 points up to the same number of sigmas in x, xp, y and yp, a 4D diagonal).
I verify that tracking in AT and MADX-PTC are consistent (radiation, exact hamiltonian, fringe fields, number of integration steps (80), etc..)
Then I take an AT lattice, convert it to MADX and build figures to compare.

For the ESRF-DBA lattice, there is a well comparing case, but works only in 4D, not in 6D.
Screenshot 2023-07-11 at 12 50 35

For the EBS lattice, I see no effect of exact pass (ok) but the AT and MADX-PTC tracking are visibly different.
Screenshot 2023-07-11 at 12 51 12

For the FCC lattice, there is a well comparing case (the same of ESRF-DBA, 4D). In 6D only the particle at 0.0 initial coordinates survives.
Screenshot 2023-07-11 at 13 10 19


For all cases I could not yet convert properly the longitudinal phase space from MADX-PTC to AT convention. I will spend some time on this issue now, so I should be able to show you also the missing phase space plot in the next message.

Conclusion for now is: 
* with at**disable**_6d using all Exact pass methods makes AT and PTC tracking close to identical.  
* with atenable_6d there are still issues

simoneliuzzo:
Dear @lfarv and @swhite2401,

I have noticed that in my simulations for EBS (only) I never actually used ExactRectangularBendPass. 

With this PassMethod in the DQ the optics become largely mis-matched. 
May be this pass method is not ready to accept Quadrupole components? 

Also the ExactRectangularBendRadPass seems to have some issue. As soon as I use it the tracking fails (only the zero amplitude particles survive).

May be I am not supposed to test with it. For the ESRF-DBA and FCC lattices, using it the 4D tracking agrees with MADX-PTC.

lfarv:
@simoneliuzzo
You should not use the RectangularBendPass for DQs for 2 reasons:
1. DQs are not straight magnets: their magnetic axis is an arc of a circle. In rectangular bends the magnetic axis is a straight line (like in a displaced quadrupole). For pure dipoles it does not matter, the magnetic field is uniform. If there are other multipoles, it does matter,
2. When discussing rectangular bends on my last visit, I pointed out that for non-uniform field, we are still missing a tuning function: if there are multipoles, the trajectory in the magnet is no more an arc of a circle. For arbitrary multipoles, it cannot be predicted. We need a tuning function which will iteratively displace the magnet until the nominal deviation is achieved. I promised you this function, but I didn't do it yet. The path length has also to be tuned. Until this is available, and unless you want iterate manually, you are restricted to pure dipolar field: ESRF-DBA is OK.

Did the RadPass methods fails on ESRF-DBA ? That's an interesting.

I did not come back the sector bends, I've been busy with tapering, but I'll switch to that soon!

simoneliuzzo:
Ok. I stop testing for now. The tools are in place on my side. 

Yes, the ESRF-DBA lattice fails with ExactRectangular`Rad`Pass

4D
Screenshot 2023-07-13 at 08 39 44

6D
Screenshot 2023-07-13 at 08 39 39

In particular the AT lattice has very strange optics, tunes, chromaticity. I use the AT lattice to generate coordinates for tracking and they are NAN (NAN emittances, energy spread, etc). Consequently the converted MADX also fails tracking, but optics are perfectly fine.

I attach here the failing lattice.
[lattice.m.zip](https://github.com/atcollab/at/files/12036174/lattice.m.zip)

and the output of ringpara and atx
```

   ******** AT Ring Parameter Summary ********
   Energy:                      6.00000 [GeV]
   Circumference:               844.39069 [m]
   Revolution time:             2816.58417 [ns] (0.35504 [MHz]) 
   Betatron tune H:             NaN ( NaN [kHz])
                 V:             NaN ( NaN [kHz])
   Chromaticity H:              NaN
                V:              NaN
   Synchrotron Integral 1:      NaN [m]
                        2:      2.60359e-01 [m^-1]
                        3:      1.09718e-02 [m^-2]
                        4:      NaN [m^-1]
                        5:      NaN [m^-1]
   Damping Partition H:         NaN
                     V:         1.00000
                     E:         NaN
   Radiation Loss:              4746.84462 [keV]
   Natural Energy Spread:       NaN
   Natural Emittance:           NaN [nm]
   Radiation Damping H:         NaN [ms]
                     V:         7.12031 [ms]
                     E:         NaN [ms]
   Momentum Compaction Factor:  NaN
   Slip factor:                 NaN

   Assuming cavities Voltage:   9000.00000 [kV]
                   Frequency:   352.19966 [MHz]
             Harmonic Number:   992
   Synchronous Phase:           2.58602 [rad] ( 148.16822 [deg])
   Linear Energy Acceptance:    NaN %
   Synchrotron Tune:            NaN ( NaN kHz or  NaN turns) 
   Bunch Length:                NaN [mm],  NaN [ps]

   Emitty from Dy:              NaN [nm]
   Emitty 1/gamma cone limit:   NaN [pm]
>> [a,b] =atx(r)
Warning: Matrix is singular, close to singular or badly scaled. Results may be inaccurate. RCOND = NaN. 
> In xorbit_6 (line 47)
In findorbit/xfindorbit (line 44)
In frequency_control (line 28)
In findorbit (line 33)
In findm66/xfindm66 (line 72)
In frequency_control (line 28)
In findm66 (line 52)
In ohmienvelope (line 27)
In atx/go (line 195)
In atx (line 116)
 
Warning: Matrix is singular, close to singular or badly scaled. Results may be inaccurate. RCOND = NaN. 
> In ohmienvelope (line 55)
In atx/go (line 195)
In atx (line 116)
 
Warning: Emittance computation failed:
First and second inputs must not contain NaN or Inf.
 
> In atx/go (line 208)
In atx (line 116) 

a = 

  1×1637 struct array with fields:

    R
    M
    A
    mu
    SPos
    ClosedOrbit
    alpha
    beta
    Dispersion
    modemit
    emit44
    emit66


b = 

  struct with fields:

              ll: 8.443906927513550e+02
           alpha: 1.779465694145638e-04
       fractunes: [0.440020226187056 0.389996716488082]
       fulltunes: [36.440020226187059 13.389996716488083]
             nuh: 36.440020226187059
             nuv: 13.389996716488083
    chromaticity: [7.254689719330779 11.832685785565880]
     dampingtime: [NaN NaN NaN]
        dampingJ: [NaN NaN NaN]
         espread: NaN
         blength: NaN
    modemittance: [NaN NaN NaN]
          energy: 6.000000000000000e+09
              fs: NaN
           eloss: 0
    synchrophase: 0
      momcompact: 1.779465694145638e-04

thank you for your work
best regards
Simone

```

lfarv:
@simoneliuzzo: got your lattice, I'm looking at it.

simoneliuzzo:
Dear @lfarv 

thank you for the fast update. In fact the ExactRectangularBendRadPass is now giving stable lattices.

However, when using the *Rad version of ExactMultipolePass  and ExactRectangularBendPass the tracking in the px, y and py planes differs more from the MADX-PTC one (the x plane tracking instead is even more similar to the MADX one than without rad) compared to the 4D case.

4D (madx does not track ct and delta)
Screenshot 2023-07-17 at 13 15 44

6D
Screenshot 2023-07-17 at 13 15 32

thank you for your help
best regards
Simone

simoneliuzzo:
Dear @lfarv 
let me know when I should run my benchmark test for this branch.
thank you!
best regards
Simone

swhite2401:
> when using the *Rad version of ExactMultipolePass and ExactRectangularBendPass the tracking in the px, y and py planes differs

It seems the Rad passes are now incorrect, have we found the problem? It would be nice to finalize the PR, it seems we are very close and working on a dev branch is really not comfortable.

Can I help in some way to speed this up?

simoneliuzzo:
Dear @lfarv,

let me know when I have to run the test. 

I have tried ExactSectorBendPass and it is in agreement with MADX
Screenshot 2023-08-30 at 16 58 19

lfarv:
Hello @simoneliuzzo,

I'll send you next Monday a review of where I am on that (I'm away for the end of the week). I ran many tests and still found a number of issues, from minor to critical, the main one being ExactSectorBendRadPass (still not working).

Your test of ExactSectorBendPass looks very nice, though this passmethod has still some small problems for me: more next Monday…

lfarv:
@simoneliuzzo

Here is the status of "exact" passmethods:

- `ExactRectangularBendPass` is OK for me,
- `ExactRectangularBendRadPass` is OK for me. I do not know how PTC handles radiation. I handle it as in other places: at each kick, compute the momentum change and apply it keeping the angles. In other passmethods, AT applies 1st the momentum change and then the kick (bending or focusing). I do not know why in that order. I can try half-kick, momentum change, half-kick to see if it makes a difference,
- `ExactSectorBendPass` looks OK. But there is a slight numerical problem. If I compare the horizontal tune, I get this:
tuneh1
  Zooming around dp=0 gives this:
tuneh2
  There is a ripple of about 2.0e-5 in horizontal tune at around dp=0. This is not tremendous but it ruins the computation of chromaticity with the standard parameters. It's a numerical problem with Forest's formulae (like subtracting two very close floating point numbers). I'm investigating this,
- `ExactSectorBendRadPass` is completely wrong. The final momentum loss is correct, but the kick are crazy…

And finally, E. Forest gives an alternative formula for RectangularPass, extrapolated from the general SectorPass. I called it ExactRectBEndPass for the moment. It has a big advantage: like `ExactSectorBendPass`, since it's an exact analytic formula, if the is no multipole or radiation, it does not need any splitting of the magnet. By setting NumIntSteps=0, we gain a factor ten to twenty in speed ! But it suffers from the same problems as SectorPass, so forget it for the moment.

simoneliuzzo:
Dear @lfarv , 

thank you for the update. Let me know if there is something I can do to help apart running the benchmark with MADX-PTC.

regards
Simone

lfarv:
This version is rebased on the last master commit, so it includes all the recent fixes of the master branch.

simoneliuzzo:
Dear @lfarv and @swhite2401 ,

@lfarv, thank you for these updates! They sounds very promising.

I have started now the benchmark with madx-ptc. 

I'll send the plots as soon as I can.

regards
Simone

simoneliuzzo:
Dear @lfarv and @swhite2401,

This is the overall summary of the benchmark for the ESRF DBA lattice:
 
Screenshot 2023-09-12 at 15 59 35

ExactSectorBendPass works as well as ExactRectangularBendPass

The *Rad are still a bit different:

BndMPoleSymplectic:
Screenshot 2023-09-12 at 15 59 56

ExactRectangularBend:
Screenshot 2023-09-12 at 15 58 55

ExactSectorBend:
Screenshot 2023-09-12 at 16 03 25

simoneliuzzo:
Dear @lfarv,

I run the same code for EBS. Contrary to the DBA lattice when setting `ExactRectangularBendPass` the emittance computation fails (ringpara gives negative emittance, atx does not work for the rad-off case) and so my definition of initial coordinates for tracking is not real. 

The optics of the lattice are mis-matched by the change of integrators in the dipoles. 

Screenshot 2023-09-18 at 11 00 21

With ExactSectorBendPass atplot does not show the optics (eig input matrix must not containt inf or nan).

lfarv:
@simoneliuzzo: thanks for feedback

1.  you cannot use `ExactRectangularBend` for DQs, since they are not rectangular magnets. But you can for DLs. Is this what you tried?
2. For `ExactSectorBend`, it seems that the problem is a septum magnet with a zero bending angle. There must be a problem with zero bending angle in sector bends. I'll look at that, but in the meantime, can you test `ExactSectorBend` with excluding the septum, or any other magnet with zero angle? Thanks

lfarv:
@simoneliuzzo : the problem with zero bending angle is corrected, you can try again.

lfarv:
@simoneliuzzo, @swhite2401 

The "Exact" passmethods are now ready for review. The tuning function for the rectangular magnets is now available and documented.

- Matlab: function `atrbendtune(elem)`. Try `>> help atrbendtune`
  ```Matlab
  % Identify the rectangular bends (for example...)
  rbends=atgetcells(ring,'PassMethod', 'ExactRectangularBendPass');
  % Set their correct attributes
  ring(rbends)=cellfun(@attunerbend,ring(rbends),'UniformOutput',false);
  ```
- python: `Dipole.rbendtune()` method. Try `>>> help(at.Dipole.rbendtune)`.
  ```python
  # Identify the rectangular bends (for example...)
  rbends = ring.get_bool_index(checkattr("PassMethod", "ExactRectangularBendPass")
  # Set their correct attributes
  for dip in ring.select(rbends):
      dip.rbendtune()
  ```
The new pass methods are also documented on the Web pages, in a [preliminary documentation](https://at.readthedocs.io/en/exact_multipole_pass/common/passmethods.html) of all passmethods. Please have
a look and post you comments.

atintegrators/BendLinearPass.c
```diff
@@ -10,7 +10,7 @@
 #include "atphyslib.c"
 
 
-#define SQR(X) X*X
```
swhite2401, line 13: Just for my culture, could you explain this modification?

docs/common/ex_passmethods.md
```diff
@@ -0,0 +1,115 @@
...
+If the magnet field includes quadrupole or higher components, the reference trajectory 
+in the magnet is no more an arc of a circle. A tuning of the `X0ref` attribute is
+necessary to get the correct output angle from the magnet. This can be seen as a
+horizontal translation of the magnet until the correct deviation angle is obtained.
+Similarly, the path lengthening must be adjusted to take into account the
+length of the non-circular trajectory.
+
+This tuning is performed using a dedicated function/method:
+
+- python
```
swhite2401, line 90: These should be mentioned in the docstrings as well not? Or at least a reference to this shuld be added.

swhite2401:
Very nice addition! I trust you on the math providing @simoneliuzzo benchmarking is ok.
Adding a documentation section is also very good and this practice should be extended, however I think it would also be very usefull to add something in the docstrings or the element class to describe the available passmethods and basic functionalities, arguments etc...
This element is becoming complicated... do you want to create a class ExactDipole?

The method to tune the dipoles could be added as a member of the class.

atintegrators/BendLinearPass.c
lfarv: @swhite: to avoid repeating 10 times the definition of SQR, it has been included in `atelem.c`, but in its correct form:
```C
#define SQR(X) ((X)*(X))
```
Keeping a duplicate definition in existing modules is allowed, provided it's exactly the same. But this one was different, so I had to remove it. Other existing files are unchanged.

swhite2401: Right, but why are the parenthesis needed?

docs/common/ex_passmethods.md
lfarv: The new tuning method is documented if you look at the `Dipole` class.

lfarv:
> This element is becoming complicated... do you want to create a class ExactDipole?

I would prefer keeping a single dipole class, so that you can easily change the passmethod keeping the same element. The only additional complexity is the new tuning method.

> The method to tune the dipoles could be added as a member of the class.

It is the case, and it's the only way to access it. As explained in the docstring, it does nothing if the passmethod is not a rectangular pass, so you can run it safely for any dipole.

Concerning docstrings, you are right, but it's a more general question. At the moment, the attributes and methods are documented, and the default passmethod is mentioned (for all classes). A simple addition could be to mention all the **acceptable** passmethods, with a link to the global passmethod descriptions that I just initiated. Can this be set as another PR? The passmethod description is still preliminary, I just took care of the new "Exact" methods

lfarv:
There are still things that can be changed:
- the name of the tuning function. Any better idea?
- in python, I tried not to lengthen the `element.py` module (already 1192 lines), so I implemented the new method in a separate file. Is it the best way? Or should we move the entire `Dipole` class to a separate file?
- The new bend classes have an optional possibility for multipole fringe field. At the moment, I kept this private until we find the best way to integrate this new flag in the existing attributes (a new value to the `FringeQuadEntrance` integer flag?).
- Same for the dipole fringe field: a new `gK` variable looks identical to the product of the existing `FullGap*FringeInt1` attributes (usually at 0). It needs to be checked, I could not find until now what gK refers to…
- There is an alternate method for rectangular bends, presently called `ExactRectBendPass`. It gives the same result as `ExactRectangularBendPass` with an advantage (a single step for pure dipole field) and a drawback (small discontinuity around the origin as for the sector bend). It is also considered private for now, but I'd like you to test it.

swhite2401:
> * in python, I tried not to lengthen the `element.py` module (already 1192 lines), so I implemented the new method in a separate file. Is it the best way? Or should we move the entire `Dipole` class to a separate file?

In fact I was about to suggest some re-organization of this part, can we separate elements by type, magnets, diagnostics, etc... ? I think this would make things more readable. But this is probably for another PR

lfarv:
>  different passmethods taken different input arguments, can we add notes as well to describe the arguments/functions specific to each passmethods?

That's exactly what I did in the passmethod description I started: it shows how the method works and what attributes it uses. The mention of the available passmethod in the element class doctoring **must** link to that. I'll try right now for the dipole class, you'll tell me what you think of it.

simoneliuzzo:
Would it be possible to add atrbendtune in the rectangular pass element creator? 

here a few proposal for different names:
at_set_rectangual_exact_bend_parameters
at_set_rbend_X0_DZ_ref

Why are the two additional fields name X0ref and RefDZ ? Is it for consistency with notation in the references?

atintegrators/BendLinearPass.c
lfarv: without parentheses:

`3*SQR(a+b)` -> `3*a+b*a+b`: completely wrong

You always need to parenthesise everything when defining macros.

lfarv:
@swhite2401: can you now look again at the doc of the `Dipole` class: it now points to the full description of all the available passmethods (at least those that I have documented…). It can still be improved and generalised, but I think it's the way to go.

lfarv:
@simoneliuzzo 
I agree that the attribute names are not very nice, but it's their name in @catohxb's `BndStrMPoleSymplectic4Pass` where they have the same role (and the same values after tuning). I borrowed them from there and I'd like to keep the same attribute names everywhere.

It would be easy to call the tuning function within the element creation, but I fear that if it's hidden, users would forget to call it after modifying `PolynomB`. That's why I preferred to keep it explicit. If there is a common agreement to include it, why not. And what about python?

About names: yours are quite long, and Matlab style for function names is to use camelCase rather that underscores (opposite to python)… Any preference @swhite2401, @carmignani ?

simoneliuzzo:
Dear @lfarv

I have run the benchmark again. However something is strange. 

with AT from 12 September I get the improvement in benchmark as in this figure, while with the latest AT (pulled today 9 October) instead I do not get it anymore.

This is true already for ExactDrift and ExactMultipole integrators.
Screenshot 2023-10-09 at 12 36 44

and even more evident when including ExactRectangularBend
Screenshot 2023-10-09 at 12 35 32

Did anything change that could explain this change?

lfarv:
@simoneliuzzo : Strange indeed

- `ExactDrift` and `ExactMultipole` did not change at all
- `ExactSectorBend` has an additional check to avoid problems with a 0 bending angle (a problem you had with EBS), but no change in the computation,
- `ExactRectangularBend` has the addition of the 2 new tuning attributes, but no change in the computation,

For me, the results did not change since long. Are you sure you are on the right branch and rebuilt everything (atmexall?). By the way, are you running Matlab or python ?

lfarv:
@simoneliuzzo : If you confirm today's problem, I'd like you to repeat you September test. You can go back to the code of September 12th by typing:
```
git checkout 1113ed0
```
You can come back to today's version with:
```
git checkout exact_multipole_pass
```
Don't forget to rebuild everything when switching versions:
for Matlab:
```Matlab
atmexall
```
For python:
```
rm -rf build
pip install . # (or pip install -e .)
```

simoneliuzzo:
Dear @lfarv ,

my matlab benchmark code does this. Every time I download a new AT, give it a name, run atmexall and run the tests. So I can go back to the frozen version I downloaded on the day of test.


```
clear
close all

format long

warning('off','all')

atfolder = 'at_9Oct2023';
% atfolder = 'at_12Sept2023';

% use local AT
restoredefaultpath;

% rmpath(genpath('/machfs/liuzzo/EBS/beamdyn/at/atintegrators'))
% rmpath(genpath('/machfs/liuzzo/EBS/beamdyn/at/atmat'))

addpath('/machfs/liuzzo/EBS/beamdyn/matlab/matlab_extensions/export_fig-master')

% add local AT folders
addpath(genpath(['./' atfolder '/atintegrators']))
addpath(genpath(['./' atfolder '/atmat']))
addpath('.');

branch = 'exact_multipole_pass';

cd(['./' atfolder ])
system(['git checkout ' branch]);
cd('..');


if ~exist(['./' atfolder '/atintegrators/AperturePass.mexa64'],'file')
    cd(['./' atfolder '/atintegrators'])
    atmexall
    warning('remember to change RF=1.0 in line 31 of at/atmat/lattice/Converters/AT2MADX/AT_2_MADX.m')
end
warning('on','all')

warning('off','all')


```

I will check if there is something that did not go well, but since I could easily recompute the data for September 2023 I am rather confident I did the correct thing. May be the  X*X modification had some unwanted effect?

lfarv:
@simoneliuzzo : sorry if I look too suspicious, but I would feel safer if you could restart from scratch from the git repository (commit 1113ed0).

I went carefully though all the modifications since September 12 and none can affect the results. I found a bug fix for "radiation" methods (*RadPass) which should be different, but not for "plain" methods.

In the worst case, if you confirm that commit 1113ed0 gives better results, I can "replay" one by one all the modifications and check all along, but that looks really unlikely…

lfarv:
@simoneliuzzo : a question about your script: in
```Matlab
cd(['./' atfolder ])
system(['git checkout ' branch]);
cd('..');
```

- If the folder is empty, `git checkout` is unable the make a new clone (no reference to the remote repository). You must use `git clone`,
- if the folder is already a git working space, `git checkout exact_multipole_pass` switches to the branch `exact_multipole_pass`, but does not update it. In addition you must call `git pull`

So there is for sure something else than what you show… I'd like to see the whole story

simoneliuzzo:
Dear Laurent,

the commit you mention does not seem to exist, but is also mentioned above in this conversation.  
Screenshot 2023-10-09 at 22 24 47

I took an other one 3bc224d. Problem was not present.
Screenshot 2023-10-09 at 22 34 10

I took an other one bbc5346. Problem was present.
Screenshot 2023-10-09 at 22 45 42


Concerning the matlab code for my tests, I definitely do not want to checkout and pull. I clone a new AT copy when I need to run a new test and that is my reference copy, no need to remember the latest commit code. Frozen forever, exactly to be able to reproduce previous results.

lfarv:
Hi @simoneliuzzo, thanks for investigating, it's indeed crucial.

https://github.com/atcollab/at/commit/3bc224dfc15b3562b3bea219bbbc68a03b6679e4 is very old (March 14), and https://github.com/atcollab/at/commit/bbc53461c9d5d662c47b77ecb17aab8af9a3f612 is very recent (October 5).

In the history, I find 4a68ffa1 from September 11, just before you ran your successful test. This one should be correct. The problem must lie after this one.

You say that the problem involves ExactDriftPass or ExactMultipolePass (at least). 
`ExactDrift` did not change since July,
`ExactMultipole` changed 3 times since September 11:

- 90d3a21d 12/09 - Send error if NumIntSteps <= 0
- 0b032310 14/09 - Add the `FieldScaling` attribute for radiation tapering. Disabled unless there is a `FieldScaling` in your  magnets.
- 488ceef7 14/09 -  Use the standard attribute names for multipole fringe fields. Disabled unless there is a `FringeQuadEntrance` or `FringeQuadExit` in your magnets. Same for bending magnets by the way: if there is some of these flags, it may have activated some fringe field effects in the last modifications

Can you check that you are not in one of there cases ?

Otherwise, you may try these commits, but don't rush, there is no pressure to merge. I continue investigating on my side but I'm leaving tomorrow for one week.

> Concerning the matlab code for my tests, I definitely do not want to checkout and pull. I clone a new AT copy when I need to run a new test and that is my reference copy, no need to remember the latest commit code. Frozen forever, exactly to be able to reproduce previous results.


Perfect. I was just surprised because the "clone" does not appear in your script. Could you just in addition record the commit hash corresponding to each of your clones ? It helps a lot in browsing the history of the code.

simoneliuzzo:
The DBA lattice I am using has FringeQuad = 1. 
In MADX I set fringe effects to FALSE for all cases.
If I remove them I am back to good agreement. 
Basically the initial version of the pass methods automatically ignored them, now I have to explicitly remove the field or set them to 0.
The comparison is working again. The difference AT-MADX for the 6D case is still present. 
4D
Screenshot 2023-10-10 at 11 55 36
6D
Screenshot 2023-10-10 at 11 46 35

lfarv:
@simoneliuzzo

Excellent ! Indeed, fringe field effects can now be activated with the standard attributes, but the default is to disable them.

I see that you tried the `ExactRectBendPass` with good results. I think I'll include it also, with the adequate documentation. 

Could you retry the test on EBS to check the dipoles with gradients (`ExactSectorBend` for the DQs)? The problem with the septum magnet should be solved. No pressure again…

Since I'll be away for one week, that leaves you this time for the last remarks (naming, documentation,…)

swhite2401:
Documentation is indeed on the right track, to be further developed in separate PRs, for me this is fine

simoneliuzzo:
Dear @lfarv,

Do you think the discrepancy in 6D is not relevant? 

Before merging I would like to test for EBS and also for FCC, that is the lattice where the EXACT methods should give a non negligible effect.

ciao
simone

simoneliuzzo:
Summary for DBA: 
The first Identity pass corresponds to comparison with MADX-PTC with exact flag = false.
all other columns corresponds to comparison with MADX-PTC with exact flag = True.

4D all seems ok is the Drift, Bend and Multipole Exact integrators are used.
6D the ExactMultipolePass seems to add some issue.

Screenshot 2023-10-10 at 14 03 12

I will now repeat for EBS

lfarv:
@simoneliuzzo :

> Do you think the discrepancy in 6D is not relevant?

The discrepancy is not huge. But what I'm saying is that I treat the radiation as it is done elsewhere in AT: at each integration step, I apply:
- 1st half the magnetic field kick
- radiation kick
- 2nd half of the magnetic field kick

That look correct to me, but I have no idea of what PTC is doing. So until we have further information, I propose to keep this: in terms of energy loss and damping time it's correct.

> Before merging I would like to test for EBS and also for FCC, that is the lattice where the EXACT methods should give a non negligible effect.

Sure, you have time! Lattices where the **dipole** exact pass methods are critical are small machines, because of a small bending radius and of large particle angles. Not the case of FCC, but I guess the problem there is not the dipoles (which should be very close to straight sections).

simoneliuzzo:
Dear @lfarv, 

indeed for FCC the change due to exact integrators originates in drifts and quads at the IP.

best regards
simone

simoneliuzzo:
Same is true for ExactRectBendPass. ExactSectorBendPass instead runs ok.

lfarv:
> the NaN in the septum are a potential candidate?

Indeed there are!

It was again a problem with 0 angle dipoles (septum). It's now corrected.

Can you test EBS again with `ExactRectangularBendPass` in the DLs?

simoneliuzzo:
Dear Laurent,

I have tested for DBA. EBS is running.

For DBA it is somehow worst than it used to be (see below)
Screenshot 2023-10-26 at 11 54 13

> Dear @lfarv and @swhite2401 ,
> 
> Here a summary for DBA:
> 
> The first column Identity pass corresponds to comparison with MADX-PTC with exact flag = false. all other columns corresponds to comparison with MADX-PTC with exact flag = True.
> 
> 4D all seems ok is the Drift, Bend and Multipole Exact integrators are used. comparison concerns only x,xp,y,yp. 6D the ExactMultipolePass seems to add some issue. comparison of all coordinates. largest discrepancies in transverse plane.
> Screenshot 2023-10-10 at 14 03 12
> 
> I will now repeat for EBS

simoneliuzzo:
Dear Laurent,

for DBA it looks ok.
Screenshot 2023-10-26 at 12 11 14

For EBS the discrepancy is not negligible in my opinion and does not improve with the Exact Integrators
Screenshot 2023-10-26 at 12 46 15

Screenshot 2023-10-26 at 12 47 01
Screenshot 2023-10-26 at 12 47 12
Screenshot 2023-10-26 at 12 47 32

lfarv:
@simoneliuzzo : for EBS, I see that you used `ExactRectangularBendPass`, but what about DQ's ?
On the other hand, yo can use `ExactSectorBendPass` for all magnets. Did you try?

simoneliuzzo:
dear Laurent,

here the test with all passmethods in dipoles and DQs swiched
Screenshot 2023-10-26 at 14 42 01

ExactRectBendRadPass gives some more problem, also in AT
Screenshot 2023-10-26 at 14 43 23

Screenshot 2023-10-26 at 14 41 45
Screenshot 2023-10-26 at 14 41 52

lfarv:
Hi @simoneliuzzo, thanks for your extensive tests. My feeling for now is that:

1. `ExactRectBendRadPass` looks wrong, seeing the energy behaviour. That's not a big deal, we can forget it, it does not bring anything,
2. For EBS, I fear that there is a difference in the descriptions of the lattice in AT and MAD-X. Looking at the DBA tests, I think that all 5 exact methods are correct (at least without focusing in dipoles): the agreement is perfect. For a test of focusing in dipoles, it would be easier with a much simpler lattice, fodo for instance, that could point out exactly where a potential difference would be: a single dipole, with a small bending radius… I can provide such a lattice, but I rely on you for PTC… At least, from linear optics analysis, we now that the focusing is correctly handled by `ExactSectorBendPass`,
3. For 6D, as I told before, there is a difference with PTC, but there is nothing I can do at the moment, without more information on what PTC does.

@simoneliuzzo, @swhite2401:
For now, I propose to take `ExactRectBendPass` out and merge the other methods: it would allow to issue a new Matlab release which is pending for too long, and concentrate on a much smaller set of modifications (we have 29 modified files in this PR). What do you think?

Side remark: apparently you are using NumIntSteps=80. Is there any reason for that? For me, the optimum is somewhere between 10 and 20, so 80 looks far too much… Are you using the same number in PTC?

swhite2401:
I agree that a release is much needed

simoneliuzzo:
Dear @lfarv and @swhite2401,

I am running the same tests for FCC@Z. I think that at least 2 lattices should be fine in the benchmark before merging. 

I will concentrate on this tomorrow as well, to try to figure out if there is a problem on the MADX-PTC side for EBS. I do not see substantial differences apart the DQs between the DBA and HMBA lattice in terms of elements and integrators.

Ok for the 6D, we will try to understand it later.

ciao
simone

lfarv:
> I am running the same tests for FCC@Z

Fine. Take your time !

simoneliuzzo:
Dear @lfarv and @swhite2401,

I have a few plots for FCC@Z (the origin of all these developments) 

in 4D without exact integration in either code AT and MADX-PTC agree well up to a given amplitdue. AT considers lost particles that are not lost for MADX.

Screenshot 2023-10-27 at 08 00 26

This is the case 4D with no exact integrators in AT and MADX-PTC with exact integrators
Screenshot 2023-10-27 at 07 58 18

Including the Exact* Pass I expect improvement in the comparison, here are the figures:
ExactDrift and ExactMultipole
Screenshot 2023-10-27 at 08 00 26

ExactDrift, ExactMultipole, ExactRectangularBend
Screenshot 2023-10-27 at 08 14 01

ExactDrift, ExactMultipole, ExactSectorBend (FAILS to compute optics)

ExactDrift, ExactMultipole, ExactRectBend 
Screenshot 2023-10-27 at 08 13 35

When including 6D motion (atenable_6d and ptc_track, ..., radiation=true,...) I get this:
Screenshot 2023-10-27 at 08 12 17
But we said we ignore 6D discrepancies for the moment.

Present status for FCC@Z:
1. for all cases, AT looses more particles than MADX.
2. 4D Drift and Multipole OK
3. 4D ExactDrift and ExactMultipole  OK
4. 4D ExactDrift and ExactMultipole and Exact*Bend  Worst than 3.
5. 4D ExactDrift and ExactMultipole and ExactSectorBend Fails to compute optics
6. 6D larger discrepancy in all cases

Present status for DBA:
1. 4D Drift and Multipole OK
2. 4D ExactDrift and ExactMultipole  NO improvement
3. 4D ExactDrift and ExactMultipole and Exact*Bend  OK
5. 6D larger discrepancy in all cases

Present status for EBS:
1. 4D Drift and Multipole OK
2. 4D ExactDrift and ExactMultipole   NO improvement
3. 4D ExactDrift and ExactMultipole  and Exact*Bend  NO improvement
4. 6D larger discrepancy in all cases

I will look further in the EBS case (worst case) today to see if there are not well converted cases / options.

In all cases: FringeQuadEntrance=0 and FringeQuadExit=0 and set fringe=false in madx-ptc.

simoneliuzzo:
Dear @lfarv  and @swhite2401 ,

May be I understood the particle lost in AT and not MADX. MADX ignores completely the 5th and 6th coordinate in tracking. While for AT even in 4D if a particle goes to far in 5th and 6th coordinate, is lost.
Am I wrong?

best regards
simone

simoneliuzzo:
Dear @lfarv and @swhite2401 ,
I think I finally found the issue with the EBS lattice. Octupole need to have class Octupole for the conversion to MADX. Else they are translated to drift-multipole-drift and the result is not the same. So, simply specifying Class Octupole the conversion defines a MADX Octupole and the comparison is much better.
I will send updated figures and state of the comparison as soon as the computations are done.
ciao
simone

simoneliuzzo:
Dear @lfarv
what is the difference between ExactRectangular and ExactSector Bend? They both replace BndMPole*Pass, which one should be chosen? For the users, may be it would be better to simply keep 1 of the two (ExactBendPass?)?
thank you
best regards
simone

simoneliuzzo:
Screenshot 2023-10-27 at 15 14 07

simoneliuzzo:
ExactRectangularBend seems the favourite for EBS

lfarv:
Hi @simoneliuzzo 

> what is the difference between ExactRectangular and ExactSector Bend?

Very important point: these are two different kinds magnets. This is explained in the[ documentation coming with this release](https://at.readthedocs.io/en/exact_multipole_pass/common/ex_passmethods.html). Please comment if you think it can be improved. To summarise:
- In a sector bending magnet, the magnetic axis is an arc of a circle following the reference trajectory. The reference trajectory sees a constant magnetic field an is effectively an arc of a circle. Example: DQs in EBS. In AT, it can be modelled by `BndMPoleSymplectic4Pass` or `ExactSectorBendPass`.
- In a rectangular bending magnet, the magnetic axis is a straight line. So the reference trajectory sees a field varying across the magnet and is not an arc of a circle. For arbitrary multipoles, it cannot be determined analytically, that's why we need `atrbendtune`. Example: off-axis quadrupole. In AT it can be modelled with `ExactRectangularBendPass`, or `BndStrMPoleSymplectic4Pass`, **but not `BndMPoleSymplectic4Pass`**

As you can see, for a pure dipole field (no multipoles), the magnetic axis is undefined, the reference trajectory is in any case an arc of a circle and you can use any passmethod. Your DBA test shows that they give identical results. But as soon as there is some focusing (or higher multipolar field), you **must choose the appropriate description**. In EBS for instance, you can take any for the DLs, but **you must take `ExactSectorBendPass` for the DQs** (and make the similar choice in MADX-PTC).

simoneliuzzo:
Dear @lfarv,

how can we distinguish in AT, given any lattice, if an element is a sector bend or rectangular bend? Only from the pass method? And only if the creator of the lattice is using Exact pass methods.
I think that both atsbend and atrbend give an element with PassMethod BndMPoleSymplectic4Pass. May be atrbend should give Passmethod BndStrMPole ?
In my comparisons above MADX uses always SBEND (sector). So the comparison should be better with ExactSectorBendPass, but infact the difference AT - MADX-PTC is a bit smaller when using  ExactRectangularBendPass.

lfarv:
>  So, simply specifying Class Octupole the conversion defines a MADX Octupole and the comparison is much better

This raises a problem: Octupole is not a predefined Matlab class. In several functions (at2str, for instance…) there is a switch  based on the Class, which will fail if it encounters an "unknown" class. Of course this will not break the tracking, which is independent of Classes, but this may give wrong plots, `atguessclass` will not detect these "custom" classes, `atwritem` will print wrong  descriptions…

So assigning non-official classes may be done (you can set any custom attribute to an element), but you take the risk of unpredictable behaviour in non-critical functions.

simoneliuzzo:
The converter AT_2_MADX uses the class information. In particular for octupole it usues the matlab octupole, rather than the multipole, and this has an impact on tracking. I opened a discussion about adding several new accepted "classes".

lfarv:
> how can we distinguish in AT, given any lattice, if an element is a sector bend or rectangular bend? Only from the pass method? And only if the creator of the lattice is using Exact pass methods.

Right, only the passmethod tells you the shape of the magnetic field. Up to now, there was no modelling of rectangular bends: `atrbend` and `atsbend` differ by the angle of the pole faces, but both define a sector bending magnet. Changing that now would change the behaviour of existing code, which is excluded in my opinion.

With the introduction of rectangular bends, it's true that there is a naming problem: "rectangular" may refer to either the shape of the magnet, or to the layout of the magnetic field, as it is the case here. I followed the nomenclature of E. Forests's book and I tried to make this clear in the documentation (check again, you can propose a better wording). And I accept any improvement!

simoneliuzzo:
Dear @lfarv and @swhite2401,

here the status for FCC@Z (HFD lattice by pantaleo)

Screenshot 2023-10-27 at 20 13 04
Screenshot 2023-10-27 at 20 13 21
Screenshot 2023-10-27 at 20 13 35
Screenshot 2023-10-27 at 20 13 45
Screenshot 2023-10-27 at 20 14 06
Screenshot 2023-10-27 at 20 14 13

In the horizontal plane it is clearly working ok. in the vertical plane, it is not clear, it does not get as sharp as for DBA and EBS, so the residual is not reducing progressively. Also, for this lattice, the AT particles are surviving more than the MADX ones... This I am not able to explain at the moment. It is not the same FCC lattice as the one above, it is the pantaleo HFD lattice.

Screenshot 2023-10-27 at 20 22 07

lcarver:
Hello @simoneliuzzo. I have no comment to make on this big work you are doing. Except that it would be very nice if you can find a way to include your benchmarking scripts in the pyat repository.

simoneliuzzo:
Dear @lcarver,

this is all in matlab at the moment. I can add it to pubtools, "the old way".  

If needed I can make the matlab scripts a bit nicer and put it together with the lattice converters.

simoneliuzzo:
Below the status for DBA, EBS, FCC-V22, FCC-HFD.

I let you (@lfarv, @swhite2401, @lcarver) decide if it is good enough agreement to merge.

DBA:

EBS:
Screenshot 2023-10-27 at 21 51 25

FCC-V22:
Screenshot 2023-10-27 at 21 50 18

FCC-HFD:


I could do the same plots with Fringe ON, if needed, easily. From the EBS test it seems that Fringe fields are somehow different in AT and madx-PTC (2nd column in the figures).

Fringe fields (FringQuadEntrance/Exit) are available in the ExactMultipole and Exact*Bend?

lfarv:
@simoneliuzzo, @swhite2401, @lcarver:

First, thanks @simoneliuzzo for the extensive tests! This allowed to fix a number of bugs (BendingAngle == 0, for instance).
Concerning the integrators themselves:
##### 4D
the DBA test clearly shows that the 5 `Exact*` 4D integrators are strictly identical to their MADX-PTC equivalent. They cannot be correct for one lattice and wrong for another one. So the moderate deviations in other lattices are presumably due to small difference in lattice descriptions (example: vertical behaviour on FCC).
##### 6D
There is some difference with MADX-PTC. As they are, the AT Exact* integrators give the correct lines optics, damping times, emittances, and a reasonably good behaviour compared to MADX-PTC. I'm in favour of integrating them as they are, except for `ExactRectBendRadPass`, clearly wrong. We'll see in the future if we can get more info on the PTC processing. However, @simoneliuzzo pointed out that there may be a problem in `atenable_6d`. This should be clarified before merging. @simoneliuzzo, can you give more details on that?

So my proposal is:
1. solve the problem of `atenable_6d`, if any,
2. remove temporarily `ExactRectBendPass`. We can rediscuss its utility later,
3. merge the other 4 integrators (and their *RadPass equivalents) in their present state.

Still remains for the future the test of fringe field effects, which are already implemented in all magnet passmethods, but not tested nor advertised in the doc.

And I have a question about the tests:
> Octupole need to have class Octupole for the conversion to MADX. Else they are translated to drift-multipole-drift and the result is not the same

Does this mean that more generally an arbitrary multipole in AT is not translated in the exactly equivalent long multipole in MADX? Is there a special case for octupoles? this looks more like a translation problem, which can probably be improved…

simoneliuzzo:
Dear @lfarv @swhite2401 @lcarver,

I think that the discrepancy visible in FCC could actually be relevant, as it is the only lattice where the Exact* effects are "strong" due to the IP optics. It is also true that FCC is a difficult lattice. I will try SuperB optics if I can recover them somewhere.

Concerning atenable_6d, here is what I have to do to try to make all elements pass methods switch to rad (as it is the default in madx)

```
ringrad = atenable_6d(ringmod,...
            'sextupass',pass_methods_switch_list{i,6},... ExactMultipoleRadPass or StrMPoleSymplectic4RadPass
            'octupass',pass_methods_switch_list{i,6});
        ringrad = atsetfieldvalues(ringrad,...
            find(atgetcells(ringrad,'PassMethod','ExactMultipolePass')),...
            'PassMethod','ExactMultipoleRadPass');
        ringrad = atsetfieldvalues(ringrad,...
            find(atgetcells(ringrad,'PassMethod','ExactRectangularBendPass')),...
            'PassMethod','ExactRectangularBendRadPass');
        ringrad = atsetfieldvalues(ringrad,...
            find(atgetcells(ringrad,'PassMethod','ExactSectorBendPass')),...
            'PassMethod','ExactSectorBendRadPass');

```

This is due to elements that are "Multipoles" or are not detected as dipoles (BendingAngle=0 ?). I agree it does not change in principle as we are specifing a Rad pass method for an element that is actually a "Drift", but it is a bit of a strange mix. 
I went to the PTC page [(ptc track radiation)](http://mad.web.cern.ch/mad/madx.old/ptc_track/ptc_track.html#RADIATION) and if I understand well, in madx **radiation on** means radiation only for dipoles! I deduce this since there is an option RADIATION_QUADR. I will try to track specifing RADIATION_QUADR=true in madx.

Now we are ought to keep the backward compatibility so we have to keep as default that radiation is only "partially on", as in madx, apparently.

I would suggest to add two options to atenable_6d: 1) multpass and 2) otherpass (or non-auto-pass) so that to get radiation on globally the command would be:

```
atenable_6d(ringmod, 'otherpass', 'StrMPoleSymplectic4RadPass');
```

in particular otherpass (or non-auto-pass), would define the default for all elements not explicitly specified (as allpass) and not 'auto', replacing the defualt 'no-change' specified by ''.

or we could make may be more simply an option 'allrad' that turns any-passmethod in any-element to its *Rad* equivalent.

I am running more tests with RADIATION_QUADR in madx and Fringe fields.

simoneliuzzo:
But, my madx apparently does not like radiation_quadr

Standard madx-ptc tracking: fringe: 0, exact: 0
+=+=+= fatal: illegal keyword: radiation_quadr
mv: cannot stat 'trackone': No such file or directory

simoneliuzzo:
Below a test for EBS:
1) I include the DQ in the pass-method conversion. Evidently DQs do not like ExactRectangularBendPass. @lfarv Is this expected?
2) I remove radiation in quadrupoles in AT. This does not solve the 6D discrepancy, actually makes it bigger... 
3) added a final test with all Exact*Pass and Fringe Fields. The case Standard + Fringe Fields is column 2 in the figures above. I could not find documentation in MADX concerning what they mean for Fringe Field the command I use is  [PTC_SETSWITCH](https://mad.web.cern.ch/mad/madX/doc/usrguide/ptc_auxiliaries/PTC_SetSwitch.html) .

Screenshot 2023-10-28 at 14 18 33

simoneliuzzo:
Dear @lfarv 
a last test, with the magic combination of commands to make all pass methods turn to Rad.
Screenshot 2023-10-28 at 15 13 12

The case of ExactDriftPass 6D is chopped, it actually goes up to 6000.
I put a few figures for the cases in the plot:

### No Exact in AT and No Exact in MADX, 6D
Screenshot 2023-10-28 at 15 21 05

### ExactDrift, 6D  <<-- very strange!
Screenshot 2023-10-28 at 15 15 12

### ExactDrift, Multipole 6D
Screenshot 2023-10-28 at 15 14 50

### ExactDrift, Multipole, Sector 6D
Screenshot 2023-10-28 at 15 14 20

### ExactDrift, Multipole, Sector 6D
Screenshot 2023-10-28 at 15 14 12

### ExactDrift, Multipole, Sector 6D +Fringe
Screenshot 2023-10-28 at 15 13 28

lfarv:
Hi @simoneliuzzo, you never stop working?

> Evidently DQs do not like ExactRectangularBendPass. @lfarv Is this expected?

As I said [here](https://github.com/atcollab/at/pull/581#issuecomment-1632905319) and [here](https://github.com/atcollab/at/pull/581#issuecomment-1783094224), **DQs are not rectangular magnets**. Their magnetic axis is a curved line. So you must use `ExactSectorBendRadPass`.

Of course, you may design a different lattice using rectangular magnets for DQs. Given the large bending radius of DQs, I expect the optics to be very similar, with a possible small change in tunes and chromaticities. But it's normal that the tracking is different.

lfarv:
Concerning `atenable_6d`:

A bending magnet with a 0 bending angle is indeed left unchanged. This is a protection for the case where its passmethod would be a `DriftPass` (normal for  0 bending), to avoid generating a `DriftRadPass`. This can be changed by simply checking the passmethod. Do you confirm that only magnets with 0 bending angle are left unchanged? If so, I will modify `atenable_6d`, in another pull request.

There is a bug in the help which mentions an "octupass" keyword. In fact, octupoles are left unchanged (but who would want radiation in octupoles?).  I will remove the sentence in the help, and reconsider handling octupoles if we decide in the future to create an Octupole class in Matlab.

Finally, about your proposal of "otherpass", there is already an "allpass" keyword which applies to all the classes handled by `atenable_6d` except those selected by a specific keyword. See the examples in "`help atenable_6d`".

lfarv:
> May be I recalled why the particles are lost in AT and not MADX. MADX ignores completely the 5th and 6th coordinate in tracking. While for AT even in 4D if a particle goes to far in 5th and 6th coordinate, is lost.
Am I wrong?

AT checks the first 5 coordinates against 1 (1m position, 1rad angle or 100% dp/p). There is no check on the 6th coordinate (ct) so that there is no problem in 4D off-momentum. So I have difficulties to explain the difference between AT and MADX

lfarv:
@simoneliuzzo, @swhite2401 : I made a small correction on the `*RadPass` methods. But I don't think it will suppress the small difference with MADX-PTC.

I also made a number of tests with a small ring (bending radius ~= 1.9 m). It's a very simple lattice, which should avoid any translation problem with the MADX format (it could even be done manually). I also enhances the differences between linearised and exact trackings. So I'd like @simoneliuzzo to run it through his tests. [Here it is](https://github.com/atcollab/at/files/13217402/fodo.mat.txt). Also, @simoneliuzzo, could you send me your test script, to avoid overloading you with bunches of tests?

At this point for me, and forgetting fringe field effects:
- without radiation, everything looks perfect. The test of the joined fodo lattice will give the final answer.
- with radiation, there seems to be a small difference with PTC. I see a possible explanation: in AT, none of the passmethods deals with radiation effects induced by dipole face angles. These are small multipole-like effects and I do not know if MADX does anything with these effects. If it does, that could explain the difference. The fodo lattice joined here has no face angles, so it should answer the question.

I found another strange point when reducing the bending radius: for damping partition numbers, both `BndMPoleSymplectic4Pass` and `ExactSectorPass` start deviating from the normal (1, 1, 2) value, ending up at (0.38, 1, 2.62) for a radius of 1.9 m. Only `ExactRectangularBendPass` sticks to (1, 1, 2). However, I could not see any difference between `ExactSectorPass` and `ExactRectangularBendPass` in horizontal or longitudinal tracking results at large amplitudes. What's happening?

simoneliuzzo:
> Concerning `atenable_6d`:
> 
> A bending magnet with a 0 bending angle is indeed left unchanged. This is a protection for the case where its passmethod would be a `DriftPass` (normal for 0 bending), to avoid generating a `DriftRadPass`. This can be changed by simply checking the passmethod. Do you confirm that only magnets with 0 bending angle are left unchanged? If so, I will modify `atenable_6d`, in another pull request.
> 
> There is a bug in the help which mentions an "octupass" keyword. In fact, octupoles are left unchanged (but who would want radiation in octupoles?). I will remove the sentence in the help, and reconsider handling octupoles if we decide in the future to create an Octupole class in Matlab.
> 
> Finally, about your proposal of "otherpass", there is already an "allpass" keyword which applies to all the classes handled by `atenable_6d` except those selected by a specific keyword. See the examples in "`help atenable_6d`".

Dear Laurent,

no I did not check if it is only for 0 bending angle. It seems to me that this topic could be big enough to claim for a separate pull request. 
I would need octupoles to radiate, if they radiate in MADX-PTC. 

The description of "allpass" is not clear in the help. Reading it I understood this was not what I wanted to use. I understood it would overwrite all passmethods to the one give to allpass.  

A more clear set of options to replace bendpass, quadpass, sextupass, allpass, for me would be:
'StrMPoleSymplectic4Pass', 'StrMPoleSymplectic4RadPass', 'BndMPoleSymplectic4Pass', 'BndMPoleSymplectic4RadPass', ... 
etc... couples of no-rad, rad passmethods. evenutally triplets:
refpts1, 'StrMPoleSymplectic4Pass', 'StrMPoleSymplectic4RadPass', refpts2, 'BndMPoleSymplectic4Pass', 'BndMPoleSymplectic4RadPass', ...

Why would I want to set radiation only partially?

thank you for your help
best regards
Simone

simoneliuzzo:
Dear @lfarv, @swhite2401 and @lcarver,
The scripts I use for benchmark are not really nice code. 
I can share them, but they are not meant for sharing. somewhere my madx is called in some of my folders.
I can make them better by translating them to python. This will not happen soon.
I will share the bench mark results also for the new lattice proposed by @lfarv.
However, I think that the "test is passed" if the new integrators work for any lattice, not for a peculiar lattice designed for the test itself. If the real lattices do not compare correctly, there is something to solve (either in AT or in MADX) and users will not feel confident in using the integrators as the benchmark is not fully ok. 
This said, I will not oppose if other developers are in favor of merging.

lfarv:
@simoneliuzzo: I don't think your test scripts need to be included in the distribution: they are very specific, and they require MADX. I just proposed to get them to relieve you from preliminary tests…

Now concerning the tests themselves, for this PR we need the check whether the new integrator work as expected: no bug in applying E. Forest's formulae. For this we do not need a lattice: a single element is enough. So yes, the test is passed if it's correct for the simplest lattice, if we can't run it on a single element.

Otherwise, we'll be debugging other things: the translation from AT format to MADX format, or the fact that MADX handles things differently on certain points (radiation for instance). That's interesting but that's another topic.

So let's proceed step by step and concentrate here on the integrators themselves: the ultimate test is the small radius fodo lattice.

lfarv:
I just got you last DBA test, and compare with https://github.com/atcollab/at/pull/581#issuecomment-1783436061.

- All blue lines should be identical (no change on non-rad integrators). That may be true
- All first 5 sets (up to +ExactMultipolePass) should be identical (the only change concerns the Dipole radiation). SO I don't understand the +ExactDriftPass reaching 3000. There must be something wrong…

lfarv:
It seems that this strange behaviour appeared (on EBS) when you introduced your _"magic combination of commands to make all pass methods turn to Rad "_.
Can you stick to radiation in dipoles only, or dipoles+quads,  depending on what PC does?

simoneliuzzo:
Dear @lfarv,

I prefer not to share my benchmark code in it's present state. I am happy to run it on demand.
I will now approve this PR as I feel I am the slowing down / stopping element. 

If you need further tests please let me know. I will state my concerns in the approval message. May be it is better to progress and have the integrators in the new release. 

Latest tests:

EBS:

Screenshot 2023-11-03 at 07 26 55


FODO (24 cells, RF 352MHz, tunes .21 .34, chroma 0.0 0.0):

Screenshot 2023-11-03 at 07 47 28
Screenshot 2023-11-03 at 07 47 54

Screenshot 2023-11-03 at 07 50 13
Screenshot 2023-11-03 at 07 50 30
Screenshot 2023-11-03 at 07 50 21

simoneliuzzo (approved):
Compares well with MADX-PTC for a FODO lattice provided by the author of the PR @lfarv.

Benchmarking AT vs MADX-PTC is ok in 4D for DBA and EBS 
Benchmarking AT vs MADX-PTC +Exact* integrators is ok in 4D for DBA and EBS lattices.
Benchmarking AT vs MADX-PTC +Exact* integrators shows some discrepancy in 6D for DBA, EBS and FCC-HFD65@Z lattices.

For FCC-HFD65@Z lattice the vertical tracking does not compare well. The Exact pass methods do make the tracking more similar. The scale of the vertical phase space are very small (~1micrometer). 

Other open issues visible in benchmarking AT vs MADX-PTC (some discrepancy between the codes):
- fringe fields (4D, no Exact pass method) --> could require a separate issue and PR
- 6D tracking (no Exact pass method) --> could require a separate issue and PR
- AT enable/disable 6d functions may need some updates --> could require a separate issue and PR

simoneliuzzo:
Dear @lfarv,

for which integrators should we use the atrbendtune function? Any bend? only ExactRectangularBendPass?

I use it for all Bends. I imagine that ExactSectorBendPass will ignore the additional fields in the elements.

simoneliuzzo:
Dear @lfarv,

here the comparisons for FODO and DBA. I added one column where I skip the atrbendtune function call for bends (the second ExactRectangularBendPass column).

FODO:
Screenshot 2023-11-03 at 09 10 09

DBA:
Screenshot 2023-11-03 at 09 11 13

simoneliuzzo:
Here the AT vs MADX-PTC benchmark function.

[Compare6DBeamATvsMADX.txt](https://github.com/atcollab/at/files/13247981/Compare6DBeamATvsMADX.txt)

comments are welcome but not requested.

simoneliuzzo:
I reduced the tracking amplitude to 5sigmas for the FODO lattice.

Screenshot 2023-11-03 at 09 28 59

Screenshot 2023-11-03 at 09 29 06

the 6D deviation is mostly in ct and is order of the energy spread of the beam.

lfarv:
@simoneliuzzo 

> I feel I am the slowing down / stopping element.

For sure not: the tests you are running are crucial, and the fodo test revealed a bug in `ExactRectangularBendPass` which I'll now correct before asking you to rerun it.

I'm also thinking of sending you a fodo lattice with focusing in dipoles, just to make sure.

simoneliuzzo:
Dear @lfarv ,

apparently there is no problem. I must have missed the atrbendtune call somehow, see the latest plots.

lfarv:
There was indeed a bug appearing for small radius and large angles of entrance/exit faces (still a high blue line: for a pure dipole, with and without `atrbendtune` must be identical). I pushed the corrected version, back to you…

simoneliuzzo:
Screenshot 2023-11-03 at 12 00 07

simoneliuzzo (approved):

simoneliuzzo:
Screenshot 2023-11-03 at 12 00 43

lfarv:
@simoneliuzzo : a last test to be complete: here is a simple
[lattice with focusing in dipoles](https://github.com/atcollab/at/files/13256311/fodon.mat.txt).

simoneliuzzo:
Dear @lfarv,

here the figures for the lattice you sent.
It seems that with radiation the difference is really large. If atrbendtune is not used, then 4D tracking also is rather different.
Screenshot 2023-11-07 at 12 29 20
ZOOM 1:
Screenshot 2023-11-07 at 12 29 39
ZOOM 2:
Screenshot 2023-11-07 at 12 30 05
notice that in this zoom, it is visible that the Exact Integrators are of no help in solving the discrepancy between AT and MADX-PTC.

Here the tracking without radiation
Screenshot 2023-11-07 at 14 15 47

and with radiation
Screenshot 2023-11-07 at 12 30 46

and the optics with and without radiation.
Screenshot 2023-11-07 at 12 42 21
notice the large change of tunes.
Screenshot 2023-11-07 at 12 43 49

best regards
Simone

lfarv:
Hello @simoneliuzzo. Thanks for the tests: this one is very interesting, as it shows clearly the difference between sector and rectangular magnets. Here, since there is focusing in the dipoles, they behave differently.
1. If you don't call `atrbendtune`, you should get a very large closed orbit, which explains the large optics distortion,
2. If you apply `atrbendtune`, you now get 2 different optics with sector bend and rectangular bends. Tunes and chromaticities are slightly different. And then, you have to make sure to use in MADX-PTC the same kind of magnet (sector or rectangular). From you results, it seems that MADX-PTC is using a sector magnet in both cases, hence the difference in tracking. 

If I summarise your whole series of tests, I conclude that, **forgetting fringe field effects**:
- in 4D, I consider the agreement with MADX-PTC perfect! The only difference appears in vertical on FCC. Since it appears only in vertical, I exclude any problem in drifts or straight multipoles (they act symmetrically). So the difference must lay in dipole pole face angles. My feeling is that there is somewhere a difference in lattice descriptions, but I may be wrong…
- in 6D, there is indeed a difference. I don't know if you have any detailed reference on the processing in PTC… However, I noticed that the agreement in longitudinal was good for the initial fodo.mat lattice, but bad in the last food.mat. This suggests two things:
  1. the missing processing of pole face angles in AT that I already mentioned (fodo.mat has a 0 angle, so the problem disappears)
  2. possibly still a bug in focusing in dipoles (fodon.mat). For this I'll compare longitudinal tracking between `BndMPoleSymplectic4Pass` and `ExactSectorBendPass` on that lattice. Though `BndMPoleSymplectic4Pass` is not well adapted to small bending radii,

Now for fringe fields, we could start by first looking at multipoles, by running a simple fodo lattice without dipoles. I'll send it to you!

lfarv:
@simoneliuzzo : concerning your optics plots, could you run a single period instead of 20 ? It would be more explicit…

swhite2401:
As a side comment, in madx the dipole face angle has the same sign for exit and entrance while in AT they have opposite signs, i.e. there is a sign difference in the exit face angle if I am not mistaken

lfarv:
No, in AT the face angles should have the same sign: for instance &theta;/2 for a rectangular magnet.

swhite2401:
So in madx it should -theta/2 to get an rbend: https://mad.web.cern.ch/mad/madX/doc/usrguide/Introduction/bend.html

swhite2401:
Oups, sorry again I read too fast... forget my messages, it is late...

simoneliuzzo:
Below I run the lattice cells proposed by @lfarv. I do not see the advantage of this test, for me a single cell is not a realistic SR lattice and I do not trust computations on such sequence of magnets. 

@swhite2401, the bend face angle conversion is included in AT_2_MADX. If it did not work the vertical tunes would not be correct.

I would like to stress that from the comparison on the FODO lattices provided by @lfarv, I can conclude that the exact integrators are not usefull to improve the comparison with MADX-PTC. The difference introduced in column 3 is never recovered.
Screenshot 2023-11-09 at 09 09 14

Screenshot 2023-11-09 at 09 09 59
Screenshot 2023-11-09 at 09 09 28
Screenshot 2023-11-09 at 09 09 19

From these figures I would have concluded that in 4D the ExactIntegrators are not helping (contrary to the case of DBA, EBS and FCC).

PR #582, opened 2023-03-17
Add files via upload

catohxb:
I uploaded a few pass methods, such as quadrupole linear pass method with fringe fields, straight-geometry bends, straight symplectic integrator with exact drift and with quadrupole fringe fields, exact drift, and crab cavity pass.

### DIFF
--- a/atintegrators/BndMPoleSymplectic2RadQEPass.c
+++ b/atintegrators/BndMPoleSymplectic2RadQEPass.c
@@ -3,109 +3,79 @@
 #include "elempass.h"
 #include "atlalib.c"
 #include "atphyslib.c"
+/*
+Modified 7/1/2013 by X. Huang for particles to lose energy stochastically. 
+*/
 
 #define DRIFT1    0.6756035959798286638
 #define DRIFT2   -0.1756035959798286639
 #define KICK1     1.351207191959657328
 #define KICK2    -1.702414383919314656
 
-/*  
- *Straight dipole w/ multipole using Symplectic Integration and rotation at dipole faces
- *Created by Xiaobiao Huang, 7/31/2018
- */
+
 #define SQR(X) ((X)*(X))
 
-/*At entrance edge: move particles to the field edge and convert coordinates to x, dx/dz, y, dy/dz, then
- convert to x, px, y, py as integration is done with px, py*/
-void E1rotation(double *r,double X0ref, double E1)
-{
-    double x0,dxdz0, dydz0, psi;
-    double fac;
+
+
+double B2perp(double bx, double by, double irho, 
+                            double x, double xpr, double y, double ypr)
+/* Calculates sqr(|e x B|) , where e is a unit vector in the direction of velocity  */
     
-    dxdz0 = r[1]/sqrt(SQR(1+r[4])-SQR(r[1])-SQR(r[3]));
-    dydz0 = r[3]/sqrt(SQR(1+r[4])-SQR(r[1])-SQR(r[3]));
-    x0 = r[0];
-    
-    psi = atan(dxdz0);
-    r[0] = r[0]*cos(psi)/cos(E1+psi)+X0ref;
-    r[1] = tan(E1+psi);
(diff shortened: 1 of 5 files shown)


catohxb:
These pass methods were used to compare with Elegant and good agreement was found. See
https://journals.aps.org/prab/abstract/10.1103/PhysRevAccelBeams.22.114601

lfarv:
Hello @catohxb, thanks for uploading these passmethods. I went rapidly through, and have a few questions:

1. `BndStrMPoleSymplectic4Pass` is already in the main branch. It is slightly different from yours because of the introduction of the new entry point trackFunction replacing PassFunction. I'll check that it's still exactly equivalent to yours.
2. `LaDriftPass`: it looks very similar to the `ExactDriftPass` proposed in #581. To be checked also, and we'll keep one.
3. `BndMPoleSymplectic2RadQEPass`: it look similar to `BndMPoleSymplectic4QuantPass`, with a few differences:
   - you replace the 4th order integrator with a 2nd order: any reason ?
   - the random generation of the energy loss is different, and I'm not sure that the approximation of the Poisson distribution with a Gaussian is valid. Could you have a look to `BndMPoleSymplectic4QuantPass` and give your comments?
   - you may have different fringe field effects, I'll come to that later
4. `QuadLinearFNPass`: Could we add the fringe fields as optional in the existing QuadLinearPass ? This is what is done in `StrMPoleSymplectic4Pass` to avoid numerous pass methods. By the way, could you check the method "2" in `StrMPoleSymplectic4Pass`, described as "elegant-like fringe field effect" and using the `fringeIntM0` and `fringeIntP0` integrals? It would be nice to have the same inputs for both linear and symplectic passmethods.
5. `StrMPoleSymplectic4NPass`: similar to `ExactMultipolePass`, and same question about fringe fields.
6. `CrabCavityPass`: no question, nice addition!

The comparison with Elegant is very interesting and encouraging. You did a lot on fringe fields, could you check what's already in the current passmethods:
- for straight elements there is now a `FringQuadEntrance` attribute (0, 1 or 2) defining the method. Are these methods correct, does any corresponds to yours, and if not, could we add yours as an additional method?
- for curved elements, it's similar with a `FringeBendEntrance` attribute.

catohxb:
Hi, @lfarv , Thanks for going through the passmethods and the suggestions. Indeed many of these features are already included. It's fine to leave them out. I'll check the fringe field model.

lfarv:
@catohxb : Anyway, `CrabCavityPass` is a nice addition. I have another question about `BndStrMPoleSymplectic4Pass`:
how should we use the `X0ref` end `RefDZ` parameters ? Could you give some example of how to use this passmethod?
Thanks!

catohxb:
@lfarv Sorry for the late reply. Basically we need to set X0ref to the proper value such that the particle comes at this position and with the correct entrance angle will have a symmetric trajectory across the magnet (exit at X0ref and with the correct angle). Then we set RefDZ to the change of z-coordinate on the magnet. 
I have a function to look for the correct X0ref value and set RefDZ. I also made an example. But I am not sure how to pass the example to you (there are 4 files).

swhite2401:
Can we close this PR now that the exact passmethods have been integrated to the distribution?

catohxb:
Yes, we can. No objections.

Xiaobiao
________________________________

lfarv:
Hello,I think there is a « crab cavity «  method that could be worth updating and added Le 29 janv. 2025 à 08:01, Xiaobiao Huang ***@***.***> a écrit :﻿
Yes, we can. No objections.



Xiaobiao

________________________________

From: Simon White ***@***.***>

Sent: Tuesday, January 28, 2025 6:55 AM

To: atcollab/at ***@***.***>

Cc: Xiaobiao Huang ***@***.***>; Mention ***@***.***>

Subject: Re: [atcollab/at] Add files via upload (PR #582)





Can we close this PR now that the exact passmethods have been integrated to the distribution?

swhite2401:
Ok then let me add a new PR for just that one.

PR #585, opened 2023-03-24
Fix atBunchLength

carmignani:
atBunchLength now checks if the ring is 4D or 6D. If it is 4D, the needed parameters are computed with ringpara, if it is 6D they are computed with atx.
The function blgrowth has been separated from BunchLength, so it can be used by both BunchLength and atBunchLength.

### DIFF
--- a/atmat/atphysics/LongitudinalDynamics/atBunchLength.m
+++ b/atmat/atphysics/LongitudinalDynamics/atBunchLength.m
@@ -6,7 +6,8 @@ function BL = atBunchLength (ring,Ib,Zn)
 %
 % Ib is the bunch current [A] (it may be a vector for multiple values)
 % Zn is the longitudinal broadband impedance [Ohms]
-% ring is the at ring without radiation
+% ring is the at lattice (4D or 6D)
+%
 % BL is the bunch length in metres 
 %
 %   see also: BunchLength, blgrowth
--- a/atmat/atphysics/LongitudinalDynamics/atBunchLength.m
+++ b/atmat/atphysics/LongitudinalDynamics/atBunchLength.m
@@ -9,14 +9,25 @@ function BL = atBunchLength (ring,Ib,Zn)
 % ring is the at ring without radiation
 % BL is the bunch length in metres 
 %
-%   see also: BunchLength
+%   see also: BunchLength, blgrowth
 
-rp=ringpara(ring);
-[E0,~,Vrf,h,U0]=atenergy(ring);
-alpha=rp.alphac;
-sigdelta=rp.sigma_E;
+is6d=check_6d(ring);
 circ=findspos(ring,length(ring)+1);
 
-BL = BunchLength(Ib,Zn,Vrf,U0,E0,h,alpha,sigdelta,circ);
-
+if ~is6d
+    [E0,~,Vrf,h,U0]=atenergy(ring);
+    rp=ringpara(ring);
+    alpha=rp.alphac;
+    sigdelta=rp.sigma_E;
+    BL = BunchLength(Ib,Zn,Vrf,U0,E0,h,alpha,sigdelta,circ);
+else
+    [~,ringdata]=atx(ring,1);
+    [~,~,Vrf,h]=atenergy(ring);
+    alpha = ringdata.alpha;
+    E0 = ringdata.energy;
+    U0 = ringdata.eloss;
+    sigdelta = ringdata.espread;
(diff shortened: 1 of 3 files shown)


atmat/atphysics/LongitudinalDynamics/atBunchLength.m
```diff
@@ -6,17 +6,29 @@
...
+    BL = BunchLength(Ib,Zn,Vrf,U0,E0,h,alpha,sigdelta,circ);
+else
+    [~,ringdata]=atx(ring,1);
+    [~,~,Vrf,h]=atenergy(ring);
+    alpha = ringdata.alpha;
+    E0 = ringdata.energy;
+    U0 = ringdata.eloss;
+    sigdelta = ringdata.espread;
+    bl0 = ringdata.blength;
+    BL = bl0 * blgrowth(Ib,Zn,Vrf,U0,E0,h,alpha,sigdelta);
```
swhite2401, line 32: There are some possible optimizations here:
alpha -> mcf(disable_6d(ring))
energy, vrf, h -> atGetRingProperties
sigdelta, bl0 -> ohmi_envelope
U0 -> atgetU0(ring, 'method', 'tracking')

lfarv: On top of @swhite2401 proposal:

`is6d` and `alpha` are also available through `atGetRingProperties`. Asking for `is6d`, `Vrf`, `E0`, `h`, `alpha` and `circ` in one call (before the `if` test) minimises the number of passes through the lattice. Be careful to use the "single cell" attributes: `cell_rf_voltage`, `cell_harmnumber`, `cell_length`.

You could also get both `bl0` and `sigdelta` from `ringpara` and completely forget `BunchLength`.

carmignani:
> alpha -> mcf(disable_6d(ring))

I was thinking to use mcf for alpha, but then it would be the on-energy value and I think here we need the off-energy one. 
I assumed that atx was giving the off-energy alpha, but apparently it is still the on-energy one. Is it a bug?

carmignani:
@lfarv : I tried atx with a 6D lattice on-energy and with Delta frequency of 300 Hz and alpha returned by atx is exactly the same.

lfarv:
@carmignani, @swhite2401:

The problem, for a given 6D lattice, is to derive its off-momentum `dp` so that it can be given to `mcf`. The solution is: get `df` = (RF frequency - nominal frequency), from `df` derive the path lengthening `dct`: `dct` = &beta;.c.`df`/f<sup>2</sup>, then compute the 4d closed obit with this `df`, this will give `dp` in the 5th coordinate. But I do not know where to put this: in `atBunchLength` ? `mcf` ? in `atx` ?

More: if the 6d lattice results from a `frequency_control` decorator (or equivalent in Matlab), the `dp` was maybe specified by the user to generate the RF frequency, and maybe not. We should avoid double computation…

Up top now, (in linopt6 I think) this was approximated by simply computing the 6d closed orbit and taking out the `dp` at the entrance of the lattice. This is approximate…

lfarv:
@swhite2401:
> For 6D, as you said this is approximate but do you have other ideas?

- The approximate method is to take the `dp` at origin of the 6d closed orbit (instead of the average). But it's correct for differential `dp` (subtracting 2 closed orbits at different frequencies), as it is done in `linopt6`,
- The exact method is the one I described above. It's not so expensive: deriving a 4d lattice (anyway necessary for `mcf`), and 4d closed orbit computation. Probably not worse than computing the integral, and works in all conditions (vertical dispersion, coupling…)

> I would implement the correction at the lowest level

This means all functions needing the `dp` value for any use… I have no exhaustive list, and it could result in multiple computations… But may be doing it in `mcf` will solve most of the needs.

swhite2401:
Ok understood, it would be nice to have this list of functions, and if they are used in core AT functions... double computations could be acceptable in ringparam that does need to be particularly fast for example but not in find_orbit

lfarv:
> it would be nice to have this list of functions

Yes, of course… But I even ignored the existence of `atBunchLength`… And I don't see how we could do an automatic search. Hence my proposal to start with `mcf`, which could fix the functions we are aware of at the moment: `ringpara`, `atsummary`, `atx`, `atBunchLength`.

`mcf` would accept 4D ot 6D lattices, with `dp`, `dct` and `df` arguments. At the cost of one conversion to 4D and 2 or 3 closed orbit computations (2 are anyway already necessary).

PR #587, opened 2023-03-28
Fixes bug producing infrequent NaN from calcnaff

### LINKED ISSUE #40: nafflib problems
Hello,
I tried to use the calcnaff function that we have in the nafflib directory and I found some problems.

If I run the function to analyze the same signal two times I get the same frequencies, but the phases are every time different.

A second issue about the function is that it requires a complex signal, so it cannot be used with real bpm measurements. We could give a vector of zeros as imaginary part of the signal, but then I don't know if we can trust the results.

pcsch:
Dear all,

This pull request concerns MATLAB.
As noted in #40 calcnaff sometimes produces NaN values even though the input is valid and a second run does result in valid frequencies. 
The problem is that the interface between MATLAB and the NAFF algorithm did no write to the last element of the
complex input to NAFF.
This pull request fixes that.

In my tests this also fixed the fluctuating results from run to run (e.g. phase) seen in #40 as well.

Fixes #40

### DIFF
--- a/atmat/atphysics/nafflib/calcnaff.m
+++ b/atmat/atphysics/nafflib/calcnaff.m
@@ -4,7 +4,7 @@ function [frequency,amplitude,phase] = calcnaff(Y, Yp, varargin)
 %
 %  INPUTS
 %  1. Y  - position vector
-%  2. Yp -
+%  2. Yp - angle vector
 %  3. WindowType  - Window type - 0 {Default} no windowing
 %                                 1 Window of Hann
 %                                 2 etc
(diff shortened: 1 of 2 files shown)


atmat/atphysics/nafflib/calcnaff.m
```diff
@@ -71,15 +65,6 @@
...
 else % Frequency map analysis
     [frequency,amplitude,phase] = nafflib(Y, Yp, WindowType,nfreq,DebugFlag);
-    %It seems there is a bug in nafflib, something returns nan even for valid data 
-    niter = 0;
-    while any(isnan(frequency)) && (niter < nitermax)
-        pause(2);
-        fprintf('Warning Nan returned by NAFF (x%d)\n', niter);
-        niter = niter +1;
-        [frequency,amplitude,phase] = nafflib(Y, Yp, WindowType,nfreq,1); % add debugging
-    end
```
lfarv, line 81: So apparently, up to now, if the output is wrong, we try again until it gets right !
This looks crazy, your fix is really welcome !

lfarv (approved):
Looks good. Any comment @lnadolski ?

atmat/atphysics/nafflib/calcnaff.m
lnadolski: Line #7 should become
 2. Yp - angle vector

Line 36 should become
NT = 9996; % divisible by 6

line #60 
elseif any(isnan(Y(1,:)))
should read
elseif any(isnan(Yp(1,:)))

lnadolski (commented):
Apart 3 small modifications in calcnaff, this looks fine for me.

lfarv (approved):
OK again !

PR #588, opened 2023-03-29
Acceptance shift zero + docs

swhite2401:
This PR fixes a couple small issues:
-acceptance shift_zero now applied to all coordinates with a larger default and documentation completed: this prevents to build artificial wings on the DA when on one coordinate is non zero
-The documentation of lattice_object.slice_elements is corrected.


lfarv (approved):
Ok for me

Issue #594, opened 2023-04-21
atreduce transforms rbend in sbend


simoneliuzzo:
```
r = {atrbend('Bend',0.1,0.1);
    atrbend('Bend',0.1,0.1);
    };

red = atreduce(r);

atwritem(red)
```

Running this small example above, it is evident that rbend are reduced to sbend. In a real lattice this generates large vertical tunes discrepancy among the intiial and reduced lattice

swhite2401:
I think there needs to be a constraint that is not included in matlab: `ExitAngle_b1=-EntranceAngle_b2` 
Adding this constraint you cannot merge 2 rectangular bends where (ExitAngle_b1=EntranceAngle_b2=theta/2).

lfarv:
Fixed in #604

PR #604, opened 2023-05-23
atreduce: check entrance and exit angles in bends

lfarv:
This fixes a problem reported in #594. A check of entrance and exit angles of adjacent bending magnets is added:
```Matlab
abs(ExitAngle(bend1)+EntranceAngle(bend2)) < 1.e-6
```
This is always true for realistic magnets, otherwise the poles would overlap. However AT also allows non-realistic magnets, where "reducing" gave wrong results.


swhite2401 (approved):

lfarv:
I'd still like to have @simoneliuzzo's advice!

lfarv:
@simoneliuzzo : does this answer your question ?

Issue #592, opened 2023-04-06
AT does not compile with OpenMP=1


swhite2401:
The C code including random number (QuantdiffPass, *QuantPass, etc...) generation does not compile anymore if OpenMP is activate below the error:

![image](https://user-images.githubusercontent.com/10014116/230361424-73b3a718-1c64-4783-b6b0-6e079b948d38.png)

It looks like the random generator has to be added to the shared variables. @lfarv do you confirm?

swhite2401:
In fact adding rng to the shared variables allows to compile correctly, I have test if it runs...

lfarv:
The `rng` variable must indeed be added to the shared variables. I then compiles and runs, but there is another problem: the average energy loss from quantum radiation is ~10% too high with OpenMP.

I'm looking at that.

lfarv:
Update: the test is done looking at the energy loss over one turn of S28F, with 1000 particles and `BndMPoleSymplectic4QuantPass` compared to `BndMPoleSymplectic4RadPass`.

The results are in excellent agreement without OpenMP and differ by ~10% with OpenMP

lfarv:
### Results without OpenMP

![histo1](https://user-images.githubusercontent.com/20038121/230598413-90c49606-a6f3-4f11-bd92-9f816fdd75c6.png)
![curve1](https://user-images.githubusercontent.com/20038121/230598424-0fe5cb5b-8c4b-4738-81cc-32ab33e3d262.png)

Here is the notebook used for the test (remove the .txt extension):
[testquantum.ipynb.txt](https://github.com/atcollab/at/files/11178428/testquantum.ipynb.txt)

lfarv:
### Results with OpenMP:
![histo2](https://user-images.githubusercontent.com/20038121/230599039-bf128c53-a43a-4523-802d-93b81a17e5c4.png)
![curve2](https://user-images.githubusercontent.com/20038121/230599054-8e9505ce-fa1a-476c-b13a-241001594c1e.png)

I do not understand what may happen with the random generator… It would be interesting to do the same test with MPI.

For the time being, it seems necessary to disable OpenMP for the 2 passmethods `BndMPoleSymplectic4QuantPass`, `StrMPoleSymplectic4QuantPass` and `QuantDiffPass`. What do you think ?

swhite2401:
Yes I agree, it has to be disabled. I will check that results are correct with MPI

PR #607, opened 2023-05-25
Fix compilation with OPENMP=1

lfarv:
Fixes a bug reported in #592.
The fix for compiling with OpenMP was easy, but it revealed a [problem with random generators](https://github.com/atcollab/at/issues/592#issuecomment-1500188375).

This PR solves the problem of compilation failure with OPENMP=1, but OpenMP is disabled in the `BndMPoleSymplectic4QuantPass`, `StrMPoleSymplectic4QuantPass` and `QuantDiffPass` passmethods until the random generation of quantum emission is validated with OpenMP.

I keep the issue #592 open because it documents the remaining problem. A dummy passmethod called `TestRandomPass` is temporarily added to the repository: it checks the random generators with and without OpenMP. The direct output of Gaussian distribution looks correct.

### DIFF
--- a/atintegrators/BndMPoleSymplectic4QuantPass.c
+++ b/atintegrators/BndMPoleSymplectic4QuantPass.c
@@ -69,12 +69,15 @@ void BndMPoleSymplectic4QuantPass(double* r, double le, double irho, double* A, 
     double pi = 3.14159265358979;
     double alpha0 = qe * qe / (4 * pi * epsilon0 * hbar * clight);
 
+/*  The behaviour of random generators with OpenMP is doubtful. OpenMP disabled until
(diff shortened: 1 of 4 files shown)


swhite2401:
Ok for for me, any idea on the problem with the RNG?

swhite2401 (approved):

lfarv:
> Ok for for me, any idea on the problem with the RNG?

It looks very strange. The results of TestRandomPass show the mean and std values of the generated numbers are correct and identical with and without OpenMP. The common and thread-specific behaviours are also correct.

A next step will be to check if the problem lies in getting the number of emitted photons, or in computing their energy…

PR #608, opened 2023-05-31
New interface to tracking functions

swhite2401:
This PR proposes to refactor the pyAT tracking functions, presently we have
- `element_pass` to track through a single element
- `lattice_pass` to track through a sequence of elements
- `patpass` to track through a a sequence of elements using python multiprocessing

All these functions are used in internal pyAT basic functionalities such as orbit search or optics calculations.This has several drawbacks: 
- the user is provided with several entry points with different interfaces
- the interface and performance have to be maintained for internal AT functions preventing any developments or modifications (see #568 for example)

This PR proposes to refactor the tracking function as follows:
- a single entry point is provided to the users for all tracking functionalities (element, lattice, parallel)
- the interface is modified to provide flexibility for future developments without affecting existing code
- protected function are used for pyAT internal use to remove inter-dependency from the user interface

**All existing functions remain valid and usable as before but are moved to the `deprecated` module: they will therefore not be update with future developments**

Presently a minimal set of output is provided (some may not be so useful while others could be added) as an example. With the new implementation it is however possible to add outputs or post processing steps without changing the interface or affecting other pyAT functions.

`


lfarv:
@swhite2401
At first glance this looks nice. It will need a more careful investigation, but before that I have a few preliminary remarks:
- I think one should add @MJGaughran to the reviewers,
- I would not modify the existing tests because it's the way to be sure that the old functions still work. Or at least keep the most critical, if adding the tests of the new interface make the sequence too heavy,
- Maybe one could make a minor release after the pending bug fixes are merged, but before this one. This one could be be first of the 1.x series.

swhite2401:
Thanks @lfarv for the feedback, concerning test this was a bit of a dilemma: I do not want to integrate deprecated functions in the test (see discussion in #601). For me it makes no sense, so test should be modified to call 'active' functions, that are in any case called by the old ones...

Then I hesitated between the new interface and the `internal_*` functions, either one is fine for me. I had the intention to add the test to check that `track_function` and `internal_*` give identical results in any case to make sure all functions are tested.

I agree about the release, this also gives time to review this one properly.

lfarv:
> I do not want to integrate deprecated functions in the test (see discussion in https://github.com/atcollab/at/pull/601)

It's different: we may decide that deprecated functions will not be upgraded any more, but still make sure that they work correctly. New code and access to new functionalities need a switch to the new interface, but **existing code must still run** ! The tests are the only way to guarantee this !

swhite2401:
> The tests are the only way to guarantee this !

Ok, I agree, but then we should provide only minimal testing for old functions: just make sure that results are identical to the new ones.

Otherwise we have to duplicate all tests for all/new functions... this does not make much sense to me since one is calling the other...

pyat/at/tracking/utils.py
```diff
@@ -2,11 +2,103 @@
...
-__all__ = ['get_bunches', 'get_bunches_std_mean', 'unfold_beam']
+__all__ = ['fortran_align', 'get_bunches', 'format_results',
+           'get_bunches_std_mean', 'unfold_beam', 'has_collective',
+           'initialize_lpass']
+
+
+DIMENSION_ERROR = 'Input to lattice_pass() must be a 6xN array.'
+
+
+def _set_beam_monitors(ring: List[Element], nbunch: int, nturns: int):
```
lfarv, line 20: To follow:
I think here a `Sequence` instead of a `list` should be enough

pyat/at/tracking/track.py
```diff
@@ -134,113 +208,57 @@ def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
...
-    :meta private:
-    """
-    warn(UserWarning("The public interface for tracking is 'lattice_pass'"))
-    return _atpass(*args, **kwargs)
-
-
-# noinspection PyIncorrectDocstring
-def elempass(*args, **kwargs):
-    """elempass(element, r_in)
+        initialize_lpass(lattice, kwargs)
```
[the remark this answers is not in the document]
swhite2401: Other option is to simply add:

```python
    if not isinstance(lattice, list):
        lattice = list(lattice)
```

in the old `lattice_pass` not such a big overhead no?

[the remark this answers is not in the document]
lfarv: And require a `Sequence` everywhere except in the old `lattice_pass` ? This removes the conversion from `_lattice_pass`, it's simpler and I don't think we loose anything…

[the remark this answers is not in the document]
swhite2401: No doesn't work, this allows to input a `numpy.array` that is not accepted by `atpass`, a check still has to take place in `_lattice_pass`

pyat/at/tracking/utils.py
```diff
@@ -2,11 +2,106 @@
 Utility functions for tracking simulations
 """
 import numpy
-from at.lattice import Lattice, DConstant
-from typing import Optional
+import functools
+from typing import List, Optional, Sequence
```
lfarv, line 6: For typing `initialize_lpass`:
```
import numpy
import functools
from collections.abc import Sequence, Iterable
from typing import Optional
```

pyat/at/tracking/utils.py
```diff
@@ -2,11 +2,106 @@
...
+    nbunch = getattr(lattice, 'nbunch', 1)
+    bunch_currents = getattr(lattice, 'bunch_currents', numpy.zeros(1))
+    if unfoldbeam:
+        bunch_spos = getattr(lattice, 'bunch_spos', numpy.zeros(1))
+    else:
+        bunch_spos = numpy.zeros(len(bunch_currents))
+    return nbunch, bunch_spos, bunch_currents
+
+
+def initialize_lpass(lattice, kwargs):
```
lfarv, line 39: To avoid warnings in `track.py`:
```
def initialize_lpass(lattice: Iterable[Element], kwargs) -> list[Element]:
```

pyat/at/tracking/track.py
```diff
@@ -1,70 +1,118 @@
...
+    refs = get_uint32_index(lattice, refpts)
+    kwargs['reuse'] = kwargs.pop('keep_lattice', False)
+    return _atpass(lattice, r_in, nturns, refpts=refs, **kwargs)
 
 
 @fortran_align
-def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
-                 refpts: Refpts = End, **kwargs):
+def _plattice_pass(lattice: Sequence[Element], r_in, nturns: int = 1,
+                   refpts: Refpts = End, pool_size: int = None,
```
lfarv, line 72:
```
def _plattice_pass(lattice: list[Element], r_in, nturns: int = 1,
                   refpts: Refpts = End, pool_size: int = None,
                   start_method: str = None, **kwargs):
```
Same thing.

Don't forget the line
```
```
for python<3.9

lfarv:
With correct type hints in `internal_lpass` and `internal_plpass`, one should see immediately if there is a problem with the many modified files using them !

pyat/at/physics/energy_loss.py
```diff
@@ -9,7 +9,7 @@
 from at.lattice import QuantumDiffusion, Collective
 from at.lattice import get_bool_index, set_value_refpts
 from at.constants import clight, Cgamma
-from at.tracking import lattice_pass
+from at.tracking.track import internal_lpass
```
lfarv, line 12:
```
from at.tracking import internal_lpass
```
One should stick to the package name (internal package organisation is private)

lfarv:
Coming now to the contents of the new interface:
I see 2 nice improvements:
- lattice tracking with single or multiple processes are unified, which makes the choice much easier. By the way, I saw that from python 3.13, the "spawn" startup mode will be the default for all platforms, including linux. Is the "fork" method going to disappear?
- the output format allows for future additional fields when necessary (the exact contents can be discussed).

On the other hand, I am sceptical about the merge of `Element` and `Lattice` tracking, for the following reasons:
- the code is mostly disjoint: almost no common part,
- the argument list is almost completely different:
   - `nturns`, `refpts`, `keep_lattice`, `keep_counter`, `turn`, `losses`, `unfold_beam` have no meaning for element tracking,
   - the two common arguments, `particle` and `energy` behave differently: they are almost useless for lattice tracking since they are part the Lattice object. They could even be removed if we would restrict the lattice tracking to `Lattice` objects.
     On the other hand, they are mandatory for element tracking
- for object methods, the type is known, but we anyway call a single function which immediately splits in two branches…
- I do not see any added value: I don't imagine any piece of code where the tracking object could be one or the other. It's always a known type.

For me the best situation would be:

a `lattice_track(lattice,...)` function with a `Lattice.track(...)` method, and 
a `element_track(element)` function with a `Element.track(...)` method

Any other advice on that?

lfarv:
Another question about the duplication of the input buffer.
I take the case of tracking with many particles for many turns (collective effects  for instance). To avoid generating a huge output buffer, a common solution is to track turn-by-turn:
```python
r_in = np.array(...).  # Large buffer
for turn in range(10000):
    r_out = lattice_pass(ring, r_in, nturns=1, turn=turn,...)
    # process r_out
```
This way, `r_out` is only single turn data, and there is no buffer copy. The overhead for entering `lattice_pass` is small enough.

With the new interface, this will duplicate the input buffer on each turn, without any benefit, and may degrade performance.

Did you think of a way to inhibit to duplication of `r_in` ?

Or, if the idea is to keep track of the input data, wouldn't it be more meaningful to make a copy of `r_in`, as you do, but track the copy and leave `r_in` unchanged ? This is the way most functions work… So it could be the default behaviour, no-one would be surprised, but an option could allow tracking `r_in` itself for cases as the one I showed above.

swhite2401:
Very good point, myself I use:

```python
r_in = np.array(...).  # Large buffer
for turn in range(10000):
    lattice_pass(ring, r_in, nturns=1, turn=turn,...)
    # process r_in
```

but it is true that the in-place modification regularly comes as a surprise for many users... we can certainly add a 'silent' option or something similar

lfarv:
What about my other alternative:

we put a new keyword `in_place: bool = False`

- if it is False (the default), you make a copy of `r_in` and track the copy. This is what everybody expects: the function does not modify its arguments. You need to set at least `refpts=End` (which could be the default also, as it is now) to get the result in `r_out`.
- if it is True (set explicitly to perform the turn-by-turn tracking described above), you track `r_in` directly, as it is done in lattice_pass.

This is for me the most understandable solution.

swhite2401:
@ lfarv, I have implemented your suggestions:

- element and lattice are treated in separate functions
- in_place flag added
- some output parameters remove (I kept rout in case at.End is not included in refpts it can be useful to restart the tracking)
- in element_track I only output rout because the other were empty so the interface is different in the end... I was hesitating on this one but as you said element_track is almost only used internally or for development purposes

Any other suggestions?

pyat/at/tracking/track.py
```diff
@@ -1,73 +1,122 @@
...
+                     refpts=refpts, **kwargs)
+    else:
+        if any_collective:
+            warn(AtWarning('Collective PassMethod found: use single process'))
+        else:
+            warn(AtWarning('no parallel computation for a single particle'))
+        return _atpass(lattice, r_in, nturns=nturns, refpts=refpts, **kwargs)
+
+
+def lattice_track(lattice: Union[Element, Iterable[Element]], r_in,
```
lfarv, line 94: `lattice: Iterable[Element]` only, no `Union` any more. A single element is no more accepted

pyat/at/tracking/track.py
```diff
@@ -1,73 +1,122 @@
...
+                  nturns: int = 1, refpts: Refpts = End,
+                  in_place: bool = True, **kwargs):
     """
-    :py:func:`lattice_pass` tracks particles through each element of a lattice
-    calling the element-specific tracking function specified in the Element's
-    *PassMethod* field.
+    :py:func:`track_function` tracks particles through each element of a
+    lattice or throught a single Element calling the element-specific
+    tracking function specified in the Element's *PassMethod* field.
```
lfarv, line 100: no single Element any more

pyat/at/tracking/track.py
```diff
@@ -1,73 +1,122 @@
...
+
+    Usage:
+      >>> track_function(lattice, r_in)
```
lfarv, line 103: `lattice_track` instead of `track_function`

pyat/at/tracking/track.py
```diff
@@ -1,73 +1,122 @@
...
+      >>> lattice.track(r_in)
 
     Parameters:
-        lattice:                list of elements
-        r_in:                   (6, N) array: input coordinates of N particles.
+        lattice: list of elements or Element
+        r_in: (6, N) array: input coordinates of N particles.
           *r_in* is modified in-place and reports the coordinates at
           the end of the element. For the best efficiency, *r_in*
           should be given as F_CONTIGUOUS numpy array.
```
lfarv, line 111: *r_in* is modified only if `in_place` is True.

pyat/at/tracking/track.py
```diff
@@ -80,52 +129,72 @@ def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
...
     If *energy* is not available, relativistic tracking if forced,
     *rest_energy* is ignored.
 
     Returns:
         r_out: (6, N, R, T) array containing output coordinates of N particles
-          at R reference points for T turns.
-        loss_map: If *losses* is :py:obj:`True`: dictionary with the
-          following key:
+          at R reference points for T turns. If *squeeze_out* is :py:obj:`True`
+          all dimensions of length 1 are removed
```
lfarv, line 151: If looks like the `squeeze_out` keyword is removed, which is a safe decision!

pyat/at/tracking/track.py
```diff
@@ -134,113 +203,101 @@ def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
...
-        reuse (bool):           if True, use previously cached description
-          of the lattice.
-        omp_num_threads (int):  number of OpenMP threads
-          (default 0: automatic)
-        losses (bool):          if True, process losses
+    Keyword arguments:
+        particle (Optional[Particle]): circulating particle.
+          Default: :code:`lattice.particle` if existing,
+          otherwise :code:`Particle('relativistic')`
+        energy (Optiona[float]): lattice energy. Default 0.
```
lfarv, line 282: There is no lattice any more, so the default for particle is relativistic

pyat/at/tracking/track.py
```diff
@@ -134,113 +203,101 @@ def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
...
-        r_out:  6 x n_particles x n_refpts x n_turns Fortran-ordered
-          numpy array of particle coordinates
+    If *energy* is not available, relativistic tracking if forced,
+    *rest_energy* is ignored.
 
-    :meta private:
+    Returns:
+        r_out: (6, N, R, T) array containing output coordinates of N particles
+          at R reference points for T turns. If *squeeze_out* is :py:obj:`True`
+          all dimensions of length 1 are removed
```
lfarv, line 290: Same comment as in `lattice_track`!

lfarv:
Apart from docstrings mentioned above, that looks nice!

swhite2401:
All docstrings corrected as suggested

pyat/at/tracking/track.py
```diff
@@ -1,73 +1,123 @@
...
-from ..lattice import Lattice, Element, Particle, Refpts, End
-from ..lattice import elements, refpts_iterator, get_uint32_index
-from typing import List, Iterable
+from .utils import fortran_align, has_collective, format_results
+from .utils import initialize_lpass
+from ..lattice import Lattice, Element, Refpts, End
+from ..lattice import get_uint32_index
+from ..lattice import AtWarning, DConstant, random
+from collections.abc import Iterable, Sequence
+from typing import Optional, Union
```
lfarv, line 10: `Sequence` and `Union` are not used any more

pyat/at/tracking/track.py
```diff
@@ -1,73 +1,123 @@
...
+
+    Usage:
+      >>> lattice_track(lattice, r_in)
+      >>> lattice.track(r_in)
 
     Parameters:
-        lattice:                list of elements
-        r_in:                   (6, N) array: input coordinates of N particles.
-          *r_in* is modified in-place and reports the coordinates at
+        lattice: list of elements or Element
```
lfarv, line 107:
```
         lattice: list of elements
```
A single Element is not valid any more

lfarv (approved):
For me that's ok. Unfortunately no news from other reviewers…

oscarxblanco:
Dear @lfarv  ,

I'll finish reviewing #624 and get back to this topic.

swhite2401:
Also @lmalina told me he is having a look

swhite2401:
No rush by the way!

pyat/at/tracking/track.py
```diff
@@ -134,113 +203,90 @@ def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
...
 
-# noinspection PyIncorrectDocstring
-def atpass(*args, **kwargs):
+def element_track(element: Element, r_in, in_place: bool = True, **kwargs):
     """
-    atpass(line, r_in, nturns, refpts=[], reuse=False, omp_num_threads=0)
+    :py:func:`element_track` tracks particles through one element of a
+    calling the element-specific tracking function specified in the
+    Element's *PassMethod* field. Particle are always considered
+    relativitic
```
lfarv, line 258: No, particle and energy may be entered as keyword arguments.

pyat/at/tracking/track.py
```diff
@@ -134,113 +203,90 @@ def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
...
-        reuse (bool):           if True, use previously cached description
-          of the lattice.
-        omp_num_threads (int):  number of OpenMP threads
-          (default 0: automatic)
-        losses (bool):          if True, process losses
+        in_place (bool): If True *r_in* is modified in-place and
+          reports the coordinates at the end of the element.
+          (default: False)
+        omp_num_threads (int):  Number of OpenMP threads
+          (default: automatic)
```
lfarv, line 275: These keyword are missing in the docstring:
```
        particle (Optional[Particle]): circulating particle.
          Default: :code:`Particle('relativistic')`
        energy (Optiona[float]): lattice energy. Default 0.
```

pyat/at/tracking/track.py
```diff
@@ -134,113 +203,90 @@ def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
...
-        r_out:              (6, N) array containing output the coordinates of
-          the particles at the exit of the element.
-    """
-    return _elempass(element, r_in, **kwargs)
+    return rout, trackparam, trackdata
 
 
-# noinspection PyIncorrectDocstring
-def atpass(*args, **kwargs):
+def element_track(element: Element, r_in, in_place: bool = True, **kwargs):
```
lfarv, line 253: `in_place` should be `False` by default.

lfarv (changes requested):
a few points to be modified in `element_track`

oscarxblanco:
@swhite2401 In pyat/at/physics/frequency_maps.py I use a parameter called ncpu to directly set the maximum number of cpus used in patpass.
https://github.com/atcollab/at/blob/e2ab70cc55917267250e566a66cae825e5b5e42f/pyat/at/physics/frequency_maps.py#L26

This line seems incompatible with the new tracking method. I get the following error
```python
  File "/home/sources/physmach/blanco-garcia/codes/pyenv/venv_pyat_from_sources/lib/python3.10/site-packages/at/physics/frequency_maps.py", line 178, in fmap_parallel_track
    patpass(ring, z0.T, nturns, pool_size=ncpu, losses=lossmap)
  File "/home/sources/physmach/blanco-garcia/codes/pyenv/venv_pyat_from_sources/lib/python3.10/site-packages/at/tracking/deprecated.py", line 281, in patpass
    return internal_plpass(lattice, r_in, nturns=nturns,
  File "/home/sources/physmach/blanco-garcia/codes/pyenv/venv_pyat_from_sources/lib/python3.10/site-packages/at/tracking/utils.py", line 86, in wrapper
    return func(lattice, r_in, *args, **kwargs)
  File "/home/sources/physmach/blanco-garcia/codes/pyenv/venv_pyat_from_sources/lib/python3.10/site-packages/at/tracking/track.py", line 84, in _plattice_pass
    return _pass(lattice, r_in, pool_size, start_method, nturns=nturns,
  File "/home/sources/physmach/blanco-garcia/codes/pyenv/venv_pyat_from_sources/lib/python3.10/site-packages/at/tracking/track.py", line 52, in _pass
    results = pool.starmap(passfunc, args)
  File "/usr/lib/python3.10/multiprocessing/pool.py", line 375, in starmap
    return self._map_async(func, iterable, starmapstar, chunksize).get()
  File "/usr/lib/python3.10/multiprocessing/pool.py", line 774, in get
    raise self._value
TypeError: 'use_mp' is an invalid keyword argument for this function
```

I wonder if it would be better to remove this parameter, and let python automatically use all parallel resources, or, to change the call for something compatible with the new tracking method.

lfarv:
The automatic resource allocation occurs when you do not specify `pool_size`. The bug you got is different and will be easily corrected.

pyat/at/tracking/deprecated.py
```diff
@@ -0,0 +1,285 @@
...
+         An empty list is returned and the tracking results of the last turn
+         are stored in *r_in*.
+       * By default, :py:func:`patpass` will use all the available CPUs.
+         To change the number of cores used in ALL functions using
+         :py:func:`patpass` (:py:mod:`~at.acceptance.acceptance` module for
+         example) it is possible to set ``at.DConstant.patpass_poolsize``
+         to the desired value.
+
+    """
+    kwargs['use_mp'] = True
```
lfarv, line 277: `patpass` calls directly `internal_plpass` which does not accept the `use_mp` argument (it's redundant). This line must be removed.

oscarxblanco: @lfarv should the parameter ncpu 'pyat/at/physics/frequency_maps.py' also be removed, ignored or default to None ? 
For example, 'get_acceptance' does not set an explicit value for the max number of cpus, and I think this was done in order to use all cpus available.

Also, 'ncpu' is just an upper limit, I didn't find a direct way to actually confirm the number of cpus that were used by patpass ( would the new pass methods allow to know ?). It seems that pyat prefers to use all cpus and let the user do the settings outside pyat, for example through slurm.

lfarv: Looking at 'pyat/at/physics/frequency_maps.py', I have the following comments:
- what is called `ncpu` here is sent as `pool_size` to `patpass`. It would have been better called `pool_size` to be consistent.
- this parameter only concerns the python multiprocessing library. It's the [number of python processes](https://docs.python.org/3.9/library/multiprocessing.html#module-multiprocessing.pool) started, independently of the real number of processors. It's the exact number of processes, not a maximum. If there are less processors, processes are started sequentially. If `pool_size`/`ncpu` is None, python chooses a number equal to the number of processors. So as for `patpass`, I would set the default to None.
-  This tuning is independent of the users's choice to use slurm,  MPI, OpenMP or other. Though it's not recommend to use several systems together. That's why I would suggest to add a `use_mp` keyword to let the user choose to use python multiprocessing or not.

lfarv:
@swhite2401:
The bug reported [above](https://github.com/atcollab/at/pull/608#issuecomment-1646754873) by @oscarxblanco should have been caught by the tests. 
But the parametrisation in `test_patpass.py` does not work !

swhite2401:
I think I have fixed everything, is this ok now?

lfarv:
> I think I have fixed everything, is this ok now?

I think so !

oscarxblanco:
@swhite2401 I have run the same test to calculate a diffussion map and it run without errors.

swhite2401:
I am ready to merge, should I wait for another approval?

lfarv:
Go on, everything is OK !

pyat/at/tracking/track.py
```diff
@@ -80,52 +130,71 @@ def lattice_pass(lattice: Iterable[Element], r_in, nturns: int = 1,
...
+       * :pycode:`track_function(lattice, r_in, refpts=0)` is a copy of *r_in*
          since the reference point 0 is the entrance of the first element.
        * To resume an interrupted tracking (for instance to get intermediate
          results), one must use one of the *turn* or *keep_counter*
          keywords to ensure the continuity of the turn number.
        * For multiparticle tracking with large number of turn the size of
          *r_out* may increase excessively. To avoid memory issues
-         :pycode:`lattice_pass(lattice, r_in, refpts=None)` can be used.
+         :pycode:`track_function(lattice, r_in, refpts=None)` can be used.
          An empty list is returned and the tracking results of the last turn
```
lmalina, line 198: I believe here, `in_place=True` should be added.

lmalina:
Following #568 , adding a test that track_function does not modify its input parameters would be nice.

swhite2401:
I have added a test of `rin` for `in_place=False` and corrected the docstring as suggested

lfarv:
@swhite2401: it seems that in `find_elem_m66`, the `particle` and `energy` keywords must be forwarded to `internal_epass`:
```python
internal_epass(elem, in_mat, **kwargs)
```

swhite2401:
> it seems that in `find_elem_m66`, the `particle` and `energy` keywords must be forwarded to `internal_epass`:

Done

lfarv (approved):
Once more ! And last one I hope.

swhite2401:
> And last one I hope.

I hope as well! I leave up for comments until the end of the week in case

swhite2401:
@lfarv can you re-approve? I had to resolve conflicts with the master...

lfarv (approved):
Yes, I can approve again !

PR #615, opened 2023-06-06
Matching fixes

swhite2401:
Couple bugs were found in the matching module:

- The ordering of positional and keyword argument in Variable class was wrong and is was not possible to pass args to the set/get functions. args is not set as a keyword with type tuple
- use_integer option in LinoptConstraint was not working do to a problem with the logic

These are now fixed.

In addition an improved handling of phase advance avoiding modulos in users script not matching integer part of the tune is added. Finally for 2D and 4D linopt calculations, the AtError 'Unstable ring' is caught and sent back as warning, in which case NaN are returned

### DIFF
--- a/pyat/at/matching/matching.py
+++ b/pyat/at/matching/matching.py
@@ -460,16 +460,16 @@ class LinoptConstraints(ElementConstraints):
             refpts = []
             self.get_chrom = True       # slower but necessary
         elif param == 'mu' or param == 'mun':
-            norm = norm_mu[param]
-            if use_integer:
-                self.refpts[:] = True # necessary not to miss 2*pi jumps
-            else:
-                target = target % (2*np.pi/norm)  
-            def fun(refdata, tune, chrom):               
+            # noinspection PyUnusedLocal
+            def fun(refdata, tune, chrom):
                 if use_integer:
-                    return getf(refdata, 'mu') / norm
+                    return getf(refdata, 'mu') / norm_mu[param]
                 else:
-                    return (getf(refdata, 'mu') % (2*np.pi)) / norm       
+                    return (getf(refdata, 'mu') % (2*np.pi)) / norm_mu[param]
+            if use_integer:
+                self.refpts[:] = True  # necessary not to miss 2*pi jumps
+            else:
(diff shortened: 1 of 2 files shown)


pyat/at/physics/linear.py
```diff
@@ -228,6 +236,8 @@ def propagate6(ri, phi, ai):
         dtype = _DATAX_DTYPE
     a0, vps = a_matrix(mt)
 
+    print(a0, vps)
+
```
lfarv, line 240: Isn't this debug output ?

lfarv (changes requested):

pyat/at/physics/linear.py
lfarv: Otherwise it's ok for me!

pyat/at/matching/matching.py
```diff
@@ -41,8 +43,8 @@ class Variable(object):
     """
     def __init__(self, setfun: Callable, getfun: Callable,
                  name: str = '',
-                 bounds: Tuple[float, float] = (-np.inf, np.inf),
-                 *args, **kwargs):
+                 bounds: tuple[float, float] = (-np.inf, np.inf),
+                 args: tuple = (), **kwargs):
```
lfarv, line 47: Detail: to avoid confusion with the usual `*args` (variable argument list), would this  be better named `fun_args`, or something better?

swhite2401:
@simoneliuzzo , is this ok for you? You will have to remove the modulo on the desired phase in your matching jobs as discussed yesterday.

lfarv (approved):
For Nans instead of errors, the rule is not clear. For safety. penalty evaluation functions should be prepared for both possibilities.

swhite2401:
> For Nans instead of errors, the rule is not clear. For safety. penalty evaluation functions should be prepared for both possibilities.

True, to be kept in mind for the new matching module

swhite2401:
I just checked with @simoneliuzzo this is ok so I merge it

swhite2401:
@lfarv could you re-approve? I just made some pep8 fixes

lfarv (approved):
Still ok !

Issue #622, opened 2023-06-21
modification to get_lifetime


bdalena:
Hello,

In order to compute the local Piwinski scattering rate we would like to have the function get_lifetime modified as follow:

def get_lifetime(ring, emity, bunch_curr, emitx=None, sigs=None, sigp=None,
                 zn=None, momap=None, refpts=None, **kwargs):
    """Touschek lifetime calculation
    
    Computes the touschek lifetime using the Piwinski formula

    args:
        ring:            ring use for tracking
        emity:           verticla emittance
        bunch_curr:      bunch current

    keyword args:
        emitx=None:      horizontal emittance
        sigs=None:       rms bunch length
        sigp=None:       energy spread
        zn=None:         full ring :math:`Z/n`
        momap=None:      momentum aperture, has to be consistent with ``refpts``
                         if provided the momentum aperture is not calculated
        refpts=None:     ``refpts`` where the momentum aperture is calculated,
                         the default is to compute it for all elements in the
                         ring, ``len(refpts)>2`` is required
        resolution:      minimum distance between 2 grid points, default=1.0e-3
        amplitude:       max. amplitude for ``RADIAL`` and ``CARTESIAN`` or
                         initial step in ``RECURSIVE``
                         default = 0.1
        nturns=1024:     Number of turns for the tracking
        dp=None:         static momentum offset
        grid_mode:       ``at.GridMode.CARTESIAN/RADIAL`` track full vector
                         (default). ``at.GridMode.RECURSIVE``: recursive search
        use_mp=False:    Use python multiprocessing (``patpass``, default use
                         ``lattice_pass``). In case multi-processing is not
                         enabled ``GridMode`` is forced to
                         ``RECURSIVE`` (most efficient in single core)
        verbose=False:   Print out some inform
        epsabs, epsrel:  integral absolute and relative tolerances

    Returns:
        tl: touschek lifetime, double expressed in seconds 
        ma: momentum aperture (len(refpts), 2) array
        refpts: refpts used for ma calcualtion (len(refpts), ) array
    """

    epsabs = kwargs.pop('epsabs', 1.0e-16)
    epsrel = kwargs.pop('epsrel', 1.0e-12)
    interpolate = kwargs.pop('interpolate', True)

    emitx, sigs, sigp = get_beam_sizes(ring, bunch_curr, zn=zn,
                                       emitx=emitx, sigs=sigs, sigp=sigp)
    if refpts is None:
[...]

PR #624, opened 2023-06-23
Scattering rate

swhite2401:
This implements the request of #622. A new function is added to derive the scattering rate. The interface is identical to the one of the lifetime calculation.

### DIFF
--- a/pyat/at/acceptance/touschek.py
+++ b/pyat/at/acceptance/touschek.py
@@ -12,7 +12,7 @@ from scipy.optimize import fsolve
 from ..constants import qe, clight, _e_radius
 
 
-__all__ = ['get_bunch_length_espread', 'get_lifetime']
+__all__ = ['get_bunch_length_espread', 'get_lifetime', 'get_scattering_rate']
 
 
 def get_bunch_length_espread(ring, zn=None, bunch_curr=None, espread=None):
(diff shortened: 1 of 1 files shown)


swhite2401:
@bdalena I have used your formula, could you please check that everything is fine on your side?
 You can call the new function as follows:

`scatter, *_ = ring.get_scattering_rate(emity=emity, bunch_curr=bunch_curr, refpts=refpts)`

bdalena (approved):
comparing the modification in get_lifetime and the new update od pyAT with get_scattering_rate

the function scattering_rate is equivalent to our modification of get_lifetime.
Many Thanks
Mihail and Barbara

swhite2401:
@lfarv , @oscarxblanco , could you please take a look at this on e for approval when you have time?

I have been asked by @bdalena to add the calculation of scattering rate, it makes use of the same algorithm as the lifetime with minor variations, so this is for the most part reshuffling code and introducing a new function to output the requested quantities.

I have checked the lifetime part and results are identical.

oscarxblanco:
@swhite2401 Using the input from @bdalena and Mihail Miceski I checked that the function returns w/o issues. Based on their test I created another test which calculates the scattering rate without the momentum aperture input using the current SOLEIL lattice, and it also returns w/o issues. I launch another test using the flag use_mp from the get_scattering_rate function call and confirm it used the available cores through htop.

In addition, I compared the rate equations with the piwinski article and they are written almost in the same way. I was not able to follow the construction of B1 and B2, I  would trust @bdalena's  and Mihail Miceski's cross check.

I add a folder with the scripts used for these tests, except the SOLEIL II lattice that I won't provide.

I think this branch could be safely merged.


[test_scattering_rate.zip](https://github.com/atcollab/at/files/11959753/test_scattering_rate.zip)

PR #625, opened 2023-06-28
Fix at disable_6d

lfarv:
When turning radiation off, at `disable_6d` considered multipoles with PolynomB == zeros as drifts, so kept the passmethod unchanged. This is corrected by taking into account the `Class` field for multipoles if it exists, before scanning PolynomB.


swhite2401 (approved):
Looks fine to me

lfarv:
@simoneliuzzo : does it solve the problem for you ?

simoneliuzzo:
tested this morning. works ok for me.

simoneliuzzo (approved):

PR #631, opened 2023-07-24
deprecates ncpu; use pool_size

oscarxblanco:
This PR deprecates the misleading parameter 'ncpu' used on frequency_maps.py, mentioned in https://github.com/atcollab/at/pull/608#discussion_r1271924788

Instead, an optional parameter pool_size is added with default value set to None.
If ncpu is used, a warning is shown, ncpu is removed from the argumemts and its value is given to pool_size.
pool_size is passed to patpass through kwargs.


swhite2401 (approved):
Looks ok to me

PR #632, opened 2023-07-24
Introduction of an "energy loss" element

lfarv:
This is a tentative answer to #629. A new 'EnergyLoss' element is added, together with its 'EnergyLossRadPass' passmethod.

This element simulates an abrupt energy loss. The longitudinal motion and damping times are affected accordingly. Equilibrium emittances also, but since there is no diffusion term associated, the computed emittances assume that the losses occur in a non-dispersive location.

The "energy loss" element is defined by its energy loss in eV.


lfarv:
Hello @lnadolski,
This is the 1st step: no new element class, so for the moment, the element must be defined as:
```python
eloss = at.Marker('name', PassMethod='EnergyLossRadPass', EnergyLoss=40000.0, Energy=6.0e9)
```
Or, in Matlab:
```Matlab
eloss=atmarker('name','PassMethod','EnergyLossRadPass','EnergyLoss',40000.0,'Energy',6e9);
```

It is not yet handled by `enable_6d`/`disable_6d`, so you should append it to a ring where 6d is already enabled.
However, I would appreciate some tests to validate the concept.

atintegrators/EnergyLossRadPass.c
```diff
@@ -0,0 +1,86 @@
...
+        if (mxGetM(prhs[1]) != 6) mexErrMsgIdAndTxt("AT:WrongArg","Second argument must be a 6 x N matrix");
+        /* ALLOCATE memory for the output array of the same size as the input  */
+        plhs[0] = mxDuplicateArray(prhs[1]);
+        r_in = mxGetDoubles(plhs[0]);
+        EnergyLossRadPass(r_in, dpp , num_particles);
+    } else if (nrhs == 0) {
+        /* list of required fields */
+        plhs[0] = mxCreateCellMatrix(2,1);
+        mxSetCell(plhs[0],0,mxCreateString("EnergyLoss"));
+        mxSetCell(plhs[0],1,mxCreateString("EnergyLoss"));
```
lnadolski, line 77: mxSetCell(plhs[0],1,mxCreateString("Energy"));

lfarv: Corrected!

lfarv:
I did some tests in python, not yet in Matlab

lnadolski:
Dear @lfarv 

I will come back to you tomorrow to review the code.
It seems to works well for matlab

image

The new damping times seem fine.
U0 for the bare lattice is 457 keV with damping times of 7.8/14.2/12.2 ms Jx is 1.83
U0 with extra loss is 946 keV with damping times of 49/69/43 ms Jx becomes 1.40 as expected from simple analytical computation

atx does not fully work, waiting to get atdisable6d fixed  ;-)
U0 is correct using the tracking option.

U0 = atgetU0(myringRadOn, 'method', 'tracking');

Regarding the 6D closed orbit, the fixed point for the energy is -201 keV for the nominal lattice and -657 keV.
Does it make sense to you not to get something closer to 0 keV ?

Regards,

Laurent.

lnadolski:
Fitting the envelop, I have got : a longitudinal damping time of 

- 10830 turns (theory at zero amplitude : 9858)
- 3746 turns (theory at zero amplitude : 3634)

image

lfarv:
@lnadolski : all this looks nice! I'm going on with the dedicated element class.

lnadolski:
Dear @lfarv 

For me, the Matlab version is fine and fulfills the goal.
To be improved: The code does not complain is the energy is missing

lfarv:
@lnadolski : the Energy field is now handled by the enable/disable_6d method, as it is for magnetic elements (dipoles)
- the element constructor sets the passmethod to 'IdentityPass': by default, it's disabled.
- the Energy field is added or removed when enabling/disabling longitudinal motion, exactly as for dipoles

Saving/loading works,
The total energy loss is correctly computed for both 'integral' and 'tracking' methods

For me, it's ready.

lnadolski (changes requested):
ringpara need some update

lnadolski (approved):
This works well for me!

lfarv:
@lnadolski: a new approval is needed after solving conflicts with the master branch…

Issue #637, opened 2023-07-28
Closed orbit in momentum aperture


oscarxblanco:
Dear all,

in `get_momentum_acceptance` pyat calculates the closed orbit on each reference point.
For a lattice with errors the closed orbit guess around zeros might not converge.

It would be better to either calculate the closed orbit at the start of the ring and propagate the coordinates througth out the accelerator up to the reference point, or to store the orbit in all reference points and use it later on in the routine.

I have tried the second approach by using the parameter offset in `get_acceptance` to pass the closed orbit, however, it seems to consider a single 6D point offset.

Is there any way to pass the closed orbit to momentum aperture ? or is offset able to work with multiple reference points ?

Best regards,
o

swhite2401:
@oscarxblanco `get_1d_acceptance` that is used in `get_momentum_acceptance` has no offset parameter.
In any case you are correct, the offset is not correctly handled in the presence of multiple `refpts` so this would need to be improved.

I can propose the following changes:
-adapt offset in `get_acceptance` to handle multiple `refpts`
-introduce offset in `get_1d_acceptance`

In this way you should be able to pre-compute the closed orbit at your observations points and pass it to the function function through the offset argument.

Would that work for you?

oscarxblanco:
Yes, this sounds great

PR #639, opened 2023-08-02
refpts offset for acceptance

swhite2401:
This PR answers #637. Individual offset can now be input for each refpts in get_acceptance in the form of an ndarray.

### DIFF
--- a/pyat/at/acceptance/acceptance.py
+++ b/pyat/at/acceptance/acceptance.py
@@ -5,7 +5,7 @@ from .boundary import GridMode
 from .boundary import boundary_search
 from typing import Optional, Sequence
 import multiprocessing
-from ..lattice import Lattice, Refpts, frequency_control
+from ..lattice import Lattice, Refpts, frequency_control, AtError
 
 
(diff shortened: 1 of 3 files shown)


swhite2401:
@oscarxblanco, everything seems to work fine on my side could you please let me know if this is what you were looking for?

oscarxblanco:
@swhite2401 I have made several tests to obtain the momentum aperture :
1) using get_momentum_acceptance and passing the closed orbit and their reference points.
2) using get_acceptance and passing the closed orbit and reference points
3) using get_acceptance and passing only the closed orbit.

All three finish without issues. The lifetime calculated from this three methods differs by less than 1%. The reason of this difference is not completely clear, but in any case, the modifications you made are OK.

I only have a doubt about a line that is removed where the variable 'istracked' is removed. It seems to have no impact but the name seems very important.

oscarxblanco:
It's OK on my side. Thank you for this modifications !

oscarxblanco (approved):

Issue #636, opened 2023-07-26
atSetRingProperties and cavpts side effect


lnadolski:
Dear @lfarv,

I am in trouble with the behavior of atSetRingProperties.

If I call the function, per default il will add a RingParam element at the beginning of the lattice. :-)

atSetRingProperties(THERING, 'FamName', AD.ATModel);

but per default, it will add the field cavpts. ;-(

This is **unwanted** since if I add a new element before the cavity in the lattice cavpts should be updated **by hand.**
Line 70 of atSetRingProperties:
`[parmelem,~]=atparamscan(ring,parmelem,'Particle','cell_harmnumber','cavpts');` 

For instance,  atGetRingProperties(myringRadOn, 'HarmNumber', 'rf_frequency') will use the cavpts if present in ringparam.

Is there any reason to keep this behavior?
I will either remove the 'cavpts' or make a sanity check that the cavpts need update.

Do you have any thoughts about it?

Best regards,

Laurent.

lfarv:
I see the problem. I think that the idea was to improve performance when acting in the cavities. The two solutions you propose may be implemented. The easiest (and safest) is to remove `cavpts` from the cached values. That's my preferred option for now.

`cavpts` will remain as a user-defined property to describe the "main cavities" in the case of complex RF systems (if they are different from the lowest frequency cavities), and for optional performance improvement. But it will then be completely handled by the user and absent by default.

The cavity "how-to" guide needs to be updated.

lfarv:
> Can be closed?

Yes ! Fixed in #640.

PR #640, opened 2023-08-06
Don't automatically cache the location of RF cavities

lfarv:
This corrects #636. Caching the location of cavities slightly improved performance, but raised problems after modifying the lattice. Setting the `cavpts` ring property still works for accelerating the access to the cavities, but it's now the user's responsibility to use it or not, it's mo more automatically set.

The guide for cavity control is corrected.


lfarv:
Hello @swhite2401. Could you please approve this one, which is pending for long now…

swhite2401:
Well this was a request of @lnadolski shouldn't he review this? I don't feel very comfortable with approving something I did not request and will not review...

swhite2401:
@lfarv, merging without approval is against what we have decided concerning the merge process and should be avoided.
I agree this one is small and not problematic, but it would still be good that everyone complies to the rules we have set, thanks.

Issue #642, opened 2023-08-08
Transverse Damping coming from EnergyLoss element


lcarver:
Hello @lfarv and @lnadolski ,

In the EnergyLoss element, the radiation damping is equal in the horizontal and vertical planes (and half in longitudinal) corresponding to damping partitions of [1, 1, 2].

It would be nice if we could specify the damping partition numbers in order to have the correct damping times given by this element. I'm not sure of the best way to implement this. 

As a reminder: the analytical formula for the damping time is given by

tau = 2 * E0 * T0 / ( j * U0 )

lcarver:
The intended usage of this is explained a bit more in PR #643 .

swhite2401:
Can we improve documentation then? Is it possible to include your comment above in the docstring?

I would not include any 'specific' feature to this element but rather add the possibility to 'lose' energy in existing ID elements.

swhite2401:
Solved in #959

Issue #606, opened 2023-05-24
include fast_ring generation from a list of machine parameters (tunes, chromaticities, ac, U0, etc)


lcarver:

swhite2401:
@lfarv this is done in other collective effects codes using just damping times + equilibrium emittances (exponential decay + white noise which amplitude depends on the distance w.r.t the target)

For the rest I think it is straight forward as long as you assume uncoupled motion.

I have somewhere an implementation in matlab that I used before we introduced the fastring.

lcarver:
So for some more information...

I want to make a function that can provide some simple longitudinal and transverse elements for a given list of global machine parameters i.e.

fring = gen_uncoupled_ring(Qx, Qy, Qpx, Qpy, axx, axy, ayy, Vrf, phi_s, U0, alphac, emittances, damping_times)

And then fring would be composed of:

a 6x6 tracking element (uncoupled and without radiation damping)
a nonlinear element (for chromaticity and amplitude detuning)
an rf cavity element
a damping element to include the radiation damping and equilibrium emittance (as Simon said, to see the pyheadtail implementation see here: https://github.com/PyCOMPLETE/PyHEADTAIL/blob/master/PyHEADTAIL/radiation/radiation.py)

If we have this, I can (for example) take a table of parameters for any other machine, and make some simulations. I don't need to have the full lattice model.

lfarv:
This seems to be answered in #643. Can we close the issue?

lcarver:
thanks to all.

PR #643, opened 2023-08-09
Simple ring model

lcarver:
This is presented as a solution to issue #606 . This PR was also a big motivation in the creation of issue #642 which has not yet been solved. 

Here I have introduced a way to create a very simple ring based on a dictionary of global machine parameters. The produced lattice has the following elements.

1. An RF Cavity element based on the provided parameters
2. An M66 matrix that provides tbt tracking for the given Qx and Qy and the path lengthening in z.
3. A detuning element that covers linear chromaticity and amplitude detuning
4. An EnergyLoss element that applies the U0.
5. A new passmethod 'SimpleQuantDiffPass' which applies random noise to xp, yp, and dp with a given sigma that can recreate the desired equilibrium emittances. 

I added the new passmethod, added an element to elements.py, and added a function to fastring.py.
When we have a solution to #642, we can simply add the damping partition numbers as a parameter in the function 'gen_simple_ring' and pass the damping partition numbers to the EnergyLoss element.

Below is a copy paste of a test file that I have written that shows good agreement for tunes, damping times, equilibrium emittances and z center of mass. 

OUTDATED. SEE NEWFILE BELOW


lcarver:
If it is not possible to easily add damping partition numbers to the EnergyLoss element. 
I can change a bit how the effects are introduced.

1. I remove the energyloss element.
2. I add the U0 as a misalignment error.
3. I provide damping 'artificially' with user specified damping  times and the damping is introduced either in SimpleQuantDiffPass or by adding a damping term to the Mat66.

swhite2401:
@lcarver, before going into the details of the implementation I would suggest to integrate this directly in the fastring function instead of adding a new one: make the new function private and call it in fastring in case the ring is not provided.

swhite2401:
Similarly, the quantum diffusion element could be integrate in radiation.gen_quantdiffelem when the ring is not provided

lfarv:
I fully agree with @swhite2401 on one point: the output should be similar to the one of `fastring`: exactly the same list of elements with the same functionality. I don't see why you need a new passmethod if a fastring can represent everything.

On the other hand, I don't see a good reason to group both in a same function: the inputs are completely different, and I don't think there is anything in common in the implementation. `fastring` and `simplering` look nice for me.

lfarv:
I have no objection to split betatron motion and damping in two different elements, but:
- then one should do the same in `fastring`: your argument about turning on/off is equally valid for `fastring`. It's trivial to split the global M66 matrix,
- damping (excluding diffusion) does not require a new pass method: M66 is ok for that, and generating the matrix from the given damping times is easy.
- diffusion and energy loss can be treated exactly as in fastring.

But still none of the proposed splits allows emittance computation with ohmienvelope… Too bad.

swhite2401:
I  prefer what is proposed by @lcarver, is it much simpler and more efficient. Having all radiation effects in a single element is a very nice and convenient feature and even if it is not used for this simple fastring this new passmethd is a nice addition that  should be integrated in AT.

lfarv:
Why not, but @lcarver proposes to include the damping in the (single) 6x6 matrix, while I thought you preferred having it separated from betatron motion.

Also, the pass method included in this PR does not deal neither with damping nor with energy loss: it only does the quantum excitation.

This has some advantages: if you disable it, the beams damps to zero but the closed orbit and linear optics still give the good results: you get the right synchronous phase and damping times.

Anyway, I don't care to much about how the transfer is split, I just think that once the optimum is chosen, there is no reason why it should be different between fastring and simplering…

carmignani:
I think it would be nice to have the damping separated from the M66 in the fast ring, so that we could simply remove the diffusion element and the damping element and have the fast ring without radiation.
The simplest way is to have an additional linear matrix doing only the damping, with the disadvantage that the tracking would be slower. Having a single synchrotron radiation element doing both damping and diffusion wouldn't allow to do simulations with only damping, that is what we often do in single particle simulations.
The diffusion matrix for the simple ring could be built from the moments matrix and the 6x6 matrix without the need of an additional passmethod.

lcarver:
The SimpleQuantDiffPass has been modified. It now requires U0 (in addition to previous arguments).
The passmethod now implements radiation damping (if tau>0) and the simple quantum diffusion (if emit>0) and it also removes U0/ring->Energy (same as applying a misalignment) to model the energy loss.

If the user asks for emit=0, then no quantum diffusion is applied.
If the user asks for emit=0 and tau=0, then radiation damping and quantum diffusion is not applied

The energyloss element has been removed as U0 is covered in SimpleQuantDiffPass with user defined damping times.  The file below is the updated one from the first comment (which I have now removed). It shows good damping times and emittances, correct tunes and beam centered on zero.

For me this solves the problem of generating a ring from a table. This is the simplest implementation and is a very straightforward way of generating a lattice that can be used for collective effects studies in the fewest elements. 
Ready for review.

```
import at
from at.physics.fastring import gen_simple_ring
import numpy as np
from scipy.signal import hilbert

dic = {
       'alpha_x': 1e-3,
       'alpha_y': 1e-4,
       'beta_x': 20,
       'beta_y': 10,
       'Qx': 74.1235,
       'Qy': 28.3,
       'Qpx': 0,
       'Qpy': 0,
       'A1': 0.,
       'A2': 0.,
       'A3': 0.,
       'circumference': 843.977,
       'alpha': 8.5e-5,
       'U0': 2.5e6,
       'Vrf': 6e6,
       'energy': 6e9,
       'harmonic_number': 992,
       'emit_x': 100e-12,
       'emit_y': 10e-12,
       'sigma_dp': 9.5e-4,
       'tau': [3000, 5000, 1000]
}
       
ring = gen_simple_ring(dic)

r_in = np.zeros((6,1))
r_in[0] = 1e-1
r_in[2] = 1e-1
r_in[4] = 1e-3

# Check radiation damping
nt = 10000
r_out = at.lattice_pass(ring, r_in, nturns=nt)

xdat = r_out[0,0,0,:]
fitx = np.polyfit(np.arange(2500), np.log(np.abs(hilbert(xdat)))[500:3000], 1)
print('tau_x', -1/fitx[0], dic['tau'][0])

ydat = r_out[2,0,0,:]
fity = np.polyfit(np.arange(2500), np.log(np.abs(hilbert(ydat)))[500:3000], 1)
print('tau_y', -1/fity[0], dic['tau'][1])

#spikes at start and finish are artefacts of hilbert transform
plt.suptitle('envelope from hilbert transform')
plt.semilogy(np.abs(hilbert(xdat)),'r', label='x')
plt.semilogy(np.abs(hilbert(ydat)),'b', label='y')
plt.xlabel('turn')
plt.ylabel('env [m]')
plt.legend()
plt.show()



r_in = np.zeros((6,1))
r_in[0] = 0.
r_in[2] = 0.
r_in[4] = 1e-3
r_out = at.lattice_pass(ring, r_in, nturns=nt)
r_in[4] += 5e-2
r_out = at.lattice_pass(ring, r_in, nturns=nt)
dpdat = r_out[4,0,0,:]
zdat = r_out[5,0,0,:]

fitz = np.polyfit(np.arange(500), np.log(np.abs(hilbert(dpdat)))[250:750], 1)
print('tau_z', -1/fitz[0], dic['tau'][2])

plt.suptitle('abs(hilbert(dp))')
plt.semilogy(np.abs(hilbert(dpdat)))
plt.show()


# Check emittances
sigma_matrix = at.sigma_matrix(betax=dic['beta_x'], betay=dic['beta_y'], emitx=dic['emit_x'], emity=dic['emit_y'], espread=dic['sigma_dp'], blength=3e-4, alphax=dic['alpha_x'], alphay=dic['alpha_y'])
nparts = 10000

r_in = at.beam(nparts, sigma_matrix)

emitx = []
emity = []
sigma_e = []
blength = []
for i in np.arange(20000):
    _ = at.lattice_pass(ring, r_in, nturns=1)

    track_emitx = np.sqrt(np.mean(r_in[0,:]**2)*np.mean(r_in[1,:]**2) - np.mean(r_in[0,:]*r_in[1,:])**2)
    track_emity = np.sqrt(np.mean(r_in[2,:]**2)*np.mean(r_in[3,:]**2) - np.mean(r_in[2,:]*r_in[3,:])**2)
    emitx.append(track_emitx)
    emity.append(track_emity)
    sigma_e.append(np.std(r_in[4,:]))
    blength.append(np.std(r_in[5,:]))
    
print('emitx', track_emitx, dic['emit_x'])
print('emity', track_emity, dic['emit_y'])
print('sigma_e', np.std(r_in[4,:]), dic['sigma_dp'])
print('blength', np.std(r_in[5,:]))


# Check tunes
freq = np.fft.rfftfreq(nt)
ffreq = np.fft.rfftfreq(nt,d=dic['circumference']/clight)
plt.semilogy(freq, np.abs(np.fft.rfft(xdat)),'r', label='x')
plt.axvline(dic['Qx']%1, color='r')
 
plt.semilogy(freq, np.abs(np.fft.rfft(ydat)), 'b', label='y')
plt.axvline(dic['Qy']%1, color='b')
plt.legend()
plt.xlabel('tune')
plt.ylabel('abs(fft)')
plt.show()

#Expected fs is about 1300 Hz.
plt.semilogy(ffreq, np.abs(np.fft.rfft(dpdat)), 'b')
plt.xlim([0,3000])
plt.xlabel('freq [Hz]')
plt.ylabel('abs(fft(dp))')
plt.show()



ring_noqd = ring.deepcopy()
#ring_noqd.pop(-1)
ring_noqd.get_energy_loss(at.energy_loss.ELossMethod.TRACKING)

# Check TimeLag of cavity
r_in = np.zeros((6,1))
r_out = at.lattice_pass(ring, r_in, nturns=100000)
print('zc', np.mean(r_out[5,0,0,:]))


```

pyat/at/lattice/elements.py
```diff
@@ -972,6 +972,45 @@ def __init__(self, family_name: str, m66=None, **kwargs):
...
+                                                     'emit_y',
+                                                     'sigma_dp',
+                                                     'tau_x',
+                                                     'tau_y',
+                                                     'tau_z',
+                                                     'beta_x',
+                                                     'beta_y',
+                                                     'U0']
+
+    def __init__(self, family_name: str, emit_x: float, emit_y: float,
```
swhite2401, line 987: Would it be simpler to default everything to None and use GetOptionalDouble in the C passmethod, this users only need to provide what is useful

pyat/at/lattice/elements.py
```diff
@@ -972,6 +972,45 @@ def __init__(self, family_name: str, m66=None, **kwargs):
...
+            tau_y:          Vertical damping time [turns]
+            tau_z:          Longitudinal damping time [turns]
+            beta_x:         Horizontal beta function at element [m]
+            beta_y:         Vertical beta function at element [m]
+            U0:             Energy Loss [eV]
+            
+        Default PassMethod: ``SimpleQuantDiffPass``
+       """
+        kwargs.setdefault('PassMethod', 'SimpleQuantDiffPass')
+        super(SimpleQuantDiff, self).__init__(family_name,
```
swhite2401, line 1006: Why pass all the specific arguments to your supper method?
You may instead declare the element's attribute in the __init__ method as:

```
if emitx is not None:
    self.emitx=emitx
```

etc...

pyat/at/lattice/elements.py
```diff
@@ -972,6 +972,45 @@ def __init__(self, family_name: str, m66=None, **kwargs):
         super(M66, self).__init__(family_name, M66=m66, **kwargs)
 
 
+class SimpleQuantDiff(Element):
+    """Linear tracking element for simplified quantum diffusion"""
+    _BUILD_ATTRIBUTES = Element._BUILD_ATTRIBUTES + ['emit_x',
```
swhite2401, line 977: This is not needed if argument are optional

pyat/at/lattice/elements.py
```diff
@@ -972,6 +972,45 @@ def __init__(self, family_name: str, m66=None, **kwargs):
         super(M66, self).__init__(family_name, M66=m66, **kwargs)
 
 
+class SimpleQuantDiff(Element):
```
swhite2401, line 975: The radon/radoff treatment is missing

pyat/at/physics/fastring.py
```diff
@@ -4,13 +4,14 @@
 from functools import reduce
 import numpy
 from typing import Tuple
-from at.lattice import RFCavity, Marker, Lattice, get_cells, checkname
-from at.lattice import get_elements
+from at.lattice import RFCavity, Element, Marker, Lattice, get_cells, checkname
+from at.lattice import get_elements, M66, EnergyLoss, SimpleQuantDiff
```
swhite2401, line 8: EnergyLoss is still used?

pyat/at/physics/fastring.py
```diff
@@ -4,13 +4,14 @@
...
 from at.physics import gen_m66_elem, gen_detuning_elem, gen_quantdiff_elem
+from at.constants import clight, e_mass
 import copy
 
 
-__all__ = ['fast_ring']
+__all__ = ['fast_ring', 'gen_simple_ring']
```
swhite2401, line 14: Why not call it simply simple_ring?

pyat/at/physics/fastring.py
```diff
@@ -100,3 +101,135 @@ def fast_ring(ring: Lattice, split_inds=[]) -> Tuple[Lattice, Lattice]:
...
+                         [0., 0., 0., 0., M44, M45],
+                         [0., 0., 0., 0., M54, M55]], order='F')
+
+    # generate the linear tracking element, we set a length
+    # which is needed to give the lattice object the correct length
+    # (although it is not used)
+    lin_elem = M66('Linear', m66=Mat66, Length=circumference)
+
+    # Generate the simple quantum diffusion element
+    quantdiff = SimpleQuantDiff('SQD', emit_x, emit_y, sigma_dp,
```
swhite2401, line 220: Since you are using dictionaries why not passing the dictionary to your new element and decoding it inside the __init__ method

The attribute declaration could then be made simpler:

```
for k,v in dictionnary:
    setattr(self, k, v)
```

Of course you have to extract only the needed values from your ring_dictionnary

It could be interesting to set default values for other parameters if they are not provided, for instance the detuning coeffcients=0, beta=1, alpha=0, such that the user may generate a standard ring with minimal inputs

atintegrators/SimpleQuantDiffPass.c
```diff
@@ -0,0 +1,169 @@
...
+}
+
+#if defined(MATLAB_MEX_FILE) || defined(PYAT)
+ExportMode struct elem *trackFunction(const atElem *ElemData,struct elem *Elem,
+                double *r_in, int num_particles, struct parameters *Param)
+{
+/*  if (ElemData) {*/
+        if (!Elem) {
+            double emit_x, emit_y, sigma_dp, tau_x, tau_y, tau_z, beta_x, beta_y, U0, EnergyLossFactor;
+            emit_x=atGetDouble(ElemData,"emit_x"); check_error();
```
swhite2401, line 72: use GetOptionalDouble for optional arguments

lcarver:
Hello,
All of your modifications were taken into account.
Default passmethod of SimpleQuantDiff is SimpleQuantDiffPass. When applying rad off it switches to IdentityPass.

Rather than implement the optional / default values in the C passmethod, I prefer to set them all on the python side. this allows me to make some assertions about what cases are allowed or not (tau with emit=0 is ok, emit with tau=0 is not).

Now, simple_ring requires 7 positional arguments: energy, circumference, harmonic_number, Qx, Qy, Vrf, alpha. The rest of the parameters are keyword arguments that have default values. 

New parsing of the function looks like this:

```
dic = {
       'alpha_x': 1e-3,
       'alpha_y': 1e-4,
       'beta_x': 20,
       'beta_y': 10,
       'Qx': 74.1235,
       'Qy': 28.3,
       'Qpx': 0,
       'Qpy': 0,
       'A1': 0.,
       'A2': 0.,
       'A3': 0.,
       'circumference': 843.977,
       'alpha': 8.5e-5,
       'U0': 2.5e6,
       'Vrf': 6e6,
       'energy': 6e9,
       'harmonic_number': 992,
       'emit_x': 100e-12,
       'emit_y': 10e-12,
       'sigma_dp': 9.5e-4,
       'tau_x': 3000,
       'tau_y': 5000,
       'tau_z': 1000
}
       
ring = simple_ring(dic['energy'], dic['circumference'], dic['harmonic_number'], dic['Qx'], dic['Qy'], dic['Vrf'], dic['alpha'], tau_x=dic['tau_x'], emit_x=dic['emit_x'], tau_y=dic['tau_y'], emit_y=dic['emit_y'], tau_z=dic['tau_z'], sigma_dp=dic['sigma_dp'], U0=dic['U0'], A1=dic['A1'], A2=dic['A2'], A3=dic['A3'], Qpx=dic['Qpx'], Qpy=dic['Qpy'])


```

pyat/at/physics/fastring.py
lcarver: modified

lcarver:
ready for review

pyat/at/lattice/elements.py
```diff
@@ -972,6 +972,67 @@ def __init__(self, family_name: str, m66=None, **kwargs):
         super(M66, self).__init__(family_name, M66=m66, **kwargs)
 
 
+class SimpleQuantDiff(_DictLongtMotion, Element):
+    """Linear tracking element for simplified quantum diffusion"""
```
swhite2401, line 976: And damping and eloss per turn?

pyat/at/physics/fastring.py
```diff
@@ -100,3 +101,120 @@ def fast_ring(ring: Lattice, split_inds=[]) -> Tuple[Lattice, Lattice]:
...
+    # compute rf frequency
+    frf = harmonic_number * clight / circumference
+
+    # compute slip factor
+    gamma = energy / e_mass
+    eta = alpha - 1/gamma**2
+
+    # compute the synchronous phase and the TimeLag
+    phi_s = numpy.arcsin(U0/Vrf)
+    TimeLag = clight * phi_s / (2 * numpy.pi * frf)
```
swhite2401, line 164: Why do you want to add a time lag? Should it at least be optional?

pyat/at/physics/fastring.py
```diff
@@ -100,3 +101,120 @@ def fast_ring(ring: Lattice, split_inds=[]) -> Tuple[Lattice, Lattice]:
...
+    # generate rf cavity element
+    rfcav = RFCavity('RFC', 0, Vrf, frf, harmonic_number, energy,
```
swhite2401, line 166: Have you thought about harmonic cavities?

pyat/at/physics/fastring.py
```diff
@@ -100,3 +101,120 @@ def fast_ring(ring: Lattice, split_inds=[]) -> Tuple[Lattice, Lattice]:
...
+                     TimeLag=TimeLag)
+
+    # Now we will use the optics parameters to compute the uncoupled M66 matrix
+    def sincos(x):
```
swhite2401, line 170: This is a bit odd.... and you are not really saving lines of code with it...

pyat/at/lattice/lattice_object.py
```diff
@@ -50,6 +50,7 @@
         ('collective_pass', elt.Collective, 'auto'),
         ('diffusion_pass', elt.QuantumDiffusion, 'auto'),
         ('energyloss_pass', elt.EnergyLoss, 'auto'),
+        ('simplequantdiff_pass', elt.SimpleQuantDiff, 'auto')
```
swhite2401, line 53: May diffusion_pass, energyloss_pass, simplequantdiff_pass can be grouped into a new radiative_pass argument?

pyat/at/lattice/elements.py
lcarver: done

pyat/at/physics/fastring.py
lcarver: made it optional with default = False. I also checked and this all still works with ring.set_cavity_phase()

pyat/at/physics/fastring.py
lcarver: implemented array or scalar input for harmonic_number and Vrf

pyat/at/physics/fastring.py
lcarver: removed. you were right

pyat/at/lattice/lattice_object.py
lcarver: To be discussed. I am not sure how best to do it

lcarver:
all implemented except the last comment

pyat/at/physics/fastring.py
lfarv: The harmonic number is a lattice property, rather than a cavity property. Harmonic cavities should be defined by their harmonic with respect to the main frequency (3, 5, …). But that would make the input more complicated, so we could keep it like that.

When building the `Lattice` object, the constructor looks for the lowest frequency cavity and derives the lattice harmonic number from its harmonic number. So in the end, it will be correct.

pyat/at/physics/fastring.py
```diff
@@ -100,3 +101,136 @@ def fast_ring(ring: Lattice, split_inds=[]) -> Tuple[Lattice, Lattice]:
                          split_inds=split_inds,
                          detuning_elem=detuning_elem)
     return fastringnorad, fastringrad
+
+
+def simple_ring(energy: float, circumference: float, harmonic_number: int,
+                Qx: float, Qy: float, Vrf: float, alpha: float,
```
swhite2401, line 107: Vrf and harmonic_number can be scalars or Sequence of scalars

pyat/at/physics/fastring.py
```diff
@@ -100,3 +101,136 @@ def fast_ring(ring: Lattice, split_inds=[]) -> Tuple[Lattice, Lattice]:
...
+    * an RF cavity,
+    * a 6x6 linear transfer map,
+    * a detuning and chromaticity element,
+    * a simplified quantum diffusion element
+        which contains equilibrium emittance and radiation damping
+
+    Positional Arguments:
+        * energy [eV]
+        * circumference [m]
+        * harmonic_number - can be scalar or 1d array
```
swhite2401, line 132: You may say here that the RF frequency is derived from this number and the ring circumference provided

pyat/at/physics/fastring.py
```diff
@@ -100,3 +101,136 @@ def fast_ring(ring: Lattice, split_inds=[]) -> Tuple[Lattice, Lattice]:
...
+        * Qx - horizontal tune
+        * Qy - vertical tune
+        * Vrf - RF Voltage set point [V] - can be scalar or 1d array
+        * alpha (momentum compaction factor)
+
+    Optional Arguments:
```
swhite2401, line 138: You have to mention the default values for optional arguments

pyat/at/lattice/elements.py
```diff
@@ -972,6 +972,70 @@ def __init__(self, family_name: str, m66=None, **kwargs):
...
+                 beta_y: Optional[float]=1.0, emit_x: Optional[float]=0.0,
+                 emit_y: Optional[float]=0.0, sigma_dp: Optional[float]=0.0,
+                 tau_x: Optional[float]=0.0, tau_y: Optional[float]=0.0,
+                 tau_z: Optional[float]=0.0, U0: Optional[float]=0.0,
+                 **kwargs):
+        """
+        Args:
+            family_name:    Name of the element
+            
+        Optional Args:
```
swhite2401, line 993: You should put the defaults values there and add that emit=0, etc..... will be ingnored (similar to the sentence you have for the ring function)
In fact it would be nice to have all docstings identical for the same input

pyat/at/lattice/lattice_object.py
swhite2401: Ok for now

pyat/at/physics/fastring.py
```diff
@@ -100,3 +101,136 @@ def fast_ring(ring: Lattice, split_inds=[]) -> Tuple[Lattice, Lattice]:
...
+
+    assert len(harmonic_number) == len(Vrf), (
+        "harmonic_number input must match length of Vrf input"
+        )
+
+    # compute rf frequency
+    frf = harmonic_number * clight / circumference
+
+    all_cavities = []
+    for icav in numpy.arange(len(harmonic_number)):
```
swhite2401, line 172: There is a mistake there, Timelag applies to all cavities and should be computed outside of the loop using Vrf[0] and frf[0].
However this is not safe: it is said nowhere that your main RF system should be at index 0. You have to improve the help and check that harmonic numbers are properly sorted. Other question, how to you set the Timelag of your harmonic cavity to tune its phase w.r.t. the main RF?
A Timelag input similar to Vrf and harmonic number (scalar or array) seems necessary... in this case do you really want to add this calculation inside the function? It looks like an unnecessary complication especially since you can use the set_cavity_phase function after your ring is created, I would just remove this part.

pyat/at/physics/fastring.py
lcarver: done

pyat/at/physics/fastring.py
lcarver: done

swhite2401 (approved):

PR #645, opened 2023-08-10
Repair the Matlab tests

lfarv:
The Matlab tests are broken since the introduction of the new python tracking interface. This PT fixes that.


swhite2401 (approved):
Sorry about that....

lfarv:
My fault: I pushed to change the function names…

Issue #649, opened 2023-08-29
Beam Moments passmethod causes core dump


lcarver:
Hello,

Lets take a ring (for example full S28F.mat), and use it to generate a fast_ring (called fring).

If I do: 

```
bm = at.BeamMoments('bm')
fring.append(bm)
_ = at.lattice_pass(fring)
```

then everything works fine. But if I do

```
bm = at.BeamMoments('bm')
fring.append(bm)
sig = at.sigma_matrix(ring)
```
I get a core dump. 

@swhite2401 believes the issue comes from the recent lattice_track interface. For now, I can avoid the problem by ensuring that the beam moments is appended to the fring right before I track. He is looking into it.

swhite2401:
Yes I have found the problem, I am on it.

lcarver:
Fixed in #650

PR #650, opened 2023-08-29
Remove collective passes from internal lattice_pass

swhite2401:
As pointed out in #649 wrongly initialized BeamMoments elements in the internal lattice_pass function was causing a segmentation fault. This PR disables all BeamMoments and Collective passmethods in the internal lattice_pass: this solves the issue and provides more efficient tracking for the pyAT functions for which these collective elements are not relevant.


swhite2401:
@lcarver this should fix your problem

lcarver:
my script that was not working, is now working.
So thanks for that.

lcarver (approved):

pyat/at/tracking/utils.py
```diff
@@ -25,6 +27,18 @@ def _set_beam_monitors(ring: Sequence[Element], nbunch: int, nturns: int):
...
         m.set_buffers(nturns, nbunch)
     return len(monitors) == 0
+    
+    
+def disable_collective(ring):
+    "Function to disable collective effects elements"
+    refs = numpy.zeros(len(ring), dtype=bool)
+    for elmt in _DISABLE_ELEMS:
+        refe = get_bool_index(ring, elmt, endpoint=False)
+        refs = refs | refe
```
lfarv, line 37: Can be shortened as:
```python
refs = ring.get_bool_index(at.checktype(_DISABLE_ELEMS))
```

lfarv:
@swhite2401 : That looks OK, but can you explicit what means "wrongly initialized BeamMoments", and how it could end up in a segmentation fault? Just to understand the root of the problem…

swhite2401:
So BeamMoment stores the beam center of masses and sizes as a function of turn number, in the the previous implementation it was initialized in lattice_pass with the correct number of turns.
In the case of internal_lpass this initialization was not done and the c code was trying to access a buffer with the wrong size.

Probably this could be handled in the C, although I am not quite sure what would be the consequence of changing the C buffer size of an existing numpy array on the fly... may you have so,e ideas on this point?

Concerning your second comment, if you look at it the changes are made on copies, so the lattice is changed only internally and there is therefore no need to re-enable it. Generalizing this disabling to all random processes is certainly something one could think of, I can temporarily add the QuantdiffElem (and others) to the list in this implementation, this is trivial. Acting on the element class is also possible but more complicated, not sure if there is any real added value to be honest... but I can do it if you think this is something needed.

lfarv:
> Acting on the element class is also possible but more complicated, not sure if there is any real added value to be honest... but I can do it if you think this is something needed.

No, I think you are right, as I understand it now, no added value.

swhite2401:
I have implemented the suggested modification:
-add all variable elements to the list of elements to disable
-rename a function and variable to make them more understandable
-disable keep_lattice flag in case variable elements are present and not disabled

lfarv:
@swhite2401 : On top of that, there is a absolute protection to be set on the C part: you can get the dimensions of the BeamMoments attributes using `atGetOptionalDoubleSz` instead of `atGetOptionalDouble`, and then make sure that you never write more than there buffer size. Whatever comes in, it's safe.

lfarv:
> -add all variable elements to the list of elements to disable

Remains the case of the `*QuantPass` elements (dipoles). For that, identifying the "variable" elements at the `Element` level becomes necessary. I think that's out of the scope of this PR (it does not crash anything).

So for me it's OK as it is, but I push for the protection in the `BeamMomentsPass.c`: it's the only completely safe solution. About that, there is some improvement to be done in `atelem.c`: it the attribute is absent, the returned pointer is NULL, but in additions the sizes should be set to 0, which is not done…

swhite2401:
How do I set the size of a NULL pointer? Can I use this:

`atGetArrayData(NULL, NULL, NPY_DOUBLE, 1, 1);`

lfarv:
In `atelem.c`, I would add 2 lines:

```C
static double *atGetOptionalDoubleArraySz(const PyObject *element, char *name, int *msz, int *nsz)
{
    PyArrayObject *array = (PyArrayObject *) PyObject_GetAttrString((PyObject *)element, name);
    if (array == NULL) {
        PyErr_Clear();
        *msz=0;           /*     <- new line     */
        *nsz=0;           /*     <- new line     */
        return NULL;
    }
    return (double *) atGetArrayData(array, name, NPY_DOUBLE, msz, nsz);
}
```
This avoids leaving `msz` and `nsz` uninitialised if the attribute is missing. For the calling function (pass method), checking that the buffer size (`msz`) is larger than the data to be written is sufficient to allow writing.

The interface deals only with the 1st two dimensions, but it's enough for what we need.

swhite2401:
Ok clear thanks!

swhite2401:
This will not work the turn is on the 3rd dimension....

swhite2401:
Ok so for the moment I implemented an in-between solution:
-In BeamMoments I initialize the nbunch to zero instead of 1
-In the passmethod I check the second dimension that should always be >=0 if the element buffers have been initialized

In principle this test should be extended to the third dimension but it is not possible with the present interface. It should be improved to array with n dimensions... but this is probably not for this PR.

I can put an issue as a reminder and once the interface has been improved update this passmethod to check both nbunch and nturns.

swhite2401:
Or in BeamMoment I change the order to (nturns, nbunch, 6)... I think only @lcarver is using this element and just only recently so in principle the impact is not enormous...
This is probably the simpler solution

swhite2401:
I can even make such that the read_only value return arrays with the present order in fact

swhite2401:
The 6 is always there it is not accessible for the users unless he starts to alter private variables.
I am also away for one week...I'll try to close this today but we'll see.

Other option is to store _nbunch and _nturns in element attributes at initialization and when calling set_buffer...

the generalized interface is indeed the best option an can be re-used for other cases...

swhite2401:
Hello, I finally found time to progress on that one, in the end I added a function in atelem.c to check the array shape.
This may not be ideal, but to do better I am afraid we would have to rewrite the interface and make it general which is out of the scope of this PR.
Let me know what you think, this is the cleanest I could think of to check all dimensions.

lfarv:
>  I added a function in atelem.c to check the array shape

Very good idea. This works perfectly if the caller knows the size it needs (which is the case here), so it solves the problem.

- as it is, `atCheckArrayDims` does not return anything, so it should be declared as `void atCheckArrayDims()` instead of `void *atCheckArrayDims()`.

- Here is a proposal for the Matlab version, necessary to pass the Matlab tests. It's not tested, but it compiles:
```Matlab
void atCheckArrayDims(const mxArray *ElemData, char *fieldname, int ndim, int *dims)
{
    const mwSize *dptr, *dlim;
    mwSize nd;

    mxArray *field=get_field(ElemData,fieldname);
    if (!field)
        mexErrMsgIdAndTxt("AT:WrongArg", "The required attribute %s is missing.", fieldname);
    nd=mxGetNumberOfDimensions(field);
    if (nd != ndim)
        mexErrMsgIdAndTxt("AT:WrongArg", "%s should have %d dimensions instead of %d.", fieldname, ndim, nd);
    dptr = mxGetDimensions(field);
    dlim = dptr+nd;
    while (dptr < dlim) {
        if (*dptr != *dims)
            mexErrMsgIdAndTxt("AT:WrongArg", "%s dimension error: %d / %d", fieldname, *dptr, *dims);
        dptr++;
        dims++;
    }
}
```

We'll see later for a general interface!

lfarv (approved):
OK !

swhite2401:
Hello @lfarv , I was trying to test this in matlab but I cannot create a struct field that starts with an underscore as requested by the element, do you have a way around this?

lfarv:
@swhite2401:
It's indeed a problem. The present solution is:

- in the C integrators, the access to the attributes of an element from Matlab is modified so that if the attribute name starts with an underscore (example `_rms`), it will look for `rms_`: a Matlab attribute with the underscore at the end.
- so when creating the element in Matlab, you have to put the underscore to the end instead of the beginning.

This works but the drawback is that such elements cannot be transferred through .mat files.

swhite2401:
Ok so few changes made to the matlab interface, in particular nbunch and num_turns had to be set in atpass.c
There is nevertheless still an issue in matlab: I could not find a way to create an array with size (6,1,1) which is a problem for simulations with one bunch and one turn.

I do not see why would one want to do this but still the check will fails with unclear messages

swhite2401:
To be honest this function should not be used with matlab, nbunch is not even defined in matlab

lfarv (approved):
OK

PR #654, opened 2023-09-06
Bug fix in atdisable_6d: keep the Energy field in cavities.

lfarv:
`atdisable_6d` removes the `Energy` field in all elements, including RF cavities. The `Energy` field is not used in cavity processing, however it is taken as the definition of the ring energy if there is no `RingParam` element. So this creates problems for rings without `RinpParam` element.

In this PR, the `Energy` field is kept only in RF cavities.


simoneliuzzo:
I have tested `atenable_6d` and `atdisable_6d` with and without ringparam element.
the command `atenergy` works. `atlinopt`/`atlinopt6` also work.

simoneliuzzo (approved):

Issue #647, opened 2023-08-28
Naming of the keyword arguments of simple_ring


lfarv:
The recent `simple_ring` function has some arguments in common with the existing `sigma_matrix` function. However, their names are slightly different. Apart from the mnemonic inconvenience, this prevents from using the same dictionary as input for both functions (I encountered the problem when trying to compare the "design" &Sigma;-matrix with the output of tracking a population through `simple_ring`: I need both functions).

So I would like to rename the `simple_ring` arguments coherently with those of `sigma_matrix`:
`beta_x` -> `betax`
`alpha_x` -> `alphax`
`emit_x` -> `emitx`
`sigma_dp` -> `espread`
For the last one, strictly speaking, it's indeed a momentum spread (not energy), so we could envisage adding an alias in `sigma_matrix` (`mspread` or even keeping `sigma_dp`). But I would not like to alias all the arguments…

Next: `simple_ring` could accept the same input as `sigma_matrix`:
- `twiss_in`, using `R` if present (this can define a coupled lattice), or `alpha` and `beta` (for uncoupled lattices),
- otherwise `alphax`, … as now
- one could even accept a &Sigma;-matrix, which can replace `alpha*`, `beta*`, `emit*`, `sigma_dp*`.

Third: we miss the Matlab simple ring: `atsimplering`

**Otherwise, simple_ring worked nicely in all my tests!**

lfarv:
Anybody willing to do something about that ? Or shall I do it ?

swhite2401:
I think @lcarver said he would do it, but most likely not for matlab...

lfarv:
> This is not clear to me.

The &Sigma;-matrix (beam matrix as returned by `ohmienvelope`, for instance) contains all the optics and emittance information. So, with &Sigma; and tunes, you can get the M<sub>66</sub> matrix and the emittances. Adding the damping times, you have everything.

Here, since the longitudinal motion is treated apart, I think we should use a 4D `twiss_in` input, generate a 4x4 transfer map and add the 44, 45, 54 and 55 terms as you do it now.

So to start from the R matrix in `twiss_in`, the easiest is first to build &Sigma; = &epsilon;<sub>1</sub>.R<sub>1</sub> + &epsilon;<sub>2</sub>.R<sub>2</sub>, then use `a_matrix` to get the A matrix, use the tunes to generate the rotation matrix S and finally build M<sub>44</sub> = A.S.A<sup>-1</sup>

[...]

PR #655, opened 2023-09-15
Fix attribute names in Simple Ring

lcarver:
This PR addresses the main comments from @lfarv in #647 . However I would like to discuss a bit the other comments.

I don't feel that there is any added value allowing a twiss_in input for simple_ring. In order to generate a twiss_in style input, you need a ring to begin with. In this case it is much easier to use fast_ring to generate your new ring and not bother using simple_ring.

I cannot really envisage a scenario where I might have only the output of some tracking results (a sigma matrix) and then I have to find an equivalent lattice to give me these parameters (where I still have to provide tunes and other non-linear parameters).

### DIFF
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -991,58 +991,58 @@ class SimpleQuantDiff(_DictLongtMotion, Element):
     _BUILD_ATTRIBUTES = Element._BUILD_ATTRIBUTES
     default_pass = {False: 'IdentityPass', True: 'SimpleQuantDiffPass'}
 
-    def __init__(self, family_name: str, beta_x: Optional[float]=1.0,
-                 beta_y: Optional[float]=1.0, emit_x: Optional[float]=0.0,
-                 emit_y: Optional[float]=0.0, sigma_dp: Optional[float]=0.0,
-                 tau_x: Optional[float]=0.0, tau_y: Optional[float]=0.0,
-                 tau_z: Optional[float]=0.0, U0: Optional[float]=0.0,
+    def __init__(self, family_name: str, betax: Optional[float]=1.0,
+                 betay: Optional[float]=1.0, emitx: Optional[float]=0.0,
+                 emity: Optional[float]=0.0, espread: Optional[float]=0.0,
+                 taux: Optional[float]=0.0, tauy: Optional[float]=0.0,
+                 tauz: Optional[float]=0.0, U0: Optional[float]=0.0,
                  **kwargs):
         """
         Args:
             family_name:    Name of the element
             
         Optional Args:
-            beta_x:         Horizontal beta function at element [m]
-            beta_y:         Vertical beta function at element [m]
-            emit_x:         Horizontal equilibrium emittance [m.rad]
-            emit_y:         Vertical equilibrium emittance [m.rad]
-            sigma_dp:       Equilibrium energy spread
-            tau_x:          Horizontal damping time [turns]
-            tau_y:          Vertical damping time [turns]
-            tau_z:          Longitudinal damping time [turns]
+            betax:         Horizontal beta function at element [m]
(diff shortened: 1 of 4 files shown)


lfarv (approved):
> In order to generate a twiss_in style input, you need a ring to begin with.

Why? It's just a generalised way of defining the optics properties: you give all the parameters you want in a single `dict` object. And since the R matrices are a generalisation of the 2D alpha and beta parameters, you can even define a fully coupled beam.

> I cannot really envisage a scenario where I might have only the output of some tracking results (a sigma matrix)...

Same remark: a sigma matrix has nothing to do with tracking. It's the simplest way to describe the beam size and divergence you want (6D ellipsoid), without having to care about beta values.

Anyway, all that is not needed, my concern was about names, so it's OK for merging.
(I did not think of changing the names of the C variables, but why not…)

I'll look at the  Matlab version.

pyat/at/lattice/elements.py
```diff
@@ -991,58 +991,58 @@ class SimpleQuantDiff(_DictLongtMotion, Element):
...
-    def __init__(self, family_name: str, beta_x: Optional[float]=1.0,
-                 beta_y: Optional[float]=1.0, emit_x: Optional[float]=0.0,
-                 emit_y: Optional[float]=0.0, sigma_dp: Optional[float]=0.0,
-                 tau_x: Optional[float]=0.0, tau_y: Optional[float]=0.0,
-                 tau_z: Optional[float]=0.0, U0: Optional[float]=0.0,
+    def __init__(self, family_name: str, betax: Optional[float]=1.0,
+                 betay: Optional[float]=1.0, emitx: Optional[float]=0.0,
+                 emity: Optional[float]=0.0, espread: Optional[float]=0.0,
+                 taux: Optional[float]=0.0, tauy: Optional[float]=0.0,
+                 tauz: Optional[float]=0.0, U0: Optional[float]=0.0,
```
lfarv, line 998: For PEP8 compliance, it should be:
```python
    def __init__(self, family_name: str, betax: float = 1.0,
                 betay: float = 1.0, emitx: float = 0.0,
                 emity: float = 0.0, espread: float = 0.0,
                 taux: float = 0.0, tauy: float = 0.0,
                 tauz: float = 0.0, U0: float = 0.0,
```
No `Optional` (it's never `None`, always float), and white space around the `=` sign.

pyat/at/lattice/elements.py
```diff
@@ -991,58 +991,58 @@ class SimpleQuantDiff(_DictLongtMotion, Element):
...
+            tauz:          Longitudinal damping time [turns]
             U0:             Energy Loss [eV]
             
         Default PassMethod: ``SimpleQuantDiffPass``
        """
         kwargs.setdefault('PassMethod', self.default_pass[True])
        
-        assert tau_x>=0.0, 'tau_x must be greater than or equal to 0'
-        self.tau_x = tau_x
+        assert taux>=0.0, 'taux must be greater than or equal to 0'
```
lfarv, line 1019: PEP8 again:
```python
assert taux >= 0.0, 'taux must be greater than or equal to 0'
```
Same for the next ones

pyat/at/lattice/elements.py
lcarver: This is taken into account. Also the same thing in simple_ring. The Optional was not necessary

pyat/at/lattice/elements.py
lcarver: Done

lcarver:
Thanks for your comments. I made the changes you suggest. I agree we can do the simple_ring modifications with twiss in and sigma_matrix at a later date. 

Ready for merge, just needs reapproval.
I will move onto your matlab branch now.
Cheers,

lfarv (approved):
All OK for me!

PR #657, opened 2023-09-17
New Matlab function atsimplering

lfarv:
The new function `atsimplering` creates a simple lattice according to a set of ring properties. A "simple ring" consists of:
- A RF cavity
- a 6x6 linear transfer map
- a detuning and chromaticity element
- a simplified quantum diffusion element which provides the equilibrium emittance, radiation damping and energy loss

It is defined by:
- the nominal energy [eV],
- the ring circumference [m],
- the harmonic number,
- the horizontal tune,
- the vertical tune,
- the RF voltage [V],
- the momentum compaction factor.

plus optional parameters:
- either vertical and horizontal alpha, beta, emittance, or sigma matrix,
- horizontal, vertical and longitudinal damping times,
- energy loss per turn,
- particle,
- lattice name.

It is similar to the python `simplering` introduced in #643


simoneliuzzo:
Dear @lfarv @lcarver and @swhite2401 ,

what is the difference with atfastring? The imputs are parameters rather than an AT lattice? 

I will try to test it

thank you,

best regards
Simone

lcarver:
The idea is to be able to generate a ring with only a table of 
parameters. No initial lattice needed.

On 2023-09-18 07:18, simone liuzzo wrote:

> Dear @lfarv [1] @lcarver [2] and @swhite2401 [3] ,
> 
> what is the difference with atfastring? The imputs are parameters 
> rather than an AT lattice?
> 
> I will try to test it
> 
> thank you,
> 
> best regards
> Simone
> 
> --
> Reply to this email directly, view it on GitHub [4], or unsubscribe 
> [5].
> You are receiving this because you were mentioned.Message ID: 
> ***@***.***>



Links:
------
[1] https://github.com/lfarv
[2] https://github.com/lcarver
[3] https://github.com/swhite2401
[4] https://github.com/atcollab/at/pull/657#issuecomment-1722769089
[5]

simoneliuzzo:
Dear @lcarver @lfarv and @swhite2401,

if the output is ought to be the same, why having two functions rather than simply adding inputs to `atfastring`?
 
For example: 
if you give a lattice it will use the lattice. (atfastring)
if you give the parameters it will use the parameters. (atsimplering)
if you give a partial amount of parameters it will use the lattice to get the defaults for the others. (atfastsimplering)

best regards
Simone

simoneliuzzo:
Dear @lfarv @lcarver and @swhite2401,

do you have a test function checking that the output elements are strictly the same for atfastring(ring) and atsimplering(parameters_of_ring) ?

lfarv:
> if the output is ought to be the same, why having two functions rather than simply adding inputs to atfastring?

@simoneliuzzo 
What you propose could be done, though I have doubts on the interest of the "mixed" input. But I'm not pushing for that because:
- generally speaking, I prefer having different functions for different tasks, rather than trying to mix things with ambiguous inputs. Parsing the different possible ways of giving the simplering parameters is already tricky,
- here, the purpose of the two functions is really different: modelling an existing lattice for `fastring`, creating a new lattice from scratch for `simplering`,
- if you look in detail, you'll see that the generated lattices are somewhat different: in `fastring` the damping is included in the linear 6x6 matrix, in `simplering` it's in the quantdiff element, the quantum diffusion is also treated differently. It's not the ideal solution, it could be possible for `fastring` to generate the same elements than `simplering` but not the other way round. So unifying the output needs a modification of `fastring`. Possible, but it's another topic.

The test you propose is a good idea. But the comparison is not easy because the generated lattice are different, though they should behave the same…

atmat/lattice/atsimplering.m
```diff
@@ -0,0 +1,124 @@
...
+[tauz,varargs]=getoption(varargs,'tauz',0.0);
+[sigma,varargs]=getoption(varargs,'sigma',[]);
+[espread,varargs]=getoption(varargs,'Espread',0.0);
+if isempty(sigma)
+    [alphax,varargs]=getoption(varargs,'alphax',0.0);
+    [betax,varargs]=getoption(varargs,'betax',1.0);
+    [alphay,varargs]=getoption(varargs,'alphay',0.0);
+    [betay,varargs]=getoption(varargs,'betay',1.0);
+    [emitx,varargs]=getoption(varargs,'emitx',0.0);
+    [emity,varargs]=getoption(varargs,'emity',0.0);
```
lcarver, line 71: if I put in only the essential arguments for atsimplering, I get an unstable ring because the defaults here for emitx and emity are 0.0.  It gives an error at line 80.

atmat/lattice/atsimplering.m
```diff
@@ -0,0 +1,124 @@
...
+%
+%RING=ATSIMPLERING(...,'emitx',emitx,'emity',emity,...)
+%       specify the equilibrium emittance. Default: 0, meaning no quantum
+%       diffusion
+%
+%RING=ATSIMPLERING(...,'taux',taux,'tauy',tauy,'tauz',tauz,...)
+%       specify the damping times. Default: 0, meaning no damping
+%
+%RING=ATSIMPLERING(...,'U0',U0,...)
+%       specify the emergy loss per turn [eV]. Default: 0
```
lcarver, line 50: You mention here that the default for U0 is 0, but there is no getoption for it below. I don't see wher eit is passed to SimpleQuantDiff

atmat/lattice/element_creation/atSimpleQuantDiff.m
```diff
@@ -0,0 +1,50 @@
...
+[betay,rsrc]      = getoption(rsrc,'Betay',1.0);
+[emitx,rsrc]      = getoption(rsrc,'Emitx',0.0);
+[emity,rsrc]      = getoption(rsrc,'Emity',0.0);
+[espread,rsrc]    = getoption(rsrc,'Espread',0.0);
+[taux,rsrc]       = getoption(rsrc,'Taux',0.0);
+[tauy,rsrc]       = getoption(rsrc,'Tauy',0.0);
+[tauz,rsrc]       = getoption(rsrc,'Tauz',0.0);
+[U0,rsrc]         = getoption(rsrc,'U0',0.0);
+elem=atbaselem(fname,method,'Class',cl,'Betax',betax,'Betay',betay,...
+    'Emitx',emitx,'emity',emity,'Espread',espread,...
```
lcarver, line 47: 'emity' should be 'Emity'

atmat/lattice/atsimplering.m
```diff
@@ -0,0 +1,124 @@
...
+[alphax,betax]=ab(am(1,1),am(2,1),am(2,2));
+[alphay,betay]=ab(am(3,3),am(4,3),am(4,4));
+emit=real(-1i*lambda);
+
+m66=eye(6);
+m66(1:4,1:4)=am*rots/am;
+m66(6,5)=eta*circumference;
+
+rf_cavity=cellfun(@makerf,num2cell(vrf),num2cell(frf),num2cell(hnum),'UniformOutput',false);
+lin_elem=atM66('Linear',m66,'Length',circumference);
```
lcarver, line 91: In the python version, I added a lenth to the M66 in order to make it such that len(ring) still gave the good value of the circumference.

In Matlab you have this first element which contains all the ring information including length, is this length the length of the element (0) or the length of the ring (circumference).

atmat/lattice/element_creation/atSimpleQuantDiff.m
lfarv: Though attribute names in Matlab elements had traditionally a capital initial letter, I finally set everything uniform, lower case letters. At both ends, that's what the user should give to atsimplering, and that's what the integrator looks for.

atmat/lattice/atsimplering.m
lfarv: Right, if an emittance is zero, the optics cannot be deduced from sigma. That's corrected

lcarver:
Looks good to me. Thanks a lot

lcarver (approved):

lfarv:
@lcarver : I had to solve a conflict with the updated master. Could you approve again? Thanks

@simoneliuzzo : any more comments on this function?

lcarver (approved):

lfarv:
After solving a conflict with the master branch (Matlab version number), I need another review.
Can @lcarver or @simoneliuzzo approve again?

lcarver (approved):

PR #658, opened 2023-09-21
Optimize atimplib

swhite2401:
This PR proposes a couple optimization/simplification of the C library for collective effects, a 10% speed-up is obtained in tracking simulations.


lcarver:
While I have the numbers and plots to hand. Here is the new branch MPI comparison.

image

Some numbers:
master branch, Nslices=300, Nparts_per_slice=2000, Nturns=10k
1 core = 1318.8
60 cores = 30.4

optimize_atimplib, Nslice=300, Nparts_per_slice=2000, Nturns=10k
1 core = 1317
60 cores = 28.6

lfarv:
@swhite2401 : while you are on that, on line 366:
```C
int i,ib,it,is;
```
Could you remove the `it` variable which is not used but generates warnings in compilation ?

swhite2401:
Hello, so this is a bit more than just optimization in the end:
-I have found a bug when changing Nslice or Nturns on the fly for collective effects elements, this is no fixed and I have added a dimension check on the turnhistory to prevent memory errors
-Minor changes relating the warning from the windows compiler
-Simplification a of weigthed average calculation of vbeam in the beam loading passmethods

I would suggest to double check this thoroughly. On my side everything looks OK but I haven't used these features as extensively as you.

lcarver:
Hello,
I run a single bunch case and compare vbeam with analytical.
But for BLMode=Wake the single bunch at high current gets destroyed and particles are lost. 
On the master the same script runs and gives good agreement.

For PHASOR mode the same values run nicely, so not a physics problem.

lcarver:
for BLMode.WAKE, the Vgen is continuously decreasing. The Vbeam computation is good but the Vgen loop is broken somehow. I continue to investigate

atintegrators/atimplib.c
```diff
@@ -352,37 +331,29 @@ static void compute_kicks_longres(int nslice,int nbunch,int nturns, double *turn
...
     MPI_Allreduce(MPI_IN_PLACE,vbi,nbunch,MPI_DOUBLE,MPI_SUM,MPI_COMM_WORLD);
+    MPI_Allreduce(MPI_IN_PLACE,&totalW,1,MPI_DOUBLE,MPI_SUM,MPI_COMM_WORLD);
+    MPI_Allreduce(MPI_IN_PLACE,totalWb,nbunch,MPI_DOUBLE,MPI_SUM,MPI_COMM_WORLD);
     MPI_Barrier(MPI_COMM_WORLD);
     #endif
 
-    vba = sqrt(vbeamk[0]*vbeamk[0]+vbeamk[1]*vbeamk[1]);   
-    vbp = atan2(vbeamk[1],vbeamk[0]);
+    vba = sqrt(vbeamk[0]*vbeamk[0]+vbeamk[1]*vbeamk[1])/totalW;
+    vbp = atan2(vbeamk[1],vbeamk[0])/totalW;
```
lcarver, line 346: here is the bug.

In the old script, I just divided vbeamk[0] and vbeamk[1] by totalW which is fine as it is real and imaginary voltage.

Line 345 is fine, but line 346 should not have the /totalW as anyway atan2 is  ratio so makes no sense

swhite2401: Good catch in fact the phase should not be normalized

lcarver:
All example files are running nicely. I am now launching MWT and TMCI (but I suspect they will be fine due to the fact that the turnhistory modifications only really impact multibunch)

lcarver:
While testing I found an unrelated bug in wake_elements.py 
```
    def set_normfactxy(self, ring):
        l0, _, _ = ring.get_optics(ring)
        self.NormFact[0] = 1/l0['beta'][0]
        self.NormFact[1] = 1/l0['beta'][1]
```

`ring.get_optics(ring)` is wrong, it should be either 

`ring.get_optics()`

or

`at.get_optics(ring)`

lcarver:
TMCI and MWT both fine.

lcarver (approved):

swhite2401:
> While testing I found an unrelated bug in wake_elements.py
> 
> ```
>     def set_normfactxy(self, ring):
>         l0, _, _ = ring.get_optics(ring)
>         self.NormFact[0] = 1/l0['beta'][0]
>         self.NormFact[1] = 1/l0['beta'][1]
> ```
> 
> `ring.get_optics(ring)` is wrong, it should be either
> 
> `ring.get_optics()`
> 
> or
> 
> `at.get_optics(ring)`

Fixed

swhite2401:
@lcarver you need to re-approve before I can merge

lcarver (approved):

PR #659, opened 2023-09-21
Update of the build process

lfarv:
Main changes:

- the `numpy` dependencies are simplified in `pyproject.toml` by using the package [oldest-supported-numpy](https://github.com/scipy/oldest-supported-numpy) which selects for each python version, platform and architecture the correct `numpy` version used for compilation.
- As a consequence, the substitution `githubproject.toml` -> `pyproject.toml` used in building wheels is no more necessary, which makes building wheels simpler
- A few GitHub actions have been updated to new versions, avoiding the warnings like "The following actions uses node12 which is deprecated and will be forced to run on node16…"

The production of documentation and wheels has been tested.


swhite2401:
Hello, since we are looking at the build I have noticed that pip requires internet access to build and install AT. Do you know any workaround? This would be very useful if it exists

pyproject.toml
```diff
@@ -1,12 +1,8 @@
 [build-system]
 requires = [
-    "numpy >= 1.16.6;python_version<'3.8'",
```
MJGaughran, line 3: This requirement is changed by the adoption of `oldest-supported-numpy`. Unfortunately, I am still stuck with Python 3.7 for the moment.

Perhaps there's an alternative way of doing this that lets us continue to use `oldest-supported-numpy `for the remaining Python versions?

MJGaughran: I have to admit I can't remember why we have 1.16.6 as the build requirement. Did you go through the previous issues / PRs to confirm this was appropriate?

lfarv:
@swhite2401:
When building from source, `pip` works in a separate environment. That means that `pip` has to fill an empty environment with the **build** dependencies (`setuptools`, `numpy` and `wheels`). On the 1st build, it will download them. On the next runs, `pip` maintains a cache with all its downloads, so internet access should not be necessary. Then for installation, pip checks if the already existing packages fullfil the **install** dependencies (`scipy` and more). If not, again it will download them on the 1st build.

So `pip` needs internet of the 1st install. Normally not on next ones, but maybe it does other things I ignore.

Without internet access, the best way is to install from binaries, after copying all the necessary wheels (AT and its install dependencies) dowloaded or created on another machine with Internet access.

pyproject.toml
lfarv: The dependencies for `oldest-supported-numpy` are roughly the same that the old ones (at least for "usual" platforms), except for 3.7. `oldest-supported-numpy` specifies 1.14.5 instead of 1.16.6.

1.16.6 comes from the fact that when I looked for supported packages, I did not look before the install dependency (1.16). As it ignores the install dependency, `oldest-supported-numpy` goes back to the real oldest compatible.

Practically, this does not change anything and I checked that AT builds and runs perfectly on python 3.7. I remind that numpy 1.14.5 is used only for building and is not related to the version used for running AT.

The full explanation is given [here](https://numpy.org/doc/1.24/dev/depending_on_numpy.html). In particular:

 _"…you must build your own binaries (wheels or other package formats) against the lowest NumPy version that you support (or an even older version)_

We are now in the case of _an older version_ !

MJGaughran: Good, that's seems fine, then.

.github/workflows/build-python-wheels.yml
```diff
@@ -51,7 +41,7 @@ jobs:
     runs-on: ubuntu-20.04
```
MJGaughran, line 41: Was this version supposed to be updated too? Or is it held back for e.g. GLIBC errors.

On that note, I just want a sanity check that none of these updates could introduce GLIBC errors.

lfarv: This line is used for building the source distribution, so it cannot have any effect on GLIBC errors.

On line 16, the one for building binaries, I updated it because at some point, it did not build python 3.10 and 3.11 and I thought that this was the problem. In fact, my mistake was somewhere else (`pyproject.toml`). So one could revert to 20.04 if you think it's better.

I have no idea of how we can deal with GLIBC problems: is it as for `numpy` by using the lowest version ? The only question is on line 16, and I let you decide what is the best ubuntu version to put there.

swhite2401:
> @swhite2401: When building from source, `pip` works in a separate environment. That means that `pip` has to fill an empty environment with the **build** dependencies (`setuptools`, `numpy` and `wheels`). On the 1st build, it will download them. On the next runs, `pip` maintains a cache with all its downloads, so internet access should not be necessary. Then for installation, pip checks if the already existing packages fullfil the **install** dependencies (`scipy` and more). If not, again it will download them on the 1st build.
> 
> So `pip` needs internet of the 1st install. Normally not on next ones, but maybe it does other things I ignore.
> 
> Without internet access, the best way is to install from binaries, after copying all the necessary wheels (AT and its install dependencies) dowloaded or created on another machine with Internet access.

Well to be more precise I was working on the C in the train with no internet, so I have to rebuilt several times but could not because pip was looking for the dependencies. I tried the --no-deps flag without success.
Anyway, this is very specific and not so important, I just wanted to know whether there existed easy solutions for that... but it seems not

.github/workflows/build-python-wheels.yml
MJGaughran: I would prefer to leave these OS versions as they were before to avoid any issues. When we have made further progress on our own internal infrastructure, we will hopefully be a bit more flexible.

Ubuntu 20 has full support until 2025, and LTS until 2030, so unless there are missing features it is best to be conservative.

lfarv: > I would prefer to leave these OS versions as they were before to avoid any issues

OK: Ubuntu version for building wheels is back to 20.04

> Ubuntu 20 has full support until 2025, and LTS until 2030,

The question is rather: how long will GitHub keep a runner with this version. It looks they keep 2 versions, so 20.04 will likely disappear when 24.04 is released. That still gives us some time…

MJGaughran: Good point, it should last until Dec '24.

lfarv:
Is there any other remark on this? Otherwise could someone approve?

swhite2401 (approved):

MJGaughran (approved):

PR #661, opened 2023-09-22
Fix bug sigma matrix

lcarver:
Little bug with passing arguments to r_4d


lcarver:
This was a mess. I will redo it cleanly.

PR #664, opened 2023-09-26
Collective bugfix

lcarver:
Hello,

Sneaky bug was introduced with the new tracking interface. Unfold beam default was wrongly set, this meant that for the last month, the bunch_spos that was passed to the PassMethod was an array of zeros. All multibunch simulations completely wrong.

I also take this opportunity to fix two other minor things.

1. removal of the couple of the array conversions in BeamLoadingElement. It was squashing the arrays to a 1d array instead of keeping the good shape. (TurnHistory and Vbunch).
2. BeamLoadingElement did not inherit from WakeElement, which is fine but there were a few features that were not taken over. I stole a couple of properties to do with Nslice / Nturns and TurnHistory and copied them onto the BeamLoadingElement.

### DIFF
--- a/pyat/at/collective/beam_loading.py
+++ b/pyat/at/collective/beam_loading.py
@@ -112,8 +112,6 @@ class BeamLoadingElement(RFCavity, Collective):
                         _beta=float, _wakefact=float, _nslice=int,
                         ZCuts=lambda v: _array(v),
                         _nturns=int, _phis=float,
-                        _turnhistory=lambda v: _array(v),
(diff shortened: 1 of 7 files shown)


lcarver:
Just to add, I have rerun many of my previous scripts from several months ago and I am getting excellent agreement once more.

swhite2401 (approved):
If you have tested that everything is ok, you can merge. This is a terrible bug thanks to @TeresiaOlsson for spotting this out and sorry about the wasted time...

swhite2401 (approved):

PR #667, opened 2023-10-05
Bug fix in atgeometry

lfarv:
Fixes a bug in `atgeometry` reported by @simoneliuzzo.

The `centered` flag now works as expected.

However, the computation of the radius is made using the 1st and last points. For a full ring, these points almost coincide, which makes the computation very imprecise. This happens for both Matlab and python.

For better results, the computation of the radius should be completely rewritten by taking all points into account (noting that this "polygonal" radius is not a constant…). This if left for another pull request.

### DIFF
--- a/atmat/lattice/survey/atgeometry.m
+++ b/atmat/lattice/survey/atgeometry.m
@@ -20,7 +20,7 @@ function  [posdata,radius] = atgeometry(ring,varargin)
 %       a scalar offset value is equivalent to [0 OFFSET]
 %
 %POSDATA=ATGEOMETRY(...,'centered')
-%       The offset is set as [0 RADIUS 0]
+%       The offset is set as [0 RADIUS]
 %
 %POSDATA=ATGEOMETRY(...,'Hangle',h_angle)
 %       Set the initial trajectory angle
@@ -28,11 +28,12 @@ function  [posdata,radius] = atgeometry(ring,varargin)
 %
 %See also: ATGEOMETRY3
 
-[thetac,args]=getoption(varargin,'Hangle',0);
-centered=getflag(args,'centered');
-[refpts,offset]=parseargs({1:length(ring)+1,[0 0]},args);
+[centered,args]=getflag(varargin,'centered');
+[theta0,args]=getoption(args,'Hangle',0);
+[refpts,offset]=getargs(args,1:length(ring)+1,[0 0]);
 xc=0;
 yc=0;
+thetac=theta0;
 if isscalar(offset)
     yc=offset;
 else
@@ -40,9 +41,10 @@ else
     yc=offset(2);
 end
 [xx,yy,txy,lg,tr]=cellfun(@incr,[{struct()};ring]);    % Add a dummy element to get the origin
-radius=-(yc+xc/tan(thetac));
+radius=-(yc+xc/tan(thetac-theta0));
 if centered
-    yy=yy+radius;
+    xx=xx-radius*sin(theta0);
+    yy=yy+radius*cos(theta0);
 end
 posdata=struct('x',num2cell(xx(refpts)),'y',num2cell(yy(refpts)),...
     'angle',num2cell(txy(refpts)),'long',num2cell(lg(refpts)),...
--- a/atmat/lattice/survey/atgeometry.m
+++ b/atmat/lattice/survey/atgeometry.m
@@ -13,14 +13,15 @@ function  [posdata,radius] = atgeometry(ring,varargin)
 %
 %[POSDATA,RADIUS]=ATGEOMETRY(RING,REFPTS)
 %       Outputs the machine radius at the beginning of the lattice.
-%       Meaningful if RING is a cell of a periodic lattice.
-%
-%POSDATA=ATGEOMETRY(RING,REFPTS,OFFSET)
-%       Start at x=offset(1), y=offset(2)
-%       a scalar offset value is equivalent to [0 OFFSET]
+%       Note: this is different from the radius usually defined as
+%       circumference/2/pi
 %
 %POSDATA=ATGEOMETRY(...,'centered')
-%       The offset is set as [0 RADIUS]
+%       The offset is set so that the origin is at the centre of the ring
+%
+%POSDATA=ATGEOMETRY(RING,REFPTS,OFFSET)
+%       Start at x=offset(1), y=offset(2). Ignored if 'centered' is set.
+%       A scalar offset value is equivalent to [0 OFFSET].
 %
 %POSDATA=ATGEOMETRY(...,'Hangle',h_angle)
 %       Set the initial trajectory angle
@@ -28,6 +29,7 @@ function  [posdata,radius] = atgeometry(ring,varargin)
 %
 %See also: ATGEOMETRY3
 
+epsil=1.e-3;
 [centered,args]=getflag(varargin,'centered');
 [theta0,args]=getoption(args,'Hangle',0);
 [refpts,offset]=getargs(args,1:length(ring)+1,[0 0]);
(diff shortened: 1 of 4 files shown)


simoneliuzzo:
The centered option seems to work a bit better, but not yet as expected.

Screenshot 2023-10-09 at 09 02 47

atmat/lattice/survey/atgeometry.m
```diff
@@ -20,29 +20,31 @@
...
 else
     xc=offset(1);
     yc=offset(2);
 end
 [xx,yy,txy,lg,tr]=cellfun(@incr,[{struct()};ring]);    % Add a dummy element to get the origin
-radius=-(yc+xc/tan(thetac));
+radius=-(yc+xc/tan(thetac-theta0));
 if centered
-    yy=yy+radius;
+    xx=xx-radius*sin(theta0);
```
lcarver, line 46: maybe i'm missing something. But why not do
```
xcenter = (max(xx) + min(xx) )/2;
ycenter = (max(yy) + min(yy) )/2;
xx=xx-xcenter;
yy=yy-ycenter;
```

simoneliuzzo: @lcarver would work only for full SR.

simoneliuzzo:
Dear @lfarv,

the centered option seems to 'center' on ~2.5*10^5 rather than on ~140. 

may be you are already looking at this issue.

best regards
simone

lfarv:
@simoneliuzzo : see my previous answer: the present geometric computation cannot work for full lattices. So 

1. we stay as now: this bug fix gives at least a decent plot, and you can still use offset to put the center where you like,
2. we make a special case for full lattices with one of the methods given above.

simoneliuzzo:
Dear @lfarv,
I have memory of this feature working correctly. In pyAT for example it is working as expected. Could we apply the same strategy?  
regards
Simone

lfarv:
pyAT suffers from exactly the same problem !

As I already wrote, for a full ring, you cannot determine the center by looking at both ends.

The radius being defined as the intersection of the perpendiculars to both end of the lattice, for a full ring it goes to infinity. If because of rounding errors the ring does not close exactly, then you get any crazy value for the radius, it depends on your ring…

If we go for a special case for full rings, it will of course be applied to both Matlab and python.

swhite2401:
Since we are looking into this it would probably make sense to fix it here.

I could not find the proposed solutions mentioned in a previous message but since the radius is not constant I suppose we would have to compute some kind of average, was this the proposed solution? This could be quite heavy calculation no?

carmignani:
I think what @lcarver proposed is the best: the centre would be the average of max and min xx and yy positions.

lfarv:
My proposal: if the input is a cell of a periodic ring, the centre is non-ambiguous and the present solution is correct, we keep it. If fails in 2 cases: full ring and half-ring, when the perpendiculars to both ends coincide. So we need 2 special cases:
1. Full ring: theta-theta0 ≈ 2pi. There is no well-defined centre, we take either Lee's solution or take the average of x values and y values. This is good enough. 
2. Half ring: theta-theta0 ≈ pi. An approximate center is the middle of both ends.

I can't help until next week, so if anyone wants to deal with that, he's welcome !

lfarv:
Finally, centring now works in most cases, including full rings and half rings. The fix is applied to both Matlab and python.

It's now ready for merging. @lcarver and @simoneliuzzo, any comment?

swhite2401:
I looked at the python and this seems fine, I just have one cosmetic comment `_EPSIL` is defined a the top level while it will be used only locally in the function.
I can imagine that other function would need an epsilon in the future so defining it at the top level may not be the best.... for now this is fine anyway so it is up to you.

lfarv:
@swhite2401: defining the threshold globally makes it visible in case one wants to modify it, and makes it assigned once at import rather that each time the function is called.

To avoid the ambiguity that you pointed out, I renamed it to `_GEOMETRY_EPSIL`.

Waiting for approval ! @simoneliuzzo ?

simoneliuzzo:
The python version gives me this error:
```
/usr/local/bin/python3.9 /Users/liuzzo/PycharmProjects/testatgeometry/testgeom.py 
  File "/Users/liuzzo/PycharmProjects/testatgeometry/testgeom.py", line 1, in <module>
    import at
  File "/Users/liuzzo/beamdyn/at/pyat/at/__init__.py", line 3, in <module>
    from ._version import __version__, __version_tuple__
ModuleNotFoundError: No module named 'at._version'

```

I installed with these commands:
python3.9 -m venv atgeo
source atgeo/bin/activate
pip install --upgrade pip
pip install git+https://github.com/atcollab/at.git@fix_atgeometry
pip install matplotlib

I will try the matlab version now.

simoneliuzzo:
The matlab version gives a circle for centered but vertically displaced by 2.5 10^5 (as in the previous version)

Screenshot 2023-10-25 at 13 48 29


to install I did
```
git clone https://github.com/atcollab/at.git at_geom_25Oct
git checkout fix_atgeometry

>> restoredefaultpath;
>> addpath(genpath('./at_geom_25Oct/atmat'))
>> addpath(genpath('./at_geom_25Oct/atintegrators'))
>> atmexall;

```

lfarv:
@simoneliuzzo: Python test

I reproduced exactly your commands, and I get no problem (see record below). On which platform did you run?
If it still fails, you can try as you did for Matlab: clone, change branch, pull and build.
```
bijou:devtest $ python3.9 -m venv atgeo
bijou:devtest $ source atgeo/bin/activate
(atgeo) bijou:devtest $ pip install --upgrade pip
Requirement already satisfied: pip in ./atgeo/lib/python3.9/site-packages (23.2.1)
Collecting pip
  Obtaining dependency information for pip from https://files.pythonhosted.org/packages/47/6a/453160888fab7c6a432a6e25f8afe6256d0d9f2cbd25971021da6491d899/pip-23.3.1-py3-none-any.whl.metadata
  Using cached pip-23.3.1-py3-none-any.whl.metadata (3.5 kB)
Using cached pip-23.3.1-py3-none-any.whl (2.1 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 23.2.1
    Uninstalling pip-23.2.1:
      Successfully uninstalled pip-23.2.1
Successfully installed pip-23.3.1
(atgeo) bijou:devtest $ pip install git+https://github.com/atcollab/at.git@fix_atgeometry
Collecting git+https://github.com/atcollab/at.git@fix_atgeometry
  Cloning https://github.com/atcollab/at.git (to revision fix_atgeometry) to /private/var/folders/kc/97y2h1wj6sl5gp6cbqzgmyfc000105/T/pip-req-build-374vtx89
  Running command git clone --filter=blob:none --quiet https://github.com/atcollab/at.git /private/var/folders/kc/97y2h1wj6sl5gp6cbqzgmyfc000105/T/pip-req-build-374vtx89
  Running command git checkout -b fix_atgeometry --track origin/fix_atgeometry
  Switched to a new branch 'fix_atgeometry'
  branch 'fix_atgeometry' set up to track 'origin/fix_atgeometry'.
  Resolved https://github.com/atcollab/at.git to commit d843aa3b7a2d5646c83d5da4afded65b9aef28a3
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting numpy>=1.16.6 (from accelerator-toolbox==0.4.1.dev32+gd843aa3)
  Using cached numpy-1.26.1-cp39-cp39-macosx_10_9_x86_64.whl.metadata (61 kB)
Collecting scipy>=1.4.0 (from accelerator-toolbox==0.4.1.dev32+gd843aa3)
  Using cached scipy-1.11.3-cp39-cp39-macosx_10_9_x86_64.whl.metadata (60 kB)
Using cached numpy-1.26.1-cp39-cp39-macosx_10_9_x86_64.whl (20.6 MB)
Using cached scipy-1.11.3-cp39-cp39-macosx_10_9_x86_64.whl (37.3 MB)
Building wheels for collected packages: accelerator-toolbox
  Building wheel for accelerator-toolbox (pyproject.toml) ... done
  Created wheel for accelerator-toolbox: filename=accelerator_toolbox-0.4.1.dev32+gd843aa3-cp39-cp39-macosx_13_0_x86_64.whl size=1853834 sha256=ed4973c9c7434aca79975290785d0bd865d882277e30c46cb1215eb27e1a1e24
  Stored in directory: /private/var/folders/kc/97y2h1wj6sl5gp6cbqzgmyfc000105/T/pip-ephem-wheel-cache-i1epn219/wheels/a5/25/4b/a191292f1c614d4ab0955010de94974e161ba85ac34c6d43a8
Successfully built accelerator-toolbox
Installing collected packages: numpy, scipy, accelerator-toolbox
Successfully installed accelerator-toolbox-0.4.1.dev32+gd843aa3 numpy-1.26.1 scipy-1.11.3
(atgeo) bijou:devtest $ python
Python 3.9.18 (main, Aug 24 2023, 21:32:48) 
[Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> import at
>>> 
(atgeo) bijou:devtest $ 
```

lfarv:
@simoneliuzzo : I may have an explanation for you python problem: in the command
```
/usr/local/bin/python3.9 /Users/liuzzo/PycharmProjects/testatgeometry/testgeom.py 
```
you are not calling the python from your virtual environment, but the system one. AT is not installed there (or badly, it seems). While your virtual environment is activated, just call
```
python /Users/liuzzo/PycharmProjects/testatgeometry/testgeom.py 
```

simoneliuzzo:
contrary to atgetgeometry,  python get_geometry, does not accept refpts as input. it is usefull, sometimes, to get the coordinates of a single point or a few, along the lattice.

simoneliuzzo:
The lattice I use is may be an old one. 

/machfs/liuzzo/EBS/beamdyn/matlab/optics/sr/S28F_restart/betamodel.mat

lfarv:
@simoneliuzzo: even old lattices should work! I corrected the &theta; threshold detecting a full ring.
It now works with you lattice.
Thnaks for your tests, back to you.

simoneliuzzo:
Thanks @lfarv ,

also matlab is ok for centered option.

Screenshot 2023-10-25 at 22 08 51

simoneliuzzo (approved):

lfarv:
> contrary to atgetgeometry, python get_geometry, does not accept refpts as input. it is usefull, sometimes, to get the coordinates of a single point or a few, along the lattice.

That's done. @simoneliuzzo, could you approve again? Thanks

simoneliuzzo:
Checked. ok.
Blue is centered all elements
orange not centered at quads
green not centered at BPMs 
Screenshot 2023-10-26 at 12 10 44

simoneliuzzo (approved):

Issue #673, opened 2023-10-16
help of tracking method + start method handle


swhite2401:
Some of the documentation of patpass was not ported correctly to the new tracking interface. This has to be fixed.

There is an interest to provide a gobal way to change the start_method, this has to be added.

swhite2401:
Solved in #675

PR #675, opened 2023-10-17
Parallel lattice_track bugfix and help

swhite2401:
This answers #673 raised by @simoneliuzzo.
The arguments pool_size and start_method were not correctly documented for the new tracking interface.
These were also not correctly handled in the new interface, leading to possible errors when set while use_mp=False.
A variable was added to at.lattice.DConstant to set the start_method at the top level.


swhite2401:
As a side comment, the default 'spawn' is non fonctionnal on MAC architectures. This had already been discussed but is still an issue in my opinion.

simoneliuzzo (approved):
thank you for this change

lfarv:
> As a side comment, the default 'spawn' is non fonctionnal on MAC architectures. This had already been discussed but is still an issue in my opinion.

I have been working 90% on macOS for years, using 3 different machines, several macOS versions and python versions, and I have never been able to reproduce the problem. `patpass` is also part of the test sequence run on GitHub (with the default "spawn"), and as far as I know, it never failed.
So I think that the problem is in the installation of python and AT, or in some configuration. I suggest that if @simoneliuzzo can reproduce the problem, I may have a look at his setup.
Also, be aware that starting at python 3.13 (the next one), "spawn" will be the default on all platforms. This may be a sign that "fork" may disappear in some time. Better make sure that "spawn" works.

pyat/at/lattice/options.py
```diff
@@ -30,6 +30,7 @@ class _Dst(object):
     OrbMaxIter = 20          # Max. number of iterations for orbit
     omp_num_threads = int(os.environ.get('OMP_NUM_THREADS', '0'))
     patpass_poolsize = multiprocessing.cpu_count()
+    patpass_startmethod = None
```
lfarv, line 33: Could you document the new `patpass_startmethod` option in the docstring of `DConstant`, a few lines later?

swhite2401: Done

swhite2401:
> > As a side comment, the default 'spawn' is non fonctionnal on MAC architectures. This had already been discussed but is still an issue in my opinion.
> 
> I have been working 90% on macOS for years, using 3 different machines, several macOS versions and python versions, and I have never been able to reproduce the problem. `patpass` is also part of the test sequence run on GitHub (with the default "spawn"), and as far as I know, it never failed.
> 
> I also googled for problems with `spawn` on macOS, and could not find anything looking like what @simoneliuzzo described.
> 
> So I think that the problem is in the installation of python or AT, or in some configuration. I suggest that if @simoneliuzzo can reproduce the problem, I may have a look at his setup.
> 
> Also, be aware that starting at python 3.13 (the next one), "spawn" will be the default on all platforms. This may be a sign that "fork" may disappear in some time. Better make sure that "spawn" works!

Well I have tried the spawn method on a MAC at home and on Linux at ESRF and both failed, I am quite surprised that it works for you in fact... you are probably the only person that know for whom it works. 
It would be good to understand what you are doing differently.... this is not for this PR, I'll add an issue on this

lfarv (commented):
While you are here, there is still a typo in the docstring of lattice_track, output section:
```
trackdata: A dictionary containinf tracking data...
```

swhite2401:
> While you are here, there is still a typo in the docstring of lattice_track, output section:
> 
> ```
> trackdata: A dictionary containinf tracking data...
> ```

Done

lfarv (approved):

PR #688, opened 2023-11-03
perf: improve by a factor ~ 10 the performance

lnadolski:
key: using atgetfieldvalues with indices

Atsummary was very slow when calling ElementRadiation.

The main reason is the call of the function atgetfieldvalues for the full ring.
It is very slow if the full ring is used instead of the indices where the required field is present.

This a quick fix
% SLOW when elements do not have the 'BendingAngle'
% angle=atgetfieldvalues(ring,'BendingAngle');
% isdipole=isfinite(angle) & (angle~=0);

% Much faster with indices
dipoleIdx=findcells(ring,'BendingAngle');
angle=atgetfieldvalues(ring, dipoleIdx, 'BendingAngle');
isdipole = atgetcells(ring,'BendingAngle');
isdipole(dipoleIdx) = isfinite(angle) & (angle~=0);

This could be maybe improved evermore.
A a SOLEIL II lattice the gain to run atsummary is a factor 10 (1.5 s instead of 15 s).

**The 2 main modifications are when calling atgetfieldvalues for dipoles and quadrupole.**

The rest of the modifications are only style.

### DIFF
--- a/atmat/atphysics/Radiation/ElementRadiation.m
+++ b/atmat/atphysics/Radiation/ElementRadiation.m
@@ -10,20 +10,18 @@ function [I1, I2, I3, I4, I5, I6, Iv] = ElementRadiation(ring, lindata, varargin
 
 [quadon, ~] = getoption(varargin, 'UseQuadrupoles', true);
 
-% Much faster with indices
-dipoleIdx = findcells(ring, 'BendingAngle');
-angle = atgetfieldvalues(ring, dipoleIdx, 'BendingAngle');
+% Look for dipole elements with non zero angles
 isdipole = atgetcells(ring, 'BendingAngle');
-isdipole(dipoleIdx) = isfinite(angle) & (angle ~= 0);
+angle = atgetfieldvalues(ring, isdipole, 'BendingAngle');
+isdipole(isdipole) = (angle ~= 0);
 
+% Look for quadrupole elements with non zero focusing
 if quadon
     isquadrupole = atgetcells(ring, 'Class', 'Quadrupole');
-    quadIdx = findcells(ring, 'Class', 'Quadrupole');
-    pb = atgetfieldvalues(ring, quadIdx, 'PolynomB', {2});
-    isquadrupole(quadIdx) = (pb ~= 0);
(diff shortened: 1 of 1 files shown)


atmat/atphysics/Radiation/ElementRadiation.m
```diff
@@ -8,140 +8,162 @@
...
-[quadon, ~]=getoption(varargin,'UseQuadrupoles', true);
-angle=atgetfieldvalues(ring,'BendingAngle');
-isdipole=isfinite(angle) & (angle~=0);
+[quadon, ~] = getoption(varargin, 'UseQuadrupoles', true);
+
+% Much faster with indices
+dipoleIdx = findcells(ring, 'BendingAngle');
+angle = atgetfieldvalues(ring, dipoleIdx, 'BendingAngle');
+isdipole = atgetcells(ring, 'BendingAngle');
+isdipole(dipoleIdx) = isfinite(angle) & (angle ~= 0);
```
lfarv, line 17: `findcells` and `atgetcells` return the same thing (`findcells` calls `atgetcells`). The only difference is that `findcells` returns integer indices while `atgetcells` return logical indices. Both can be used here, but logical indexing is more efficient, according to Mathworks. So a single call is enough:

```Matlab
isdipole = atgetcells(ring, 'BendingAngle');
angle = atgetfieldvalues(ring, isdipole, 'BendingAngle');
isdipole(isdipole) = (angle ~= 0);

if quadon
    isquadrupole = atgetcells(ring, 'Class', 'Quadrupole');
    pb = atgetfieldvalues(ring, isquadrupole, 'PolynomB', {2});
    isquadrupole(isquadrupole) = (pb ~= 0);
```

lfarv:
@lnadolski
I get a factor 4 speedup for my EBS lattice: still surprisingly good.
Thanks for contributing!

atmat/atphysics/Radiation/ElementRadiation.m
lnadolski: This is even better

lnadolski (commented):
We can merge

lfarv (approved):
Up to you !

PR #689, opened 2023-11-05
Matlab release 2.5

lfarv:
A new Matlab release including more that 1 year of updates.

There is no code change in this PR, only documentation and configuration changes.

A new tag and release will be published as soon as this PR is approved. The distribution on Matlab Central will follow immediately.


lnadolski (changes requested):
I will update in the next commit 2 files for documentation.

lnadolski:
As described in https://github.com/atcollab/at/wiki

1/ Contents can be generated
2/ html documentation can once again be generated using using atm2html (part of MML) or to be installed from
https://github.com/gllmflndn/m2html
graph dependence using https://graphviz.org/

lnadolski:
Dear @lfarv,

All Contents are generated using atupdateContents.
Do you want to include them in the release or should they be added during the installation of AT following the wiki page?

lnadolski:
In the release not, could add the CorrectorPass bug fix.
- Misalignments and aperture are now considered

lfarv:
@lnadolski 

> In the release not, could add the CorrectorPass bug fix.

You are right. I'll do that manually. Most of release notes is generated automatically, using the **title** of each pull request. Unfortunately, the CorrectorPass bug fix was not mentioned in the title… This shows how the PR titles must be very carefully chosen, and why mixing several things in a single PR may lead to problems.

lfarv:
@lnadolski: I ignored the existence of the wiki (my fault, it's indeed referred in the [README](https://github.com/atcollab/at/blob/master/README.md)), and the instructions are somewhat different from the ["official" ones](https://atcollab.github.io/at/m/Installation.html#).
This raises a few problems:
1. using `atupdateContents` breaks the version number by rewriting the 1st two lines of `atmat/Contents.m`. According to the [developer notes](https://github.com/atcollab/at/blob/master/developer/matlab/README.md) this file (its 1st 2 lines, in fact) is handled by the `setversion` command. `atupdateContents` should be modified to preserve these 1st 2 lines.
2. I don't think it's up to the user to generate the documentation: it should be ready and available in each release. I don't think that atm2html is the solution to the Matlab documentation problem, but if it works, why not keep it… In this case, the generation of documentation should be added to the developer notes, and removed from the installation process. And we should find a way to host the html doc somewhere on the web site. A local version is difficult to handle for a standard installation from Matlab Central because the installation location depends on the platform, version, etc. For instance, for me (macOS), `atroot='/Users/laurent/Library/Application Support/MathWorks/MATLAB Add-Ons/Toolboxes/Accelerator Toolbox/atmat'`.

swhite2401:
Shall we make a new python release at the same time?

lfarv:
> Shall we make a new python release at the same time?

PyAT 0.4.0 is almost 5 months old. Why not… But exact synchro is not necessary.

lfarv:
Added the CorrectorPass bug fix in the [release notes](https://github.com/atcollab/at/releases/edit/untagged-2ba8c95d588043722fb9).

lfarv:
@lnadolski : I ran `atupdateContents` and committed the created `Contents.m` files, so that a user does not need to run it. But I see some changes to be done in the future:
- is the double recursion needed? Contents files are recursively generated in all subfolders (as normally in Matlab), but in addition each of these files describes recursively everything below. This ends up with many repetitions. While normally a Matlab contents describes only its containing folder.
- One should skip the functions declared private or obsolete. Unless we move them in specific folders

I think this should be part of another dedicated PR for documentation. For the moment, all contents files are there, that should be enough for a release. Do you agree with the present state?

lfarv:
I'd like to wait for the upgrade of the C constants (#691 ) before releasing the version. Otherwise, any comment on this release?

lfarv:
This release is ready. Please comment or approve.

simoneliuzzo (approved):

simoneliuzzo (changes requested):
`cavpts` missing from atSetRingParameters atGetRingParameters

lfarv:
> cavpts missing from atSetRingParameters atGetRingParameters

Can you give more details on your problem: which platform, lattice example ? 
I can't reproduce the problem, for me atGetRingProperties(ring, 'cavpts') returns the cavities with the lowest frequency, as expected.

lfarv:
@simoneliuzzo : Sorry. I missed your message, but I have now all the info, thanks.

lfarv:
Hello @simoneliuzzo. The bug you reported in now corrected. Unless you have other remarks, can you approve again the release?

simoneliuzzo (approved):

PR #691, opened 2023-11-08
Update and centralize the C constants

JeanLucPons:
Hello,

I centralized constants used in at in an include file [atconstants.h](https://github.com/atcollab/at/blob/centralize_constants/atintegrators/atconstants.h)

I'm facing an issue concerning the CGAMMA constant defined as below: 
Physics of Electron Storage Ring, M. Sands (4.2)
image

In the atconstants header file, the result of the formula gives: 8.846273834e-05 and it AT the old define was 8.846056192e-05.
The new cgamma constant results in at_physics test failure:

```
test_physics.py ...........FFF.....F..............FF
```

The best match i can get (8.846092697e-05) is by using 2.8179e-15m for the classical e- radius and 511keV for the rest energy.

Should we modify the test_physics.py or rather adjusting the constants above ?

Note: The very small correction of symplectic integrator constants does not affect the result of the tests.
Thanks


swhite2401:
Hello @JeanLucPons , this is a nice re-organization.

Concerning the constant, if it is wrong (@lfarv @lnadolski , any idea why??) it has to be corrected and the unit tests modified accordingly.

lfarv:
> Should we modify the test_physics.py or rather adjusting the constants above ?

I am rather skeptical about "adjusting" the physical constants…

PyAT (in python) uses the correct value of C&gamma;, computed in `at/pyat/at/constants.py`. It uses the last definition of the physical constants (2018) from [SciPy](https://docs.scipy.org/doc/scipy/reference/constants), based on [CODATA 2018](https://physics.nist.gov/cuu/Constants/) values. And the result is 8.846273822420375e-32, you are almost right, @JeanLucPons… The value in C must date from the previous release of the physical constants (2014) or even before…

So the correct value should of course be set, and the tests corrected accordingly. But the fact that the AT results will change must be clearly documented, it may raise complains…

lfarv:
The physical constants were updated in Matlab in #437, and I remember it raised many questions. But they probably were never updated in C

JeanLucPons:
Thanks Laurent, I also prefer to update the physical constants to the good values and update the unit tests.
I will copy the constants from the Pr you mentioned.

lnadolski (approved):

lfarv:
The tests are modified using results from Matlab. No tolerance had to be modified (good point!).

Please approve, to unlock the pending Matlab and python releases.

simoneliuzzo (approved):

PR #698, opened 2023-11-14
Add Cavity monitor

lcarver:
Modified the beam loading pass methods and functions to allow an optional cavity monitor. For each turn, the Vgen, Vbeam and Vbunch will all be stored in a buffer. This makes more detailed verification of what is happening a lot easier and simplifies the final scripts.

There are currently 2 issues with this implementation that I need some help solving.

1. In tracking.utils there is the function set_beam_monitors. I do not have a nice way to check if an element is a beam loading element or not. This is because BeamLoadingElement is not a part of lattice.elements, so when I try ti import it I get an error for circular import. The implementation in now works but is not very good.

2. All of the buffer setting works for Vbeam and Vgen, but I am finding it difficult to properly set the elements for the vbunch as I don't fully understand the ordering of the C buffer.


lcarver:
The reason there is so much additional stuff, is actually because I cloned the branch from the SliceMoments which is a few commits behind. I'm not so good at gitting, so sorry if it's a bit of a mess.

PR #699, opened 2023-11-15
Add vbeam vgen vbunch buffers for beam loading element

lcarver:
NOTE: I cleaned up the PR. Previous one is closed as it was confusing to read.

Modified the beam loading pass methods and functions to allow an optional cavity monitor. For each turn, the Vgen, Vbeam and Vbunch will all be stored in a buffer. This makes more detailed verification of what is happening a lot easier and simplifies the final scripts.

There are currently 2 issues with this implementation that I need some help solving.

~~1. In tracking.utils there is the function set_beam_monitors. I do not have a nice way to check if an element is a beam loading element or not. This is because BeamLoadingElement is not a part of lattice.elements, so when I try to import it I get an error for circular import. The implementation in now works but is not very clean and needs to be modified~~

clear_history function is now used and a handles initialization, the buffersize has to be predefined

~~2. All of the buffer setting works for Vbeam and Vgen, but I am finding it difficult to properly set the elements for the vbunch as I don't fully understand the ordering of the C buffer. I define the vbunch buffer in python with fortran ordering, but then when I loop through the vbunch to set to the voltage and phsae in the right slot for the vbunch_buffer, i can't make it work~~

Reshaping done as follows:
```
Vbeam_buffer: (2, buffersize)
Vgen_buffer: (2, buffersize)
Vbunch_buffer: (2, buffersize,nbunch)
```


swhite2401:
Hello @lcarver, before jumping into this I would really like that we evaluate the situation and try to come up with a generalized attribute history capability.
Adding monitors all over the place is getting very messy and I would not want that this grows to something unmanageable.

This will certainly slow down your developments but we have to find a better solution.

swhite2401:
@lcarver this is ready for review

simoneliuzzo (approved):

PR #704, opened 2023-11-21
Support of python 3.12, 64-bit wheels only 

lfarv:
python 3.12 is officially released, so it's now included in tests and in release builds.

Numpy does not provide 32-bit wheels (binary builds) for recent versions. It's therefore impossible to build 32-bit wheels for PyAT starting with python 3.12. We still provide 32-bit Windows and Linux builds for the previous python versions. But we may consider removing completely 32-bit builds in the future (building from source is still possible if needed).

### DIFF
--- a/.github/workflows/build-python-wheels.yml
+++ b/.github/workflows/build-python-wheels.yml
@@ -28,7 +28,7 @@ jobs:
           python-version: '3.9'
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.16.0
+        uses: pypa/cibuildwheel@v2.16.2
 
       - name: Upload wheels
         uses: actions/upload-artifact@v3
--- a/.github/workflows/build-python-wheels.yml
+++ b/.github/workflows/build-python-wheels.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, macos-12, windows-2022]
+        os: [ubuntu-22.04, macos-12, windows-2022]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


lfarv:
Hello @MJGaughran. From what I could understand, the choice of the GitHub runner has no influence on the generated wheels: they are built in Docker containers, based on 2 different Linux distributions. This depends on the tags given to `cibuildwheel`. These tags are in the form [{python_tag}-{platform_tag}](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip).

The **python tag** is obvious: cp38 -> cp for CPython  followed by the version number,

The **platform tag** for linux has 2 flavours corresponding to two distributions. [`manylinux*`](https://peps.python.org/pep-0600/) for distributions based on "glibc", [`musllinux*`](https://peps.python.org/pep-0656/) for distributions based on "musl". Each runs in a Docker container based on the corresponding Linux distribution. So the distribution used be the GitHub runner itself has no influence.

This is what I understand at the moment. I tested the wheel generation, it's OK, but I cannot install and run them all…

Concerning the Linux GitHub runner, all we ask it is to run python3.9 and run `cibuildwheel`. Today, I think Ubuntu20.04 is still able to do that. But we'll have to move at some point. New wheels will be generated on the next PyAT release, in some time from now (we just released 0.5). But in the mean time, you can explicit ask for wheel generation by triggering manually the ["Build and upload wheels and sdist"](https://github.com/atcollab/at/actions/workflows/build-python-wheels.yml) workflow, and have a look.

What's you advice about the distribution of 32-bit wheels?

lfarv:
Is there still any question about this one?

MJGaughran:
Hi @lfarv, sorry for not getting back to you. I can see from the built wheels that it is still compatible with the same GLIBC versions as before.

I'm not sure what I can say about 32-bit wheels other than that it is probably sensible to drop them if they become annoying to maintain.

lfarv:
@MJGaughran : so we keep 32-bit wheels as the are, that is for python < 3.12. For now, it's harmless.

PR #706, opened 2023-11-22
use np.float64 instead of np.float in function get_bunch_length_espread of pyat/at/acceptance/touschek.py module

simoneliuzzo:
use np.float64 instead of np.float in function get_bunch_length_espread of pyat/at/acceptance/touschek.py module

This solves an error of deprecation given by numpy (since numpy 1.20). An alternative is to upgrade the package requirement for scikit-learn

### DIFF
--- a/pyat/at/acceptance/touschek.py
+++ b/pyat/at/acceptance/touschek.py
@@ -54,7 +54,7 @@ def get_bunch_length_espread(ring, zn=None, bunch_curr=None, espread=None):
         nus = ep.f_s/ring.revolution_frequency
         cst = (-0.5 * numpy.sqrt(numpy.pi) * bunch_curr * zn /
                (vrf * h * cs * (abs(etac) * espread / nus)**3))
-        bl = numpy.float(bl0*fsolve(haissinski, (1,), args=cst))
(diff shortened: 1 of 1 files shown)


simoneliuzzo:
this pull request would solve issue 704

lfarv:
The reason for this cast seems to be the fact that `fsolve` works on an array variable (array of 1 element here). To convert an array of one element into a scalar, the clean solution is:
```python
bl = bl0*fsolve(haissinski, numpy.array([1.0]), args=cst)[0]
```
Note that the initial value (1.0) should also be an array (otherwise PyCharm gives a warning)

simoneliuzzo:
Dear @swhite2401,

what is your opinion? You are the author of the code, may be there are good reasons for the cast instead of indexing proposed by @lfarv ?

thank you
best regards
Simone

swhite2401:
Hello, sorry for the late reply, I have been very busy lately. I think I agree with @lfarv on this one, no need to cast. Indexing makes it clear to the reader that fsolve returns an array so it seems more logical to me.

simoneliuzzo:
The solution proposed by @lfarv is adopted. I tested on my side and it is ok.

swhite2401 (approved):

PR #707, opened 2023-11-22
return ZERO acceptance if no tracked particle survives in at/pyat/acceptance/boundary.py

simoneliuzzo:
return ZERO acceptance if no tracked particle survives in at/pyat/acceptance/boundary.py

In case of strongly varying acceptances along s, the  grid may not apply for all points. Zero could a good approximation (by defect) rather than an error.

### DIFF
--- a/pyat/at/acceptance/boundary.py
+++ b/pyat/at/acceptance/boundary.py
@@ -4,13 +4,14 @@ calculate the loss boundary for different
 grid definitions
 """
 
-from at.lattice import Lattice, AtError
+from at.lattice import Lattice, AtError, AtWarning
 from typing import Optional, Sequence
 from enum import Enum
 import numpy
 from scipy.ndimage import binary_dilation, binary_opening
 from collections import namedtuple
 import time
+import warnings
 
 __all__ = ['GridMode']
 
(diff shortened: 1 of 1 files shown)


pyat/at/acceptance/boundary.py
```diff
@@ -259,14 +262,22 @@ def radial_boundary(mask, grid):
             bnd[:, i] = search_bnd(ma, sa)
         return bnd
 
+    some_survived = True
     if not numpy.any(mask):
```
swhite2401, line 266: Implementing this check in search_bnd function could potentially make the code more compact what do you think?

simoneliuzzo: Dear @swhite2401,

no objection. My only issue is to make it proceed even if it finds a ZERO. I am open to any alternative that avoids the error. I do not know the code enough so I proposed something that could work, simply. 
Please push your proposed solution, or a modified code. I will test it!

thank you

swhite2401: Ok I take over

simoneliuzzo:
Dear @swhite2401, I tested this version and it works for my simulations (FCC vertical angular acceptance with errors and radiation).

lfarv (approved):
I trust @simoneliuzzo's tests, so its OK for me

simoneliuzzo:
Dear @swhite2401,

I merged too soon... I realized that the default output [0.0, 0.0] should be of shape (2, 1) to be consistent with the rest. it is instead (2,) 
numpy.atleast_2d returns shape (1, 2).

swhite2401:
So you want me to change the output to `numpy.zeros((2, 1))` that is easily done no problem.
Let me know!

swhite2401:
See #714 for fix

Issue #694, opened 2023-11-09
beamloading set point not correct for simple_ring


lcarver:
Hello,
phi_s is not set correctly when adding beam loading with a `simple_ring`. Example below:

```
import numpy as np
from at import simple_ring
import at
from at import add_beamloading, 

Vc = 5.5e6

ring_full = at.load_mat('../Repo/pyce/machine_data/esrf.mat', mat_key='LOW_EMIT_RING_INJ')
ring_full.set_rf_voltage(Vc)
ring_full.set_cavity_phase()
_, ring_pyat = at.fast_ring(ring_full)

U0 = ring_full.get_energy_loss()

ring_simple = simple_ring(ring_full.energy, ring_full.circumference, ring_full.harmonic_number, 0.1, 0.1, Vc, at.get_mcf(ring_full.disable_6d(copy=True)), U0=U0, TimeLag=100)

add_beamloading(ring_pyat, 100, 1, cavpts=[0], Nslice=50, VoltGain=1e-3, PhaseGain=1e-3)
print(ring_pyat[0]._phis)

add_beamloading(ring_simple, 100, 1, cavpts=[0], Nslice=50, VoltGain=1e-3, PhaseGain=1e-3)
print(ring_simple[0]._phis)

```

Output:
```
0.47833528
0.0

```
No matter what I put for TimeLag in simple_ring, the phi_s is always zero.

In `beam_loading.py`, we can see in line 188, in the function `add_beamloading`
`_, ts = get_timelag_fromU0(ring)`

This value of `ts` is not computed correctly for a simple_ring. Whatever I put for timelag, ts comes out to be -TimeLag which always gives a phi_s of 0.

lcarver:
in get_energy_loss, for the fast ring it returns the correct U0, for simple ring it doesn't find U0

lcarver:
So I know the problem, but no good idea of the solution.

To compute the U0, you must give get_energy_loss a lattice with rad on. It then switches off all collective, diffusion and cavity elements and tracks, element by element and computing sum of all energy changes.

For simple_ring, these energy changes always come out to be 0, as the energy loss is handled in the quantum diffusion element. I'm always not entirely sure if, when I added SimpleQuantDiff, if it was properly added to the is_collective group.

lcarver:
Also as a separate smaller and stranger bug

if I take my ring_simple and run (like in the get_energy_loss function)
[...]

PR #710, opened 2023-11-24
Simple Ring modification to solve energy loss problems

lcarver:
This is proposed as a solution to #694 .

Old simple ring contained:
`ring = [RF Cavity, M66 no rad, Nonlinear element, SimpleQuantDiff]`
where `SimpleQuantDiff `handled all radiation effects (damping, diffusion and U0)
This setup was problematic due to the fact that `get_energy_loss` could not compute the `U0` without the damping and diffusion, which had some other knock-on problems. 

Now simple ring contains:
`ring = [RF Cavity, M66 no rad, SimpleRadiation, Nonlinear element, SimpleQuantDiff]`
where `SimpleRadiation` is a new PassMethod that only performs U0 and radiation damping and `SimpleQuantDiff` has been modified to only act as the diffusion source for emittance. 

One possible problem that I want to draw to your attention, is that now we have 2 diffusion elements (QuantumDiffusion and SimpleQuantDiff) and I wonder if it is now more convenient to create a `Diffusion` class/group (not sure of the correct noun) and then each element has the property `is_diffusion` to enable cleaner code (for example you can see in `get_energy_loss` I have to mention both elements specifically.

### DIFF
--- a/atintegrators/SimpleRadiationPass.c
+++ b/atintegrators/SimpleRadiationPass.c
@@ -11,7 +11,7 @@ struct elem
   double EnergyLossFactor;
 };
 
-void SimpleRadPass(double *r_in,
+void SimpleRadiationPass(double *r_in,
            double taux, double tauy, double tauz,
            double EnergyLossFactor, int num_particles)
 
@@ -59,11 +59,11 @@ ExportMode struct elem *trackFunction(const atElem *ElemData,struct elem *Elem,
             Elem->U0=U0;
             Elem->EnergyLossFactor=U0/Param->energy;
         }
-        SimpleRadPass(r_in, Elem->taux, Elem->tauy, Elem->tauz, Elem->EnergyLossFactor, num_particles);
+        SimpleRadiationPass(r_in, Elem->taux, Elem->tauy, Elem->tauz, Elem->EnergyLossFactor, num_particles);
     return Elem;
 }
 
-MODULE_DEF(SimpleRadPass)        /* Dummy module initialisation */
+MODULE_DEF(SimpleRadiationPass)        /* Dummy module initialisation */
 
 #endif /*defined(MATLAB_MEX_FILE) || defined(PYAT)*/
 
@@ -85,7 +85,7 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
         /* ALLOCATE memory for the output array of the same size as the input  */
         plhs[0] = mxDuplicateArray(prhs[1]);
         r_in = mxGetDoubles(plhs[0]);
-        SimpleRadPass(r_in, taux, tauy, tauz, EnergyLossFactor, num_particles);
+        SimpleRadiationPass(r_in, taux, tauy, tauz, EnergyLossFactor, num_particles);
     }
     else if (nrhs == 0) {
         /* list of required fields */
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -1050,10 +1050,10 @@ class SimpleQuantDiff(_DictLongtMotion, Element):
         self.betay = betay
         super(SimpleQuantDiff, self).__init__(family_name, **kwargs)
 
-class SimpleRadiation(Radiative, Element):
+class SimpleRadiation(_DictLongtMotion, Element):
     """Simple radiation damping and energy loss"""
     _BUILD_ATTRIBUTES = Element._BUILD_ATTRIBUTES
-    default_pass = {False: 'IdentityPass', True: 'SimpleRadPass'}
+    default_pass = {False: 'IdentityPass', True: 'SimpleRadiationPass'}
 
     def __init__(self, family_name: str, 
                  taux: float = 0.0, tauy: float = 0.0,
@@ -1069,7 +1069,7 @@ class SimpleRadiation(Radiative, Element):
             tauz:          Longitudinal damping time [turns]
             U0:            Energy loss per turn [eV]
             
-        Default PassMethod: ``SimpleRadPass``
+        Default PassMethod: ``SimpleRadiationPass``
        """
         kwargs.setdefault('PassMethod', self.default_pass[True])
        
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -990,7 +990,10 @@ class M66(Element):
 class SimpleQuantDiff(_DictLongtMotion, Element):
     """
     Linear tracking element for a simplified quantum diffusion,
-    radiation damping and energy loss
+    radiation damping and energy loss.
+    
+    Note: The damping times are needed to compute the correct
+    kick for the emittance. Radiation damping is NOT applied.
     """
     _BUILD_ATTRIBUTES = Element._BUILD_ATTRIBUTES
     default_pass = {False: 'IdentityPass', True: 'SimpleQuantDiffPass'}
@@ -999,7 +1002,7 @@ class SimpleQuantDiff(_DictLongtMotion, Element):
                  betay: float = 1.0, emitx: float = 0.0,
                  emity: float = 0.0, espread: float = 0.0,
                  taux: float = 0.0, tauy: float = 0.0,
-                 tauz: float = 0.0, U0: float = 0.0,
+                 tauz: float = 0.0,
                  **kwargs):
         """
         Args:
@@ -1014,7 +1017,6 @@ class SimpleQuantDiff(_DictLongtMotion, Element):
             taux:          Horizontal damping time [turns]
             tauy:          Vertical damping time [turns]
             tauz:          Longitudinal damping time [turns]
-            U0:             Energy Loss [eV]
             
         Default PassMethod: ``SimpleQuantDiffPass``
        """
@@ -1044,11 +1046,44 @@ class SimpleQuantDiff(_DictLongtMotion, Element):
         if espread > 0.0:
             assert tauz > 0.0, 'if espread is given, tauz must be non zero'
             
-        self.U0 = U0
         self.betax = betax
         self.betay = betay
         super(SimpleQuantDiff, self).__init__(family_name, **kwargs)
 
+class SimpleRadiation(Radiative, Element):
+    """Simple radiation damping and energy loss"""
+    _BUILD_ATTRIBUTES = Element._BUILD_ATTRIBUTES
+    default_pass = {False: 'IdentityPass', True: 'SimpleRadPass'}
+
+    def __init__(self, family_name: str, 
+                 taux: float = 0.0, tauy: float = 0.0,
+                 tauz: float = 0.0, U0: float = 0.0,
+                 **kwargs):
+        """
+        Args:
+            family_name:    Name of the element
+            
+        Optional Args:
+            taux:          Horizontal damping time [turns]
+            tauy:          Vertical damping time [turns]
+            tauz:          Longitudinal damping time [turns]
+            U0:            Energy loss per turn [eV]
+            
+        Default PassMethod: ``SimpleRadPass``
+       """
+        kwargs.setdefault('PassMethod', self.default_pass[True])
+       
+        assert taux >= 0.0, 'taux must be greater than or equal to 0'
+        self.taux = taux
+            
+        assert tauy >= 0.0, 'tauy must be greater than or equal to 0'
+        self.tauy = tauy
+
+        assert tauz >= 0.0, 'tauz must be greater than or equal to 0'
+        self.tauz = tauz
+
+        self.U0 = U0            
+        super(SimpleRadiation, self).__init__(family_name, **kwargs)
 
 
 class Corrector(LongElement):
--- a/atintegrators/SimpleQuantDiffPass.c
+++ b/atintegrators/SimpleQuantDiffPass.c
@@ -14,13 +14,11 @@ struct elem
   double betay;
   double sigma_xp;
   double sigma_yp;
-  double U0;
-  double EnergyLossFactor;
 };
 
 void SimpleQuantDiffPass(double *r_in,
            double sigma_xp, double sigma_yp, double espread,
-           double taux, double tauy, double tauz, double EnergyLossFactor,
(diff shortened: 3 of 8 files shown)


lfarv:
Hello @lcarver and @swhite2401 

The idea is nice but I think that the implementation is wrong. More exactly, `SimpleRadiation` is wrong, its more complicated that that.

In the general 6D case; the 6x6 one-turn transfer matrix **M**<sub>rad</sub> can be expressed using the A matrix as:

**M**<sub>rad</sub> = **A** . **R** . **&Lambda;** . **A**<sup>-1</sup>

with **R** block diagonal rotation matrix (phase advance) and **&Lambda;** diagonal damping matrix.

so we have:

**M**<sub>rad</sub> = (**A** . **R** . **A**<sup>-1</sup>) . (**A** . **&Lambda;** . **A**<sup>-1</sup>)

**M**<sub>rad</sub> = **M**<sub>norad</sub> . **DAMP**

The damping matrix **DAMP** is **A** . **&Lambda;** . **A**<sup>-1</sup> !

**&Lambda;** is similar to what you wrote in `SimpleRadiation`, but:
1. The horizontal damping applies to both x and x' (and same for the other planes),
2. It has to be applied on the "normalised" coordinates. That's what the **A** matrice does. In the uncoupled case, **A** is the classical [ [1/sqrt(&beta;)  0][ &alpha;/sqrt(&beta;) sqrt(&beta;)]] matrix, but in the end, the `SimpleRadiation` is a 6x6 block diagonal matrix.

In conclusion, this option (once corrected) costs an additional matrix multiplication.

swhite2401:
@lfarv, this algorithm is implemented in many well know codes, is well benchmarked and gives the correct result in term of damping time and equilibrium emittance. This is exactly what we are looking in multi-particle simulation where the full ring is represented by a matrix.

We could go for your solution but then the algorithm becomes more complicated since you have to derive the 6x6 damping matrix, providing the huge approximation we are already making in modeling the ring I do not see much interest in this.

lfarv:
@swhite2401 : I have no doubt that you get the correct damping times and emittances, but for sure the transverse motion is wrong. The damping is applied as if the betas were 1.0. You maybe don't care about transverse optics: you can run multi particle tracking were the turn is defined by simply a phase advance, no need for beta functions, the one turn map is a simple rotation, and then your formula is correct. You just define tunes, chromaticities, and that's it. But if for some reason you want to introduce betas and alphas, it's different.

I'm 100% sure the formulas above are correct, so anything else is wrong. But I understand the need for speed, so I had a look numerically: in my notations,

**&Lambda;** = diag([exp(-1/&tau;<sub>x</sub>), exp(-1/&tau;<sub>x</sub>), exp(-1/&tau;<sub>y</sub>), ...]) with &tau; in turns. **&Lambda;**

For a sample ring (`hmba.mat`), the damping matrix **DAMP**= **A** . **&Lambda;** . **A**<sup>-1</sup> is still close to the diagonal matrix **&Lambda;**. The main terms are dispersive ones, their order of magnitude is about 10<sup>-8</sup>. So that's OK for the approximation. Ignoring them, 

**DAMP** ≈ diag([exp(-1/&tau;<sub>x</sub>), exp(-1/&tau;<sub>x</sub>), exp(-1/&tau;<sub>y</sub>), ...])

**DAMP** ≈ diag([(1-1/&tau;<sub>x</sub>), (1-1/&tau;<sub>x</sub>), (1-1/&tau;<sub>y</sub>), (1-1/&tau;<sub>y</sub>), (1-1/&tau;<sub>z</sub>), (1-1/&tau;<sub>z</sub>)])       for &tau; >> 1

Compared to what is in `SimpleRadiationPass`,
1. multiplying by the exact exponential (computed once and stored) would be faster (`r6[1] *= xdamp` compared to `r6[1] -= r6[1]/taux`), with `xdamp = exp(-1.0/taux)`
2. the damping should be applied to both x and p<sub>x</sub> instead of doubling the effect on p<sub>x</sub>. Why is it done like that?

lfarv:
Coming back to more practical problems: I agree that a `Diffusion` class would help. But at the same time, the `SimpleRadiation` element should also be handled automatically in `enable_6d`, without explicitly specifying it. So it should be declared as 
```python
class SimpleRadiation(_DictLongtMotion, Radiative, Element):
```
in that order, so that `_DictLongtMotion` provides the right methods, and `Radiative` is just used for grouping. That's not very elegant, and there could be some reorganisation of these mixin classes, to separate functionality and grouping.
But that's another topic…

lcarver:
I modified `SimpleRadiation` in the following way:

Added `Radiative` to element definition.
Passed `beta`, `alpha`, `disp `and `disp'` to the `SimpleRadiationPass.c`
In the python, use `taux,y,z` to construct 6d damping matrix.

Assuming the following formula to move from real to normalised coordinates:

```
X = x/sqrt(beta) - Dx*dpp
XP = xp*sqrt(beta) + x * alpha / sqrt(beta) - DPx*dpp
```

The particles are converted, damping applied, then reverted. U0 is applied at the end after all changes.

lcarver:
Below is a test file for your convenience:

```
import at
from at.physics.fastring import simple_ring
import numpy as np
from scipy.signal import hilbert

dic = {
       'alphax': 1e-3,
       'alphay': 1e-3,
       'betax': 20,
       'betay': 10,
       'dispx': 0.1,
       'dispxp': 1e-3,#0.0,
       'dispy': 0.1,
       'dispyp': 1e-3,#,
       'Qx': 74.1235,
       'Qy': 28.3,
       'Qpx': 0,
       'Qpy': 0,
       'A1': 0.,
       'A2': 0.,
       'A3': 0.,
       'circumference': 843.977,
       'alpha': 8.5e-5,
       'U0': 2.5e6,
       'Vrf': 6e6,
       'energy': 6e9,
       'harmonic_number': 992,
       'emitx': 100e-12,
       'emity': 10e-12,
       'espread': 9.5e-4,
       'taux': 3000,
       'tauy': 5000,
       'tauz': 1000
}
       
ring = simple_ring(dic['energy'], dic['circumference'], dic['harmonic_number'], dic['Qx'], dic['Qy'], dic['Vrf'], dic['alpha'], taux=dic['taux'], emitx=dic['emitx'], tauy=dic['tauy'], emity=dic['emity'], tauz=dic['tauz'], espread=dic['espread'], U0=dic['U0'], A1=dic['A1'], A2=dic['A2'], A3=dic['A3'], Qpx=dic['Qpx'], Qpy=dic['Qpy'], alphax=dic['alphax'], alphay=dic['alphay'], betax=dic['betax'], betay=dic['betay'], dispx=dic['dispx'], dispy=dic['dispy'], dispxp=dic['dispxp'], dispyp=dic['dispyp'])


r_in = np.zeros((6,1))
r_in[0] = 1e-1
r_in[2] = 1e-1
r_in[4] = 1e-3

# Check radiation damping
nt = 10000
r_out = at.lattice_pass(ring, r_in, nturns=nt)

xdat = r_out[0,0,0,:]
fitx = np.polyfit(np.arange(2500), np.log(np.abs(hilbert(xdat)))[500:3000], 1)
print('taux', -1/fitx[0], dic['taux'])

ydat = r_out[2,0,0,:]
fity = np.polyfit(np.arange(2500), np.log(np.abs(hilbert(ydat)))[500:3000], 1)
print('tauy', -1/fity[0], dic['tauy'])

#spikes at start and finish are artefacts of hilbert transform
plt.suptitle('envelope from hilbert transform')
plt.semilogy(np.abs(hilbert(xdat)),'r', label='x')
plt.semilogy(np.abs(hilbert(ydat)),'b', label='y')
plt.xlabel('turn')
plt.ylabel('env [m]')
plt.legend()
plt.show()



r_in = np.zeros((6,1))
r_in[0] = 0.
r_in[2] = 0.
r_in[4] = 1e-3
r_out = at.lattice_pass(ring, r_in, nturns=nt)
r_in[4] += 5e-2
r_out = at.lattice_pass(ring, r_in, nturns=nt)
dpdat = r_out[4,0,0,:]
zdat = r_out[5,0,0,:]

fitz = np.polyfit(np.arange(500), np.log(np.abs(hilbert(dpdat)))[250:750], 1)
print('tauz', -1/fitz[0], dic['tauz'])

plt.suptitle('abs(hilbert(dp))')
plt.semilogy(np.abs(hilbert(dpdat)))
plt.show()



```

lcarver:
Ready for re-review.

lcarver:
On my side, 
for betax and betay provided, alphax=alphay=0 and dispersion 0. Damping times are perfect.
For an enormous alphax=alphay=1, I see perfect damping in V and 2997 turns (compared to 3000) for H.

For alphax=alphay=0 and Dx=Dy=0.1, I see a deviation in H. 
image
Hdamping = 3215 turns (instead of 3000)
Vdamping = 5031 turns (instead of 5000).

So I think there is a mistake in how I implement the dispersion.

pyat/at/lattice/elements.py
```diff
@@ -1156,11 +1158,63 @@ def __init__(self, family_name: str, betax: float = 1.0,
...
+
+        self.U0 = U0
+        
+        self.damp_mat = numpy.zeros((6,6), order='F')
+        self.damp_mat[0,0] = dampx
+        self.damp_mat[1,1] = dampx
+        self.damp_mat[2,2] = dampy
+        self.damp_mat[3,3] = dampy
+        self.damp_mat[4,4] = dampz
+        self.damp_mat[5,5] = dampz
```
lfarv, line 1214: The damping matrix is diagonal. There is not need to store lots of zeros, you could only keep `self.dampx`, `self.dampy` and `self. dampz` and let `SimpleRadiationPass` deal with that.

pyat/at/lattice/elements.py
```diff
@@ -1156,11 +1158,63 @@ def __init__(self, family_name: str, betax: float = 1.0,
...
             assert tauz > 0.0, 'if espread is given, tauz must be non zero'
 
-        self.U0 = U0
         self.betax = betax
         self.betay = betay
         super(SimpleQuantDiff, self).__init__(family_name, **kwargs)
 
+class SimpleRadiation(_DictLongtMotion, Radiative, Element):
+    """Simple radiation damping and energy loss"""
+    _BUILD_ATTRIBUTES = Element._BUILD_ATTRIBUTES
```
lfarv, line 1167: With this, `repr(simple_radiation_element)` will not work. To make it work, you would need
```python
_BUILD_ATTRIBUTES = Element._BUILD_ATTRIBUTES + ['taux', 'tauy', 'tauz']
```
And have `taux`, `tauy` and `tauz` properties of the element (deduced from dampx,...).

This is not critical, we can live without it… Up to you.

lfarv (changes requested):
It's now correct (at least I think so), but can be simplified.

pyat/at/lattice/elements.py
lcarver: done, i now define damp_mat_diag which is an array of 6 elements.

atintegrators/SimpleRadiationPass.c
```diff
@@ -0,0 +1,189 @@
...
+        dpp *= dmatdp[4];
+      }
+
+      if(dmatz[5]!=1.0) {
+        z *= dmatz[5];
+      }
+
+      r6[4] = dpp;
+
+      x = X*sqrt(betax);
```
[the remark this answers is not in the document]
lcarver: I removed beta and alpha completely. Keeping dispersion as it will be needed when we eventually redo fast ring using this passmethod. 

All other changes implemented.

pyat/at/lattice/elements.py
lcarver: I added the taux,y,z to build attributes but

>  And have taux, tauy and tauz properties of the element (deduced from dampx,...).

this requires some additional action?

lcarver:
all changes implemented, but one outstanding question to be resolved (I made a comment in the code)

lcarver:
> @lcarver : I just found out that the linear matrix is missing the dispersion. One must add:
> 
> ```python
>     M04 = (1.0 - M00) * dispx - M01 * dispxp
>     M14 = -M10 * dispx + (1.0 - M11) * dispxp
> 
>     M24 = (1.0 - M22) * dispy - M23 * dispyp
>     M34 = -M32 * dispy + (1.0 - M33) * dispyp
> ```

Well noted. I added your modifications to the M66.

atintegrators/SimpleRadiationPass.c
```diff
@@ -0,0 +1,148 @@
...
+            double U0, EnergyLossFactor;
+            double dispx, dispxp, dispy, dispyp;
+            double *damp_mat_diag;
+            
+            damp_mat_diag=atGetDoubleArray(ElemData,"damp_mat_diag"); check_error();
+            U0=atGetDouble(ElemData,"U0"); check_error();
+            dispx=atGetDouble(ElemData,"dispx"); check_error();
+            dispxp=atGetDouble(ElemData,"dispxp"); check_error();
+            dispy=atGetDouble(ElemData,"dispy"); check_error();
+            dispyp=atGetDouble(ElemData,"dispyp"); check_error();
```
lfarv, line 91: Dispersions are not in the arguments of the `SimpleRadiation` constructor: it's possible to create an element without them. The `simple_ring` utility puts them, but someone could create a `SImpleElement` directly. So it would be safer to declare the dispersions here as optional with default value 0:
```C
dispx=atGetOptionalDouble(ElemData,"dispx", 0.0); check_error();
```

lfarv:
> ...but one outstanding question to be resolved (I made a comment in the code)

If this was a test, I admit that I failed… What is this question ?

pyat/at/lattice/elements.py
lfarv: It implies a significant refactoring of the constructor. Example: tests on argument validity are done only at element creation. It is better to do them in `_conversions` functions, so that they are also done if someone modifies an attribute. Here, `taux` would be a property, with a `_conversions` function requiring it to be positive. A user could modify `taux`, and it would be checked the same way.

For now, you can forget it, we can do that later.

lfarv (approved):
Provided your test still succeeds, that's OK for me.

PR #711, opened 2023-11-28
Added KStep and HStep keyword arguments to fit_tune and fit_chrom

simoneliuzzo:
use available scaling for gradient variation used by fit_tune and fit_chrom in at/pyat/matching/globalfit.py

### DIFF
--- a/pyat/at/matching/globalfit.py
+++ b/pyat/at/matching/globalfit.py
@@ -54,7 +54,7 @@ def _fit_tune_chrom(ring: Lattice, index: int, func,
         return sumsq
 
     if delta is None:
-        delta = 1.e6 * 10 ** index
+        delta = 1.e-6 * 10 ** index
         
     dq1 = _get_resp(ring, index, func, refpts1, 'PolynomB',
                     delta, dp, regex=regex, **kwargs)
--- a/pyat/at/matching/globalfit.py
+++ b/pyat/at/matching/globalfit.py
@@ -110,7 +110,7 @@ def fit_tune(ring: Lattice, refpts1: Refpts, refpts2: Refpts, newval,
 def fit_chrom(ring: Lattice, refpts1: Refpts, refpts2: Refpts, newval,
               tol: Optional[float] = 1.0e-12,
               dp: Optional[float] = 0, niter: Optional[int] = 3, regex=False,
-              KStep: Optional[float] = None,
+              HStep: Optional[float] = None,
               **kwargs) -> None:
     """Fit the chromaticities using 2 families
 
@@ -126,11 +126,11 @@ def fit_chrom(ring: Lattice, refpts1: Refpts, refpts2: Refpts, newval,
         niter:      Maximum number of iterations. Default 3
         regex:      Using regular expressions for refpt string matching;
                     Default: False
-        KStep: gradient variation applied to magnets. Default 1e-4
+        HStep: gradient variation applied to magnets. Default 1e-4
 
     Typical usage:
     at.fit_chrom(ring, refpts1, refpts2, [10,5])
     """
     print('\nFitting Chromaticity...')
     _fit_tune_chrom(ring, 2, _get_chrom, refpts1, refpts2, newval, tol=tol,
-                    dp=dp, niter=niter, regex=regex, delta=KStep, **kwargs)
+                    dp=dp, niter=niter, regex=regex, delta=HStep, **kwargs)
--- a/pyat/at/matching/globalfit.py
+++ b/pyat/at/matching/globalfit.py
@@ -107,7 +107,7 @@ def fit_tune(ring: Lattice, refpts1: Refpts, refpts2: Refpts, newval,
 def fit_chrom(ring: Lattice, refpts1: Refpts, refpts2: Refpts, newval,
               tol: Optional[float] = 1.0e-12,
               dp: Optional[float] = 0, niter: Optional[int] = 3, regex=False,
-              dk_multiplier_exp = 2,
+              deltaK = 1.0e-6,
               **kwargs) -> None:
     """Fit the chromaticities using 2 families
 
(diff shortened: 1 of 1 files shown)


simoneliuzzo:
An other alternative could be to add the DK to the list in at.lattice.options.DConstants

lfarv:
@simoneliuzzo : the `index` argument used in `_fit_tune_chrom` is the order of the multipole to be fitted. **It has to be 1 for tunes and 2 for chromaticity**, you cannot put anything else, it's the index in `PolynomB` which will be varied.

Then, this index is used to compute the **step in strength** to compute the Jacobian. If this step is what you want to control, rather than using the defaults value of `delta = 1e-6 * 10 ** index`, you need another input.

lfarv:
> An other alternative could be to add the DK to the list in at.lattice.options.DConstants


Right, it could be 2 new constants `DK` and `DH`, and corresponding keyword arguments in `fit_tune` and `fit_chrom`.

simoneliuzzo:
Dear @lfarv,

I modified as you suggested.

pyat/at/matching/globalfit.py
```diff
@@ -25,7 +25,7 @@ def _fit_tune_chrom(ring: Lattice, index: int, func,
                     refpts1: Refpts, refpts2: Refpts, newval,
                     tol: Optional[float] = 1.0e-12,
                     dp: Optional[float] = 0, niter: Optional[int] = 3,
-                    regex=False, **kwargs):
+                    regex=False, deltaK=1.0e-6, **kwargs):
```
lfarv, line 28: should be `delta: Optional[float] = None`

pyat/at/matching/globalfit.py
```diff
@@ -52,7 +52,7 @@ def _fit(ring, index, func, refpts1, refpts2, newval, J,
         sumsq = numpy.sum(numpy.square(numpy.subtract(val, newval)))
         return sumsq
 
-    delta = 1e-6 * 10 ** index
+    delta = deltaK * 10 ** index
```
lfarv, line 55: and here:
```
if delta is None:
   delta = 1.e-6 * 10 ** index
```

lfarv (changes requested):

pyat/at/matching/globalfit.py
```diff
@@ -52,7 +53,9 @@ def _fit(ring, index, func, refpts1, refpts2, newval, J,
         sumsq = numpy.sum(numpy.square(numpy.subtract(val, newval)))
         return sumsq
 
-    delta = 1e-6 * 10 ** index
+    if delta is None:
+        delta = 1.e6 * 10 ** index
+
```
lfarv, line 58: Sorry, my fault: it should be 1.e-6 instead of 1.e6…

pyat/at/matching/globalfit.py
```diff
@@ -73,7 +76,8 @@ def _fit(ring, index, func, refpts1, refpts2, newval, J,
 
 def fit_tune(ring: Lattice, refpts1: Refpts, refpts2: Refpts, newval,
              tol: float = 1.0e-12,
-             dp: Optional[float] = 0, niter: int = 3, regex=False,
+             dp: Optional[float] = 0, niter: int = 3, regex=False, 
+             delta: Optional[float] = None,
```
lfarv, line 80: Here, I would call it `KStep`, by analogy with `XYStep` and `DPStep` already defined, and call `_fit_tunechrom(..., delta=KStep,...)`

If you want to add these constants in `at.DConstant`, it will be easier to remove the explicit argument and use `kwargs`:
```python
delta = kwargs.pop('KStep', DConstant.KStep)    # if defined in at.DConstant
...
_fit_tunechrom(..., delta=delta)
```

pyat/at/matching/globalfit.py
```diff
@@ -91,6 +95,7 @@ def fit_tune(ring: Lattice, refpts1: Refpts, refpts2: Refpts, newval,
         fit_integer: bool (default=False), use integer tune
         regex:      Using regular expressions for refpt string matching;
                     Default: False
+        delta: gradient variation applied to magnets. Default 1e-6
```
lfarv, line 98: If I'm right, the default will be 1.e-5 for the tune fit: (`1.e-6 * 10 ** 1`)
and 1.e-4 for chromaticity (`1.e-6 * 10 ** 2`)

pyat/at/matching/globalfit.py
```diff
@@ -99,12 +104,13 @@ def fit_tune(ring: Lattice, refpts1: Refpts, refpts2: Refpts, newval,
...
         kwargs['fit_integer'] = True
     _fit_tune_chrom(ring, 1, _get_tune, refpts1, refpts2, newval, tol=tol,
-                    dp=dp, niter=niter, regex=regex, **kwargs)
+                    dp=dp, niter=niter, regex=regex, delta=delta, **kwargs)
 
 
 def fit_chrom(ring: Lattice, refpts1: Refpts, refpts2: Refpts, newval,
               tol: Optional[float] = 1.0e-12,
               dp: Optional[float] = 0, niter: Optional[int] = 3, regex=False,
+              delta: Optional[float] = None,
```
lfarv, line 113: As for the tune fit, I would call it `HStep`.

lfarv:
By the way, did you have problems with the default step values ?

If you'd like to add the two default values in `at.DConstant`, it's up to you !

And a final remark: the future  release notes will be based on the titles of PRs, so it could be clearer to have something like "Added KStep and HStep arguments to fit_tune and fit_chrom".

simoneliuzzo:
Dear @lfarv,

thank you for your input! I tried to put everything in.

I prefer not to touch on DConstants. I have nothing against adding those parameters there.
I did try to change the DConstants and that is how I found that the tune KStep was not available. So I opened this pull request. 

Since `fit_chrom` was missing `**kwargs` I added it as in `fit_tune`.

thank you
best regards
Simone

lfarv (approved):
OK for me

lfarv:
Hi @simoneliuzzo. Looks good to me.

You did not tell why you had to use non-default values for these steps. Playing with `niter` was not enough to get convergence?

simoneliuzzo:
Dear @lfarv,

for the FCC-ee lattice, changing quadrupoles by 1e-6 makes the optics unstable. At the moment the tolerances for FCC are "unrealistic" to be kind.

best regards
Simone

swhite2401:
Can we delete the branch?

PR #719, opened 2023-12-09
Update atsimplering in Matlab

lfarv:
In #710, the passmethods used in "simple rings" were modified. This branch corrects the Matlab function `atsimplering`, to generate the same elements as in python: a `SimpleRadiation` and a `SimpleQuantDiffPass`.

These two elements are correctly handled in `atenable_6d`/`atdisable_6d`.


lcarver:
Of course, my matlab skills are poor to say the least, so user error is likely. Nonetheless, I generate the following ring:

```
ring = atsimplering(6e9, 844, 992, 0.1, 0.2, 6e6, 1e-4, 'U0', 2.5e6);
ringrad = atenable_6d(ring);

```
then:
`atgetU0(ringrad)`
returns 0

lfarv:
@lcarver: 

As in python, the default method for computing the energy loss is "integral" (analytic). And in a simple ring, there is nothing to integrate !

But if you try:
```Matlab
atgetU0(ringrad,method='tracking')
```
you get the expected value of 250000. I'm pretty sure it's exactly the same in python.

lcarver:
You are if course correct. Everythings seems ok to me

lcarver (approved):

PR #720, opened 2023-12-09
Fix a few "simple ring" problems

lfarv:
This fixes a few problems with simple rings:

#### 1. Momentum compaction
The momentum compaction is not exactly as specified: let's generate a ring with `alpha = 1.e-4`:
```python
ring=at.simple_ring(6e9, 844.39, 992, 0.21, 0.34, 6e6, 1.e-4)
ringnorad = ring.disable_6d(copy=True)
ringnorad.mcf
9.999274666869721e-05
```
The small difference comes from the fact that the M54 term of the linear transfer matrix is built using `eta`, slip factor. But the 6th coordinate is the path lengthening &beta;c&tau;, and not the delay. See how `alpha` is computed in `get_mcf`. Using alpha gives the correct result.

#### 2. `repr()` output of `SimpleRadiation` elements
Corrected so that evaluating the output string generates the correct element.

#### 2. Transfer of "simple rings" in .mat files
Simple rings are now correctly transferred through` .mat` files

### DIFF
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -1082,7 +1082,7 @@ class RFCavity(LongtMotion, LongElement):
 
 class M66(Element):
     """Linear (6, 6) transfer matrix"""
-    _BUILD_ATTRIBUTES = Element._BUILD_ATTRIBUTES
+    _BUILD_ATTRIBUTES = Element._BUILD_ATTRIBUTES + ["M66"]
     _conversions = dict(Element._conversions, M66=_array66)
 
     def __init__(self, family_name: str, m66=None, **kwargs):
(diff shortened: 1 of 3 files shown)


swhite2401:
Ah nice! In fact I was writinh a converter to another code an was seeing some discrepancies in the synchrotron plane. I still see some very small differences but this certainly improves this improves it.

swhite2401:
I noted some issues with the typing hints in simple_ring() as well -> vrf, harmonic_number and TimeLag can be float of Sequence of floats (certainly not bool). Could you fix these while you are  at it?

lfarv:
> I noted some issues with the typing hints

Good remark. it should be correct now.

swhite2401:
Shouldn't harmonic_number be a int? It is for the cavity element... 
btw why are your coding on such a nice day? (I am on shift, so I am stuck in front of my computer...)

lfarv:
> Shouldn't harmonic_number be a int? It is for the cavity element...

It does not matter. It's converted to int and stored in the cavity element, but not used !

> why are your coding on such a nice day?

Unexpectedly nice! And I'm recovering from covid… I was about to ask the same question, but I have the answer.

swhite2401 (approved):
Ok for me

lcarver:
Thank you!

lcarver (approved):

lcarver:
Also, can we sneak in a modification to remove the print from at.beam? (print(h,v,delta)). When running with mpi it prints on every thread and is quite annoying.

lfarv:
@lcarver 

>  remove the print from at.beam

good point, I take note

Issue #722, opened 2023-12-14
synoptics not visible in atplot when vertical values are above zero


oscarxblanco:
Dear all,

there is a minor bug when plotting the synoptics using `atplot.m` because the vertical axis limits are not checked.

By default the synoptics flag is true and the lattice synoptics is plotted at zero in the vertical axis, but, depending on the optics the vertical values might not be close to zero and the height of the blocks is not high enough, leaving the synoptics out of the plot. For example,
![image](https://github.com/atcollab/at/assets/108396048/38f7696e-c88a-45d4-a2e0-fa1ddd18cfd5)

One easy fix could be the following. Change this
https://github.com/atcollab/at/blob/abc7aebaaf763614fff7a5c68a9d190285b9a62a/atmat/atplot/xplot.m#L67-L69

to this, or something similar :
```matlab
if synopt
    if ax.YLim(1) > 0, ax.YLim(1) = 0; end % change YLim to see synopt
    curve.lattice=atplotsyn(ax,ring,varargs{:});  % Plot lattice elements
end
```
![image](https://github.com/atcollab/at/assets/108396048/c1c7e2b4-2b07-4a69-8f14-328f90070608)
In the last image the synoptics is visible even if the grid is only around the plotted values.

lfarv:
fixed in #726

PR #726, opened 2023-12-21
set YLim when synopt is true

oscarxblanco:
Dear all,

this Pull Request addresses a small issue in AT matlab reported in #722 where the synoptics is not visible.

This PR sets the vertical Axis limits in order to display the synoptics.


lfarv (approved):
Looks nice

PR #734, opened 2024-01-28
Update GitHub actions

lfarv:
- The action `setup-python` is updated v5 to avoid a warning:
  `Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/setup-python@v4.`
- The `flake8` options are modified for compatibility with the `black` code formatter. In particular, the recommended line length is set to 88,
- `pytest` is restricted to release <8 because the new version 8.0.0 make the test sequence crash.

Note that the Matlab actions (`setup-matlab` and `run-command`) are not yet upgraded to use Node.js 20, so they still trigger the warning.

### DIFF
--- a/.github/workflows/build-python-wheels.yml
+++ b/.github/workflows/build-python-wheels.yml
@@ -23,7 +23,7 @@ jobs:
           fetch-depth: 0
 
       - name: Install Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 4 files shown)


lfarv:
The problem with pytest has been solved in #735, so it's no more mentioned here.

@swhite2401 , @MJGaughran : this PR is not critical, but I would like to move on. Do you have comments?

MJGaughran (approved):

PR #735, opened 2024-01-31
Upgrade the test sequence for pytest >= 8

lfarv:
The recent release of `pytest 8.0.0 `broke the test sequence. This is due to an incompatibility of the `pytest_lazy_fixture` package with `pytest 8.0.0`. Looking around, it seems very unlikely that `pytest_lazy_fixture` will be updated: it's now abandoned for 5 years.

A temporary solution in #734 was to pin pytest to <8.0. In this PR we get rid of `pytest_lazy_fixture`.

**Note to developers:** when upgrading to pytest 8.0.0 (now possible with this branch). you must **uninstall pytest_lazy_fixture**.

**Note to reviewers:** formatting the modified files with ["black"](https://black.readthedocs.io/en/stable/) unfortunately introduced a lot of cosmetic changes. The only real change was to replace all occurrences of lines like:
```python
@pytest.mark.parametrize('lattice',
                         [pytest.lazy_fixture('dba_lattice'),
                          pytest.lazy_fixture('hmba_lattice')])
def test_linopt6_norad(lattice):
```
with:
```python
@pytest.mark.parametrize("lattice", ["dba_lattice", "hmba_lattice"])
def test_linopt6_norad(request, lattice):
    lattice = request.getfixturevalue(lattice)
```
Using the `request` fixture.

### DIFF
--- a/pyat/test/test_lattice_object.py
+++ b/pyat/test/test_lattice_object.py
@@ -1,53 +1,63 @@
 import numpy
-from numpy.testing import assert_allclose, assert_equal
 import pytest
+from numpy.testing import assert_allclose, assert_equal
+
 from at import elements
 from at.lattice import Lattice, AtWarning, AtError
 
 
 def test_lattice_creation_gets_attributes_from_arguments():
-    lat = Lattice(name='lattice', energy=3.e+6, periodicity=32, an_attr=12)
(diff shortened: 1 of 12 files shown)


swhite2401:
Would it possible to merge this rapidly? Thanks!

Issue #739, opened 2024-02-19
python: ringparam not well identified by lattice_load


lnadolski:
Dear All,

the at.save_lattice and at.load_lattice do not behave well in presence of atringparam

# loading a mfile
When using 
`ring = at.load_lattice('ESRF_EBS.m', keep_all=True)
`
the first element of the ring is identified as a marker instead of atringparam.
`atringparam('EBS',6000000000,1);...`

```
In [53]: ring[0]
Out[53]: Marker('EBS', Periodicity=1)
```

# saving a mfile
When saving the lattice, the code adds each time an extra element (atringparam)

```
function ring = newlattice()
ring = {...
    atringparam('EBS', 6000000000.0, 1, 'Particle', struct('name', 'relativistic', 'rest_energy', 0.0, 'charge', -1.0), 'HarmNumber', 992, 'Beam_Current', 0.0, 'Nbunch', 1, 'M_File', '/lustre/work/sources/physmach/nadolski/codes/pyenv/collimation/scripts/pyat/touschekscatteringlosses/test_crotch/ESRF_EBS.m');...
       atmarker('EBS', 'Periodicity', 1);...

```

lfarv:
Hi Laurent

The `RingParam` element does not exist in python. So when reading a .mat to .m file, the element is removed and its information is moved to the python `Lattice` object.

In the other way, when saving the lattice in .mat or .m file, a `RingParam` element is added in the 1st position because Matlab needs it.

Here, you are using the `keep_all` keyword. This is intended for tests where you want to have the same number of elements in python and Matlab. To achieve this, the `RingParam` element is replaced by a `Marker` placeholder. But this does not prevent from adding the `RingParam` element when saving back the file.

A solution could be to memorise its origin (RingParam) in the Marker so that it can be deleted when saving the file. For the time being, try avoiding the `keep_all` keyword unless it's really necessary.

lnadolski:
Dear Laurent,

Thabks for your reply.
This is not convenient when we go back and forth between files.
In my example, I want to do the following

script1.py reads lattice.m and produces outputlattice.m
script2.py reads outputlattice.m and does other computations

What is the good way to do it in Python if I want to keep the same number of elements?
Should I test each time if RingParam exists and remove it? This is not convenient at all...

lfarv:
[...]

PR #741, opened 2024-02-20
Better handling of Matlab RingParam element

lfarv:
Matlab uses a `RingParam` element to store lattice-related parameters. This element does not exist in python since these informations are stored in the `Lattice` object. When reading a .mat or .m file, python moves its information to the `Lattice` and discards the element. On saving, python creates a `RingParam` element and prepends it to the real lattice elements.

But the `keep_all` keyword argument in `load_lattice` allows, for special cases, to put a `Marker` placeholder replacing the `RingParam`, in order to keep the same number of elements as in the Matlab lattice. When saving, a new `RingParam` is created and inserted before the `Marker`, so that after a number of load/save sequences, Markers accumulate at the beginning of the lattice. This was reported by @lnadolski in #739.

In this PR, the Marker placeholder is identified as such by a `tag` attribute, and is removed when saving the file. Markers don't accumulate any more.

Note however that the `keep_all` keyword should not be used unless really necessary: keeping a useless element in the lattice is disturbing and inefficient.

Note: most of the changes are due to formatting the file with [black](https://black.readthedocs.io/en/stable/). This gives a derministic format enabling easier code comparison in the future. I'm starting using it when modifying a file.

### DIFF
--- a/pyat/at/load/matfile.py
+++ b/pyat/at/load/matfile.py
@@ -1,34 +1,41 @@
 """
 Load lattices from Matlab files.
 """
-from __future__ import print_function
+
+from __future__ import print_function, annotations
+
+__all__ = ["load_mat", "save_mat", "load_m", "save_m", "load_var"]
+
 import sys
 from os.path import abspath, basename, splitext
+from typing import Optional
+from collections.abc import Sequence, Generator
 from warnings import warn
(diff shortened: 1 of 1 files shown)


lnadolski (approved):
Tested, it works well

Issue #757, opened 2024-04-23
fast_ring when having harmonic cavities in the lattice


TeresiaOlsson:
Hi,

We have a lattice with both main and harmonic cavities and I have problem with generating the fast_ring. Internally the function get_timelag_fromU0 is called and that fails because the cavities don't have the same frequency.

I have set the cavpts attribute on the lattice to only point towards the main cavities and this works when calling get_timelag_fromU0 directly, but not when it's called inside fast_ring because this attribute is never used inside the function. Perhaps fast_ring also needs a keyword to specify which cavity reference points to use like on find_orbit6?

swhite2401:
Hi @TeresiaOlsson ,
I've had similar comment from another user, most likely the same issue.
Could you please send me your script by email? I'll take a look.

Normally, get_timelag_fromU0 should be find a solution with multiple RF systems

TeresiaOlsson:
Thank you :)

I have attached my files. The script is super simple but I have some parts at the end of the lattice file which maybe could be the cause.
```
    # Turn cavity and radiation on
    ring.enable_6d()
    
    # Set cavpts to point to main cavities
    ring.cavpts = ring.get_uint32_index('CAV*')
    
    # Set only main cavity phases
    ring.set_cavity_phase(cavpts=ring.cavpts)
```

[fast_ring_test.zip](https://github.com/atcollab/at/files/15077069/fast_ring_test.zip)

swhite2401:
Fixed in #758

PR #758, opened 2024-04-24
BugFix: normalization by 0 in get_timelag_fromU0

swhite2401:
This PR is a follow-up of #757 
For multiple RF systems the synchronous phase is solved numerically. The system was normalized to U0 to prevent tolerance issues for the search of the stable point in the the presence of large energy loss per turn.
This normalization resulted in issue in get_timelag_fromU0 (called by findorbit6 and fastring) in the presence of multiple RF systems when radiations are off.
This issue is resolve by applying the normalization only when U0>0.

### DIFF
--- a/.github/workflows/matlab-tests.yml
+++ b/.github/workflows/matlab-tests.yml
@@ -20,16 +20,16 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-13, windows-latest]
 
     steps:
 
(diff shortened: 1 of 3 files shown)


swhite2401:
@lfarv, the macos-latest runs on the M1 chip and does not support python version < 3.10. Also matlab does not run on arm64 architecture.

For the tests to complete I have changed the runner config such that macos-13 is used for python <3.10 and for the matlab tests
Is this ok for you?

lfarv (approved):
@swhite2401:
I could not find documentation about the changes GitHub made on macOS runners…
Anyway your solution looks good.

Issue #749, opened 2024-03-20
Bug: atwiggler


swhite2401:
We are still suffering from the bug reported in #190. `atwiggler` uses `GWigSymplecticPass` as default, which will then point to GWigSymplecticRadPass when 6d is enabled. The latter has a bug and shall not be used. Not only does it have a bug but now fails to run with `atpass` complaining about a missing field energy even though it is defined.

We have 2 severe issues here:
1. We are distributing a passmethod with a know bug that users still use and suffer from (I was recently contacted by an NSLSII colleague regarding this issue)
2. The default at wiggler creation function points by default to this passmethod with know bugs

This situation is really not appropriate has to be solved rapidly by either removing the function with bugs (or at least make the necessary changes to not have it as default wiggler function) or by fixing it which obviously take much more efforts.

I have no problem removing it, as far as I am concerned we should not distribute function giving the wrong results.
Could you please advise?

Thanks!

ZeusMarti:
Hi @swhite2401,

As I mentioned earlier we are working on this point for our implementation at ALBA. We are basically adapting @mashl version which for him is working. As I said [before](https://github.com/atcollab/at/discussions/715), my understanding is that the current GWigSymplecticRadPass is not accompanied with the corresponding additional files and modifications in the ohmienvelop.m function. I guess @ahmad should update his pull requests but he seems quite busy and actually I'm not sure his is reading this (despite my efforts).

In summary, in case it helps we will let you know if it finally also works for us. 

Best,

Zeus

swhite2401:
Thanks @ZeusMarti it is in fact excellent news that you are tackling this issue. The question is now whether we should disable this feature until a working solution is found or at least issue a clear warning message to prevent any confusion on the users side.

ZeusMarti:
I believe a warning should suffice, just in case my opinion counts.

lfarv:
Hello @swhite2401 and @ZeusMarti,

There are 2 problems with `GWigSymplecticRadPass`:
1. For **tracking**, the present implementation is wrong: it's missing the change in momentum. As it is, there is no momentum change due to radiation, which is obviously wrong. This is easy to correct, since `GWigSymplecticRadPass` is based on a drift-kick sequence in cartesian geometry, the same as what is used if `StrMPoleSymplectic4RadPass`. So all ingredients are available,
[...]

PR #759, opened 2024-04-25
Bug Fix: Wiggler diffusion matrix

joanarenillas:
Dear all,

This is an ongoing Work In Progress, which aims to correct some bugs found in `GWigSymplecticPass.c` and  `GWigSymplecticRadPass.c`, as well as updating `gwigR.c` and `ohmienvelope` to include the computation of wiggler diffusion matrices. To do so, a new function (`FDW.c`) is currently being developed, following the work of @mashl. 

More on current and past issues on wiggler passmethods can be found in #715 and #749.

**No need to review yet.**


joanarenillas:
Dear @swhite2401,

I am currently cleaning up the code in `FDW.c`, which implements the computation of diffusion matrices for wigglers. I am having a little bit of trouble understanding how to code the Matlab mex function and the corresponding Python version. My guess is that it should look similar to the code in `findmpoleraddiffmatrix.c`, but I am not entirely sure whether I can just use the same lines of code or a more detailed makeover is necessary.

Could I get some advice with this?

Regards,

Joan

lfarv:
Hello @joanarenillas,

Sorry for my late intervention, but I was away until now and just had a look to your proposal.

There is already a git branch dedicated to the computation of diffusion matrices, called `diffusion_matrices`, with a corresponding pull request: #742. In its discussion, you will find the motivation and the proposed solution.

In short, the idea is to have the computation of diffusion matrices as modular as the tracking itself, so that introducing a new kind of element, including its contribution to equilibrium emittances,  can be made by simply dropping a new C file in the `atintegrators` directory,

Since the computation of diffusion matrices involves the tracking of the closed orbit through the element, a simple way is to add the diffusion as an optional computation in the integrator C file itself.

For backward compatibility, we cannot change the signature of the integrator, so to indicate the need for the computation of the diffusion matrix, we added a new variable in the `struct parameters` structure, called `bdiff`. `bdiff` is a `(double *)` pointer. If `bdiff == NULL`, it means that no computation of diffusion matrix is needed (simple tracking). Otherwise, `bdiff` points to an array of 36 doubles initialised with zeros. The integrator is expected to fill this array with the diffusion matrix of the element.

Existing integrators ignore this new variable, so are automatically considered as not contributing to diffusion. In the  `diffusion_matrices` branch, most of the `*RadPass` have been modified to include the computation of diffusion matrices (still missing the so-called “Exact” integrators).

I think you should go directly to this new scheme by basing your working branch on `diffusion_matrices` instead of `master`. Then your modifications will concern only a single file: `GWigSymplecticRadPass.c`.

For an example, look at `StrMPoleSymplectic4RadPass,c` in the `diffusion_matrices` branch: it's drift-kick sequence in cartesian geometry, like the wiggler.

From what I can see, your new `FDW.c` file is a good starting point for this upgraded `GWigSymplecticRadPass.c`. I does both tracking and computation of the diffusion matrix. You just need to make the computation of diffusion optional by enclosing it in something like:
```
if (bdiff) {
    wigglerM(...);
    wigglerB(...);
    ATsandwichmmt(...);
    ATaddmm(...);
}
```
and to add a loop on particles for the "normal" tracking (of course, the computation of diffusion matrices is requested only with a single particle: the closed orbit).

ZeusMarti:
Hi @joanarenillas, @lfarv and @swhite2401,

Besides Joan contributing to the diffusion_matrices branch, which I think it is a good idea, the present MR solves the #749 bug which I think is already a good reason to merge it, after your supervision of course.

Best,

atmat/atphysics/Radiation/ohmienvelope.m
```diff
@@ -24,6 +24,12 @@
 NumElements = length(ring);
 if nargin<3, refpts=1; end
 
+Wig=atgetcells(ring,'Bmax');
+Wigidx=find(Wig(:)==1);
+
+radindex(Wigidx) = false; % Erase wigglers from the radiative element list.
```
lfarv, line 30:
```Matlab
radindex = radindex & ~Wig
```
is more efficient and easier to understand

atmat/atphysics/Radiation/FDW.c
```diff
@@ -0,0 +1,572 @@
+/* FDW.c
+
+   mex-function to calculate integrated radiation diffusion matrix B defined in [2] 
+   for wiggler elements in MATLAB Accelerator Toolbox
+   O.Jimenez 3/08/18
```
lfarv, line 5: Has it really been initiated in 2018 ? You could mention your contribution…

lfarv:
@ZeusMarti, @joanarenillas:

> Besides Joan contributing to the diffusion_matrices branch, which I think it is a good idea, the present MR solves the https://github.com/atcollab/at/issues/749 bug which I think is already a good reason to merge it, after your supervision of course.

Ok, no objection! I have a few comments:
- it would be nice to add a small description in the [documentation of passmethods](https://atcollab.github.io/at/common/sa_passmethods.html) (`at/docs/common/sa_passmethods.md)`. If you don't want to dive into Markdown, send me a few lines, I can do it…
- Is there any reference to the use of Hessian matrices in this kind of computation?
- It looks difficult to me, but did you think of  way to validate the results of this new feature?

Otherwise, for me we could merge this branch.

Also, one should think soon of converting this to the new architecture for computation of diffusion matrices: #742. This branch was designed to integrate easily wigglers but also any future kind of element. Though apparently it did raise much interest since it was opened…

atmat/atphysics/Radiation/FDW.c
```diff
@@ -0,0 +1,572 @@
...
+               orbit_in[1] += ax;
+               orbit_in[3] += ay;	
+		}      
+        
+        flag=-1;
+        GWigGauge(&pWig, orbit_in, flag);
+       
+		
+        if(PR2)
+            ATmultmv(orbit_in,PR2);
```
lfarv, line 338: You should propagate the diffusion matrix through the rotation matrix. You are right that at input, propagating a zero matrix is useless, but I think here it should be done.

atmat/atphysics/Radiation/ohmienvelope.m
joanarenillas: Thanks for the comment, I'll change this.

atmat/atphysics/Radiation/FDW.c
joanarenillas: I will modify it and add my contribution to all files in the next commit.

joanarenillas:
Dear @lfarv,

Regarding your comments:

-  I will add a few lines to the [documentation of passmethods](https://atcollab.github.io/at/common/sa_passmethods.html). If I have any trouble doing so I will let you know.
- The computation of the Hessian matrices was done by @mashl. I am not sure whether he used any reference for the computation or he did it all himself. I will look into this and let you know if I find something.
- Together with @ZeusMarti, we were thinking of doing an slicing of the wiggler into many multipoles to get a very precise model for it, and then compare the emittance contributions of this model with that of `FDW.c`. We are still not completely sure if this is the right approach.

After this branch is merged, we are planning on adapting the computation of wiggler diffusion matrices of this branch to #742, so that wigglers are easily integrated. 

Regards,

Joan

atmat/atphysics/Radiation/FDW.c
lfarv: Be careful, Yrot is a rotation around the Y axis. Here we are talking of a rotation around the Z axis. I think that the propagation consists just in applying the "ATsandwich" to the rotation matrix (R1 or R2).

atmat/atphysics/Radiation/FDW.c
```diff
@@ -0,0 +1,465 @@
...
+    GWigGauge(&pWig, orbit_in, flag);
+		
+    if(R2) { 
+	   ATmultmv(orbit_in,R2);
+	   ATsandwichmmt(R2,BDIFF);
+	}
+    if(T2) {
+		ATaddvv(orbit_in,T2);	
+		ATsandwichmmt(T2,BDIFF);
+	}
```
lfarv, line 283: T2 is not a 6x6 transfer matrix, it's a vector ! You cannot send it to `ATsandwichmmt`. Did you try and check the result ? It should crash…

A translation has no effect on beam matrices. There is nothing to do to BDIFF.

joanarenillas: That is right. It is now corrected.

joanarenillas:
Dear @lfarv,

We are currently working on a benchmark to test our results. We hope to complete it soon. Once we have it, we will let you know the results so you can approve and merge.

Regards,

joanarenillas:
Dear @lfarv,

We are having difficulties in completing the benchmark for the new set of integrators, since the particle dynamics are more complex than we anticipated. We have tried solving the Lorentz equation including energy loss (gamma factor is not constant) and taking into account the [Abraham-Lorentz force](https://en.wikipedia.org/wiki/Abraham%E2%80%93Lorentz_force). Nevertheless, the results we obtain are not 100% consistent.

Regarding the references on the Hessian matrices, we have not managed to get any from @mashl, who initially did the computation. I have however reviewed and summarised his calculation in a report which I am happy to share with you if its is of your interest. I have also included some documentation on the wiggler passmethods as you suggested.

Although the benchmark is not yet finished, I think it is still worth merging this branch, since it fixes many reported bugs with these passmethods, and includes the diffusion matrix computation for the wiggler. Moreover, the tracking in the new passmethods do not introduce significant deviations from the current version (they are smaller than 10^(-8) m).

My internship period at ALBA will finish soon and @ZeusMarti will take over from now.

Kind regards,

Joan

swhite2401:
I agree with the merge, this is a major improvement. However, before that it would be nice to add simple unit tests.
Is your report publicly available on the web? If not adding it, or a summary of it to the documentation pages would be very nice

ZeusMarti:
Hi @swhite2401, by "unit tests" you mean adding an additional test definition in at/pyat/test? What kind of check do you have in mind? a tracking result through the element with and without radiation could be enough?

ZeusMarti:
Sorry for the delay, Joan left in July and then I went on vacations. We spend the last days trying to bench mark the integrators with a Runge Kutta tracking. This was promising at first but finally we were not able to make it precise enough to be sure that the modification in the radiation kicks formula was better so I just removed the modification in the previous commit.

Joan wrote a note that is still being corrected (sorry, we are quite busy with the upgrade simulations). Unfortunately ALBA has now no way to publish open access documentation, is there any interest for us to publish it in arxiv (or similar) once we go though it in detail?
 
Without this modification this branch represents just a couple of bug corrections so I think it is totally save and necessary to merge. 

@swhite2401 @lfarv , any comment before I merge it?

swhite2401:
@ZeusMarti , this is fine for me, I trust you with the calculations. It is also interesting to merge it now, in this way @lfarv can potentially integrate it in #742

ZeusMarti:
Yes, I guess this has implications for #742, the wiggler has its own diffusion matrix calculation included in this PR in function FDW.c.

swhite2401 (approved):
Ok I approve, I trust you on this one

PR #764, opened 2024-05-02
make_copy decorator not working for lattice defined as list

swhite2401:
The make copy decorator in lattice/utils.py accepts list but uses the function ring.replace() that only works for lattice objects.
This was causing problems in the disable_varelem function in tracking/utils.py

Two corrective actions are proposed:
-disable element only when they vary and their passmethod is not already identitypass
-catch error in case the lattice is a list and do not use the replace() function


lfarv (approved):
not very elegant, but it works !

swhite2401:
I agree... you have better ideas?

lfarv:
None better, it does not really matters…

swhite2401:
OK I merge

PR #773, opened 2024-05-30
Add momap multirefpts

oscarxblanco:


oscarxblanco:
This is a work in progress proposal to calculate the momentum aperture of a ring where particles from all reference points are tracked at the same time, while the energy offset step is reduced on every iteration.

It is expected to be faster due to the possibility of tracking all particles at once, and the reduced number of iterations on the stable energy boundary search.

It differs from the acceptance method where every refpts is tracked through separated call to tracking functions.

swhite2401:
Hello @oscarxblanco very interesting development, this was in fact on my todo list to allow LT calculation with GPU from @JeanLucPons.

To be noted: in his improved implement of at.c @JeanLucPons has introduced tracking from a starting refpts, but since this is not yet distributed your solution is quite nice.

First comment: could you separate functionalities? Especially the initial tracking to move all particle to a single refpts is extremely useful. If it could be use separately we could use its output to feed the functions already available in acceptance.

oscarxblanco:
Citing the work done by @ZeusMarti here #92.

JeanLucPons:
Hello,

As said by Simon, having the way to having all points "projected" (tracked) up to the end of the turn is a nice idea and may be integrated in C code. 

More details on GPU code:
For the GPU code, starting refs are handled by doing modulos on particle indices at the tracking level, the overhead is clearly not visible. However to get best gpu performance there are some constraints to fill GPU cores usage that you don't have on CPUs.
There no are no re-arrangement of particles when some of them are lost and an extra modulo is needed to get back refpts of elements where particle are lost (if needed).
The re-arrangement of stride is a tricky task not yet done and can bring benefit only when the number of stride exceed number of GPU cores (I mean actual GPU cores, not CUDA cores). At the moment, a GPU stride is automatically freed when all its particles are lost.

                          "    tracking_starts: numpy array of indices of elements where tracking should start.\n"
                          "       len(tracking_start) must divide the number of particle and it gives the stride size.\n"
                          "       The i-th particle of rin starts at elem tracking_start[i/stride].\n"
                          "       The behavior is similar to lattice.rotate(tracking_starts[i/stride]).\n"
                          "       The stride size should be multiple of 64 for best performance.\n"

oscarxblanco:
Dear @JeanLucPons in my opinion particle losses is a must also in the GPU version.

I worked yesterday on the tracking of particles from multiple points to a single reference point. It seems to work, but, I realized that there are two cases: 
1) The information of the starting reference point need to be preserved.
2) The information of the starting reference point may be removed.
I've been thinking on how to cover both in a simple way, but, at the moment I would need to create a flag called 'group' to reduce case 1) to case 2).

I will push later today a new version when I feel satisfied by the tests. 

The version I'm writing works on python. As you said, it could be implemented in directly in C but right now I don't see the advantage for a CPU because it is still a process where you need to iterate over the reference points either in C or python. I might be missing some memory allocation control where C could be of advantage.

swhite2401:
@oscarxblanco moving all the particles to a single point allows to track them all together in a single ring.track() call, no iteration is needed after that. This is the ideal case for parallelization whether it is GPU/CPU python or C.

A simple solution concerning your 2 cases is to track the same number of particles per refpts, it is then easy to split the results at the end of your tracking

oscarxblanco:
> @oscarxblanco moving all the particles to a single point allows to track them all together in a single ring.track() call, no iteration is needed after that. This is the ideal case for parallelization whether it is GPU/CPU python or C.

I agree,  all comments I did to @JeanLucPons  above refer to the tracking of multiple reference points to a single reference point in a function I called `projectrefpts` which is still under debugging.

> A simple solution concerning your 2 cases is to track the same number of particles per refpts, it is then easy to split the results at the end of your tracking

Right, I think the splitting or grouping should be provided inside the function because it prevents user mistakes during the reshaping.

JeanLucPons:
It would be nice to have the same interface for both the CPU and GPU tracking functions.
If you have the startrefpts vector then you can call projectrefpts() and then track as before.
Then I can also remove the modulo in the GPU tracking code.
Adding the limitation that Simon suggests is nice because it also helps to cover the GPU constraints.
It will just return a survived turn map with a +/- 1 turn offset (don't think it is a big issue) and lost_at_element indices stay the same as in the input ring which is intuitive.
This is my humble opinion but i may missed something.

swhite2401:
No idea...@lfarv do you know what could be the issue?

oscarxblanco:
Great, that solved the building issue. Thank you @lfarv

lfarv:
@oscarxblanco:
By the way, in your new functions, you declare "ring" as a `list`. It should rather be declared as `Lattice`, since you are using may `Lattice` methods (`ring.uint32_refpts`, `ring.radiation_parameters`, `ring.find_orbit`…).

oscarxblanco:
@lfarv , noted. It will be solved in the next push.

JeanLucPons:
> > It would be nice to have the same interface for both the CPU and GPU tracking functions.
> 
> Dear @JeanLucPons , do you mean the whole momentum aperture function? or the `projectrefpts` function ?
> 

Dear @oscarxblanco 
Yes I speak about the `projectrefpts` function.
IMHO it would be nice to have this feature available directly in the `ring.track` function allowing parallel tracking of particles starting at different refpts independently of the type of calculation you want to make.
if ``startrefpts`` is not set nothing change.

oscarxblanco:
Dear @JeanLucPons ,
a second possibility would be to have (6,NxR) particles and do modulo arithmetic when the 'startrefpts' argument is passed to `lattice_track`.

If you want to make CPU and GPU compatible, what is more convenient inside `lattice_track` ? (6,R) and one particle per reference point ? or (6,NxR) and do modulos ?

JeanLucPons:
Yes I think it could be a good idea to move ``projectrefpts`` to track.py.
I don't think that modulos are needed (I do this at the moment but your solution is simpler)

The idea is:

Tracking using one pre-``projectrefpts`` and one post-``projectrefpts`` calls:
```
startingElem = np.array([0,100,200,300,400,500,600,700],dtype=np.uint32)
rout, *_ = ring.track(rin, nturns=nturns, startrefpts=startingElem)
```
Limitations:
- The  size of ``startingElem`` (R) should divide the total number of particle in ``rin``, it gives N.
- Each particle group contains N particles.
In case of len(startingElem) does not divide the total number of input particle an exception should be thrown.

Standard tracking (all particle start at first element):
```
rout, *_ = ring.track(rin, nturns=nturns)
```

Potential issue: when several observation points are given, it may exists "hole" that should be filled but how ?

swhite2401:
Hello @oscarxblanco , @JeanLucPons ,

Maybe it is needed to take a little step back back, the PR seems to introduce 2 new features, please correct me if I misunderstood something:
-a dicotomic search of the chaotic boundary. This method is faster but cannot profit from parallelization and is known to overestimate the aperture which is fine for me but has to be clearly documented.
-for the lifetime calculation it proposes to pre-track particles to projects them at a single observation point. This is interesting to allow for LT calculation with a single call to the C tracking engine and simplifies CPU/GPU calculations. This is nevertheless only useful for lifetime calculation, at least I cannot think of other applications....

This leads to several questions:
-Should we integrate the dicotomic boundary search in the existing aperture module as another grid method?
-Should we integrate something that is only used for LT calculations in the general tracking function?

For info, I had in mind to completely refactor the aperture sub-package to optimize it for GPU. This includes the multi-refpts mode but also multi-grid for a single refpts (tracking X-Y and X-DP DA in on shot for example). This is a significant development and unfortunately I have had not time to start it for the moment....

My proposal would then be to keep this PR independent in a separate module for the moment as initially proposed, to allow to quickly merge it and then at a later stage fully integrate it in existing modules when we refactor the aperture sub-package.

What do you think?

swhite2401:
Sure but since you need to iterate I can't be fully parallelized as opposed to the case where you just send a vector of particles that you know contains the boundary which I think in the end will always be more efficient providing you have sufficient number of CPU/GPU.

oscarxblanco:
Dear @swhite2401 and @JeanLucPons 
This is ready to review, who should be doing it ?
On my side I have to note that the matlab and python functions have different names.
In AT matlab : `MomAperture_Project2Start`
In pyat : `momaperture_project2start`

Let me know if you prefer to change the names.

Here is a piece of code that could be use to run the test.
```python
"""
This script tests the bipartition algorith developed by Z. Marti
2024may08 oblanco
"""

import at
import numpy
import time

print(f'at.__version__ = {at.__version__}')

refpts = numpy.array(range(100))
ring =  at.load_mat('THERING2024may16.mat')

nturns = 1000

spos = ring.get_s_pos(refpts)

oo, orbit = ring.find_orbit(refpts)

start_time = time.time()
dnp = at.momap_alternative.momaperture_project2start( ring,
                                                refpts=refpts,
                                                nturns=nturns,
                                                dptol=1e-3,
                                                use_mp=True,
                                                verbose=True)
print("--- %s seconds ---" % (time.time() - start_time))
boundary = dnp
```

oscarxblanco:
And here is the matlab test code. I add one to `positions` to skip the Ring Parameters element, this in order to calculate the momentum aperture on the same reference points in both, matlab and python.

```matlab
%%
positions = [1:100]+1;
Nturns = 1000;
tic
[ETN, ETP]=MomAperture_Projecect2Start(THERING,positions,Nturns,1e-3,[],[1e-5 1e-5],true);
toc
```

swhite2401:
Ok thanks @oscarxblanco! I can review it, is this urgent?

JeanLucPons:
Hello @oscarxblanco , @swhite2401,

> -Should we integrate something that is only used for LT calculations in the general tracking function?
> 
> What do you think?

For me the main objective is to have a compatible low level tracking function for GPU/CPU and Matlab/Python.
Having the ``projectrefpts`` related stuff inside the acceptance module is not an issue for me but I need to know in order to remove modulos in the GPU code.

pyat/at/acceptance/momap_alternative.py
```diff
@@ -0,0 +1,350 @@
...
+      >>> multirefpts_track_islost(ring, refpts, energysetpt, orbit, initcoord)
+
+    Parameters:
+      ring: list of elements
+
+    Keyword Arguments:
+      refpts: Selects the locations.
+      energysetpt: enery set point for tracking.
+      orbit: (6,N) orbit to be added to the N refpts.
+      initcoords: (2,N) hor. and ver. transverse offsets in m.
```
swhite2401, line 289: You could maybe call this offsets

pyat/at/acceptance/momap_alternative.py
```diff
@@ -0,0 +1,350 @@
...
+      Lostpart: (N) bool array. True if the particle is lost.
+    """
+    rps = refpts
+    nrps = len(rps)
+    lostpart = numpy.ones((nrps), dtype=bool)
+    zin = numpy.zeros((6, 1, nrps, 1))
+    zout = numpy.zeros((6, 1, nrps, 1))
+    issmall = 1e-6
+    eps = numpy.finfo(float).eps
+    istiny = 100 * eps
```
swhite2401, line 301: Similar comment on these tolerances as in the matlab

atmat/atphysics/TouschekPiwinski/MomAperture_Project2Start.m
```diff
@@ -0,0 +1,193 @@
...
+
+% first track the remaining portion of the ring
+for ii=1:nposs
+    Line=THERING(refpts(ii):end);
+    Rin(:,ii)=orbit(:,ii)+[initcoord(1) 0 initcoord(2) 0 e(ii) 0.0]';
+    Rout(:,ii)=ringpass(Line,Rin(:,ii));
+    Loste(ii)=sign(sum(isnan(Rout(:,ii))));
+end
+ngood1=nposs-sum(Loste);
+Rin1=Rout(:,Loste==0);
```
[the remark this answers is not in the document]
oscarxblanco: Dear @swhite2401 , this should stay because particles in `Rin1` are later filtered through comparison. Only particles differing by some $\epsilon$ are tracked, but, this comparison is valid only if values are not `NaNs`.
See below,
```matlab
ringpass(THERING,Rin1(:,non_rep_poss),nturns,'reuse');
```
where `non_rep_poss`  means non-repeated positions. We may rewrite the variable names to make them more transparent when reading the code.

atmat/atphysics/TouschekPiwinski/MomAperture_Project2Start.m
```diff
@@ -0,0 +1,193 @@
...
+%Rin1(6,:)=0;
+    
+%Just track particles that have survived to the ring end 
+if ngood1==1
+    [~, LOSS1] =ringpass(THERING,Rin1,nturns);
+    Loste(Loste==0)=LOSS1; 
+elseif ngood1>1
+    % search for non numerically similar (100 x eps) particles
+    DiffR=squeeze(std(repmat(Rin1,[1 1 ngood1])-repmat(reshape(Rin1,[6 1 ngood1]),[1 ngood1 1])));
```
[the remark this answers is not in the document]
oscarxblanco: The goal is to avoid tracking particles that do not contribute significantly to the boundary search. For example, two particles starting at different reference points separated by a tiny (e.g. 1 mm) drift. We now that although they are indeed two different places along the accelerator, the particle coordinates at the end of the first turn will be too similar.

You are right is saying that it might be lattice dependent, so, I agree with the optional implementation of the particle filtering.

pyat/at/acceptance/momap_alternative.py
```diff
@@ -0,0 +1,350 @@
...
+      nturns: number of turns to be tracked. Default 1000
+      dptol: energy offset resolution. Default 1e-4
+      add_offset: (2, N) offsets to be added to the transverse coordinates
+        on the N reference points. Default 1e-5 m
+      eu_ini: maximum energy boundary. Default: the rf bucket height
+      orbit: (N,6) offsets to be added on the N reference points.
+        Default, the closed orbit
+      verbose: print in the standard output additional info. Default False
+      use_mp (bool): Default True. See :py:func:.`lattice_track`
+      omp_num_threads (int): See :py:fun:`lattice_track`
```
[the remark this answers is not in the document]
oscarxblanco: Different to the tracking function I wanted the default of this momentum aperture calculation to be `use_mp = True`. However, this option could be removed from the help message if you think it is better to leave the default  `False` behavior already documented in `lattice_track

oscarxblanco:
Dear @swhite2401 , could you check the matlab version ?
it has been modified to track positive and negative energy thresholds in the same iteration. And the $\epsilon$ is now an option.
 Here is an example, with $\epsilon=10^{-9}$.
NOTE: I have not yet used any parallelization function in matlab. I believe this depends on a toolbox installation and having the license to use it. Should I added any test about it?
```matlab
tic
[ETN, ETP] = MomAperture_Project2Start(THERING, ...
        'refpts' , positions, ...
        'nturns', Nturns, ...
        'detole', 1e-3, ...
        'euguess', [], ...
        'troffset', [1e-5 1e-5], ...
        'verbose', true, ...
        'epsilon6D', 1e-9);
toc
```

oscarxblanco:
Dear @swhite2401 ,
I have modified the python and matlab versions of the bipartition algorithm to find the momentum aperture, tracking for multiple turns both positive and negative test energies on a single call on every energy step.

The criteria to track only particles differing by some $\epsilon$ is now an option, and not the default behavior.

Wrt to the parallelization in matlab, it is not implemented. It seems to require the Parallel Computing Toolbox, and at least a test with ``[tf = canUseParallelPool()](https://es.mathworks.com/help/matlab/ref/canuseparallelpool.html#d126e161847). Should I add it ?

Wrt to parallelization in python, all messages in the doc have been removed. The calculation is done in serial, but one could still set `use_mp=True` through keyword arguments that are passed to `at.lattice_track`.

I leave here a call example in python and in matlab. The matlab arguments are not longer positional.
```python
start_time = time.time()
dnp = at.momap_alternative.momaperture_project2start( ring,
                                                refpts=refpts,
                                                nturns=nturns,
                                                dptol=1e-3,
                                                use_mp=True,
                                                epsilon6d=1e-9,
                                                verbose=True)
print("--- %s seconds ---" % (time.time() - start_time))
```
```matlab
tic
[ETN, ETP]=MomAperture_Project2Start(THERING, ...
        'refpts' , positions, ...
        'nturns', Nturns, ...
        'dptol', 1e-3, ...
        'euguess',[], ...
        'troffset', [1e-5 1e-5], ...
        'verbose', true,...
        'epsilon6D', 1e-9);
toc
```

oscarxblanco (commented):
All changes are implemented, except the parallelization in matlab.

pyat/at/acceptance/momap_alternative.py
```diff
@@ -0,0 +1,402 @@
...
+        Other arguments could be passed, check :py:func:`lattice_track`.
+      * This function does a quick search, but, it is succeptible to miss
+        islands of instability due to the varying energy step.
+    """
+    # verboseprint to check flag only once
+    verbose = kwargs.pop("verbose", False)
+    verboseprint = print if verbose else lambda *a, **k: None
+
+    rps = kwargs.pop("refpts", ring.uint32_refpts(range(len(ring))))
+    nrps = len(rps)
```
lfarv, line 68: `refpts` may be a string (element name pattern), or an element class, or a boolean mask. Before considering it as a list of locations, you must convert it with:
```python
rps = ring.get_uint32_index(rps)
```

pyat/at/acceptance/momap_alternative.py
```diff
@@ -0,0 +1,402 @@
...
+
+    Parameters:
+      ring: list of elements
+
+    Keyword Arguments:
+      refpts: Selects the location of coordinates output.
+        See ":ref:`Selecting elements in a lattice <refpts>`"
+      nturns: number of turns to be tracked. Default 1000
+      dptol: energy offset resolution. Default 1e-4
+      euguess: maximum energy boundary. Default: the rf bucket height.
```
lfarv, line 42: Is it energy or momentum ?

pyat/at/acceptance/momap_alternative.py
oscarxblanco: @lfarv , thank you. It has been modified in the last commit
```bash
     rps = kwargs.pop("refpts", ring.uint32_refpts(range(len(ring))))
+    rps = ring.get_uint32_index(rps)
     nrps = len(rps)
```

pyat/at/acceptance/momap_alternative.py
oscarxblanco: @lfarv , you are right in asking, it is momentum.

oscarxblanco: Solved in the last commit
```bash
-      dptol: energy offset resolution. Default 1e-4
+      dptol: momentum offset resolution. Default 1e-4
```

lfarv:
@oscarxblanco: Sorry to intervene again… Still 2 details: I get 2 warnings:

_PEP 8: E302 expected 2 blank lines, found 1: 18
Local variable 'zin' value is not used: 245_

oscarxblanco:
> @oscarxblanco: Sorry to intervene again… Still 2 details: I get 2 warnings:
> 
> _PEP 8: E302 expected 2 blank lines, found 1: 18 Local variable 'zin' value is not used: 245_

Dear @lfarv , all comments are welcome. I still don't know why flake8 missed them.
I did the changes even if I don't see those two warnings.
```bash
@@ -15,6 +15,7 @@ __all__ = ["momaperture_project2start", "projectrefpts"]
 #           Based on the MATLAB implementation by Z.Marti at ALBA
 #           See https://github.com/atcollab/at/pull/773
 
+
 def momaperture_project2start(ring: Lattice, **kwargs: Dict[str, any]) -> numpy.ndarray:
     """
     :py:func:`momap_project2start` calculates the local momemtum aperture.
@@ -242,7 +243,6 @@ def projectrefpts(
     if nparticles == 1:
         use_mp = False
 
-    zin = numpy.zeros((6, nparticles, nrps, 1))
     if groupparts:
```

oscarxblanco:
Dear @lfarv ,
the doc string in python and matlab has been modified to say momentum, not energy. Also the parameter `euguess` is now `dpuguess`

See the last commit.

lfarv (approved):
That's OK for me. Anything else, @swhite2401 ?

swhite2401:
All good for me

PR #774, opened 2024-06-03
Fix a regression bug in atavedata

lfarv:
After merging #678, a crash occurs in `atavedata.m` for some lattices. This is fixed here.


ZeusMarti:
Thanks Laurent, we noticed the same problem when polynomB was defined (in our case it was an accident, it was not supose to be defined for that element) but not a vector.

For me it is ok.

ZeusMarti (approved):

PR #744, opened 2024-02-28
Preserve the case of RingParam attribute names

lfarv:
This corrects a bug reported in #743: the name all unexpected attributes in a Matlab `RingParam` element is converted to lower case in the python `Lattice` object. When saving back to .mat or .m file, the attribute name is restored to a new `RingParam` element by capitalising its 1st letter. So the restored name may differ from the original name.

Now, the names of all unexpected attributes are kept unchanged between the Matlab `RingParam` element and the python `Lattice` object.

PR #778, opened 2024-06-12
Regression bug from PR #774: beam_current lattice attribute not propagated correctly

swhite2401:
This PR fixes a regression bug from #744. Modification in the `lattice_object` initialization resulted in wrong propagation of the `_beam_current` attribute.
This was discovered when performing consecutive `add_beamloading()` calls on the same lattice.
The `_std_attribtues` were modifed to propagate the attribute correctly, the specific case that was failing before is now working properly.

### DIFF
--- a/pyat/test/test_collective.py
+++ b/pyat/test/test_collective.py
@@ -169,6 +169,10 @@ def test_buffers(hmba_lattice):
     for i in numpy.arange(1, nturns):
         dth = numpy.sum(th[i-1,(nturns-i)*ns:]
                         -th[i,(nturns-i-1)*ns:(nturns-1)*ns]) - i*ls  
-        dvbh = numpy.sum(dvbh[i-1,(nturns-i)*2:]
-                         -dvbh[i,(nturns-i-1)*ns:(nturns-1)*2])     
-    assert_allclose([dthm dvbh], 0.0, atol = 1e-9)
+        dvbh = numpy.sum(vbh[i-1, :, (nturns-i):]
+                         -vbh[i, :, (nturns-i-1):(nturns-1)])  
+        dvgh = numpy.sum(vgh[i-1, :,(nturns-i):]
+                         -vgh[i, :,(nturns-i-1):(nturns-1)]) 
+        dvbbh = numpy.sum(vbbh[i-1, :,(nturns-i):]
+                         -vbbh[i, :,(nturns-i-1):(nturns-1)])           
+    assert_close([dth, dvbh, dvgh, dvbbh], numpy.zeros(4), atol = 1e-9)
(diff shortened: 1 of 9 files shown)


swhite2401:
@lcarver there is an issue with the `rotate_table_history` in `atimplib.c`

This dates from a while back.... most likely it was not thoroughly tested. I have restored the old version and it works now. There are some obvious differences in the way the buffers are written. 

I have not time to finish tonight, this is just an update in case you want to have a look later today.

swhite2401:
@lcarver, the present version seems to be working fine but I haven't done extensive tests. In the end, I had to rotate the turnhistory buffer manually, maybe it could be simplified but it is not so important for the moment.

Could you please run you tests in various configurations?
Could you make sure that the Vbeam, Vgen and Vbunch buffers that still use memmove are working properly? We may adopt the same method as in rotate_turnhistory.

Then we would need to add unit tests to make sure that this cannot break again. For this, I would suggest to track with various # of slices, turns and bunches and make sure the turnhistory, and voltage buffers rotate correctly.

swhite2401:
@lcarver I haver added a teste of the circulating buffers as discussed

lcarver:
So this bug was highlighted by jimin seok who was encountering some problems. This PR solves the problem of the segmentation faults with Nturns>1. 

I am truly shocked that this was somehow missed. Typically when making changes to any of the collective effects PassMethods, I run the example scripts to make sure they are all working. In this case, it was not done. The testing clearly needs to be improved so the tests proposed here are a good start. More tests of all CE or multi bunch related parameters will be put in in the future.

In the meantime, all example scripts are working well (and some other previously used scripts for beamloading calculations). Both files from Jimin are now running with no issues. 

So for me it can be merged, but we should add more tests for these modules.

pyat/test/test_collective.py
```diff
@@ -138,4 +137,42 @@ def test_track_beamloading(hmba_lattice, func):
                                         0.000000e+00,  0.000000e+00,
                                         -1.313306e-05, -1.443748e-08]
                                         ), atol=5e-10)
-    
+ 
+
+def test_buffers(hmba_lattice):
+    ring = hmba_lattice.radiation_on(copy=True)
```
lcarver, line 143: enable_6d

lcarver (approved):

lcarver (approved):

swhite2401:
@lcarver you suggestions were implemented, this is is ready for merge on your side?

lcarver:
Yes, for me there was no real changes, I just added a few extra comments in one of the example scripts to make it clear that it was not doing the 'most efficient' way of coding. All good for me.

PR #779, opened 2024-06-13
Fix bug reading atM66 elements from matlab lattice

pcsch:
Dear all, 

when loading a lattice in python from a matlab ".m" file all class names are converted to lower case.
As almost all matlab element creation functions are lower case this works, with the exception of atM66.
Therefore a lattice with an atM66 element raises a warning and the element is not included in the python lattice.

This pull requests adds "M66" with an upper case M to the alias map in at/load/utils.py to correctly read this element.


lfarv (approved):
OK, thanks for the fix.

PR #780, opened 2024-06-17
Urgent: require numpy < 2.0

lfarv:
The release of numpy 2.0 breaks the compatibility with previous versions. In this PR, we react by:
1. correcting the python code to be compatible with both numpy 1.x and 2.0
2. since the C code is not yet adapted to numpy 2.0, adding the requirement `numpy<2.0` to `pyproject.toml`


lfarv:
> will you adapt the C?

We'll have to, but I don't know yet what this implies… `numpy 2.0` what released yesterday and resulted in a crash in the tests!

swhite2401:
Ok let's see

PR #783, opened 2024-06-18
Also scaling families in tune and chromaticity fits

swhite2401:
This PR introduces the possibility to fit tunes and chromaticities by scaling the families instead of adding a delta. This is useful in case a family is defined with several magnet types in order to keep their ratio constant.
The default behavior is unchanged

### DIFF
--- a/pyat/at/matching/globalfit.py
+++ b/pyat/at/matching/globalfit.py
@@ -4,12 +4,26 @@ such as the tune and the chromaticity
 """
 from typing import Optional
 import numpy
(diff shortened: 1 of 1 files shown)


simoneliuzzo:
dear @swhite2401 
this works perfectly for me!

simoneliuzzo (approved):

simoneliuzzo:
The same applies for chromaticity?

swhite2401:
Yes the same applies for chromaticity

simoneliuzzo:
Tested also for chromaticity, I approve!

PR #784, opened 2024-06-18
Introduction of the function plot_RF_bucket_hamiltonian

SebastienJoly:
The function iterates through all the lattice RF cavities to calculate and plot the longitudinal Hamiltonian. The Hamiltonian is computed for a momentum compaction factor up to $\alpha_2$.
It can also be used with harmonic cavities.

It yields figures such as the one below:
image

### DIFF
--- a/pyat/at/plot/standalone.py
+++ b/pyat/at/plot/standalone.py
@@ -233,7 +233,7 @@ def plot_RF_bucket_hamiltonian(ring, ct_range=None, dp_range=None, num_points=40
     eta_delta = eta[0] / 2 + eta[1] / 3 * DP + eta[2] / 4 * DP**2
     T = ring.beta**2 * ring.energy * eta_delta * DP**2
 
-    hamiltonian = 0
+    hamiltonian = T
     # Iteration over all lattice cavities
     for rfcav in ring[RFCavity]:
         Voltage = rfcav.Voltage
@@ -246,8 +246,8 @@ def plot_RF_bucket_hamiltonian(ring, ct_range=None, dp_range=None, num_points=40
         # Second term of the Hamiltonian
         U = Voltage / (2 * numpy.pi * HarmNumber) * \
         (numpy.cos(phi) - numpy.cos(phi_s) + (phi - phi_s) * numpy.sin(phi_s))
-        # Total Hamiltonian
-        hamiltonian += T + U
+        # Add to total Hamiltonian
+        hamiltonian += U
 
     fig, ax = plt.subplots(1)
     lim_range = numpy.max((numpy.abs(hamiltonian).min(), numpy.abs(hamiltonian).max()))
--- a/pyat/at/plot/standalone.py
+++ b/pyat/at/plot/standalone.py
@@ -221,7 +221,7 @@ def plot_RF_bucket_hamiltonian(ring, ct_range=None, dp_range=None, num_points=40
 
     hamiltonian = 0
     # Iteration over all lattice cavities
-    for rfcav in ring.get_elements(RFCavity):
+    for rfcav in ring[RFCavity]:
         Voltage = rfcav.Voltage
         HarmNumber = rfcav.HarmNumber
         TimeLag = rfcav.TimeLag
--- a/pyat/at/plot/standalone.py
+++ b/pyat/at/plot/standalone.py
@@ -240,8 +240,8 @@ def plot_RF_bucket_hamiltonian(ring, ct_range=None, dp_range=None, num_points=40
         HarmNumber = rfcav.HarmNumber
         TimeLag = rfcav.TimeLag
 
-        phi_s = TimeLag * 2 * numpy.pi * ring.get_revolution_frequency() * HarmNumber / ring.beta / clight
-        phi = (numpy.pi - phi_s) + CT * 2 * numpy.pi * ring.get_revolution_frequency() * HarmNumber / ring.beta / clight
+        phi_s = TimeLag * 2 * numpy.pi * rfcav.Frequency / ring.beta / clight
+        phi = (numpy.pi - phi_s) + CT * 2 * numpy.pi * rfcav.Frequency / ring.beta / clight
 
         # Second term of the Hamiltonian
         U = Voltage / (2 * numpy.pi * HarmNumber) * \
--- a/pyat/at/plot/standalone.py
+++ b/pyat/at/plot/standalone.py
@@ -1,8 +1,8 @@
 """AT plotting functions"""
 from __future__ import annotations
 from at.lattice import Lattice, axis_
-from at import RFCavity
-from at.physics.revolution import get_mcf
+from at.lattice import RFCavity
+from at.physics import get_mcf
 from at.constants import clight
 import matplotlib.pyplot as plt
 from matplotlib.axes import Axes
--- a/pyat/at/plot/standalone.py
+++ b/pyat/at/plot/standalone.py
@@ -182,6 +182,8 @@ def plot_RF_bucket_hamiltonian(ring, ct_range=None, dp_range=None,
     r"""Plot the resulting longitudinal Hamiltonian of a ring (defining the RF
     bucket). The Hamiltonian is calculated by summing all the cavities in the
     ring. Harmonic cavities are supported by the function.
+    A perfectly tuned lattice is assumed, the cavities' frequency is nominal
+    and the TimeLag is set in a way ensuring ct=0 for the synchronous phase.
 
     Parameters:
         ring: Lattice description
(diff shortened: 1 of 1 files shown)


pyat/at/plot/standalone.py
```diff
@@ -172,6 +175,112 @@ def plot_sigma(sigma, axis: tuple[str, str] = ('x', 'xp'), scale: float = 1.0,
...
+                      (numpy.sqrt(overvoltage**2 - 1) - numpy.arccos(1 / overvoltage)))
+        ring.enable_6d()
+        dp = numpy.linspace(-2 * rfa, 2 * rfa, num=num_points)
+    else:
+        dp = numpy.linspace(dp_range[0], dp_range[1], num=num_points)
+    CT, DP = numpy.meshgrid(ct, dp)
+
+    hamiltonian = 0
+    # Iteration over all lattice cavities
+    for rfcav in ring.get_elements(RFCavity):
```
lfarv, line 224: Can be simplified as:
```python
for rfcav in ring[RFCavity]:
```

pyat/at/plot/standalone.py
```diff
@@ -172,6 +175,112 @@ def plot_sigma(sigma, axis: tuple[str, str] = ('x', 'xp'), scale: float = 1.0,
...
+        eta[0] = alpha[0] - 1 / ring.gamma**2
+        eta[1] = 3 * ring.beta**2 / 2 / ring.gamma**2 + alpha[1] - alpha[0] * eta[0]
+        eta[2] = -ring.beta**2*(5*ring.beta**2 - 1) / (2 * ring.gamma**2) + \
+        alpha[2] - 2 * alpha[0] * alpha[1] + alpha[1] / ring.gamma**2 + alpha[0]**2 * eta[0] - \
+        (3 * ring.beta**2 * alpha[0]) / (2 * ring.gamma**2)
+
+        eta_delta = eta[0] / 2 + eta[1] / 3 * DP + eta[2] / 4 * DP**2
+
+        # Hamiltonian divided by harmonic number to have U = U(V_rf, h, phi_s)
+        T = numpy.pi * eta_delta * DP**2
```
lfarv, line 247: Maybe I missed something, but it looks like `alpha`, `eta`, `eta_dp` and `T` can be computed only once, outside of the loop on cavities.

SebastienJoly: Indeed!
Thanks for pointing it out, I moved these quantities outside the loop.

pyat/at/plot/standalone.py
```diff
@@ -1,6 +1,9 @@
 """AT plotting functions"""
 from __future__ import annotations
 from at.lattice import Lattice, axis_
+from at import RFCavity
+from at.physics.revolution import get_mcf
```
lfarv, line 5: A detail about conventions: we assume that the internal organisation of subpackages may change at any time. So here we would write:
```python
from at.lattice import RFCavity
from at.physics import get_mcf
```

pyat/at/plot/standalone.py
```diff
@@ -172,6 +175,116 @@ def plot_sigma(sigma, axis: tuple[str, str] = ('x', 'xp'), scale: float = 1.0,
...
+        ring.enable_6d()
+        dp = numpy.linspace(-2 * rfa, 2 * rfa, num=num_points)
+    else:
+        dp = numpy.linspace(dp_range[0], dp_range[1], num=num_points)
+    CT, DP = numpy.meshgrid(ct, dp)
+
+    # Momentum compaction/phase slip factors computed up to third order
+    ring.disable_6d()
+    alpha = get_mcf(ring, fit_order=3, n_step=10)
+    ring.enable_6d()
```
lfarv, line 225: Now that the code is simpler, I come to the main point: I would prefer that the function does not modify its input arguments. Here, if a user enters a lattice with id_6d False, he will find 6d turned on after calling the function, which is unexpected. It would be safer to use copy=True when enabling/disabling 6d (while trying to minimise the number of switches…)

lfarv:
@SebastienJoly:I forgot to say that the function is very nice !

pyat/at/plot/standalone.py
SebastienJoly: Good point! I adapted the modules' import accordingly.

pyat/at/plot/standalone.py
SebastienJoly: I did not know the could return a shallow copy of a ring using this argument, thanks for bringing this to my attention. I reshuffled some parts of my function to compute the momentum compaction factor at the beginning and only use once the disable_6d() method on a shallow copy of the original ring. I agree the function should not modify the ring argument.

lfarv:
@SebastienJoly : ok for the last version, no coming to code style, I get the following PEP8 errors
```
PEP 8: E302 expected 2 blank lines, found 1:178
Parameter 'kwargs' value is not used :179
PEP 8: E122 continuation line missing indentation or outdented :212
PEP 8: E225 missing whitespace around operator :212
PEP 8: E122 continuation line missing indentation or outdented :213
PEP 8: E128 continuation line under-indented for visual indent :218
PEP 8: E502 the backslash is redundant between brackets :224
PEP 8: E128 continuation line under-indented for visual indent :225
PEP 8: E122 continuation line missing indentation or outdented :248
PEP 8: E261 at least two spaces before inline comment :256
PEP 8: E261 at least two spaces before inline comment :258
PEP 8: E122 continuation line missing indentation or outdented :268
PEP 8: E306 expected 1 blank line before a nested definition, found 0:269
PEP 8: E122 continuation line missing indentation or outdented :271
PEP 8: E122 continuation line missing indentation or outdented :275
PEP 8: E231 missing whitespace after ',' :277
PEP 8: E305 expected 2 blank lines after class or function definition, found 1:284
```
I suggest you use a code formatter like [black](https://black.readthedocs.io/en/stable/), or a linter like [flake8](https://flake8.pycqa.org/en/latest/)

pyat/at/plot/standalone.py
```diff
@@ -172,6 +175,112 @@ def plot_sigma(sigma, axis: tuple[str, str] = ('x', 'xp'), scale: float = 1.0,
...
+    # Hamiltonian (H=U+T) divided by harmonic number to have U = U(V_rf, h, phi_s)
+    # First term of the Hamiltonian
+    eta_delta = eta[0] / 2 + eta[1] / 3 * DP + eta[2] / 4 * DP**2
+    T = ring.beta**2 * ring.energy * eta_delta * DP**2
+
+    hamiltonian = T
+    # Iteration over all lattice cavities
+    for rfcav in ring[RFCavity]:
+        Voltage = rfcav.Voltage
+        HarmNumber = rfcav.HarmNumber
```
lfarv, line 240: I started using your function with different lattices, and found that this should be:
```python
    Voltage = rfcav.Voltage * ring.periodicity
    HarmNumber = rfcav.HarmNumber * ring.periodicity
```
Alternatively you can replace
`ring.get_revolution_frequency() * HarmNumber` by` rfcav.Frequency`

lfarv:
I have a doubt on the energy and time references:
- if I modify the `TimeLag` of all cavities by the same amount, nothing changes: it's just a shift of the time reference. So the plot should just shift left of right by the same amount. Here it gets modified.
- Similarly, if I scale the frequency of all cavities by the same amount, the plot should move up or down showing the off-momentum. Here, it's independent of the frequency

I have the feeling that the plot shows what happens on a perfectly tuned lattice, where the frequency is nominal, and where the TimeLag is set by `set_cavity_phase` so that the 0 is on the synchronous phase. That's OK for me, but maybe it should be mentioned in the docstring.

pyat/at/plot/standalone.py
SebastienJoly: Regarding periodicity, isn't it counterintuitive to set the cavity with an harmonic number divided by the periodicity? Meaning it could lead to non integer harmonic numbers in some cases.
However I think this is a more general comment and not specifically related to this pull request.
I will change the code accordingly.

SebastienJoly:
Indeed! This is an important point I did not consider.
I assumed to ct=0 reference to be at the center of the RF bucket, which is related to $\phi_s$. The ct grid is computed based on this assumption thus adding a timelag does not shift the plot but rather impact the $\phi_s$ value and change the Hamiltonian.
The plotting routine is therefore only valid for a perfectly tuned lattice as you said. Now the questions is, would people be interested in the Hamiltonian contours of a not perfectly tuned lattice? If yes I can spend a bit more time to solve these minor issues, if no then I follow you on modifying the docstring to warn the user.

SebastienJoly:
Thank you for all the explanations, the function docstring should be sufficiently clear now!
The request is ready to merge from my point of view.

lfarv (approved):
OK for me!

swhite2401:
@SebastienJoly you don't have merge rights? Please go ahead if the PR is approved

PR #791, opened 2024-06-25
Fix load mat

lfarv:
This fixes a bug reported by @swhite2401 in #766.


lfarv:
When this is approved, I will immediately publish a bug fix release 0.6.1. 
For this, the license in the project metadata is also fixed.

lfarv:
For info, the .mat file contains strange attributes in the RingParam element. @swhite2401, can you remember if this file was created by a recent version?
Anyway these strange attributes are now ignored.

swhite2401 (commented):
Thanks for the quick fix!
Question, can't we just assume that the lattice can have any attribute?

swhite2401:
> For info, the .mat file contains strange attributes in the RingParam element. @swhite2401, can you remember if this file was created by a recent version? Anyway these strange attributes are now ignored.

This is an old version because this is the environment I create for FCC studies a while ago, can remember exactly when

swhite2401 (approved):

lfarv:
>  can't we just assume that the lattice can have any attribute?

Yes, that's what it should be. The problem appeared with "deprecated attributes" which were kept for compatibility as properties. This created the bug and is now solved.

Issue #792, opened 2024-06-26
Bug in Lattice.avlinopt


lfarv:
The results of `atavedata` (Matlab) and `Lattice.avlinopt` (python) are different.
Here are the <&beta;<sub>x</sub>> for the first 10 elements of the test file hmba.mat.

| Matlab | python |
| ------ | ------|
|    6.9000 | 6.9000 |
|    6.9000 | 6.9000 |
|    6.9000 | 6.9000 |
|    **7.2396** | **7.2132** |
|    7.9188 | .9188 |
|    **7.9353** | **7.8489** |
|    7.4451 | 7.4451 |
|    **5.8628** | **7.3854** |
|    4.9281 | 4.9281 |
|    **3.6244** | **4.5013** |
| ------ | -------|

The same appears for <&beta;<sub>z</sub>>.
From a 1st analysis, Matlab results are correct and it seems that the problem is in the average value in **drifts** in python.

ZeusMarti:
Thanks for pointing it out, sorry I did not compare numerically the results, just qualitatively. I'll look at it.

ZeusMarti:
Hi @lfarv ,

I think the problem affects only the drifts (which actually I did not consider in my initial tests):
given this definition:

https://github.com/atcollab/at/blob/cb61c7aa6f1696bb4a6e52399907db62f2e36aca/pyat/at/physics/linear.py#L1141-L1146
https://github.com/atcollab/at/blob/cb61c7aa6f1696bb4a6e52399907db62f2e36aca/pyat/at/physics/linear.py#L1235-L1235

the booleans that select the focusing long elements are including the drifts

https://github.com/atcollab/at/blob/cb61c7aa6f1696bb4a6e52399907db62f2e36aca/pyat/at/physics/linear.py#L1268-L1271

https://github.com/atcollab/at/blob/cb61c7aa6f1696bb4a6e52399907db62f2e36aca/pyat/at/physics/linear.py#L1278

https://github.com/atcollab/at/blob/cb61c7aa6f1696bb4a6e52399907db62f2e36aca/pyat/at/physics/linear.py#L1289

This could be solved by just using  `k=0.0`  instead of `numpy.finfo(numpy.float64).eps` in line 1145. In this case the behavior is the same as in Matlab. Actually in Matlab, the equivalent of `fff` is `foc`:
[...]

PR #793, opened 2024-06-27
Bug issue792

ZeusMarti:
This PR fixes the problem described in issue  #792 .

### DIFF
--- a/pyat/at/physics/linear.py
+++ b/pyat/at/physics/linear.py
@@ -1139,7 +1139,7 @@ def avlinopt(ring, dp, refpts, **kwargs):
         :py:func:`linopt4`, :py:func:`get_optics`
     """
     def get_strength(elem):
-       try:
+        try:
             k = elem.PolynomB[1]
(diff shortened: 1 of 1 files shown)


pyat/at/physics/linear.py
```diff
@@ -1142,7 +1142,7 @@ def get_strength(elem):
         try:
             k = elem.PolynomB[1]
         except (AttributeError, IndexError):
-            k = numpy.finfo(numpy.float64).eps
+            k = 1e-7
         return k
```
lfarv, line 1146: I would rather set:
```python
try:
    k = elem.PolynomB[1]
    if k < 1.e-7:
        k = 0.0
except (AttributeError, IndexError):
    k = 0.0
return k
```
to avoid having drifts having a small k and using `betafoc`

ZeusMarti: sure, sorry I totally miss-understood your comment.

lfarv (approved):
OK

PR #794, opened 2024-07-07
New test of the average optics values

lfarv:
This PR restores the comparative tests between python and Matlab in the `pyat/test_matlab` directory, which was failing for the average beta and dispersion checks.

Unfortunately, this test cannot be run in GitHub actions because a problem of Matlab license, so the problem was not immediately spotted. We introduce here a new test in the "Matlab tests" script, which is run on GitHub, to prevent any further problem.

The introduction of this new test revealed a number of residual errors in the computation of average values. Among others:
- dipoles without gradient were considered as drifts, ignoring the intrinsic horizontal focusing (1/&rho;<sup>2</sup> term). This has a small effect on &beta;<sub>x</sub>, but more on the dispersion,
- the inverse radius effect was applied for dispersion on both horizontal and vertical planes. This results in a wrong vertical dispersion term,
- the dispersion derivative &eta;’ term was wrong (index error),
- the entrance pole face angle in dipoles was ignored in python, but applied in Matlab,
- more important: since the last release, defocusing magnets were wrong: the `k < 1.e-7` test should be `abs(k) < 1.e-7` (my fault),
- other small bugs

However all these ended up in relatively small errors. But the making it completely correct implied rather large modifications: mainly because we cannot assume that we have either focusing in both planes, or no focusing at all. Planes must now be dissociated.

The main improvement here is that the average values will be automatically checked on each commit.

@ZeusMarti: in your 1st version, you showed a validation, mainly focused on quadrupoles (which were right). Maybe you could you check that this is still ok ?

### DIFF
--- a/.github/workflows/matlab-tests.yml
+++ b/.github/workflows/matlab-tests.yml
@@ -46,7 +46,7 @@ jobs:
       run: python -m pip install .
 
     - name: Run tests
-      uses: matlab-actions/run-command@v1
+      uses: matlab-actions/run-command@v2
       with:
         command: githubrun
 #
--- a/atmat/atphysics/atavedata.m
+++ b/atmat/atphysics/atavedata.m
@@ -109,7 +109,5 @@ end
         avedisp(:,[2 4])=(disp0(:,[2 4]).*(sin(sqrt(K).*L)./sqrt(K))-...
             disp0(:,[1 3]).*(1-cos(sqrt(K).*L))+...
             ir.*(1-cos(sqrt(K).*L))./K)./L;
-        disp([disp0 ir K L]);
-        disp(avedisp);
     end
 end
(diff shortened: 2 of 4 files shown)


ZeusMarti:
Results look OK for me,

matlab_atavedata_test_errors_alba2

ZeusMarti:
I compared the result with a numerical average slicing the elements, it gives errors  at the 0.05% with up to 800 slices for each element. I haven't tried slicing further bit it seems to monotonically improve.
matlab_atavedata_test_diff

lfarv:
Hi @ZeusMarti: this looks nice. It it's OK for you, can you approve the PR? Hoping it's the end of a long story! By the way, you plots are very nice.

ZeusMarti (approved):
OK for me!

PR #802, opened 2024-07-22
use _Complex_I instead of I

swhite2401:
This PR is a workaround regarding issue #800. Following compilation issues, the I macro is replaced with _Complex_I in `atimplib.c`.
This is a **temporary** fix that will need to be looked at in more details as it is for the moment not understood and could be a symptom for more severe problems.


lfarv (approved):
OK!

swhite2401:
@TeresiaOlsson ok for you?

TeresiaOlsson (approved):
Also ok for me.

PR #806, opened 2024-07-29
New matching

lfarv:
This introduces the new matching using the new Variable definition and the newly introduced Observables.

Tests are implemented for both the old and new matching methods.


pyat/at/latticetools/matching.py
```diff
@@ -0,0 +1,110 @@
...
+        dp:             Momentum deviation.
+        dct:            Path lengthening.
+        df:             Deviation from the nominal RF frequency.
+        **kwargs:       Keyword arguments sent to
+          :py:func:`~scipy.optimize.least_squares`
+    """
+    if copy:
+        # Check that there is no ElementVariable
+        for var in variables:
+            if isinstance(var, ElementVariable):
```
swhite2401, line 104: Why do you have to impose this constraint now?.... this is a bit problematic as it was allowed in the past

swhite2401 (commented):
@lfarv I had a quick look and this seems to be very nice. I like the new syntax. Now I have to test this against real matching studies which will take time.

One thing that you may want to integrate are the response matrices that we had in there at some point, or maybe in another PR? Up to you, but it would be very useful to have these RM corrections available

pyat/at/latticetools/matching.py
lfarv: As you pointed out previously, an ElementVariable does not survive a copy. ElementVariables are allowed for in-place matching, but for copy matching, you need RefptsVariable.

swhite2401:
It would be useful to provide additional GlobalObservable like max/min, averages

swhite2401:
Allow to use "greater than" and "lower than" operators for the target would be extremely useful

lfarv:
Hi @swhite2401 

Did you go through the notebooks on variables and observables ? They show how to partially answer you questions:

> It would be useful to provide additional GlobalObservable like max/min, averages

Each Observable has this functionality to provide statistics on the results at different locations: examples from the [Observables notebook](https://atcollab.github.io/at/p/notebooks/observables.html):
```python
LocalOpticsObservable(at.Monitor, "beta", plane="v", statfun=np.amax). #  maximum of vertical beta on all monitors
LatticeObservable(at.Sextupole, "H", statfun=np.mean). # Average of all the sextupole strengths
```
Making statistics on unrelated quantities does not seem so useful, but if you have a realistic example, one can look at it.

> Allow to use "greater than" and "lower than" operators for the target would be extremely useful

Another example from the notebook:
```python
LocalOpticsObservable(at.Monitor, "beta", plane=1, target=7.0, bounds=(-np.inf, 0.0)
```
The bounds define an interval **around** the target. So here this means  beta <= 7.0

Most of what you need to perform a match is defined in variables and observables, but I agree that a notebook specific for matching is useful. If you have a realistic example, I can build a notebook.

swhite2401:
Ok thanks, in fact I may have gone through all this a bit too fast but this is a lot at once... and the link to matching is not obvious at first sight. btw I build the documentation page on readthedocs

I will continue this exercise and I can help building a step by step user guide. I may add my comments on syntax or possible improvements directly in it for the discussion.... this other reviewers will not need to start from zero

lfarv:
@swhite2401 : your help for documentation is welcome !

swhite2401:
> The bounds define an interval **around** the target. So here this means beta <= 7.0

I know this, this was already the case in the old version, but this is not very intuitive and shold be improved

swhite2401:
The residuals do not seems to be claculated corectly... I tried to look at the source but could not find the reason for the moment

The evaluate function does nopt return any value, this is also not very intuitive, it would be nice that it return constraints.value

swhite2401:
There is an issue with transfer line matching: it still tries to compute a closed orbit even though twiss_in is provided

swhite2401:
> The residuals do not seems to be claculated corectly... I tried to look at the source but could not find the reason for the moment

Ok now I understood, it would be nicer to print out the deviation rather than the residual in the summary table, this is more "human readable"

swhite2401:
Great thanks! Can I pull the latest version?
I saw this warning too... I agree this can be handled in a separate PR.

I will provide more inputs as they come...

lfarv:
> Can I pull the latest version?

Yes, please try it.

swhite2401:
@lfarv I just did a git pull on this branch to have the latest version but this resulted in a merge and now my local branch is ahead by 13 commits... I think your rebase action is the reason as I did not pull the master branch first....

This is not very nice in my opinion, merging the master into the dev branch appears to be a cleaner way to handle these things, no?

swhite2401:
> @lfarv I just did a git pull on this branch to have the latest version but this resulted in a merge and now my local branch is ahead by 13 commits... I think your rebase action is the reason as I did not pull the master branch first....
> 
> This is not very nice in my opinion, merging the master into the dev branch appears to be a cleaner way to handle these things, no?

I did a hard reset to origin and it is fine now.... not sure what happened in fact

lfarv:
@swhite2401: I don't think the rebase can have any effect on the problem you got. Your master branch does not need to be up-to-date to update the matching branch, you don't even need to have master checked out.

swhite2401:
Hello @lfarv, for me we can merge this I have been using it for few projects and it seems to work fine. Now I have to keep using it for further improvements., however it would be much more convenient to have this on the master.

I haven't had time to work on the docs that  promised... maybe we can include it later on?

lfarv:
@swhite2401 : I agree to merge it as it is and wait for the doc: at least it's a step forward. Can you approve?

swhite2401 (approved):

PR #809, opened 2024-08-02
Add tune resonances plot

oscarxblanco:
Dear all,
this PR is a work in progress.
It implements a tune resonance plot mentioned in the discussion #808 .
Here is an example of the default plot.
```python
import at, at.plot
at.plot.standalone.plot_tune2D_resonances()
```
image
Here is an example with fifth orders resonances in symmetry 7.
```python
at.plot.standalone.plot_tune2D_resonances(orders=[5],period=7,window=numpy.array([71 ,73, 45, 48]))
```
image
This is similar  to the result of `reson` from MML.
```matlab
reson(5,7,[71 73 45 48])
```
image

I think this could be useful, would you consider to add it @swhite2401 ?
I have put it for the moment in at.plot.standalone. Is this the right place @swhite2401 ?


lfarv:
Hi @oscarxblanco,
Nice addition ! I think that `at.plot.standalone` becomes larger and larger, with functions unrelated to each other. Why don't you put your function in its own module?

swhite2401:
@lfarv, sorry to go a bit off topic but it has always been assumed that importing matplolib introduces significant performance degradation.
This in return introduces counter intuitive import a plot libraries, that have already been raised by @lnadolski (can't find the discussion back...).
The question is: has this been evaluated? All the other code I can think off import matplotlib at initialization so I am wondering if we are not adding complexity for marginal gain...

lfarv:
@swhite2401:  There are in fact 2 questions raised by your comment:

1. `matplotlib` is an **optional** dependency of PyAT. You don't get it installed with:
   ```python
   pip install accelerator-toolbox
   ```
   But you get it installed with:
   ```python
   pip install "accelerator-toolbox[plot]"
   ```
   This is intended to minimise the dependencies when PyAT is installed in a virtual environment to power a virtual accelerator, for instance. Here is a comparison of the installation of PyAT in a fresh virtual environment, with and without the `[plot]` modifier:
   - without plot
     ```
     (noplot)  laurent@garou ~ % pip install accelerator-toolbox
     ...
     (noplot)  laurent@garou ~ % pip list
     Package             Version
     ------------------- -------
     accelerator-toolbox 0.6.1
     numpy               2.0.1
     pip                 24.1.2
     scipy               1.14.0
     setuptools          70.3.0
     (noplot)  laurent@garou ~ % 
     ```
   - with plot
     ```
     (plot)  laurent@garou ~ % pip install "accelerator-toolbox[plot]"
     ...
     (plot)  laurent@garou ~ % pip list
     Package             Version
     ------------------- -------
     accelerator-toolbox 0.6.1
     contourpy           1.2.1
     cycler              0.12.1
     fonttools           4.53.1
     kiwisolver          1.4.5
     matplotlib          3.9.0
     numpy               2.0.1
     packaging           24.1
     pillow              10.4.0
     pip                 24.1.2
     pyparsing           3.1.2
     python-dateutil     2.9.0.post0
     scipy               1.14.0
     setuptools          70.3.0
     six                 1.16.0
     ```
   This makes sense and can be kept, but the installation instructions should be improved to explain the use of the `[plot]` modifier.


2. `at.plot` is **not automatically imported** when importing `at`. Here the only question is the import time. I agree that this should not be a problem, and that we could import `at.plot` by default.

So my proposal is:
- keep the installation process as it is, with improved documentation for the `[plot]`, `[dev]`, `[mpi]` and `[doc]` modifiers,
- import `at.plot` by default. If `matplotlib` has not been initially installed, this just prints a message saying: _'Plotting functions require matplotlib, to use them run "pip install matplotlib"'_. This message is harmless for a non-graphics installation, and has no time penalty.

swhite2401:
@lfarv, if we allow importing matplotlib, each module could now host its own plotting function, right? Then we don't have to pile-up things in `at.plot.standalone`.
Maybe this PR could be used to start this new organization?

oscarxblanco:
Dear @lfarv and @swhite2401 ,
this PR is ready for review, who should do it?
I have also included the change you mentioned to import at.plot by default.

lfarv:
@oscarxblanco 

>  if we allow importing matplotlib, each module could now host its own plotting function, right? Then we don't have to pile-up things in at.plot.standalone.
> Maybe this PR could be used to start this new organization?

You are right, there is no need to pile up things in `standalone.py`. The point is to keep all graphics functions in a single package `at.plot`, so that a single check for the presence of matplotlib in `at.plot.__init__.py` is enough to skip the whole package. No need to check in each module.

Concerning this PR, it's better keeping it centered on the resonance plot, as it is, avoiding mixing other topics. I would have put the modification of `at.__init__.py` in another PR, as discussed above. But since it's done, keep it, don't worry !

pyat/at/plot/resonances.py
```diff
@@ -0,0 +1,505 @@
...
+        dfarey = int(daux)
+        farey.append(cfarey / dfarey)
+        fracfarey.append(Fraction(cfarey, dfarey))
+    verboseprint(f"farey_float{nthorder}= {farey}")
+    verboseprint(f"farey_frac{nthorder} = {fracfarey}")
+    return farey, fracfarey
+
+
+def plot_tune2d_resonances(
+    **kwargs: dict[str, any],
```
lfarv, line 61: Most of the keyword arguments have an immutable default value (period, verbose, includelegend, block,…). In such a case, it's better to declare them in the function declaration, if possible with a type hint. Then they appear with their type and default value in the documentation, in the signature of the function.

pyat/at/plot/resonances.py
```diff
@@ -0,0 +1,505 @@
...
+    10th to 15th: RGB increased in steps of 0.1, width 1
+
+    Custom Style:
+    You could pass a custom line style in a dictionary as
+    custom_linesty = mydictionary,
+    where mydictionary should contain two entries
+    dict(0: prop_normal, 1: prop_skew).
+    prop_normal and prop_skew are also dictionaries starting at zero,
+    i.e. the_nth_order-1. Each entry contains the line properties of the nth-1
+    resonance to plot.
```
lfarv, line 114: Why not simply `linestyles` for the keyword? (by the way, in the code, you use `customlinesty` without underscore)

Then you could use "normal" and "skew" for the keys of the top-level directory, and "order" instead of "order-1" for the keys of the subdirectories;

pyat/at/plot/resonances.py
```diff
@@ -0,0 +1,505 @@
...
+        8: prop9s.copy(),
+        9: prop10s.copy(),
+        10: prop11s.copy(),
+        11: prop12s.copy(),
+        12: prop13s.copy(),
+        13: prop14s.copy(),
+        14: prop15s.copy(),
+    }
+    # assemble final dict
+    defaultlprop = {0: propn.copy(), 1: props.copy()}
```
lfarv, line 432: The default lifestyle directory could be initialized at the module level, so that it's done once for ever at import time. And are all the `copy()` necessary ?
However, it's a detail, I don't care too much…

lfarv:
I'd suggest to simplify the keywords:
- "legend" instead of "includelegend"
- "linestyles" instead of "custom_linesty"
- "debug" instead of "debugverbose"

Any advice, @swhite2401 ?

swhite2401:
> I'd suggest to simplify the keywords:
> 
>     * "legend" instead of "includelegend"
> 
>     * "linestyles" instead of "custom_linesty"
> 
>     * "debug" instead of "debugverbose"
> 
> 
> Any advice, @swhite2401 ?

I agree on the simpler / shorter keywords. What @lfarv proposes looks fine to me

lfarv:
And what about `plot_tune_diagram` ?

I have another question about systematic and non-systematic resonances. As the function is now, you have to alternate 2 plots:
- if you set period = _machine_perodicity_,  you get only the systematic resonances,
- if you set period = 1, you get both systematic and non-systematic ones, but you cannot distinguish systematic ones.
Could there be a line property allowing to enhance systematic resonances? The line width is already used for the order…

Anyway, if there is no obvious solution, forget it, we take the function as it is, it can be improves later !

oscarxblanco:
> @oscarxblanco
> 
> > if we allow importing matplotlib, each module could now host its own plotting function, right? Then we don't have to pile-up things in at.plot.standalone.
> > Maybe this PR could be used to start this new organization?
> 
> You are right, there is no need to pile up things in `standalone.py`. The point is to keep all graphics functions in a single package `at.plot`, so that a single check for the presence of matplotlib in `at.plot.__init__.py` is enough to skip the whole package. No need to check in each module.
> 
> Concerning this PR, it's better keeping it centered on the resonance plot, as it is, avoiding mixing other topics. I would have put the modification of `at.__init__.py` in another PR, as discussed above. But since it's done, keep it, don't worry !

Dear @lfarv , I have removed the .plot import from __init__.py. I think you are right about having a separted PR because it will be clear when the change was introduced.

pyat/at/plot/resonances.py
oscarxblanco: Done.

pyat/at/plot/resonances.py
oscarxblanco: Done.

pyat/at/plot/resonances.py
oscarxblanco: All call to copies where removed. They were not really needed.
Also, the dictionary was moved out of the function.

oscarxblanco:
> I'd suggest to simplify the keywords:
> 
>     * "legend" instead of "includelegend"
> 
>     * "linestyles" instead of "custom_linesty"
> 
>     * "debug" instead of "debugverbose"
> 
> 
> Any advice, @swhite2401 ?

Done.

oscarxblanco:
> And what about `plot_tune_diagram` ?
> 
> I have another question about systematic and non-systematic resonances. As the function is now, you have to alternate 2 plots:
> 
>     * if you set period = _machine_perodicity_,  you get only the systematic resonances,
> 
>     * if you set period = 1, you get both systematic and non-systematic ones, but you cannot distinguish the systematic ones.
>       Could there be a line property allowing to enhance systematic resonances? The line width is already used for the order…
> 
> 
> Anyway, if there is no obvious solution, forget it, we take the function as it is, it can be improves later !

Dear @lfarv ,
maybe something like this ?
systematic resonances in variable width, normal are continuous (-), skew are dashed (- -).
non-systematic resonances in width=1, normal are dashdot (-.), skew are dotted (. . . ).
image

oscarxblanco:
Another example,
image

pyat/at/plot/resonances.py
```diff
@@ -0,0 +1,505 @@
...
+    9: prop9s,
+    10: prop10s,
+    11: prop11s,
+    12: prop12s,
+    13: prop13s,
+    14: prop14s,
+    15: prop15s,
+}
+# assemble final dict
+defaultlprop = {0: propn, 1: props}
```
lfarv, line 251: Still: I think it would be more clear to have
```python
defaultlprop = {"normal": propn, "skew": props}
```

oscarxblanco: Done.

lfarv:
Your last plots look nice: that would be perfect.

A last detail: your 'period' argument corresponds to the `periodicity` attribute of the lattice object. Could you switch to the same name? Sorry to intervene on such details, but I'm trying to get uniform naming across all modules…

I tried your function in a Jupyter notebook, and I get a strange thing: when validating the cell, I get two identical figures instead of one. I've never seen this. Did you try that? Could it be a problem with my setup?

oscarxblanco:
Dear @lfarv ,
it sounds like something related to this thread
https://stackoverflow.com/questions/62322083/why-on-calling-plt-figure-in-jupyter-notebook-two-figure-appear

Wrt the names, there is no problem, I will do the changes.

I think that it might be better to have the option to use axes created by the user, similar to synoptics. It will allow to plot and and the lines.

oscarxblanco:
> Your last plots look nice: that would be perfect.
> 
> A last detail: your 'period' argument corresponds to the `periodicity` attribute of the lattice object. Could you switch to the same name? Sorry to intervene on such details, but I'm trying to get uniform naming across all modules…
> 
> I tried your function in a Jupyter notebook, and I get a strange thing: when validating the cell, I get two identical figures instead of one. I've never seen this. Did you try that? Could it be a problem with my setup?

I did the change about periodicity.
Wrt to jupyter I will have to install it and test.

oscarxblanco:
> Your last plots look nice: that would be perfect.

Dear @lfarv , I have modified the function to overlay plots using the same axes, and I have added options to slightly modify the default line properties to avoid the need for the user to create another complete custom dictionary.

With those modifications one could overlay the plot in symmetry $n$ to a plot in symmetry $1$.
Here is an example using dots, width of 1, and black color.
```python
>>> import at, at.plot
>>> ax = at.plot.resonances.plot_tune_diagram(periodicity=3,window=[0,0.45,0,0.48])
>>> ax2 = at.plot.resonances.plot_tune_diagram(axes=ax,linewidth=1,linecolor='k',linestyle='dots')
```
image
And here another example, the one I showed yesterday with only third order resonances.
```python
ax = at.plot.resonances.plot_tune_diagram(orders=3,periodicity=3,window=[23,23.5,40,40.5])
vv = at.plot.resonances.plot_tune_diagram(orders=3,axes=ax,linewidth=1,linestyle='dots')
```
image

oscarxblanco:
Dear @lfarv ,
I have tried to reproduce the behavior of the double plot you see using jupiterlab. In fact, I don't get a double plot but I have the impression that jupyterlab creates its own figure and forces the plots to be visible, ignoring what is in the function.

I manage workaround it by adding another flag called `show` set to `True` by default in order to get the plot in any python3 session, and should be set to `False` when used in jupyterlab. When working in jupyterlab one should create the figure, set the axes limits and then use those axes for the plot.

Here is a screenshot and the code:
```python
fig = plt.figure();
ax = fig.add_subplot(111);
ax.set_xlim([0,0.45]);
ax.set_ylim([0,0.45]);
ax, hs, ls = at.plot.resonances.plot_tune_diagram(orders=3,axes=ax,periodicity=3,window=[23,23.5,40,40.5],legend=True,addtolabel=" syst.",show=False);
vx, h, l = at.plot.resonances.plot_tune_diagram(orders=3,axes=ax,linewidth=1,linestyle='dots',legend=True, addtolabel=" non syst.",show=False);
```
image

pyat/at/plot/resonances.py
```diff
@@ -0,0 +1,420 @@
...
+    linecolor = kwargs.pop("linecolor", None)
+    linewidth = kwargs.pop("linewidth", None)
+    addtolabel = kwargs.pop("addtolabel", None)
+    defaultlprop = create_linepalette(
+        linestyle=linestyle,
+        linecolor=linecolor,
+        linewidth=linewidth,
+        addtolabel=addtolabel,
+    )
+    lprop = kwargs.pop("linedict", defaultlprop)
```
lfarv, line 336: A suggestion (forget it if you find it irrelevant):
```python
    lprop = create_linepalette(
         linestyle=linestyle,
         linecolor=linecolor,
         linewidth=linewidth,
         addtolabel=addtolabel,
     )
    userprop = kwargs.pop("linedict", {})
    lprop["normal"].update(userprop.get("normal", {}))
    lprop["skew"].update(userprop.get("skew", {}))
```
The present behavior is kept, but it allows the user to overwrite only the lines he wants to change, others are unchanged.

pyat/at/plot/resonances.py
```diff
@@ -0,0 +1,420 @@
...
+    miny = numpy.floor(the_axeslims[1, 0])
+    miny = miny - periodicity - numpy.mod(miny, periodicity)
+    maxy = numpy.ceil(the_axeslims[1, 1])
+    maxy = maxy + periodicity - numpy.mod(maxy, periodicity)
+    minmaxydist = maxy - miny
+
+    linestyle = kwargs.pop("linestyle", None)
+    linecolor = kwargs.pop("linecolor", None)
+    linewidth = kwargs.pop("linewidth", None)
+    addtolabel = kwargs.pop("addtolabel", None)
```
lfarv, line 329: These could also be moved to the function declaration (better documentation).

lfarv:
> I have tried to reproduce the behavior of the double plot you see using jupiterlab. In fact, I don't get a double plot but I have the impression that jupyterlab creates its own figure and forces the plots to be visible, ignoring what is in the function.

Sorry the my late reply: what I get is described in the link you posted: if I don't store the function return value in a variable, Jupyter notebook displays the function result, which creates another figure. This is the normal notebook behavior. By the way, it disappeared with the new return values. So your "show" keyword is not necessary, though it can be kept, of course.

pyat/at/plot/resonances.py
oscarxblanco: Done.

pyat/at/plot/resonances.py
oscarxblanco: Done.

lfarv (approved):
OK for me, nice addition!

swhite2401:
Hell @oscarxblanco ,
On my python3.8 environment at crashes with the following error:
image

it seems ` from __future__ import annotations` is missing in `resonance.py`, at least it fixes the problem for me.

How could this pass the tests? This is a bit puzzling to me...

Should I revert and implement the fix?

oscarxblanco:
I'll fix it. I don't know how it passed that test. I work on python 3.11.2 and didn't notice it.

PR #812, opened 2024-08-11
Always attempt to import at.plot

lfarv:
As discussed [here](https://github.com/atcollab/at/pull/809#issuecomment-2268502581), `at.plot` had to be explicitly imported to access plotting functions and methods. 

Now `at.plot` will be recursively imported when importing `at`. It is still possible to run at without matplotlib: in such a case, an explicit message will be printed on importing `at`:

_matplotlib is unavailable => plotting is disabled.
To enable plotting functions, run "pip install matplotlib"._

The installation instructions are updated to describe the optional dependencies.

### DIFF
--- a/docs/m/index.rst
+++ b/docs/m/index.rst
@@ -33,3 +33,4 @@ higher-level functions to provide physics results.
    releases/r2.3
    releases/r2.4
    releases/r2.5
+   releases/r2.6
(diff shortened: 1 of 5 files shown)


oscarxblanco (approved):
Tested without errors in a python virtual environment and in jupyterlab.

lfarv:
@oscarxblanco : I had to merge your recent additions. Can you approve again?

oscarxblanco (commented):
Checked again in a python virtual environment. All ok.

oscarxblanco (approved):

PR #813, opened 2024-08-13
fix plot_tune_diagram from __future__ import annotations

oscarxblanco:
Dear @swhite2401 ,
this should solve the issue in python3.8 when using the function plot_tune_diagram.
Reported here: 
https://github.com/atcollab/at/pull/809#issuecomment-2285510511

### DIFF
--- a/pyat/at/plot/resonances.py
+++ b/pyat/at/plot/resonances.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import warnings
 from fractions import Fraction
-from typing import Dict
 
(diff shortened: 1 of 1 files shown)


swhite2401:
If you import from `__future__ import annotations` you can used `dict`, no need to used the "old" `Dict` notation. Once we stop supporting python 3.8 we can then simply remove the` __future__` import and we are good

swhite2401 (approved):
All good! @lfarv any idea why the test went through?

Issue #781, opened 2024-06-17
Rotation of dipoles


TeresiaOlsson:
Hi,

We have done measurements of the orbit distortion and dispersion when realigning a dipole magnet and I wanted to compare to what our model in pyAT predicts. I used the ``rotate_elem`` function (including the changes that have been done since the last release), but the results made no sense. Then I discovered this very nice paper explaining that the reason for it might be because applying rotations to a dipole also changes the reference system.

[https://accelconf.web.cern.ch/ipac2017/papers/wepik061.pdf](https://accelconf.web.cern.ch/ipac2017/papers/wepik061.pdf)

Is there already a ``atsettiltdipole`` function in pyAT as in matlab AT or it has to be implemented? If I need to implement it, is there any reason to not add it directly to the ``rotate_elem`` function instead of having a separate function for dipoles?

elafmusa:
Dear Teresia,
If I understood your comment correctly, using at.rotate_elem(lattice[i], tilt=a, pitch=b, yaw=c, relative=relative) for dipoles does not lead to proper physics results. is it correct ?
I have been using this function to insert a tilt on dipoles (with pitch and yaw set to 0) together with tilt errors on other magnets, and i have noticed that there is almost no any impact of the dipole tilt on orbit distortion.

Best Regards,
Elaf

TeresiaOlsson:
Hi Elaf,

Yes, that is what I think. Like you, I also got a tiny impact in the simulation but in our measurement where we went into the tunnel and physically realigned the magnet we could see a bigger impact. We did a quick calculation of how we expected the magnetic fields to change due to the realignment and that seemed to agree well with what we measured.

If I understood the paper from @simoneliuzzo correctly, the reason is because the dipoles also are setting the reference orbit. So when we change the rotation using the method in `rotate_elem` we are also changing the reference. So for the dipoles we can only get correct results if we change `PolynomA` and `PolynomB`. And that functionality has so far only been implemented in the matlab version of AT as far as I have understood. But maybe it exists in pySC? Do you know?

[...]

PR #815, opened 2024-08-14
Add feeddown polba

oscarxblanco:
Dear all,

I have created a function in pyat that calculates the feeddown PolynomB and Polynom from a transverse offset on a magnet.
It requires to provide the polynom(s) of the magnet and the transverse offset. For example, a quadrupole with transverse offset of 1 mm in horizontal and 2 mm in vertical:
```python
>>> import numpy, at
>>> at.feeddown_polynomba(polb=numpy.array([0,10]),xoffset=1e-3,yoffset=2e-3)
{'PolynomB': array([-0.01]), 'PolynomA': array([-0.02])}
```

### DIFF
--- a/pyat/at/physics/magnet_tools.py
+++ b/pyat/at/physics/magnet_tools.py
@@ -45,33 +45,32 @@ def feeddown_polynomba(
     # verify polynoms length
     maxorda = len(pola)
     maxordb = len(polb)
-    if maxorda == 0 and maxordb == 0:
-        verboseprint("Both polynoms are zero.")
-        return numpy.array([])
     maxord = max(maxorda, maxordb)
-
-    polbpad = numpy.pad(polb, (0, maxord - maxordb), "constant", constant_values=(0, 0))
-    polapad = numpy.pad(pola, (0, maxord - maxorda), "constant", constant_values=(0, 0))
-
-    verboseprint(f"polb={polb},pola={pola}")
-    verboseprint(f"xoffset={xoffset},yoffset={yoffset}")
     polasum = numpy.zeros(maxord - 1)
     polbsum = numpy.zeros(maxord - 1)
-    for ith in range(2, maxord + 1):
-        polbaux_b, polaaux_b = feeddown_from_nth_order(
-            ith, polbpad[ith - 1], xoffset, yoffset, poltype="B"
-        )
-        polbaux_a, polaaux_a = feeddown_from_nth_order(
-            ith, polapad[ith - 1], xoffset, yoffset, poltype="A"
-        )
-        polbshort = polbaux_b + polbaux_a
-        polashort = polaaux_b + polaaux_a
-        polbsum = polbsum + numpy.pad(
-            polbshort, (0, maxord - ith), "constant", constant_values=(0, 0)
-        )
-        polasum = polasum + numpy.pad(
-            polashort, (0, maxord - ith), "constant", constant_values=(0, 0)
-        )
+    if maxorda == 0 and maxordb == 0:
+        verboseprint("Both polynoms are zero.")
+    else:
+        polbpad = numpy.pad(polb, (0, maxord - maxordb), "constant", constant_values=(0, 0))
+        polapad = numpy.pad(pola, (0, maxord - maxorda), "constant", constant_values=(0, 0))
+
+        verboseprint(f"polb={polb},pola={pola}")
+        verboseprint(f"xoffset={xoffset},yoffset={yoffset}")
+        for ith in range(2, maxord + 1):
+            polbaux_b, polaaux_b = feeddown_from_nth_order(
+                ith, polbpad[ith - 1], xoffset, yoffset, poltype="B"
+            )
+            polbaux_a, polaaux_a = feeddown_from_nth_order(
+                ith, polapad[ith - 1], xoffset, yoffset, poltype="A"
+            )
+            polbshort = polbaux_b + polbaux_a
+            polashort = polaaux_b + polaaux_a
+            polbsum = polbsum + numpy.pad(
+                polbshort, (0, maxord - ith), "constant", constant_values=(0, 0)
+            )
+            polasum = polasum + numpy.pad(
+                polashort, (0, maxord - ith), "constant", constant_values=(0, 0)
+            )
     poldict = {"PolynomB": polbsum, "PolynomA": polasum}
     verboseprint(f"poldict={poldict}")
     return poldict
--- a/pyat/at/physics/magnet_tools.py
+++ b/pyat/at/physics/magnet_tools.py
@@ -10,8 +10,34 @@ from scipy.special import comb
 __all__ = [
     "feeddown_polynomba",
     "feeddown_from_nth_order",
+    "feeddown_pol_from_element",
 ]
 
+
+def feeddown_pol_from_element(ele):
+    """
+    Return the feed down polynoms due to a transverse offset on an element.
+
+    Parameters:
+        element: a ring element.
+        polb: PolynomB.
+        pola: PolynomA.
+        xoffset: Default zero. Horizontal offset in meters.
+
+    Returns:
+        Dictionary with PolynomB and PolynomA feeddown components.
+
+    Note: Thin lengs approxiamtion, T2=-T1.
+          Bending angles are ignored.
+    """
+    # check verbose flags only once
+    verboseprint = print if verbose else lambda *a, **k: None
+    # handle cases where polynom a and b are not defined
+    # get polynoms
+    # check if element has T1 and T2. Use one.
+    # Return the polynoms
+
+    return 0
 
 def feeddown_polynomba(
     polb: Sequence[float],
@@ -36,7 +62,7 @@ def feeddown_polynomba(
     Note:
         If only one polynom is passed, it is assumed to be PolynomB.
     """
-    # check and verbose flags only once
+    # check verbose flags only once
     verboseprint = print if verbose else lambda *a, **k: None
 
     # verify polynoms length
--- a/pyat/at/physics/magnet_tools.py
+++ b/pyat/at/physics/magnet_tools.py
@@ -160,26 +160,27 @@ def feeddown_from_nth_order(
     fakeimag = {0: 1, 1: 0, 2: -1, 3: 0}
     polbaux = numpy.zeros(nthorder - 1)
     polaaux = numpy.zeros(nthorder - 1)
-    for kterm in range(1, nthorder):
-        ichoosek = comb(nthorder - 1, kterm)
-        pascalsn = comb(kterm, numpy.arange(kterm + 1))
-        powk = numpy.arange(kterm + 1)
-        powkflip = numpy.arange(kterm, -1, -1)
-        recoefs = numpy.array([fakeimag[numpy.mod(idx, 4)] for idx in powk])
-        imcoefs = numpy.array([fakeimag[numpy.mod(idx + 3, 4)] for idx in powk])
-        commonfactor = nthpolcomp * ichoosek * (-1) ** kterm
-        repart = (
-            recoefs
-            * commonfactor
-            * (pascalsn * (xoffset**powkflip * yoffset**powk))
-        )
-        impart = (
-            imcoefs
-            * commonfactor
-            * (pascalsn * (xoffset**powkflip * yoffset**powk))
-        )
-        polbaux[nthorder - kterm - 1] = polbaux[nthorder - kterm - 1] + repart.sum()
-        polaaux[nthorder - kterm - 1] = polaaux[nthorder - kterm - 1] + impart.sum()
+    if nthpolcomp != 0:
+        for kterm in range(1, nthorder):
+            ichoosek = comb(nthorder - 1, kterm)
+            pascalsn = comb(kterm, numpy.arange(kterm + 1))
+            powk = numpy.arange(kterm + 1)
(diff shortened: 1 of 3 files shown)


oscarxblanco:
Dear @swhite2401 ,
is this interesting or useful enough to be added ? Who should review it ?

o

swhite2401:
@oscarxblanco yes this can be useful, does this relate to the discussion you had with @TeresiaOlsson and @simoneliuzzo in #781 ?

oscarxblanco:
> @oscarxblanco yes this can be useful, does this relate to the discussion you had with @TeresiaOlsson and @simoneliuzzo in #781 ?

No, these are not rotations, it is the effect of the magnet transverse offsets on the polynoms B and A.
At the beginning I wanted to have a theoretical value of the effect of multipole errors in quads with large offsets (anti-bends), but, it is now extended to any magnet and any transverse offset.

For an ideal quad an offset will feed down a dipole kick.
For magnets with more components (sextupoles, octupoles, etc. and/or multipole errors) the feed down is still straight forward to calculate but takes time to verify it is correct. Such calculation could be provided in AT and was missing in the library.

I add two functions: `feeddown_from_nth_order` and `feeddown_polynomba`.
`feeddown_from_nth_order` is the core. It takes as input the order of the magnet component, the value of the magnet component, a type of magnet 'B or A', and the transverse offsets. It returns the feed down polynoms B and A.
`feeddown_polynomba` takes the complete polynoms B and A and uses iteratively the former function. The total feeddown polynoms are returned in a dictionary to avoid order errors when using the output.

The alternative of tracking through the magnet with T1 and T2 offsets is equivalent. The functions mentioned above provide only a direct view of the feed down effect. They are complementary.

oscarxblanco:
Here is an example with a quad, a sext , and a combination with offsets.
```python
# quad (order 2), Kq=10, hor. offset=1 mm
>>> at.feeddown_from_nth_order(2,10,1e-3,0)
(array([-0.01]), array([0.]))
>>> at.feeddown_polynomba(polb=numpy.array([0,10]),xoffset=1e-3)
{'PolynomB': array([-0.01]), 'PolynomA': array([0.])}

# sext (order 3), Ks=10, hor. offset=1 mm
>>> at.feeddown_from_nth_order(3,10,1e-3,0)
(array([ 1.e-05, -2.e-02]), array([0., 0.]))
>>> at.feeddown_polynomba(polb=numpy.array([0,0,10]),xoffset=1e-3)
{'PolynomB': array([ 1.e-05, -2.e-02]), 'PolynomA': array([0., 0.])}

# a quad and sextupole combination with horizontal offset
>>> at.feeddown_polynomba(polb=numpy.array([0,10,10]),xoffset=1e-3,yoffset=-0e-3)
{'PolynomB': array([-0.00999, -0.02   ]), 'PolynomA': array([0., 0.])}

# a quad and sextupole combination with horizontal and vertical offsets
>>> at.feeddown_polynomba(polb=numpy.array([0,10,10]),xoffset=1e-3,yoffset=-2e-3)
{'PolynomB': array([-0.01003, -0.02   ]), 'PolynomA': array([0.01996, 0.04   ])}

**Comment by swhite2401** [comment]
Ok got it, I can review it but not before next week.

**Comment by oscarxblanco** [comment]
@swhite2401 , ok, thank you.

**Comment by swhite2401** [comment]
@oscarxblanco I had a first look and your function works well.
I will add a review in the code.

I have a general question on organization, does this really belong to `non-linear.py`? If you look at other functions in the module they compute ring properties while this one is a more general analytical derivation.

Maybe we could think of starting a new module for such utility functions? We have used utils.py in other subpackages but I don't really like to put this everywhere, any idea?

Then, do you think it would be useful to allow providing a lattice element? The feed-down would then be calculated for the pol* and T1/2 attributes?

File: pyat/at/physics/nonlinear.py [hunk 1]
```diff
@@ -37,30 +53,46 @@ def tunes_vs_amp(ring: Lattice, amp: Optional[Sequence[float]] = None,
...
     def _gen_part(ring, amp, dim, orbit, ld, nturns):
-        invx = numpy.array([[1/numpy.sqrt(ld['beta'][0]), 0],
-                            [ld['alpha'][0]/numpy.sqrt(ld['beta'][0]),
-                            numpy.sqrt(ld['beta'][0])]])
-
-        invy = numpy.array([[1/numpy.sqrt(ld['beta'][1]), 0],
-                            [ld['alpha'][1]/numpy.sqrt(ld['beta'][1]),
-                            numpy.sqrt(ld['beta'][1])]])
-        part = numpy.array([orbit, ] * len(amp)).T + 1.0e-6
+        invx = numpy.array(
```
swhite2401, line 56: I would turn of black for formulas, this makes them very difficult to read

pyat/at/physics/nonlinear.py
```diff
@@ -193,12 +235,165 @@ def gen_detuning_elem(ring: Lattice, orbit: Optional[Orbit] = None) -> Element:
...
+    xoffset: float = 0,
+    yoffset: float = 0,
+    verbose: bool = False,
+    debug: bool = False,
+) -> dict[str, numpy.ndarray]:
+    """
+    Return the feeddown due to a transverse offset.
+
+    Parameters:
+        polb: numpy array. PolynomB.
```
swhite2401, line 271: Specifying the type here may not be needed

pyat/at/physics/nonlinear.py
```diff
@@ -193,12 +235,165 @@ def gen_detuning_elem(ring: Lattice, orbit: Optional[Orbit] = None) -> Element:
...
+    maxorda = len(pola)
+    maxordb = len(polb)
+    if maxorda == 0 and maxordb == 0:
+        raise ValueError("At least one polynom is needed")
+    maxord = max(maxorda, maxordb)
+    debugprint(f"maxord={maxord},maxorda={maxorda},maxordb={maxordb}")
+
+    debugprint("Padding polynoms")
+    polbpad = numpy.pad(polb, (0, maxord - maxordb), "constant", constant_values=(0, 0))
+    polapad = numpy.pad(pola, (0, maxord - maxorda), "constant", constant_values=(0, 0))
```
swhite2401, line 301: It seems that even if the input polynom is empty you are padding it with zeros and going through the whole calculation.
Maybe this could be avoided?

In this case you could even avoid to raise exception: an empty input polynom returns an empty output polynom

Up to you...

pyat/at/physics/nonlinear.py
```diff
@@ -193,12 +235,165 @@ def gen_detuning_elem(ring: Lattice, orbit: Optional[Orbit] = None) -> Element:
...
+    # check debug and verbose flags only once
+    debugprint = print if debug else lambda *a, **k: None
+    verboseprint = print if verbose else lambda *a, **k: None
+
+    # verify polynoms length
+    maxorda = len(pola)
+    maxordb = len(polb)
+    if maxorda == 0 and maxordb == 0:
+        raise ValueError("At least one polynom is needed")
+    maxord = max(maxorda, maxordb)
```
swhite2401, line 296: Is this really useful? Why not keep the A/B calculation independent and allow for polynoms with different length
This way you can skip the whole zero padding part and simplify your code

pyat/at/physics/nonlinear.py
```diff
@@ -193,12 +235,165 @@ def gen_detuning_elem(ring: Lattice, orbit: Optional[Orbit] = None) -> Element:
...
+    return poldict
+
+
+def feeddown_from_nth_order(
+    nthorder: int,
+    nthpolcomp: float,
+    xoffset: float,
+    yoffset: float,
+    poltype: str = "B",
+    debug: bool = False,
```
swhite2401, line 336: Same comment here

pyat/at/physics/nonlinear.py
```diff
@@ -193,12 +235,165 @@ def gen_detuning_elem(ring: Lattice, orbit: Optional[Orbit] = None) -> Element:
...
     return nonlin_elem
+
+
+def feeddown_polynomba(
+    polb: numpy.ndarray or tuple = (),
+    pola: numpy.ndarray or tuple = (),
+    xoffset: float = 0,
+    yoffset: float = 0,
+    verbose: bool = False,
+    debug: bool = False,
```
swhite2401, line 265: Do you want to keep this debug option in the distribution?

pyat/at/physics/nonlinear.py
```diff
@@ -193,12 +235,165 @@ def gen_detuning_elem(ring: Lattice, orbit: Optional[Orbit] = None) -> Element:
...
+        A2=r1[0][1],
+        A3=r1[1][1],
+        T1=-orbit,
+        T2=orbit,
+    )
     return nonlin_elem
+
+
+def feeddown_polynomba(
+    polb: numpy.ndarray or tuple = (),
```
swhite2401, line 260: Don;t we want to allow a list as well? Maybe the type should be Sequence[float]

oscarxblanco:
> @oscarxblanco I had a first look and your function works well. I will add a review in the code.
> 
> I have a general question on organization, does this really belong to `non-linear.py`? If you look at other functions in the module they compute ring properties while this one is a more general analytical derivation.
> 
> Maybe we could think of starting a new module for such utility functions? We have used utils.py in other subpackages but I don't really like to put this everywhere, any idea?

I could create a file called `miscellaneous.py` with this function inside the physics folder. Is that better ?
Otherwise, I have created this function to calculate the effect of multipole errors on magnets, and therefore, I could put it on a new file called `magnet_tools.py` for example.

> 
> Then, do you think it would be useful to allow providing a lattice element?

It could be OK to have an element as input.
> The feed-down would then be calculated for the pol* and T1/2 attributes?

Sounds OK.

pyat/at/physics/nonlinear.py
oscarxblanco: non linear is restored from master w/o black.

oscarxblanco: the new functions are moved to `magnet_tools.py`

pyat/at/physics/nonlinear.py
oscarxblanco: Removed debug option

pyat/at/physics/nonlinear.py
oscarxblanco: debug option removed

pyat/at/physics/nonlinear.py
oscarxblanco: Changed to Sequence[float]

pyat/at/physics/nonlinear.py
oscarxblanco: type removed

pyat/at/physics/nonlinear.py
oscarxblanco: Code reordered to avoid padding zero-length polynoms

pyat/at/physics/nonlinear.py
oscarxblanco: Sorry, I don't understand. The input polynom a and b already could be of different lengths.
On the other side, I think the output polynom a and b must be of the same length to avoid problems with the integrators. Therefore, I am returning something that could be assigned as Polynoms for an element. If I left the output polynoms with different length, it will be cumbersome to use them later.

pyat/at/physics/magnet_tools.py
```diff
@@ -0,0 +1,162 @@
...
+from scipy.special import comb
+
+__all__ = [
+    "feeddown_polynomba",
+    "feeddown_from_nth_order",
+    "feeddown_pol_from_element",
+]
+
+
+def feeddown_pol_from_element(ele,verbose: bool=False):
```
oscarxblanco, line 17: Dear @swhite2401 ,
I have put the skeleton of a new function that takes an element and returns the polynoms. Is this what you were asking ?
If so, I will implement the steps in the commented lines.

oscarxblanco:
@swhite2401 , I have added the function `feeddown_pol_from_element`. See the example below:
```python
qf = at.Quadrupole('QF',0.1,-10)
at.shift_elem(qf,deltax=1e-3,deltaz=2e-3)
at.feeddown_pol_from_element(qf)
```

pyat/at/physics/nonlinear.py
swhite2401: Ok up to you. Just so you know this "equalization" is already done in the element class when it is initialized see `ThinMultipole`

swhite2401:
> @swhite2401 , I have added the function `feeddown_pol_from_element`. See the example below:
> 
> ```python
> qf = at.Quadrupole('QF',0.1,-10)
> at.shift_elem(qf,deltax=1e-3,deltaz=2e-3)
> at.feeddown_pol_from_element(qf)
> ```

Looks good!

pyat/at/physics/nonlinear.py
oscarxblanco: @swhite2401 , thank you, I think I found the lines 
https://github.com/atcollab/at/blob/a409dd675bfa2075304049503c2c99da0726b3fb/pyat/at/lattice/elements.py#L781-L785.
I think I will keep the zero padding, and to avoid doing long calculations over a component that is equal to zero I have included an `if nthpolcomp != 0` in the last commit.

swhite2401:
@oscarxblanco are you done with this one? I have no further comments

oscarxblanco:
@swhite2401 , yes. I have nothing to add now.

swhite2401 (approved):
Ok for me

Issue #268, opened 2021-07-20
RDT derivatives


ZeusMarti:
Hi all,

I think it is useful to implement the RDT first order derivatives respect to quadrupole, sextupole and octupole strengths. The code would be quite similar to computeRDT.m and RDTelegantAT.cpp but storing the element contributions in an array instead of summing them up. My c programing skills are a bit rusty, is anyone willing to share it?

Thanks a lot,

Zeus

PR #821, opened 2024-08-25
Compute RDTs

swhite2401:
A computation of coupling, quadrupole, sextupole and octupole RDTs is proposed in pyat.
This is based on the ELEGANT implementation already present in matlab AT.

This implementation is done in python, for the EBS lattice it is about 4 times slower than the cpp code from matlab in the second order RDTs are calculated. For first order RDTs the computation time is equivalent.

However a parallel calculation is proposed that scales almost linearly (factor 30 speed-up on 64 cores).

While implementing and testing this PR, a bug phase found in the average phase calculation in matlab that is also fixed in this PR.


swhite2401:
Comparison between matlab and python

chromatic
detuning
geometric1
geometric2

swhite2401:
@ZeusMarti in #268 you mentioned the need for RTD derivatives, is this still something you would be interested in? 
It should be possible to implement as part of this PR.

ZeusMarti:
Yes, @swhite2401, thanks for remembering that, I think the derivates are useful when implementing any correction method. Also, if I my memory serves me well , we do not need to add additional calculations, just store the derivates values that are implicitly being used in the RDT calculations.

Regarding the average phase bug, I'll give it a look asap.

swhite2401:
Yes at first sight it look straight forward. Ok then I will work a bit more on this PR and resubmit for approval when this is added

oscarxblanco:
Dear @swhite2401 , is this the expected behaviour ? The function requires a ring with 6d disabled. If I do it in one line I get the following error.
```python
>>> at.get_rdts(ring.disable_6d(),0,at.RDTType.ALL)
  File "/home/oblanco/Documents/public/progs/pyenv/testrdts/lib/python3.11/site-packages/at/physics/rdt.py", line 485, in get_rdts
    nperiods = ring.periodicity
               ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'periodicity'
```

oscarxblanco:
@swhite2401, please ignore the previous question. I didn't know I needed to set `copy=True`.
```python
at.get_rdts(ring.disable_6d(copy=True),0,at.RDTType.ALL)
```
I will continue checking.

swhite2401:
> @swhite2401, please ignore the previous question. I didn't know I needed to set `copy=True`.
> 
> ```python
> at.get_rdts(ring.disable_6d(copy=True),0,at.RDTType.ALL)
> ```
> 
> I will continue checking.

Correct in case you do not set this option to true the ring is modified in place and the function returns None

oscarxblanco:
Dear @swhite2401 , 
I compute the RDTs in pyat and AT matlab for the same lattice, and check the differences in absolute value and angle.
`h01111` agrees well, `h00201` **does not**. ( I used pyat at.__version=0.6.2.dev23+gdcb1f875 )
image

Here is the matlab and python code:
```matlab
clear; close all;
%% matlab
RINGload = load("ring.mat");
RINGtest = atdisable_6d(RINGload.THERING);
%%
refpts = 1:length(RINGtest);
theRDT = computeRDT(RINGtest,refpts);
%%
fieldlist = fieldnames(theRDT(1));
for ii = 1:length(fieldlist)
    RDT.(fieldlist{ii}) = cat(1,theRDT.(fieldlist{ii}));
end
%%
save('RDTsmatlab.mat','RDT');
```
```python
"""This file compares the RDTs from a ring."""

# 2024aug29 oblanco at ALBA CELLS

import at
import numpy
import scipy

print(f'at.__version={at.__version__}')

# get pyat rdts
ring = at.load_mat('ring.mat',keep_all=True)
ring.disable_6d()
refpts = numpy.arange(len(ring))
rdts = at.get_rdts(ring,refpts,at.RDTType.ALL)
thepyrdts = vars(rdts)

# matlab
matdata = scipy.io.loadmat('RDTsmatlab.mat')

# plot
#thepyrdts.pop('refpts')
#keylist = list(thepyrdts.keys())
keylist = ['h00111','h00201']
fig = plt.figure()
ax = fig.add_subplot(211)
for idx in range(len(keylist)):
    pyrdts = abs(numpy.array(thepyrdts[keylist[idx]]))
    matrdts = abs(numpy.squeeze(matdata['RDT'][keylist[idx]][0][0]))
    #ax.plot(pyrdts,label=keylist[idx]+'python')
    invpyrdts = 1.0/pyrdts
    ax.plot((pyrdts-matrdts)*invpyrdts,label=keylist[idx]+' (py-mat)/py')
plt.xlabel('index')
plt.ylabel('abs')
plt.legend()
ax = fig.add_subplot(212)
for idx in range(len(keylist)):
    pyrdts = numpy.angle(numpy.array(thepyrdts[keylist[idx]]))
    matrdts = numpy.angle(numpy.squeeze(matdata['RDT'][keylist[idx]][0][0]))
    invpyrdts = 1.0/pyrdts
    ax.plot(pyrdts-matrdts,label=keylist[idx]+' python-matlab')
plt.xlabel('index')
plt.ylabel('angle [rad]')
plt.legend()
plt.show(block=False)
```

swhite2401:
@oscarxblanco I ran your scripts on my side and everything looks fine.

I did it on the EBS lattice however... but I don't see how this could be the problem.

Are you sure you are using the matlab from this branch? I corrected the average phase claculation there too.

It could be interesting to show matlab and python on top of each other, for a periodic lattice rdts are periodic too, it may point to where the issue is.

swhite2401:
In fact there something strange with the phase... the first few points are wrong on my side...

image

swhite2401:
@oscarxblanco on my side it is all fine now.
I had uncomitted changes + a bug for the refpts located before the first magnet. I think we looking at different versions... sorry about that

image

oscarxblanco:
@swhite2401 , great, now the results in matlab and python agree. I'll continue checking.
image

swhite2401:
By the way, when you will include second order terms I strongly suggest to use `use_mp=True` otherwise the calculation may take quite some time...

oscarxblanco:
@swhite2401 , thank you for the advise on the parallel calculation.

`h22000` returns a phase that oscillates around a multiple of $2\pi$. This is because the imaginary part is tiny in both matlab and python, most likely numerical noise. 
```python
>>> thepyrdts['h22000'][3]
np.complex128(-66002.51469360325+1.4551915228366852e-11j)
```
```matlab
>> RDT.h22000(4)
ans =
  -6.6003e+04 - 9.3621e-11i
```
Could we introduce a rounding to $\epsilon\approx10^{-9}$ ? in both python and matlab ? Or is there any other better strategy to get rid of the phase changes ? 
image

oscarxblanco:
The case I show for `h22000` is a real number with a tiny imaginary value. But, there could be cases with the opposite combination, tiny real with an imaginary value. If you opt for rounding it should be applied to both components of the complex value.

swhite2401:
Ah this is annoying, we could introduce an option to set the precision yes.
Do you think this is something useful?

swhite2401:
did you try numpy.unwrap?

oscarxblanco:
@swhite2401 , `numpy.unwrap` works well on my case. Thank you !
Here is the plot of the RDTs. Looks ok. I continue checking.
image

oscarxblanco:
Dear @swhite2401 , `h00310` is giving me a different result in matlab AT an pyat.
image

swhite2401:
h00310 fixed

oscarxblanco:
@swhite2401 , ok, I confirm it ! Thank you. It looks like this now. I'll continue.
image

oscarxblanco:
I doubt about the sign of `h11001`.
In the SLS note  9/97, https://ados.web.psi.ch/slsnotes/sls0997.pdf , the term is written as:
image
I have made a fodo cell with only quads, no sextupoles, no dispersion. The quad length is the same, the strength is only different in sign. $\beta_x$ is much larger in QF than in QD, so I expect a positive rdt value.
image
I made a numerical approximation which is correct in magnitude, but, differs in sign. The numerical approximation I did is positive and agrees with the equation, and what I would expect.
```python
at/lattice/lattice_object.py:1514: AtWarning: No bending in the cell, set "Periodicity" to 1
  warn(AtWarning('No bending in the cell, set "Periodicity" to 1'))
h11001approx=2.105722090438852
rdtsfododict['h11001'][0]=-2.108058736989468
```
Here is the python code:
```python
"""This file test the equations of the RDTs"""
# 2024aug30 oblanco at ALBA CELLS
import at
import numpy

# from AT primer, edited
dipoleon = 0
QF=at.Quadrupole('QF',0.01,1.2*50)
Dr = at.Drift('Dr', 1.0)
HalfDr = at.Drift('Dr2', 0.5)
QD = at.Quadrupole('QD', 0.01, -1.2*50)
Bend = at.Dipole('Bend', 1, dipoleon*2*numpy.pi/40)

# add sextupole
sxt = at.Sextupole('SXT',0.01,000)

fodo = at.Lattice([HalfDr, Bend, Dr, QF, Dr, Bend,sxt, Dr, QD, HalfDr],
                      name='Simple FODO cell', energy=1E9)
# get rdts from function
refpts = numpy.arange(len(fodo))
rdtsfodo = at.get_rdts(fodo,refpts,at.RDTType.ALL,second_order=True,use_mp=True)
rdtsfododict = vars(rdtsfodo)

# get rdts from my numerical approximation
el0, rd0, elall = at.linopt4(fodo,refpts=at.All)
idxqf = int(fodo.get_refpts('QF')[0])
idxqd = int(fodo.get_refpts('QD')[0])
idxsxt = int(fodo.get_refpts('SXT')[0])
h11001approx = 0.25*( QF.K*QF.Length*elall['beta'][idxqf][0] + QD.K*QD.Length*elall['beta'][idxqd][0] + sxt.H*sxt.Length*elall['dispersion'][idxsxt][0]*elall['beta'][idxsxt][0])
print(f'h11001approx={h11001approx}')
print(f"rdtsfododict['h11001'][0]={rdtsfododict['h11001'][0]}")
```

swhite2401:
It is true that the formula used in ELEGANT has opposite sign as the Bengston paper (see line 200 of RDTelegantAT.cpp, same formula used in python).

It looks like a bug to me, so I would be tempted to correct it everywhere, @carmignani do you have any opinion on this?

carmignani:
The elegant code was taken many years ago from Elegant. I can check if the RDT computation has been changed in the more recent versions.

carmignani:
I checked in the current version of Elegant (2023.4) and the line 4889 of twiss.cc, in the function 
`void computeDrivingTerms(DRIVING_TERMS *d, ELEMENT_LIST *elem, TWISS *twiss0, double *tune, long nPeriods)
`
is `h11001 += (b3L * betax1 * etax1 / 2 - b2L * betax1 / 4) * nPeriods;` 
So it is exactly like the one in RDTelegantAT.cpp.
Maybe it's a bug in elegant. Is h11001 the only one with opposite sign?

oscarxblanco:
Dear @carmignani ,
I have not checked the other terms yet. This is the first one I tested against a fodo lattice (which you could check above) and the first one of the two articles mentioned in the DocString.

In any case, I see that the elegant file `oag/apps/src/elegant/twiss.cc` in lines 4887 to 4897 has the formulas coded below. All chromatic terms seem to have opposite sign to what is written in the article .
```c++
        /* first-order chromatic terms */¶
        /* h11001 and h00111 */¶
        h11001 += (b3L * betax1 * etax1 / 2 - b2L * betax1 / 4) * nPeriods;¶
        h00111 += (b2L * betay1 / 4 - b3L * betay1 * etax1 / 2) * nPeriods;¶
¶
        /* h20001, h00201 */¶
        h20001 += (b3L * betax1 * etax1 / 2 - b2L * betax1 / 4) / 2 * ed[nE].px[2] * PF(2, 0);¶
        h00201 += (b2L * betay1 / 4 - b3L * betay1 * etax1 / 2) / 2 * ed[nE].py[2] * PF(0, 2);¶
        /* h10002 */
        h10002 += (b3L * ed[nE].rbetax * ipow2(etax1) - b2L * ed[nE].rbetax * etax1) / 2 * ed[nE].px[1] * PF(1, 0);¶
```

oscarxblanco:
@carmignani , all geometric terms also seem to have opposite sign wrt to the article.

oscarxblanco:
@swhite2401 ,
I think this should be mentioned in the doc string. The use will know in advance that one period of the machine is needed.
```python
at/physics/rdt.py", line 488, in get_rdts
    assert nperiods == 1, "Second order available only for ring.periodicity=1"
           ^^^^^^^^^^^^^
AssertionError: Second order available only for ring.periodicity=1

```

oscarxblanco:
To make it more explicit, one could rewrite Eq. (10) in Bengtsson for a sextupole
$B_y + iB_x = -\frac{P_0}{q}(b_3 + ia_3)(x+iy)^2$.
Assuming a normal sextupole ($a_3=0$), charge $q$, and $B\rho=|P_o/q|$, one could write
$B_y + iB_x = sgn(q) B\rho (x+iy)^2$, with $sgn$ the sign function.
Therefore,
$B_y=sgn(q)B\rho*b_3(x-y)^2$ and 
$B_x=sgn(q)B\rho b_3(2xy)$
In at, the third component in the polynom B is
$PolB3=\frac{1}{B\rho}\frac{\partial^2}{\partial x^2}By = 2 sgn(q) b_3$.
Here, apart from the sign of the charge, a factor 2 appears in front of Bengtsson's $b_3$.

pyat/at/physics/linear.py
```diff
@@ -8,7 +8,6 @@
 from math import sqrt, pi, sin, cos, atan2
 
 import numpy
```
lfarv, line 10: While you are there, it would be better to use the [recommended way](https://numpy.org/doc/2.1/user/absolute_beginners.html#how-to-import-numpy) of importing numpy and use `np`.

Note: there was no problem with the 2 import lines. The import is done once, you just get 2 different names to refer to the same thing: quick and dirty trick to minimise the modifications.

oscarxblanco:
@swhite2401, I retract about the factor two I thought was missing, I'm sorry for the misleading review.
But, in the way I might have the explanation on the minus sign between the the equations and the code.
...
**Eq. (10) in Bengtsson**
$B_y + iB_x = -\frac{P_0}{q}\sum_{n=1}^\infty(b_n + ia_n)(x+iy)^{n-1}$.
Assuming a normal sextupole ($a_3=0$), charge $q$, and $B_\rho=|P_o/q|$, one could write
$B_y + iB_x = -sgn(q) B_\rho \sum_{n=1}^\infty(b_n + ia_n)(x+iy)^{n-1}$, with $sgn$ the sign function.

**In AT**
$B_y + iB_x = B_\rho \sum_{n=1}^\infty(b_n + ia_n)(x+iy)^{n-1}$, which differs from the paper in a minus sign and the sign of the charge. I suppose the sign of the charge is not relevant for this calculation in AT. 

To conclude, there is a minus between the polynom terms used by Bengtsson and the polynom in the code.

oscarxblanco (changes requested):

carmignani:
> As a side note, I wonder how the same equation works in Elegant and AT if the field representation is different. Do they scale the field ?

In Elegant they divide by 2 the K2 of the sextuples. For example, in line 4814, 4815 and 4816:
`    case T_CSBEND:`
`      b2L = ((CSBEND *)eptr1->p_elem)->k1 * ((CSBEND *)eptr1->p_elem)->length;`
`      b3L = ((CSBEND *)eptr1->p_elem)->k2 * ((CSBEND *)eptr1->p_elem)->length / 2;`

oscarxblanco:
oh ok @carmignani , thank you  ! That explains why it works. I'll continue reviewing.

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+
+    def pf(i, j):
+        return periodicfactor[4 + i][4 + j]
+
+    if nperiods != 1:
+        for i in range(9):
+            for j in range(9):
+                a1 = numpy.pi * 2 * (tune[0] * (i - 4) + tune[1] * (j - 4))
+                a2 = a1 / nperiods
+                periodicfactor[i][j] = (numpy.exp(1j * a1) - 1.0) / (
```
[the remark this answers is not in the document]
oscarxblanco: Also, this expression is invalid for $x=1$. You get a `numpy.nan`  inside the `periodicfactor` array. See here below
```python
array([[ 0.11995293,  0.71065224,  1.14536311,  0.86937818,  0.02918188,
        -0.51789102,  0.09712676,  1.92673149,  4.00214385],
       [ 0.8729706 ,  1.15079497,  0.67337023, -0.18950072, -0.4889994 ,
         0.49218752,  2.5152446 ,  4.42114406,  4.96390501],
       [ 1.09715717,  0.44600387, -0.36645857, -0.36237996,  0.96769714,
         3.095958  ,  4.73701569,  4.8011494 ,  3.24475565],
       [ 0.20445964, -0.48205199, -0.13425928,  1.50518507,  3.64058563,
         4.93338244,  4.51554829,  2.67085531,  0.61087161],
       [-0.51931852,  0.19301704,  2.08165682,  4.12193798,         **nan**,
         4.12193798,  2.08165682,  0.19301704, -0.51931852],
       [ 0.61087161,  2.67085531,  4.51554829,  4.93338244,  3.64058563,
         1.50518507, -0.13425928, -0.48205199,  0.20445964],
       [ 3.24475565,  4.8011494 ,  4.73701569,  3.095958  ,  0.96769714,
        -0.36237996, -0.36645857,  0.44600387,  1.09715717],
       [ 4.96390501,  4.42114406,  2.5152446 ,  0.49218752, -0.4889994 ,
        -0.18950072,  0.67337023,  1.15079497,  0.8729706 ],
       [ 4.00214385,  1.92673149,  0.09712676, -0.51789102,  0.02918188,
         0.86937818,  1.14536311,  0.71065224,  0.11995293]])
```

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+    """
+    Original implementation from ELEGANT
+    Based on J.Bengtsson, SLS Note 9 / 97, March 7, 1997, with corrections per W.Guo (NSLS)
+    Revised to follow C.X.Wang AOP - TN - 2009 - 020 for second - order terms
+    """
+    periodicfactor = numpy.ones((9, 9))
+    rdts = {}
+
+    def pf(i, j):
+        return periodicfactor[4 + i][4 + j]
```
oscarxblanco, line 107: Consider to remove the `4` offset. This is unnecessary in numpy arrays because you can use negative indexes.
This offset is probably inherited from C, right ?

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+
+    if nperiods != 1:
+        for i in range(9):
+            for j in range(9):
+                a1 = numpy.pi * 2 * (tune[0] * (i - 4) + tune[1] * (j - 4))
```
oscarxblanco, line 112: Consider to remove the `4` offset. This is unnecessary in numpy arrays because you can use negative indexes.
This offset is probably inherited from C, right ?

oscarxblanco (changes requested):

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+    GEOMETRIC2 = 5
+    TUNESHIFT = 6
+
+
+@dataclass
+class _RDT:
+    # Location
+    refpts: Sequence[int] | int = None
+    # Normal quadrupoles rdts
+    h20000: Sequence[complex] | complex = None
```
oscarxblanco, line 32: h20000 is missing in the matlab implementation. Why is it left out ?

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+    h00200: Sequence[complex] | complex = None
```
oscarxblanco, line 33: h00200 is missing in the matlab implementation. Why is it left out ?

oscarxblanco (changes requested):

pyat/at/physics/rdt.py
carmignani: They are missing in the elegant implementation, that's why they are missing in the Matlab implementation!

oscarxblanco:
Dear @swhite2401 , 
I didn't find h10010 nor h10100 on the articles cited in the DocString. I might be wrong, but, do they only exist in code ?

Also, there are 13 non-zero components from C.X. Wang that are not implemented. Should they be added later ? or are not useful ? I list them here below :
1. h11200
2. h00220
3. h21001
4. h30001
5. h10021
6. h10111
7. h10201
8. h11002
9. h20002
10. h00112
11. h00202
12. h10003
13. h00004

pyat/at/physics/rdt.py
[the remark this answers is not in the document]
oscarxblanco: `periodicfactor` is defined by the machine tune and does not depend on the driving term.
 It is now calculated again per driving term.
Consider to move it up in the function call to avoid repeating the same complex calculation at the expense of an extra variable of size 9x9 numpy.complex128 memory cost.

oscarxblanco (changes requested):

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+        =================   ======
+        **refts**           location of the rdt
+
+        **h20000**          Normal quadrupole RDTS
+        **h00200**
+
+        **h10010**          Coupling RDTs
+        **h10100**
+
+        **h11001**          Chromatic RDTs
```
oscarxblanco, line 456: Consider :
```python
'''
        RDT              Order         Type             Magnetic Components used:
...
        **h11001**       3rd        Chromatic           Quadrupoles and Sextupoles
'''
```

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+        rdts: rdt data (complex) at refpts
+
+        **rdts** is a dataclass with fields:
+        =================   ======
+        **refts**           location of the rdt
+
+        **h20000**          Normal quadrupole RDTS
+        **h00200**
+
+        **h10010**          Coupling RDTs
```
oscarxblanco, line 453: Consider
```python
'''
        RDT              Order         Type             Magnetic Components used:
...
        **h10010**       2th        Coupling           Skew quadrupoles
'''
```

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+        pool_size: Number of processes used for parallelization
+
+    Returns:
+        rdts: rdt data (complex) at refpts
+
+        **rdts** is a dataclass with fields:
+        =================   ======
+        **refts**           location of the rdt
+
+        **h20000**          Normal quadrupole RDTS
```
oscarxblanco, line 450: Consider:
```python
'''
        RDT              Order         Type             Magnetic Components used:
...
        **h20000**       2nd       Normal           Quadrupoles
'''
```

oscarxblanco (changes requested):

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+
+class RDTType(Enum):
+    """Enum class for RDT type"""
+
+    ALL = 0
+    FOCUSING = 1
+    COUPLING = 2
+    CHROMATIC = 3
+    GEOMETRIC1 = 4
+    GEOMETRIC2 = 5
```
oscarxblanco, line 23: GEOMETRIC1 and GEOMETRIC2 include different orders. But it is not described.
Please add the details on what is calculated on each.

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+
+    if (RDTType.TUNESHIFT in rdttype) or (RDTType.ALL in rdttype):
+        b4lm = b4l[mask_b4l]
+        betaxm = betax[mask_b4l]
+        betaym = betay[mask_b4l]
+        rdts["dnux_dJx"] = sum(3 * b4lm * betaxm * betaxm / (8 * numpy.pi) * nperiods)
+        rdts["dnux_dJy"] = sum(-3 * b4lm * betaxm * betaym / (4 * numpy.pi) * nperiods)
+        rdts["dnuy_dJy"] = sum(3 * b4lm * betaym * betaym / (8 * numpy.pi) * nperiods)
+
+    if (RDTType.GEOMETRIC2 in rdttype) or (RDTType.ALL in rdttype):
```
oscarxblanco, line 195: The geometric rdts of octupoles agree with my theoretical calculation.
There is yet no reference on where are they taken from.
Also, there is not explanation on the difference between GEOMETRIC1 and GEOMETRIC2 RDTTYPEs. Improve the DocString.

I attach the script I wrote to calculate the geometric RDTs of octupoles in Maxima.
If anybody would like also to check it, you could run it as 
```bash
maxima -q -b fourth_order.mc
```
[fourth_order.mc.txt](https://github.com/user-attachments/files/16910908/fourth_order.mc.txt)
Maxima https://maxima.sourceforge.io/docs/manual/maxima.html

oscarxblanco (changes requested):

pyat/at/physics/linear.py
swhite2401: @lfarv rather that a recommendation this is a convention, there is absolutely no difference in effect.
Since the beginning of pyAT we used `import numpy` and not `np` as convention.

I have personally no preference but the convention has to be the same for the whole code, if you now want to switch to `import numpy as np` this has to be done for the whole code in a separate PR.

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+    mask_b4l = numpy.absolute(b4l) > 1.0e-6
+
+    if (RDTType.FOCUSING in rdttype) or (RDTType.ALL in rdttype):
+        b2lm = b2l[mask_b2l]
+        betaxm = betax[mask_b2l]
+        betaym = betay[mask_b2l]
+        pxm = px[mask_b2l]
+        pym = py[mask_b2l]
+        rdts["h20000"] = sum((b2lm / 8) * betaxm * pxm * pxm * pf(2, 0))
+        rdts["h00200"] = sum((b2lm / 8) * betaym * pym * pym * pf(0, 2))
```
oscarxblanco, line 136: I have derived `h00200` and it should have a negative sign in AT.
Here below the explanation.
I believe Eq(94) in Bengtsson has a typo, there should be a minus in front of $b_2$ to be consistent with Eq(10).
This typo has no effect in third order RDTs, therefore, no other changes are needed.
I show part of the derivation in case you would like to check it.
$V_i = \frac{-b_2}{2(1+\delta)}(x^2-y^2)=\frac{-b_2}{2}(1-\delta)(x^2-y^2)$
Ignoring the terms in $\delta$
$V_i=\frac{-b_2}{2}x^2+\frac{b_2}{2}y^2$
Ignoring the terms in $x$, and using Eq(72)
$\hat V_i = R_{o\rightarrow i}A_iV_i = \frac{b_2}{2}\beta_yR_{o \rightarrow i} (\frac{h_y^+ + h_y^-}{2})^2$,
 which is the equivalent to Eqs(89,90,91) for the $y$ plane.
We expand  the polynom on the resonant base and write the term with only $h_y^+$
$\hat V_i=\frac{b_2}{2}\beta_y \{{\frac{1}{4}(h_y^+ + h_y^- )^2}} = \frac{b_2}{2}\beta_y \frac{1}{4} ((h_y^+)^2 e^{i2\mu_y} + \cdots )$
Integrating as in Eq(63)
$h_{00200}=\frac{1}{8} \beta_y b_2 L  e^{i2\mu_y}$
Changing the sign to match the AT magnetic field definition(Bengtsson has minus in Eq(10)).
$h_{00200,AT}=-\frac{1}{8} \beta_y b_2L  e^{i2\mu_y}$

oscarxblanco (changes requested):

pyat/at/physics/linear.py
lfarv: At the moment, 39 AT modules use` import numpy`, and 14 modules use `import numpy as np` (mostly recent ones). So the situation is mixed. My preference for `np` comes from the fact that all the usage examples (in numpy documentation and in other sources) use `np`. This is not critical, so my suggestion is:
- use `np` for all new developments,
- preferably switch to `np` when significantly modifying an existing module.

This is similar to what we are doing with "black" formatting.

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+        return periodicfactor[4 + i][4 + j]
+
+    if nperiods != 1:
+        for i in range(9):
+            for j in range(9):
+                a1 = numpy.pi * 2 * (tune[0] * (i - 4) + tune[1] * (j - 4))
+                a2 = a1 / nperiods
+                periodicfactor[i][j] = (numpy.exp(1j * a1) - 1.0) / (
+                    numpy.exp(1j * a2) - 1.0
+                )
```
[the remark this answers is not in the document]
swhite2401: I initialized `periodicfactor` with `dttype=complex` to fix this

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,545 @@
...
+            _get_polynom(e, "PolynomA", 1) * e.Length,
+            _get_polynom(e, "PolynomB", 1) * e.Length,
+            _get_polynom(e, "PolynomB", 2) * e.Length,
+            _get_polynom(e, "PolynomB", 3) * e.Length,
+        ]
+        for e in ring[idx_mag]
+    ]
+
+    mu = lo[-1].mu
+    tune = mu / 2.0 / numpy.pi
```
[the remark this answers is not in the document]
swhite2401: Corrected

pyat/at/physics/rdt.py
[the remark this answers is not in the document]
swhite2401: Done, `x=1` correspond to a `tune=0.0` you will get an unstable ring error before entering this function

pyat/at/physics/rdt.py
swhite2401: Same holds for h20000 in this case right?

swhite2401: In fact no... this was a serious bug... thanks!

swhite2401:
@oscarxblanco I think I am through with all your requested changes. 
One important points comes to mind in the refs from Franchi et al. a generic formula is provided for on momnetum RDTs, we may consider to use this instead of the Bengston formulas for these terms...
This will however take quite some time to implement and test...

Other option is to merge as is an keep it as a future possible improvement.

What do you think?

oscarxblanco:
Dear @swhite2401 , 
I have tested the first order (wrt sextupoles) RDTS and thus far I see no issues, although I get a small deviation from h10200 that I need to check more in detail.
Periodicity seems now to be taken into account correctly.

I now have a question on the expression used for the octupole detuning. 
I have compared with tracking and it is giving me a different result. I may be wrong in what is actually derived for 'dnux_dJx'. See the difference between `dnu_track`  and `rdtsfododict[dnux_dJx][0]`

Here below is what I obtain
```python
>>> exec(open('testtuneshift.py').read())
at.__version__=0.6.2.dev32+g9a7303c7
Octupole:
       FamName: OCT
        Length: 0.01
      MaxOrder: 3
   NumIntSteps: 10
      PolynomA: [0. 0. 0. 0.]
      PolynomB: [   0.    0.    0. 2000.]
/home/oblanco/Documents/public/progs/pyenv/testrdts/lib/python3.11/site-packages/at/lattice/lattice_object.py:1514: AtWarning: No bending in the cell, set "Periodicity" to 1
  warn(AtWarning('No bending in the cell, set "Periodicity" to 1'))
dnu_track= 
2462 J + 0.2974
rdtsfododict[dnux_dJx][0]=257.4617381695154
```


And here is the code I used for the comparison
```python
"""This file test the equations of the RDTs, in particular the octupole tuneshift"""

# 2024sep15 oblanco at ALBA CELLS

import at
import numpy

def tuneshiftwithamplitude(ring,scanaxis,amin=-0.001,amax=0.001,asteps=200,nturns=1024):
    iarray = numpy.linspace(amin, amax, asteps)
    leniarray = len(iarray)
    orbit, _ = at.find_orbit(ring)
    # add offset
    yoffset=0
    add_offset6D = numpy.zeros(6)
    add_offset6D[0] = 0e-3
    z0 = numpy.zeros((leniarray, 6))  # transposed, and C-aligned
    z0 = z0 + add_offset6D + orbit
    # check plane
    if scanaxis=='x':
        enableaxis=0
    if scanaxis=='y':
        enableaxis=2
    z0[:, enableaxis] = z0[:, enableaxis] + iarray + 1e-9
    z0[:, 2] = z0[:, 2] + 1e-9 + yoffset
    zOUT, tpar, tdl = ring.track(z0.T, nturns=nturns,use_mp=True,losses=True)
    tunesx = at.get_tunes_harmonic(zOUT[0,:,:,:],use_mp=True)
    tunesy = at.get_tunes_harmonic(zOUT[2,:,:,:],use_mp=True)
    return iarray,tunesx,tunesy


print(f'at.__version__={at.__version__}')
# from AT primer, edited
dipole_on = 0.00
QF=at.Quadrupole('QF',0.01,1.2*50)
Dr = at.Drift('Dr', 1.0)
HalfDr = at.Drift('Dr2', 0.2)
QD = at.Quadrupole('QD', 0.01, -1.25*50)
Bend = at.Dipole('Bend', 1, dipole_on*2*numpy.pi/40)

octu = at.Octupole('OCT',0.01,(0,0,0,0),(0,0,0,2000))
print(octu)
fodo = at.Lattice([HalfDr, Bend, Dr, QF,octu, Dr, Bend, Dr, QD, HalfDr],
                      name='Simple FODO cell', energy=1E9)

idxoct = int(fodo.get_refpts('OCT')[0])
fodo = fodo.rotate(idxoct)

refpts = numpy.arange(len(fodo))
rdtsfodo = at.get_rdts(fodo,refpts,at.RDTType.ALL,second_order=False,use_mp=True)
rdtsfododict = vars(rdtsfodo)


el0, rd0, elall = at.linopt4(fodo,refpts=at.All)
idxqf = int(fodo.get_refpts('QF')[0])
idxqd = int(fodo.get_refpts('QD')[0])
idxoctu = int(fodo.get_refpts('OCT')[0])


wholetune = elall['mu'][-1]/(2*numpy.pi)*fodo.periodicity
nperiods = fodo.periodicity


xmax =1e-3
xarray, nuxarray, nuyarray = tuneshiftwithamplitude(fodo,'x',amin=0,amax=xmax)
jarray = xarray**2 / (2*elall['beta'][idxoctu][0])
polynufromfit = numpy.polyfit(jarray, nuxarray, 1)
#plt.plot(xarray,nuxarray)
#plt.show(block=False)
#dnufromtrack = numpy.diff(nuxarray)
#polydnudJxfromfit = numpy.polyfit(xarray,dnufromtrack,1)
#plt.plot(dnufromtrack)
#plt.show(block=False)
p = numpy.poly1d(polynufromfit,variable='J')
#dpol = numpy.polyder(p)
#pol.variable='J'
print(f'dnu_track={p}')

# Equation (2.245) from S.Y.Lee. Accelerator Physics. Fourth Edition.
# Section VII.3 Nonlinear detuning from sextupoles and octupoles
# B3 = 3*PolynomB[3], i.e. the octupole force scale to AT
# the minus sign in front of B3 I still cannot explain.
myrsdict = {}
mykey = 'dnux_dJx'
myrsdict[mykey] = nperiods*3*octu.PolynomB[3]*octu.Length* (elall['beta'][idxoctu][0]**2)*1./(8*numpy.pi)

#print(f'{mykey}approx={myrsdict[mykey]}')
print(f"rdtsfododict[{mykey}][0]={rdtsfododict[mykey][0]}")
#print(f'diff:{rdtsfododict[mykey][0] - myrsdict[mykey]}')
##print()

```

swhite2401:
@oscarxblanco I think the problem is with the definition of the coefficients

AT has already a detuning coefficient calculation implemented that you can call using:
`at.detuning(fodo)`

Please have a look at `physics.nonlinear.detuning` for the source

The output of that function is consistent with the formulas implemented in rdt.py (I have checked only the octupole term)

Concerning the sign, I am not sure where the problem is, but we get the same result from tracking and analytical calculation which is good enough for me.

oscarxblanco:
@swhite2401, thank you. 
You are right, `detuning` and `dnux_dJx` match for a fodo with periodicity 1, and one octupole.
However, they disagree for periodicity larger than one. I would guess that `detuning` is wrong, correct ?

P.S:
The help message in `detuning` is misleading. This is deriving $d\nu/dJ$, not $d\nu/dx$. I will open a separate issue.
```python
>>>help(at.detuning)
detuning(ring: at.lattice.lattice_object.Lattice, xm: Optional[float] = 3e-05, ym: Optional[float] = 3e-05, npoints:
...
        q1 (ndarray): Amplitude detuning coefficients with shape (2, 2)
                      ``[[dQx/dx, dQy/dx], [dQx/dy, dQy/dy]]``
```

oscarxblanco:
> @oscarxblanco I think I am through with all your requested changes. One important points comes to mind in the refs from Franchi et al. a generic formula is provided for on momnetum RDTs, we may consider to use this instead of the Bengston formulas for these terms... This will however take quite some time to implement and test...
> 
> Other option is to merge as is an keep it as a future possible improvement.
> 
> What do you think?

Comming back to your question.
I have heard about Franchi's article (https://journals.aps.org/prab/abstract/10.1103/PhysRevSTAB.17.074001) and it seems possible to use. I noticed that the resonant base is different, (above Eq.(A1)), so I expect that at least some of  the terms need to be changed by their complex conjugate. I would need to check them in more detail to know if there is something else that needs to be adapted wrt the notation already in use.

Apart from that,
S. C. Leemann, and A. Streun. 'Perspectives for future light source lattices incorporating yet uncommon magnets'. PRAB  14, 030701 (2011), show in Eq(C12-C16) the terms for octupoles and Eq(D5-D7) some terms for decapoles. The resonant base is different to Bengtsson.

1 Bingfeng Wei , Zhenghe Bai, et. al. 'Minimizing the fluctuation of resonance driving terms
in dynamic aperture optimization' PRAB  26, 084001 (2023), show the sextupole cross terms including the lattice periodicity factor. See Eq. (A4). The notation needs to be checked to be compatible with Bengtsson.

With respect to Wan's article I managed to derive h22000, h31000 and h40000 which differ by a factor $1/2J$ wrt the equation. I think this is not an issue, it might be absorbed somewhere else as is the case we discussed for 'dnux_dJx'.
All other crossed terms I haven't yet verified.

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,575 @@
...
+        rdts["h20000"] = sum((b2lm / 8) * betaxm * pxm * pxm * pf(2, 0))
+        rdts["h00200"] = -sum((b2lm / 8) * betaym * pym * pym * pf(0, 2))
+
+    if (RDTType.COUPLING in rdttype) or (RDTType.ALL in rdttype):
+        a2lm = a2l[mask_a2l]
+        rbetaxm = rbetax[mask_a2l]
+        rbetaym = rbetay[mask_a2l]
+        pxm = px[mask_a2l]
+        pym = py[mask_a2l]
+        rdts["h10010"] = sum((a2lm / 4) * rbetaxm * rbetaym * pxm / pxm * pf(1, -1))
```
oscarxblanco, line 159: This expression should be 
 ```python
 rdts["h10010"] = sum((a2lm / 4) * rbetaxm * rbetaym * pxm / pym * pf(1, -1))
```
The multiplication and division by `pxm` is a bug.

**Review by oscarxblanco — CHANGES_REQUESTED** [review]

File: pyat/at/physics/rdt.py [hunk 16]
```diff
@@ -0,0 +1,575 @@
...
+        rdts["h10100"] = sum((a2lm / 4) * rbetaxm * rbetaym * pxm / pym * pf(1, 1))
```
oscarxblanco, line 160: This expression should be
```python
 rdts["h10100"] = sum((a2lm / 4) * rbetaxm * rbetaym * pxm * pym * pf(1, 1))
```
The division by `pym` is a bug.

oscarxblanco (changes requested):

pyat/at/physics/rdt.py
swhite2401: ok

pyat/at/physics/rdt.py
swhite2401: ok

swhite2401:
> Comming back to your question.
> I have heard about Franchi's article (https://journals.aps.org/prab/abstract/10.1103/PhysRevSTAB.17.074001) and it seems possible to use. I noticed that the resonant base is different, (above Eq.(A1)), so I expect that at least some of the terms need to be changed by their complex conjugate. I would need to check them in more detail to know if there is something else that needs to be adapted wrt the notation already in use.

@oscarxblanco , @lfarv, @carmignani  is it really necessary that matlab and python have the same convention? As long as the source is properly referenced maybe it is ok to accept difference? We could think of re-implementing the matlab part in case this is a strong requirement...

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,575 @@
...
+        rdts["h20000"] = sum((b2lm / 8) * betaxm * pxm * pxm * pf(2, 0))
+        rdts["h00200"] = -sum((b2lm / 8) * betaym * pym * pym * pf(0, 2))
+
+    if (RDTType.COUPLING in rdttype) or (RDTType.ALL in rdttype):
+        a2lm = a2l[mask_a2l]
+        rbetaxm = rbetax[mask_a2l]
+        rbetaym = rbetay[mask_a2l]
+        pxm = px[mask_a2l]
+        pym = py[mask_a2l]
+        rdts["h10010"] = sum((a2lm / 4) * rbetaxm * rbetaym * pxm / pym * pf(1, -1))
```
oscarxblanco, line 159: There should be a minus sign in front of the sum.
Here is the explanation:
From Bengtsson's Eq(15) $H\propto-A_s$.
In Eq(63) $M_{\xi0\rightarrow1} = \exp(-\int H ds$).
For a thin lens magnet this implies $M_{\xi0\rightarrow1} = exp(LA_s)$, which corresponds to the expansion terms in Eq(71).
Therefore, for the kth element $V_i = L_kA_{s,k}$.
Using Eq(8) for a skew quad $V_i=-L_k\mathbb{Re}\frac{ia_2}{2}(x+iy)^2=a_2xyL_i$.
Using the resonant base in Eq(87) and applying the rotation and amplitude transformations as in Eq(72) to find $\hat V_i$, we get
$a_2L_k\sqrt{\beta_x\beta_y}(\frac{1}{2}h_x^+e^{i\mu_x}+\frac{1}{2}h_x^- e^{-i\mu_x})(\frac{1}{2}h_y^+e^{i\mu_y}+\frac{1}{2}h_y^- e^{-i\mu_y}).$
We take only the term for `h10010`.
$h_{10010}=\frac{1}{4}a_2L_k\sqrt{\beta_x\beta_y}e^{i\mu_x}e^{-i\mu_y}$.
Bengtsson defines the magnetic field in Eq(10 ) with opposite sign wrt AT. Therefore,
$h_{10010,AT}=-\frac{1}{4}a_2L_k\sqrt{\beta_x\beta_y}e^{i\mu_x}e^{-i\mu_y}$.

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,575 @@
...
+        rdts["h10100"] = sum((a2lm / 4) * rbetaxm * rbetaym * pxm * pym * pf(1, 1))
```
oscarxblanco, line 160: There should be a minus sign in front of the sum.
Here is the explanation:
From Bengtsson's Eq(15) $H\propto-A_s$.
In Eq(63) $M_{\xi0\rightarrow1} = \exp(-\int H ds$).
For a thin lens magnet this implies $M_{\xi0\rightarrow1} = exp(LA_s)$, which corresponds to the expansion terms in Eq(71).
Therefore, for the kth element $V_i = L_kA_{s,k}$.
Using Eq(8) for a skew quad $V_i=-L_k\mathbb{Re}\frac{ia_2}{2}(x+iy)^2=a_2xyL_i$.
Using the resonant base in Eq(87) and applying the rotation and amplitude transformations as in Eq(72) to find $\hat V_i$, we get
$a_2L_k\sqrt{\beta_x\beta_y}(\frac{1}{2}h_x^+e^{i\mu_x}+\frac{1}{2}h_x^- e^{-i\mu_x})(\frac{1}{2}h_y^+e^{i\mu_y}+\frac{1}{2}h_y^- e^{-i\mu_y}).$
We take only the term for `h10100`.
$h_{10100}=\frac{1}{4}a_2L_k\sqrt{\beta_x\beta_y}e^{i\mu_x}e^{i\mu_y}$.
Bengtsson defines the magnetic field in Eq(10 ) with opposite sign wrt AT. Therefore,
$h_{10100,AT}=-\frac{1}{4}a_2L_k\sqrt{\beta_x\beta_y}e^{i\mu_x}e^{i\mu_y}$.

oscarxblanco (changes requested):

pyat/at/physics/rdt.py
swhite2401: @oscarxblanco I have changed the sign of the coupling rdt but now matlab and python are inconsistent, should I also modify matlab? In which case this pauses a problem with backward compatibility.... any suggestion? @carmignani @lfarv ?

oscarxblanco (changes requested):

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,575 @@
...
+            pxm3 = pxm2 * pxm
+            cpxm = np.conj(pxm)
+            cpxm3 = np.conj(pxm3)
+            pym = py[mask_b3l]
+            pym2 = pym * pym
+            cpym2 = np.conj(pym2)
+            cpxym2 = np.conj(pxm * pym2)
+            # fmt: off
+            bsign = np.array([np.sign(sm[i] - sm) * 1j * b3lm[i] * b3lm
+                              for i in range(nelem)])
```
oscarxblanco, line 263: This sign terms are reversed. It should be positive for all the jth elements anywhere downstream wrt to the ith element.
`np.sign(sm-sm[i])`

Junha9609:
Dear @swhite2401 

Hello, Dr. White! 

I encontered the indext error. I using for computeRDT np.arange(len(ring)) and np.arange(len(ring)-1)

these two command get same errors. Can I get your advice?



github1
image

pyat/at/physics/rdt.py
swhite2401: @oscarxblanco , now I am confused. For me the absolute sign does not matter but if I change it in here I also have to change it for the first order terms otherwise the result will be wrong.

More generally don't we want to keep the same as in matlab? Already the coupling terms will differ from one interface to the other...

pyat/at/physics/rdt.py
oscarxblanco: Dear @swhite2401 ,
I see the problem with backwards compatibility.
@carmignani and @lfarv might have a more relevant opinion wrt to the usage of the rdts in matlab.
If I may suggest I would keep the matlab code as it is and include a note in the matlab help message. For example.
```
help computeRDT
...
NOTE: 'coupling' terms are multiplied by a factor -1. This has not been modified to keep backwards compatibility.
```

It is understandable if you still doubt about the correct sign.  If it gives you more confidence on this modification you could compare it with the expression for the vector potential $A$ in Exercise 2.6 - 1, after section VI.5 Linear Coupling Using Transfer Matrix Formalism, in S.Y.Lee Accelerator Physics, Fourth edition.

pyat/at/physics/rdt.py
oscarxblanco: Dear @swhite2401 ,
the sign matters because you are adding the components coming from octupoles, otherwise they could be relative.

Unfortunately, the derivations are giving me the sum term as it is written in Wang's article, and it is different in the code. I will later attach the notes of this derivation in a pdf. They are too long to be written down in markup confortably.

I understand if you prefer to keep it as in matlab, I believe nobody has complained, I guess it has been checked before.

oscarxblanco: Also, the first order terms do not change.
The change is in the crossed-terms from sextupoles.

The error comes when implementing the summation in Eq. (46) on Wang's article. It is positive for $j>i$ and negative for $i > j$.

swhite2401:
> Dear @swhite2401
> 
> Hello, Dr. White!
> 
> I encountered an index error. I am using np.arange(len(ring)) and np.arange(len(ring) - 1) for computeRDT, but both commands give the same error. Could I get your advice?
> 
> github1 image

Dear @Junha9609 

I have tested your lattice file and could not reproduce this error. Below the script I used to compute chromatic RDTs.
Have you tried pulling the latest version of the branch?
Could you please send me your full script?

Thanks,
Simon
```


import at
import numpy as np

ring = at.load_lattice('K4GSR_ideal.mat')
ring.enable_6d()
print(ring.envelope_parameters())
ring.disable_6d()

print(len(ring))

rdts = ring.get_rdts(np.arange(len(ring)), rdt_type=at.RDTType.CHROMATIC, second_order=True, use_mp=True)


n_ignore = 0
for i, k in enumerate(rdts.__annotations__.keys()):
    try:
        if k =='refpts':
            pass
        else:
            plt.plot(ring.get_s_pos(rdts.refpts), np.absolute(rdts[k]), 'o', mfc='none',
                     label=k, color='C{0}'.format(i-n_ignore), linestyle='dashed')
    except (KeyError, TypeError, AttributeError):
        print(k + ' ignored')
        n_ignore += 1
    #plt.xlim([0, 50])
    plt.xlabel('s [m]')
    plt.ylabel('Abs(RDT)')


plt.legend()
plt.show()
```

pyat/at/physics/rdt.py
swhite2401: Sure, what I mean is that this is more than a sign convention as opposed to a change of the sign of the total sum. If signs are not consistent between first and second orders this is a bug that needs to be corrected everywhere (but also reported to the ELEGANT developers?)

Junha9609:
Dear @swhite2401,

Thank you for your comment.

I solved my problem. The issue was that my AT version was too old (0.5.1). I updated my pyAT version to 0.6.1, and now the index error no longer occurs.

However, I encountered a new error when using the avlinopt function in get_rdts. The error message is:
```ruby

TypeError: avlinopt() missing 1 required positional argument: 'dp'
```

So I tried changing my input to the avlinopt function in get_rdts like below:
```ruby
 lo, avebeta, avemu, avedisp, *_ = ring.avlinopt(refpts=All,dp=0)
```
Then, I got another error message:
```ruby
ValueError: NumPy boolean array indexing assignment cannot assign 3481 input values to the 3482 output values where the mask is true
```
As far as I know, my lattice length is 3481. I don't understand why 3482 appeared.

Could I get your comment?

Here is my script I used : 
```
import at
import numpy as np


ring = at.load_lattice('K4GSR_ideal.mat')
ring.enable_6d()
print(ring.envelope_parameters())
ring.disable_6d()

print(len(ring))

rdts = ring.get_rdts(np.arange(len(ring)), rdt_type=at.RDTType.CHROMATIC, second_order=True, use_mp=True)
```
 
> > Dear @swhite2401
> > Hello, Dr. White!
> > I encountered an index error. I am using np.arange(len(ring)) and np.arange(len(ring) - 1) for computeRDT, but both commands give the same error. Could I get your advice?
> > github1 image
> 
> Dear @Junha9609
> 
> I have tested your lattice file and could not reproduce this error. Below the script I used to compute chromatic RDTs. Have you tried pulling the latest version of the branch? Could you please send me your full script?
> 
> Thanks, Simon
> 
> ```
> 
> 
> import at
> import numpy as np
> import matplotlib.pyplot as plt
> 
> ring = at.load_lattice('K4GSR_ideal.mat')
> ring.enable_6d()
> print(ring.envelope_parameters())
> ring.disable_6d()
> 
> print(len(ring))
> 
> rdts = ring.get_rdts(np.arange(len(ring)), rdt_type=at.RDTType.CHROMATIC, second_order=True, use_mp=True)
> 
> 
> n_ignore = 0
> for i, k in enumerate(rdts.__annotations__.keys()):
>     try:
>         if k =='refpts':
>             pass
>         else:
>             plt.plot(ring.get_s_pos(rdts.refpts), np.absolute(rdts[k]), 'o', mfc='none',
>                      label=k, color='C{0}'.format(i-n_ignore), linestyle='dashed')
>     except (KeyError, TypeError, AttributeError):
>         print(k + ' ignored')
>         n_ignore += 1
>     #plt.xlim([0, 50])
>     plt.xlabel('s [m]')
>     plt.ylabel('Abs(RDT)')
> 
> 
> plt.legend()
> plt.show()
> ```

swhite2401:
> However, I encountered a new error when using the avlinopt function in get_rdts. The error message is:
> 
> ```ruby
> TypeError: avlinopt() missing 1 required positional argument: 'dp'
> ```

Dear @Junha9609 this is  very strange, `dp` is not a positional argument in `avelinopt` but a keyword argument. I suspect you installation is still not correct.

Could please try to reinstall from the source from scratch using the following commands:
```
git clone https://github.com/atcollab/at.git
cd at
pip install -e .
git checkout pyat_computeRDT
```
This will get the AT master branch, install it in you active python environment and then switch to the RDT dev branch.

More instructions can be found in [here](https://atcollab.github.io/at/p/howto/Installation.html).

For information the command at.refpts.All adds an additional observation point at the end of your ring this is why you get more indexes than elements. Only some function accept it but `avelinopt` should so I am not sure why it complains, I'll check that on my side.

pyat/at/physics/rdt.py
oscarxblanco: I put the question in the elegant forum.
https://www3.aps.anl.gov/forums/elegant/viewtopic.php?f=17&t=1402

swhite2401: Ah great thanks a lot! Sorry I haven't had time to come back to this lately.... meanwhile should we make a commit that gives the same results as matlab?

oscarxblanco: Dear @swhite2401 ,
Michael Borland has answered today, he confronts the results in Elegant with OPA https://ados.web.psi.ch/opa/ . The sign that matches OPA and elegant is the same you wrote, but opposite to Chunxi's Equation.

I propose the following:
1. Document in https://atcollab.github.io/at/p/index.html all the new formulas  you are adding for the linear terms that are not in any of the citations. Mention them in the DocString.
2. Mention in  https://atcollab.github.io/at/p/index.html the difference in sign between Bengtsson's article and AT. That will clarify why first order terms are written as such.
3. Provide separately the terms produced by octupoles from the second order terms by sextupoles, i.e. do not add them. The user will add or substract them based on his own knowledge.
4. Document in  https://atcollab.github.io/at/p/index.html the difference between your summation and Chun'xi's.

Michael's response does not confirms nor denies this is an error, it is based on what another code does,  therefore, I could not approve this PR without item 3.

swhite2401: Dear @oscarxblanco , I would prefer to document that we have set the signs to follow the convention used in ELEGANT for all terms and that there is descrepencies w.r.t. to the reference, such that results are identical in matlab and python.

I will provide first and second order terms separately as suggested. 

In addition, I will verify that the detuning coefficient are consistent with tracking, this should clear any ambiguity on the sign.

For what concerns RDTs it should be possible to reconstruct them using tbt tracking data but this is not so easy... I don't really have time to do that now. We can keep it for later to confirm all the signs.

Would that be acceptable for you?

oscarxblanco: > Dear @oscarxblanco , I would prefer to document that we have set the signs to follow the convention used in ELEGANT for all terms and that there is descrepencies w.r.t. to the reference, such that results are identical in matlab and python.

OK.
> 
> I will provide first and second order terms separately as suggested.

Perfect, thank you.
> 
> In addition, I will verify that the detuning coefficient are consistent with tracking, this should clear any ambiguity on the sign.

If you refer to compare the output of `at.detuning` wrt to the terms `dnux_dJx`, `dnuy_dJy` and `dnux_dJy` this is already done, the values are correct but lead to issue #836 asking to fix the documentation only.
In addition, Bengtsson's Eq (119) depends on the absolute value of the phase advance between sextupoles, so different to the Poisson bracket the detuning has no difference for sextupoles $i,j$ or $j,i$. And the equation itself has a different notation using $j,k$ as indexes which are not explained and refer to another public document which has a different notation.
I think one would need to derive the expression for the detuning following Bengtsson definitions.


> 
> For what concerns RDTs it should be possible to reconstruct them using tbt tracking data but this is not so easy... I don't really have time to do that now. We can keep it for later to confirm all the signs.

Great, that would solve it !
> 
> Would that be acceptable for you?

Yes, of course. Thank you.

swhite2401:
@oscarxblanco , I think I have implemented all the changes discussed:
1-All the signs are consistent between MATLAB and python
2-First and second order terms are provided separately, as well as as their sum
3-docstrongs were improved to mention the sign convention descrepency and ad reference was added for the focusing rdts.

Also I remove the dataclass implementation, in the end I find a dictionary to be more convenient.

Anything else?

@lfarv, any comments?

oscarxblanco:
@swhite2401 . The numpy error has disappeared, thank you.

I find the help message unclear. Here below is my suggestion, let me know if you prefer not to modify it.
While rewriting the help message I found a mistake, in **refts** instead of **refpts**, that should be corrected.

After these I could enable the merge.
```
    :py:func:`get_rdts` computes the ring RDTs based on the original implementation from ELEGANT.
    For consistency, pyat keeps the sign convention of the AT MATLAB interface.

    Refs [1]_ and [2]_ were used to calculate the magnitude of first and second order rdts and
    Ref. [3]_ for the focusing rdt, however, different sign conventions are used along the references
    and in the original implementation. To overcome this issue, first and second order rdts
    are provided as a total and also separately so the user can redefine these conventions if needed.
    
    The periodicity property of the lattice is automatically taken into account in the rdt
    calculation, except for the GEOMETRIC2 and DETUNING types which are only implemented for
    periodicity=1 (i.e. full ring).

    The resonance base of GEOMETRIC2 rdts is explained in Eq. (54) of Ref [4]_.
    ...
        **refpts**           location of the rdt
    ...

       **[4]** Chunxi Wang and Alex Chao. Notes on Lie algebraic analysis of achromats. SLAC/AP-100. Jan/1995
```

o

oscarxblanco:
@swhite2401 , if I leave the default or set `second_order=False`, it seems the second order values are still calculated but the phases are wrong. See the plot below were I execute those lines and I plot all first order rdts common to matlab and `h20110`.
Is there anyway to avoid having wrong phase results ? I think that zeros on non-requested rdts, or non-defined rdts if not requested is less prone to misuse.
```
rdts = at.get_rdts(ring,refpts,at.RDTType.ALL,second_order=False)
thepyrdts = rdts[0]
```

image

oscarxblanco (approved):

swhite2401:
I will add some tests before merging

swhite2401:
Dear all,
I think this is ready for a (hopefully) final review, I have added some tests and fixed a bug on the periodic factor calculation w.r.t. previous version.
Everything else is the same.

pyat/at/physics/rdt.py
```diff
@@ -0,0 +1,569 @@
...
+
+    Returns:
+        rdts: rdt data (complex) at refpts
+        rdts2: (complex) contribution from sextupole second order terms
+          Available only for GEOMETRIC2 terms
+        rdttot: (complex) total rdts
+
+        **rdts** is a dictionary with keys:
+        =================   ======
+        **refpts**           location of the rdt
```
oscarxblanco, line 455: There is one extra space before location that does not match the column alignment.

oscarxblanco: Dear @swhite2401, all other test are  correct. After fixing this typo you could merge.

oscarxblanco (approved):

PR #822, opened 2024-08-31
Fix wiggler radiation

lfarv:
The diffusion matrix of wigglers was recently introduced (#759) but has a few problems:

#### For Matlab
- FDW.c is not compiled in `atmexall`, and so is missing when needed,
- Compilation fails when requested on MacOS (wrong ordering of #include lines).

#### For python
- FDW.c is not compiled in `setup.py` (same as for Matlab),
- FDW is not imported in the `at.physics` module,
- Wigglers are not considered in the computation of emittances.

This PR corrects all this. As mentioned before, we have no way of cross-checking the results (specially for the python implementation), so we trust the computation. This is anyway better than before when it was wrong.

### DIFF
--- a/atintegrators/gwigR.c
+++ b/atintegrators/gwigR.c
@@ -124,7 +124,6 @@ static void GWigPass_4th(struct gwigR *pWig, double *X)
   double dl, dl1, dl0;
   double B[3];
   double ax, ay, axpy, aypx;
-  double factorvect;
 
   Nstep = pWig->PN*(pWig->Nw);
   dl = pWig->Lw/(pWig->PN);
@@ -450,7 +449,7 @@ static void GWigB(struct gwigR *pWig, double *Xvec, double *B)
 {
   int    i;
   double x, y, z;
-  double kx, ky, kz, tz, kw;
+  double kx, ky, kz, tz;
   double cx, sx, chx, shx;
   double cy, sy, chy, shy;
   double cz, sz;
@@ -460,8 +459,6 @@ static void GWigB(struct gwigR *pWig, double *Xvec, double *B)
   x = Xvec[0];
   y = Xvec[2];
   z = pWig->Zw;
-
-  kw   = 2e0*PI/(pWig->Lw);
 
   B[0] = 0;
   B[1] = 0;
@@ -596,9 +593,9 @@ static void GWigRadiationKicks(struct gwigR *pWig, double *X, double *Bxy, doubl
 static void Hessian(struct gwigR *pWig, double *Xvec, double *H2)
 {
 	int i,j,k;
-	double x,px,y,py,D;
-	double ax,axx,axxx,axy,axyy,axxy;
-	double ay, ayx, ayxx, ayy, ayyy, ayxy;
+	double px, py,D;
+	double ax,axx,axxx,axy,axxy;
+	double ay, ayx, ayxx, ayy, ayxy;
 	double Pax[6];
 	double Pay[6];
 	double H[6][6];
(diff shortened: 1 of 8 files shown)


ZeusMarti:
@lfarv ,

Not sure How @joanarenillas was compiling it, probably manually. He just worked on the Matlab part, sorry I should have stated that clearly. 

I'm not able to review this, specially not the compilation in the python part, but thanks for including me, I'll try to go through it and learn as much as possible.

lfarv:
Hi all. Please have a look to this PR: without these small fixes, all the work done by @joanarenillas and @ZeusMarti is still not usable, the wiggler contribution still fails.

There is no change in the wiggler code except for fixing a compilation bug on MacOS, removal of compiler warnings and triggering the compilation in atmexall.

oscarxblanco:
Dear @lfarv ,
it seems that the wiggler element in pyat is created without 'Energy'. See here below when I insert a fake wiggler at the end of the ring.
```python
>>> ring[-1]
  File "/home/oblanco/Documents/public/progs/pyenv/testwiggler/lib/python3.11/site-packages/at/lattice/elements.py", line 311, in __repr__
    clsname, args, kwargs = self.definition
                            ^^^^^^^^^^^^^^^
  File "/home/oblanco/Documents/public/progs/pyenv/testwiggler/lib/python3.11/site-packages/at/lattice/elements.py", line 397, in definition
    arguments = tuple(attrs.pop(
                ^^^^^^^^^^^^^^^^
  File "/home/oblanco/Documents/public/progs/pyenv/testwiggler/lib/python3.11/site-packages/at/lattice/elements.py", line 398, in <genexpr>
    k, getattr(self, k)) for k in self._BUILD_ATTRIBUTES
       ^^^^^^^^^^^^^^^^
AttributeError: 'Wiggler' object has no attribute 'Energy'
```

Here is the python code:
```python
"""This script test the wiggler use"""
# 2024sep06 oblanco
import at
print(f'at.__version__={at.__version__}')
ring = at.load_mat('ring.mat')
# create wiggler
wlength = 1e-4
wperiod = 1e-5
wbmax = 1e-3
wenergyineV=3e9 #eV
WIG = at.Wiggler('WIG',wlength,wperiod,wbmax,wenergyineV)
ring.insert(len(ring),WIG)
# ring[-1] # this line fails if lauched interactively
```

In matlab AT I can create, add to ring, plot, track and use ohmienvelope. Here is the code I wrote for the test.
It runs without errors but I haven't checked the physics.
```matlab
clear; close all;
%% oblanco 2024sep06 test wiggler use
load('ring.mat');
%%
wlength = 0.00001;%3*i; % wiggler length
periodlength = 0.000001; % wiggler period
bmax = 1.6e-3;%1.5; % max field in Tesla
beamenergy  = 3e9; % in eV
passmethod = 'GWigSymplecticRadPass'; % rad or no rad
thewiggler = atwiggler('WIG',wlength,periodlength,bmax,beamenergy,passmethod);
%%
[~,radindex,~] = atenable_6d(THERING);
%%
[ENVELOPE, RMSDP, RMSBL, M66, T, ORBIT] = ohmienvelope(THERING,radindex);
%%
THERING2 = [THERING; thewiggler];
%%
atplot(THERING2);
%%
Z0=ringpass(THERING2,zeros(6,1),1000);
%%
plot(Z0(1,:));
%%
[~,radindex2,~] = atenable_6d(THERING2);
[ENVELOPE_2, RMSDP_2, RMSBL_2, M66_2, T_2, ORBIT_2] = ohmienvelope(THERING2,radindex2);
```

lfarv:
@oscarxblanco : thanks for testing. As far as I can see, Matlab is correct. In python, the wiggles indeed created with an Energy attribute, but it is deleted when inserted in ring. This is because in python, the energy is never taken from the element, but from the lattice. The bug was in the display of the wiggler element, but should be solved now.

oscarxblanco:
pyat now prints the wiggler without errors. I tested also at.plot_beta and ohmienvelope, both seem ok.

oscarxblanco (approved):

PR #824, opened 2024-09-03
Fix error messages in load_madx

lfarv:
This PR fixes `load_madx` with in certain cases prints error messages about "unknown commands" even if the processing is correct.

This was reported in #820

### DIFF
--- a/pyat/at/load/file_input.py
+++ b/pyat/at/load/file_input.py
@@ -438,14 +438,17 @@ class BaseParser(DictNoDot):
         self,
         lines: Iterable[str],
         final: bool = True,
+        **kwargs,
     ) -> None:
         """Process input lines and fill the database
 
         Args:
             lines: Iterable of input lines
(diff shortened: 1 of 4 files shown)


simoneliuzzo (approved):

lfarv:
@simoneliuzzo : The last modifications solve the problems reported in #820:
- the "const", "real" and "int' MADX keywords are interpreted (and ignored). This solves the problem of zero-length magnets
- saving in .mat format works again

@simoneliuzzo : can you approve again (if you are happy with this version)?

simoneliuzzo:
All seems ok!
conversion is correct on the first pass.
t =

    0.2250    0.2900

c =

    2.0447    1.7691

Screenshot 2024-09-05 at 09 27 23

also save_mat is working correctly.

all ok for me!

simoneliuzzo (approved):

Issue #825, opened 2024-09-04
ohmienvelope with radindex length smaller than the ring produces an error


oscarxblanco:
Dear all,

it seems to me there is a bug in at matlab `ohmienvelope`.
The following lines use the variable `Wig` to mask `radindex`.
The length of `Wig` is equal to the length of the ring, while, the length of `radindex` could be equal or smaller than the length of the ring.
Matlab does not throw an error, but, does an unexpected operation with the lists of different lengths. Could you confirm it ?
https://github.com/atcollab/at/blob/8ba4eb17ea22c839ab13f9d916b2d548cdc02152/atmat/atphysics/Radiation/ohmienvelope.m#L29-L30

Best regards,
o

lfarv:
`radindex` is normally provided by `atenable_6d`:
```
if nargout > 1
    cavitiesIndex=atgetcells(newring,'PassMethod',@(elem,pass) endsWith(pass,'CavityPass'));
    radelemIndex=atgetcells(newring,'PassMethod',@(elem,pass) endsWith(pass,'RadPass'));
end
```
and has the same length as `Wig`, so for me there is no problem.

oscarxblanco:
Oh, I see, `radindex` is actually a mask.
I mean not [1,2,3,...] but zeros and ones [0,1,0,1,...]

PR #828, opened 2024-09-04
Fix ohmienvelope wiggler mask

oscarxblanco:
This PR solves issue #825 .
The wiggler indexes and radiation element indexes are intersected. Common elements, if any, are removed from the radiation element list.
This is done in order to keep indexes compatible with the separation of the diffusion matrix calculation for the wigglers.

Additionally, help is modified and non-strict check_6d is included.

I also found out that the atmexall was not compiling `FDW.c`. I added it.

o

### DIFF
--- a/atmat/atmexall.m
+++ b/atmat/atmexall.m
@@ -101,6 +101,7 @@ oldwarns=warning('OFF','MATLAB:mex:GccVersion_link');
 % Diffusion matrices
 cdir=fullfile(atroot,'atphysics','Radiation');
 compile([alloptions, {passinclude}], fullfile(cdir,'findmpoleraddiffmatrix.c'));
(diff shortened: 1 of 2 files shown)


oscarxblanco:
Dear @swhite2401 and @lfarv ,
I have included `FDW.c` in `atmexall.m` . It seems to have issues in macos-13. Do you know what could be happening ?

lfarv:
As I mentioned `here`, there seems to be no problem with indexes.
Concerning FDW.c and `ohmienvelope`, this is corrected in #822, which supersedes this PR.

@oscarxblanco: I suggest that you rather look at #822 (I add you to the reviewers). You'll see by the way that radindex is now useless.

PR #830, opened 2024-09-04
matlab ohmienvelope add help message; add check_6d

oscarxblanco:
This PR modifies the help message displayed in matlab for the function `ohmienvelope`.
It also adds a check_6d set to strict level zero.


oscarxblanco:
Dear @swhite2401  and @lfarv ,
would you consider ok to add this modifications ?
o

lfarv:
@oscarxblanco:
- Ok for the help message (you're encouraged to improve he help everywhere!)
- The check_6d is not strictly necessary: `ohmienvelope` behaves correctly for any lattice and gives the correct diffusion matrix according to the passmethods it encounters, possibly zeros(6,6) if there is no radiating element. In such a case, the envelope matrix is NaN(6,6), which is also correct. But such a case is of no interest, so why not a warning.

That's ok for me.

lfarv (approved):
OK for me

Issue #829, opened 2024-09-04
PyAT lattices saved in .m format are not recognised by Matlab


lfarv:
Since numpy 2.0 was released, saving lattice in .m format generate files which cannot be interpreted by Matlab. The reason is that some values are written as `"np.float64(1.5)"` instead of `"1.5"` as before. Until this is solved, the recommendation is to use the JSON format, another text format recognised by both Matlab and python

lfarv:
fixed in #832

PR #832, opened 2024-09-06
Fix saving lattices in m-files

lfarv:
This fixes a problem described in #829: the string representation of numpy arrays changed in numpy 2.0, resulting in m-files refused by Matlab.

This is corrected here.

### DIFF
--- a/pyat/at/load/madx.py
+++ b/pyat/at/load/madx.py
@@ -713,7 +713,7 @@ class _MadParser(UnorderedParser):
             if cavities:
                 cavities.sort(key=lambda el: el.Frequency)
                 c0 = cavities[0]
-                params["_harmnumber"] = getattr(c0, "HarmNumber", np.nan)
+                params["_cell_harmnumber"] = getattr(c0, "HarmNumber", np.nan)
 
         part = kwargs.get("particle", None)
         if isinstance(part, str):
--- a/pyat/at/load/matfile.py
+++ b/pyat/at/load/matfile.py
@@ -2,14 +2,14 @@
 Load lattices from Matlab files.
 """
 
-from __future__ import print_function, annotations
+from __future__ import annotations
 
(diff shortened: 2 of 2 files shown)


pyat/at/load/madx.py
```diff
@@ -713,7 +713,7 @@ def beta() -> float:
             if cavities:
                 cavities.sort(key=lambda el: el.Frequency)
                 c0 = cavities[0]
-                params["_harmnumber"] = getattr(c0, "HarmNumber", np.nan)
+                params["_cell_harmnumber"] = getattr(c0, "HarmNumber", np.nan)
```
simoneliuzzo, line 716: this modification does not seem to belong to the same pull request. is it needed? what is _cell_harmnumber?

pyat/at/load/matfile.py
```diff
@@ -157,7 +157,10 @@ def ringparam_filter(
...
 
     if len(ringparams) > 1:
-        warn(AtWarning("More than 1 RingParam element, the 1st one is used"))
+        warn(
+            AtWarning("More than 1 RingParam element, the 1st one is used"),
+            stacklevel=2,
+        )
 
 
 def load_mat(filename: str, **kwargs) -> Lattice:
```
simoneliuzzo, line 166: load_mat tested, OK. also tested load of .mat file in Matlab. OK.

pyat/at/load/matfile.py
```diff
@@ -418,21 +421,29 @@ def scan(d):
                     yield convert(k)
                     yield convert(v)
 
-            return "struct({0})".format(", ".join(scan(pdir)))
+            return "struct({})".format(", ".join(scan(pdir)))
```
simoneliuzzo, line 424: tested save_mat. OK, file is saved and can be loaded in python and in matlab yelding the same lattice.

simoneliuzzo (approved):

Issue #836, opened 2024-09-16
pyat detuning help message and lattice periodicity


oscarxblanco:
Dear all,
the function `at.detuning` in pyat calculates the derivative wrt action $J=\frac{x^2\gamma}{2}$ where $\gamma=\frac{1+\alpha^2}{\beta}$.
The help message for `q1` should be `[[dQx/dJx, dQy/dJx], [dQx/dJy, dQy/dJy]]`.
Instead, the help message states it is the derivative wrt to $x$ and $y$.
```python
>>>help(at.detuning)
detuning(ring: at.lattice.lattice_object.Lattice, xm: Optional[float] = 3e-05, ym: Optional[float] = 3e-05, npoints:
...
        q1 (ndarray): Amplitude detuning coefficients with shape (2, 2)
                      ``[[dQx/dx, dQy/dx], [dQx/dy, dQy/dy]]``
```
it also seems to fail when periodicity is larger than one.
See https://github.com/atcollab/at/pull/821#issuecomment-2353087386

PR #837, opened 2024-09-17
Tune, chromaticity and detuning periodicty

swhite2401:
This PR was initially answering #836, however by doing so I realized that the function `get_tune()` and `chromaticity()` were also not taking into account the ring periodicity.

This is now fixed, black and `import numpy as np` were applied to all the modules that are modified in this PR, most of the changes come from these....

I think this is a necessary fix but is not backward compatible and may affect scripts using a periodicity different from 1, is this acceptable?


pyat/at/physics/linear.py
```diff
@@ -1370,36 +1457,46 @@ def gen_centroid(ring, ampl, nturns, remove_dc, ld):
...
+        assert method == "linopt", "Integer tune only accessible with method=linopt"
+    if method == "linopt":
         if get_integer:
-            _, _, c = get_optics(ring, refpts=range(len(ring)+1),
-                                 dp=dp, dct=dct, df=df, orbit=orbit)
-            tunes = c.mu[-1]/(2*numpy.pi)
+            _, _, c = get_optics(
+                ring, refpts=range(len(ring) + 1), dp=dp, dct=dct, df=df, orbit=orbit
+            )
+            tunes = c.mu[-1] / (2 * np.pi) * ring.peridicity
```
oscarxblanco, line 1474: There is a typo in `ring.peridicity`

oscarxblanco (changes requested):

pyat/at/physics/linear.py
swhite2401: Done

swhite2401:
@oscarxblanco thanks for the check, I have added a unit test to check that the periodicity is correctly implemented in get_tune(), get_chrom(), detuning() and chromaticity()

swhite2401:
@lfarv I need advice here:  the `periodicity` is also not implemented in matlab for the `tunechrom` function. This function may be used in many places and I am really not sure what the consequences would be to correct it.

Now I am going to complain sorry...: this `periodicity` parameter is presently a disaster we have no idea where it applies and where it does not and no one I know of is using it. I did put an issue in the past on this topic and it was not fixed.
So now either we completely remove it, which would simplify a lot the code or we fix it once and for all, but this hybrid situation is really very very bad an unacceptable.

My perference is to remove it off course... now this is up for discussion...

oscarxblanco:
Dear @swhite2401 ,
the last commit is giving me the total chromaticity and detuning as expected when including the periodicity.
About the tune, the fractional part given by `get_tune` matches what I expected when including the periodicity. This is OK for me.

It could be somehow unexpected that a fodo returns such a big chromaticity and tune that is far from the one that could be estimated for a single cell.
 
If I may give an opinion on the periodicity, I think it should be stated in the DocString when and when is not considered. And also somewhere in the documentation if periodicity is fixed or could be override or overwritten. This because it is sometimes is useful to have the values for a single fodo cell.

swhite2401:
> This because it is sometimes is useful to have the values for a single fodo cell.

How to set periodicity is probably documented together with the lattice object, I have to check.... but in any case I agree that it is clearly not adequate and has to be improved

However, the feedback that I got (and that I share) is that the periodicity is an annoying feature rather than something convenient. In the users end up proceeding like this: when they want one cell they load one cell with periodicity=1 and when they want the full ring they do ring.repeat(ncells).
This is simple, well understood and you know what you are doing and what to expect. Anything else produces random results depending on what function is called or how you loaded you lattice.

The only useful case I could identify is for `radiation_parameters()`, `envelope_parameters()` and eventually to generate a `fast_ring`. Periodicity could be set locally in these functions rather than polluting the whole code...

Anyway,  this is my users and developers opinion not necessarily the one of the majority...

lfarv:
@swhite2401, @oscarxblanco:

My comments on `periodicity`:
- `periodicity` is really useful for periodic machines: when interested in the full ring, properly using it saves lot of computing time compared to the `ring.repeat(cells)` solution (a factor `ncells` in all optics computations),
- For those finding it useless or even disturbing, a simple solution is to set it to 1. It's simple and you can forget about it,
- @oscarxblanco: yes, the `periodicity` attribute can be changed at any time to switch between single cell and full ring.

However I agree that as it is now, it may create problems and I agree with most of you proposals:
- Document where `periodicity` is used and when not. As a simple rule, it should not be taken into account in functions using `refpts`: you can index elements only in a single cell, so all results should be consistent with that, including global ones.
- A modification could be to set `periodicity` to 1 when creating a lattice from a list of elements or from a foreign data file. At the moment, it is automatically set so that the bending angle for the full ring is $2 \pi$ (clever, no?). Instead, it could be set to 1, so that the user would need to explicitly set it to use it. This is a behaviour change, but I think it's worth it.

lfarv (commented):
Good point that you "blackened" those large files !

While you are on it, you could add a line in the docstring of `linopt2`, `linopt4`, `linopt6`, `linopt`, `get_optics`:

```
r"""Linear analysis of a fully coupled lattice using normal modes

:py:func:`linopt6` computes the linear optics parameters on a single cell
(*periodicity* is not taken into account).
...
```
For `avlinopt`:
```
r"""Linear analysis of a lattice with average values

:py:func:`avlinopt` returns average beta, mu, dispersion over the lattice
elements. :py:func:`avlinopt` acts on a single cell (*periodicity* is not taken
into account).
...
```
For `get_tune` and `get_chrom`:
```
r"""Computes the tunes of the full ring, taking *periodicity* into account

:py:func:`get_tune` may use several methods depending on a *method* keyword.
...
```

This is already a step to a better documentation!

pyat/at/physics/nonlinear.py
```diff
@@ -37,49 +42,65 @@ def tunes_vs_amp(ring: Lattice, amp: Optional[Sequence[float]] = None,
...
+        )
+        part = (
+            np.array(
+                [
+                    orbit,
+                ]
+                * len(amp)
+            ).T
+            + 1.0e-6
+        )
```
lfarv, line 66: Surprising result !

Apparently this is due to the comma after `orbit`. Why this comma ? in a list, is useless, but I don't understand the whole statement…

swhite2401: Yes it was the comma. You are right, I am not so sure why this 1.0e-6 was added to all coordinates and how its value was determined, maybe to be sure you get some signal from the cross terms, @lcarver any idea?

At least its value could be determined to adapt to the beta functions....

swhite2401:
Any further comments for this branch? Can I merge?

lfarv (approved):
OK for me.
Any further comment, @oscarxblanco ?

oscarxblanco:
No, it is OK for me too.

oscarxblanco (approved):

Issue #838, opened 2024-10-02
pyat fails to load .mat from matlab with element `VariableMultipole`


oscarxblanco:
Dear all,
I get the following error when loading a .mat file saved in matlab.
```python
>>> ring = at.load_mat('ring.mat')
Traceback (most recent call last):
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/lattice/elements.py", line 296, in __setattr__
    value = self._conversions.get(key, _nop)(value)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: only length-1 arrays can be converted to Python scalars

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/load/matfile.py", line 202, in load_mat
    return Lattice(
           ^^^^^^^^
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/lattice/lattice_object.py", line 191, in __init__
    super(Lattice, self).__init__(elems)
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/lattice/lattice_object.py", line 1478, in params_filter
    for elem in elem_filter(params, *args):
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/load/matfile.py", line 140, in ringparam_filter
    for elem in elem_iterator(params, *args):
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/load/matfile.py", line 105, in _matfile_generator
    yield Element.from_dict(kwargs, index=index, check=check, quiet=quiet)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/load/utils.py", line 365, in element_from_dict
    element = cls(*(arg for arg in elem_args if arg is not None), **elem_dict)
[...]

PR #839, opened 2024-10-03
`VariableMultipole` compatibility between matlab AT and pyat

oscarxblanco:
This pull request address issue #838 .
The pyat element class `VariableMultipole` is modified.
The attribute `to_dict` is checked in order to determine if the element is already created, if it has not been created the attributes are assigned as before.
This also fixes a bug in the determination of `MaxOrder`. Before it could be None, now it is limited to be at least zero.


ZeusMarti:
Not  totally related, but I think it is also an small bug in In Matlab, I get this weird behavior:

KIDDK = atvariablemultipole('KIDDK','arbitrary','VariableThinMPolePass',...
    'AmplitudeB',[0 0 0 0],...
    'AmplitudeA',[0 0 0 0],...
    'FuncA',[1 0],...
    'FuncB',[1 0])

gives:

 struct with fields:

       FamName: 'KIDDK'
        Length: 0
         Class: 'VariableMultipole'
          Mode: 2
      PolynomA: []
      PolynomB: []
    AmplitudeB: [0 0 0 0]
    AmplitudeA: [0 0 0 0]
         FuncA: [1 0]
         FuncB: [1 0]
      MaxOrder: -2
     NSamplesA: 2
     NSamplesB: 2

If MaxOrder is fixed to a value then it is ok.

oscarxblanco:
Dear @ZeusMarti ,
I confirm it. The same weird behaviour is in pyat, I'm working on a solution in pyat compatible with matlab. Once I have it I will fix the bug in matlab.

oscarxblanco:
Dear @swhite2401 ,
who could review this pull request ?

swhite2401:
@oscarxblanco , I can do it. In fact there are some changes that you did that I don't understand.
Also I like that important input argument are not integrated in `kwargs` so that they appear in the API doc. I may propose changes to the `__init__()`

pyat/at/lattice/variable_elements.py
```diff
@@ -68,74 +89,103 @@ def __init__(self, family_name, AmplitudeA=None, AmplitudeB=None,
...
+            if modefromdict is not None:
+                mode = modefromdict  # matlab class
+            else:
+                mode = modepyatinput
+            self.Mode = int(mode)
+            # MaxOrder is later overwritten by lenth of amplitudes
+            self.MaxOrder = kwargs.get("MaxOrder", 0)
+            amplitudea = None
+            amplitudeb = None
+            if "AmplitudeA" in kwargs:
```
swhite2401, line 118: This lol ne and the ones below can be removed

pyat/at/lattice/variable_elements.py
```diff
@@ -68,74 +89,103 @@ def __init__(self, family_name, AmplitudeA=None, AmplitudeB=None,
...
             delta = self.MaxOrder + 1 - len(ampb)
             if delta > 0:
-                ampb = numpy.pad(ampb, (0, delta))
-            setattr(self, 'AmplitudeB', ampb)
+                ampb = np.pad(ampb, (0, delta))
+            self.AmplitudeB = ampb
 
-    def _set_params(self, amplitude, mode, ab, **kwargs):
+    def _set_params(
+        self, amplitude: int or str, mode, a_b: str, **kwargs: dict[str, any]
```
swhite2401, line 165: No need to define the type of kwargs

swhite2401 (commented):
Hi @oscarxblanco , here is a first round of comments

pyat/at/lattice/variable_elements.py
```diff
@@ -1,64 +1,85 @@
...
+        Seed=int,
+        NSamplesA=int,
+        NSamplesB=int,
+        FuncA=_array,
+        FuncB=_array,
+        Ramps=_array,
+        Periodic=bool,
+    )
+
+    def __init__(self, family_name: str, **kwargs: dict[str, any]):
```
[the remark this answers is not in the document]
oscarxblanco: Dear @swhite2401 , if I remove the annotation I get the following flake8 warning. This is the reason why I included, should I ignore it ?  I may be using flake8 with a wrong configuration or I may use another tool ?
```
/variable_elements.py:42:44: ANN003 Missing type annotation for **kwargs
```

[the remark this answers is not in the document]
oscarxblanco: > Could you please restore the defaults for AmplitudeA/B and the mode? I think it is better if they appear in the API doc, in fact I would be tempted to add more of these keyword arguments rather than removing them

Dear @swhite2401 , the reason why I remove it is because of an incompatibility between AT matlab and this implementation of the element in pyat.
If I save a lattice in matlab and try to read it in pyat I get the error here below, I'll try to explain the reason in  detail, but, in few words: **one cannot have args and kwargs with the same name**

When reading an element from matlab it is initially put in a dictionary `elem_dict`, and then the `_BUILD_ATTRIBUTES` are stripped to pass them to the class instantiation as `elem_args`. See the lines below
https://github.com/atcollab/at/blob/268994f8214cb758714eae5a3f55e95e5e915ad4/pyat/at/load/utils.py#L368-L369

I could not put the AmplutudeA, AmplitudeB and Mode in the element BUILD_ATTRIBUTES because they are not garanteed to exist in matlab, see the following line
https://github.com/atcollab/at/blob/268994f8214cb758714eae5a3f55e95e5e915ad4/atmat/lattice/element_creation/atvariablemultipole.m#L1
therefore, the call to cls with args in necessarily incomplete and the AmplitudeA, AmplitudeB and Mode are passed in a dictionary together with args that have the same name. Python does not handle this mix well, it removes the elements from the kwargs and leaves unset the args.

There are ways to workaround this situation when args and kwargs have the same name, but they require changing the way the class is instantiated in utils.py. I decided not to do it because that will change how all elements are instantiated.

A second possibility was to rename the element args in pyat, for example amplitudeA instead of AmplitudeA i.e. lower case in the first letter to make one small change. However, this would break compatibility.

I decided to keep the element without any requirement as in AT matlab, which forced me to do checks on the input parameters before assigning them inside `__init__`. This wouldn't break compatibility with previous versions of pyat, and AT matlab files could be read because the element also has no requirements except for the FamName.

If you prefer, I could do as you asked. Keep the args and break backwards compatibility. Or force the user to provide AmplitudeA, AmplitudeB and Mode in matlab, which will make possible to include them in BUILD_ATTRIBUTES that will be removed in the input dictionary and set before the class is instantiated.

Or maybe there is another solution ?

```python
>>> ring = at.load_mat('test1.mat')
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/load/matfile.py", line 202, in load_mat
    return Lattice(
           ^^^^^^^^
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/lattice/lattice_object.py", line 191, in __init__
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/lattice/lattice_object.py", line 1478, in params_filter
    for elem in elem_filter(params, *args):
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/load/matfile.py", line 140, in ringparam_filter
    for elem in elem_iterator(params, *args):
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/load/matfile.py", line 105, in _matfile_generator
    yield Element.from_dict(kwargs, index=index, check=check, quiet=quiet)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/load/utils.py", line 365, in element_from_dict
    element = cls(*(arg for arg in elem_args if arg is not None), **elem_dict)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/lattice/variable_elements.py", line 111, in __init__
    raise AtError("Please provide at least one amplitude for A or B")
at.lattice.utils.AtError: Please provide at least one amplitude for A or B
```

pyat/at/lattice/variable_elements.py
```diff
@@ -68,74 +89,103 @@ def __init__(self, family_name, AmplitudeA=None, AmplitudeB=None,
...
-        self.PolynomB = numpy.zeros(self.MaxOrder+1)
-        ramps = kwargs.pop('Ramps', None)
-        if ramps is not None:
-            assert len(ramps)==4, \
-                'Ramps has to be a vector with 4 elements'
-            self.Ramps = ramps
-        super(VariableMultipole, self).__init__(family_name, **kwargs)
-
-    def _setmaxorder(self, ampa, ampb):
+        if len(kwargs) > 0:
```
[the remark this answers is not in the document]
oscarxblanco: This was needed in order to separate the case when the element is called inside pyat, therefore the object already exists and there are no kwargs, from the case when the element is instantiated using a dictionary (the dictionary comming from reading a .mat, or from user settings).

Sometimes the object is read, for example when printing in a terminal, the element is passed again (I guess this happens because it inherits `__str__` from other element) and no modifications in the object are needed, I just need to pass it to super and find its way to `__str__`.

pyat/at/lattice/variable_elements.py
```diff
@@ -68,74 +89,103 @@ def __init__(self, family_name, AmplitudeA=None, AmplitudeB=None,
...
+            self.FamName = family_name
```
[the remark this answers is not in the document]
oscarxblanco: I explained before, but, when the element comes from AT matlab, a dictionary is created but all BUILD_ATTRIBUTES are stripped. 'FamName' is a a BUILD_ATTRIBUTE of the class element and therefore it is not in kwargs. One has to take it from args.

pyat/at/lattice/variable_elements.py
```diff
@@ -68,74 +89,103 @@ def __init__(self, family_name, AmplitudeA=None, AmplitudeB=None,
...
+            if "AmplitudeA" not in kwargs and "AmplitudeB" not in kwargs:
```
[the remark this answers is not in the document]
oscarxblanco: Sorry @swhite2401 , I don't understand. What you say is implemented in line 103 above.
You mean individually ? I thought this element should accept to have at least one of the two.

pyat/at/lattice/variable_elements.py
oscarxblanco: What do you mean by lol ?

pyat/at/lattice/variable_elements.py
```diff
@@ -68,74 +89,103 @@ def __init__(self, family_name, AmplitudeA=None, AmplitudeB=None,
...
+            # check ramps
+            ramps = kwargs.pop("Ramps", None)
+            if ramps is not None:
+                if len(ramps) != 4:
+                    raise AtError("Ramps has to be a vector with 4 elements")
+                self.Ramps = ramps
+        # fill in super class
+        super().__init__(family_name, **kwargs)
+
+    def _setmaxorder(self, ampa: np.ndarray, ampb: np.ndarray):
```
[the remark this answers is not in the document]
oscarxblanco: I have to repeat the test but suppose you have the current implementation available in the master. You will have this behaviour with zeros.
```python
>>> numpy.max(numpy.nonzero([0,0,0,0]))
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/numpy/core/fromnumeric.py", line 2810, in max
    return _wrapreduction(a, np.maximum, 'max', axis, None, out,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/numpy/core/fromnumeric.py", line 88, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: zero-size array to reduction operation maximum which has no identity
```

I changed to ` mxa = np.max(np.append(np.nonzero(ampa), 0))` which is the maximum between the nonzero in ampa and zero, and I get sure ampa exists before doing it.

pyat/at/lattice/variable_elements.py
oscarxblanco: It is flake8 who suggested. Should I ignore it ? Do you work with some custom flake8 settings for pyat development ?

pyat/at/lattice/variable_elements.py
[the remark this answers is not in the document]
oscarxblanco: Dear @swhite2401 , I asked @ZeusMarti  and he was not able to use it effectively because he intends to have a custom function where the time of the particle is also considered, i.e. not only a kick every turn for all particles.
I have started modifying the C code today to make some test, is this feature something you would accept for this element ? Or should I keep the custom function option as it is ?

o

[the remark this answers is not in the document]
swhite2401: > I have started modifying the C code today to make some test, is this feature something you would accept for this element ?

Sure! In fact I developed this for @ZeusMarti initially, so any improvement is perfectly fine.
You can already do a kick every n turns by using the table by the way, just fill zeros where you do not want any kick.

Do you want help for this?

oscarxblanco:
Dear @swhite2401 , I think this is ready for a second review.
Who should review it ?

I force the user to define name and mode.
The use of one amplitude value or array for A or B is checked at the element setup.
The MaxOrder option has been removed as it seemed to be overwritten in most cases by the element setup.
The seed option has been removed. See this discussion https://github.com/atcollab/at/discussions/879.
The modes SIN and ARBITRARY now consider the particle time delay. Additionally a Sinlimit option has been added to create a half-sin when setting it to zero.
Mode WHITENOISE, ignores the particle time delay. MEAN and STD have been added as options with default values 0 and 1.
Ramping has not been modified.
In mode ARBTRARY an additional time delay has been added as an option, default value zero. Also, the array of turn by turn values was extended to the 4th Taylor expansion wrt $\tau$ (the time delay, or the sixth coordinate divided by the speed of light).
Periodic is now only used in ARBITRARY mode.

I have done test creating the elements in matlab, saving and running in python. Also, creating the elements in python, saving and runing in matlab. And also, creating and running the elements on the same environment they are created. They all work, but as the number of options is large because the pass method combines three elements in one, I have checked only kicks and skew quadrupole components with varying modes.

swhite2401:
I can review it but I will be away for 2 weeks starting tomorrow.

Discussing with colleague here at ESRF we thought it would be nice that this element takes into account the time delay in multi-bunch mode (for instance to simulate an injection kicker pulse), I think this is possible with the present implementation since the multi-bunch beam is unfolded by default (see tracking/track.py) but it would be nice to confirm it.

I provide a review next year... unless @lfarv would like to take a look earlier?

oscarxblanco:
I'm not sure how multi bunch simulations are done, but, I have heard something similar from Gabriele at ALBA. So, I'll have a look.

I am also leaving today, and I'll be back the 7th of Jan/2025.

Happy hollydays.
o

swhite2401:
I can send you an example, happy holidays!

pyat/at/lattice/variable_elements.py
```diff
@@ -1,141 +1,280 @@
...
+        if mode == ACMode.ARBITRARY:
+            kwargs = self._set_arb(a_b, **kwargs)
+        return kwargs
+
+    def _set_white_noise(self, a_b: str, **kwargs: dict[str, any]) -> dict[str, any]:
+        if "Mean" + a_b not in kwargs:
+            kwargs["Mean" + a_b] = kwargs.get("Mean" + a_b, 0)
+        if "Std" not in kwargs:
+            kwargs["Std" + a_b] = kwargs.get("Std" + a_b, 1)
+        return kwargs
```
lfarv, line 242: I think that `Mean` and `Std` do not bring anything : the static part "Amplitude*Mean" can be provided by a standard `ThinMultipole` element, and for the dynamic part, `Amplitude` and `Std` are redundant. I would simplify both python and C b removing these attributes: `Amplitude` is sufficient.

pyat/at/lattice/variable_elements.py
```diff
@@ -1,141 +1,280 @@
...
+            if ramps is not None:
+                kwargs["Ramps"] = ramps
+        super().__init__(family_name, **kwargs)
+
+    def _check_amplitudes(self, **kwargs: dict[str, any]) -> dict[str, any]:
+        amp_aux = {"A": None, "B": None}
+        all_amplitudes_are_none = True
+        for key in amp_aux:
+            if "Amplitude" + key in kwargs and "amplitude" + key in kwargs:
+                raise AtError("Duplicated amplitude " + key + "parameters.")
```
lfarv, line 204: Why are you looking for "amplitudeA/B" if the attribute name is AmplitudeA/B" ?

pyat/at/lattice/variable_elements.py
```diff
@@ -1,141 +1,280 @@
...
-            self.Seed = kwargs.pop('Seed', datetime.now().timestamp())
-        self.PolynomA = numpy.zeros(self.MaxOrder+1)
-        self.PolynomB = numpy.zeros(self.MaxOrder+1)
-        ramps = kwargs.pop('Ramps', None)
-        if ramps is not None:
-            assert len(ramps)==4, \
-                'Ramps has to be a vector with 4 elements'
-            self.Ramps = ramps
-        super(VariableMultipole, self).__init__(family_name, **kwargs)
+        kwargs["Mode"] = kwargs.get("Mode", mode)
```
lfarv, line 180: better written as:
```
kwargs.setdefault("Mode", mode)
```

pyat/at/lattice/variable_elements.py
```diff
@@ -1,141 +1,280 @@
...
+            for key in amp_aux:
+                kwargs["Polynom" + key] = kwargs.get(
+                    "Polynom" + key, np.zeros(maxorder + 1)
+                )
+            ramps = self._check_ramp(**kwargs)
+            if ramps is not None:
+                kwargs["Ramps"] = ramps
+        super().__init__(family_name, **kwargs)
+
+    def _check_amplitudes(self, **kwargs: dict[str, any]) -> dict[str, any]:
```
lfarv, line 199: All the following methods can be defined as static methods, or nested in __init__

pyat/at/lattice/variable_elements.py
```diff
@@ -1,141 +1,280 @@
...
+        if "Std" not in kwargs:
+            kwargs["Std" + a_b] = kwargs.get("Std" + a_b, 1)
+        return kwargs
+
+    def _set_sine(self, a_b: str, **kwargs: dict[str, any]) -> dict[str, any]:
+        frequency = kwargs.get("Frequency" + a_b, None)
+        if frequency is None:
+            raise AtError("Please provide a value for Frequency" + a_b)
+        kwargs["Phase" + a_b] = kwargs.get("Phase" + a_b, 0)
+        kwargs["Sinlimit" + a_b] = kwargs.get("Sinlimit" + a_b, -1)
```
lfarv, line 249: Same comment as above:
```
kwargs.setdefault("Phase", 0.0)
kwargs.setdefault("Sinlimit", -1)
```

pyat/at/lattice/variable_elements.py
```diff
@@ -1,141 +1,280 @@
...
+                raise AtError("Duplicated amplitude " + key + "parameters.")
+            lower_case_kwargs = {k.lower(): v for k, v in kwargs.items()}
+            amp_aux[key] = lower_case_kwargs.pop("amplitude" + key.lower(), None)
+            if amp_aux[key] is not None:
+                all_amplitudes_are_none = False
+        if all_amplitudes_are_none:
+            raise AtError("Please provide at least one amplitude for A or B")
+        return amp_aux
+
+    def _set_amplitude(self, amplitude: float or _array or None) -> _array:
```
lfarv, line 213: _array is a function:
```python
def _array(value, shape=(-1,), dtype=numpy.float64): ...
```
You cannot use it as a type. Here, the type could be `float | Sequence[float]`

pyat/at/lattice/variable_elements.py
```diff
@@ -1,141 +1,280 @@
...
+        if np.isscalar(amplitude):
+            amplitude = [amplitude]
+        return np.asarray(amplitude)
```
lfarv, line 216: Anyway the whole function is useless: when setting the `Amplitude` attribute, the conversion function (`_array`) will convert the value into a Fortran-aligned array of floats with shape (-1,) which is exactly what you are doing here.

lfarv:
@oscarxblanco : the d[ocumentation of the VariableMultipole](https://at.readthedocs.io/en/fix_bug_pyat_variablemultipole/p/api/at.lattice.variable_elements.html#at.lattice.variable_elements.VariableMultipole) class is a bit strange. Sphinx is tricky with indents and formatting…

pyat/at/lattice/variable_elements.py
oscarxblanco: `Mean` and `Std` have been removed.

pyat/at/lattice/variable_elements.py
oscarxblanco: OK. I put them inside __init__

pyat/at/lattice/variable_elements.py
oscarxblanco: OK

pyat/at/lattice/variable_elements.py
oscarxblanco: The function has been removed

pyat/at/lattice/variable_elements.py
oscarxblanco: OK, removed.

oscarxblanco:
> @oscarxblanco : the d[ocumentation of the VariableMultipole](https://at.readthedocs.io/en/fix_bug_pyat_variablemultipole/p/api/at.lattice.variable_elements.html#at.lattice.variable_elements.VariableMultipole) class is a bit strange. Sphinx is tricky with indents and formatting…

I haven't checked this before locally. I now compile with the 'doc' option and it looks ok in my local copy. It should have improved.

swhite2401:
Hi @oscarxblanco , I see many commits, are you ready for another review or do you plan to make further modifications?

oscarxblanco:
Hi !
there are two pending things.
- The custom discrete function was modified to include time derivatives, and it is now being tested here at ALBA but I still have no feedback. It might take a few days to see if there is something to change. For example, I wrote the derivatives wrt $\tau$, not $c\tau$ (the sixth coordinate). It might be unusual for AT users.
- You mentioned to incorporate the effect of delay in multi-bunch mode. I modify the code to take into account the particle delay in SINE and ARBITRARY mode, and to ignore the particle delay in WHITENOISE mode. I don't  know how the multi-bunch simulations are done, so, I couldn't check it.

Excluding that I think the rest could be reviewed.

swhite2401:
Good! I can test the multi-bunch

pyat/at/lattice/variable_elements.py
```diff
@@ -1,141 +1,259 @@
...
+            FuncBderiv2(list): tbt list for PolynomB 2st derivative wrt tau.
+                Default: zeros array of the custom function length.
+            FuncAderiv3(list): tbt list for PolynomA 3st derivative wrt tau.
+                Default: zeros array of the custom function length.
+            FuncBderiv3(list): tbt list for PolynomB 3st derivative wrt tau.
+                Default: zeros array of the custom function length.
+            FuncAderiv4(list): tbt list for PolynomA 4st derivative wrt tau.
+                Default: zeros array of the custom function length.
+            FuncBderiv4(list): tbt list for PolynomB 4st derivative wrt tau.
+                Default: zeros array of the custom function length.
```
lfarv, line 151: The `list` type hint looks too restrictive: for me, it should rather be `Sequence[float]`,
with
```python
from collections.abc import Sequence
```

oscarxblanco:
Dear @swhite2401 ,
There's been two remarks, thus far, from people who is using this element :
- They prefer `Periodic=False` as default behavior of the custom function. It is considered to be less dangerous  because the user would assume that only the defined turns are applied, otherwise the user could set it to `True`.
- They consider dangerous leaving the PolynomA or PolynomB at the state given by the calculation on the last turn. This could lead to confusion because the ring before and after tracking are different, which is atypical for AT. They prefer the variable element to be automatically reset after tracking. Doing a reset would be OK for the reusage of the same ring without any other action, but, it means we remove any trace of what is actually applied to the beam.

Do you have any preference/opinion on these two remarks ?

swhite2401:
> Dear @swhite2401 and @lfarv , There's been two remarks, thus far, from people who is using this element :
> 
>     * They prefer `Periodic=False` as default behavior of the custom function. It is considered to be less dangerous  because the user would assume that only the defined turns are applied, otherwise the user could set it to `True`.
> 
>     * They consider dangerous leaving the PolynomA or PolynomB at the state given by the calculation on the last turn. This could lead to confusion because the ring before and after tracking are different, which is atypical for AT. They prefer the variable element to be automatically reset after tracking. Doing a reset would be OK for the reusage of the same ring without any other action, but, it means we remove any trace of what is actually applied to the beam.
> 
> 
> Do you have any preference/opinion on these two remarks ?

First point is fine for me. The other one is a bit more tricky, because you want to be able to access what is applied to the beam. However, I understand the complaints, we could instead think of adding a history buffer attribute that is filled during the tracking. This already exists for other elements like the beam monitor or beam loading for example, what do you think?
I you decide to go for this option you have to be very careful: the buffer needs to be properly initialized depending on the number of turns otherwise you will get segmentation faults

oscarxblanco:
Periodic is now set to False by default.
I will have a look on the buffer construction.

oscarxblanco:
Dear @swhite2401 and @lfarv ,
there a third comment from users:
- Could it be a thick element instead of a thin element ? Some pulsed element have a non-negligible length.

I guess the element could be changed, but, should it keep the same name ?
Is there any reason why it was not implemented as a thick element ?
In case this is done, should we keep both ? (thin and thick) variable elements.

lfarv:
@oscarxblanco and @swhite2401

> They prefer the variable element to be automatically reset after tracking.

I am clearly in favour of resetting the PolynomA/B to the initial value. Anything else is confusing, keeping only the last value does not help in checking the turn-by-turn evolution.

Now for a way to check what is applied to the beam, I would prefer a specific method of the VariableMultipole element which would visualise the turn-by-turn evolution of PolynomB/A, independently of any tracking: either by outputting a table of values or with a plot. This avoids all the memory management problems of handling a buffer, keeps the C code simpler and cleaner, and is easier to use.

lfarv:
> Could it be a thick element instead of a thin element ?

We already have a `ThinMultipole` and a `Multipole` (thick). It could be the same for variable elements, right? The big difference is that the PolynomA/B describe the **integrated** field expansion in the 1st case and the local field expansion in the 2nd case.

There are at the moment discussions in the "standard lattice description" working group about a convention to use the integrated field expansion in both cases, which could avoid confusion.

However, AT being as it is, I would rather make the variable elements following exactly the static ones,

lfarv:
In any case, I think that implementing a thick element deserves a dedicated PR, otherwise this one may delayed again for long…

swhite2401:
@oscarxblanco , finally I found some time to test this. All seems to be working fine, here a test script and output.

`import at
import numpy as np

filename = '../lattice/hmba.mat'
ring = at.load_lattice(filename, periodicity=1)
ring = ring.repeat(32)
ring.radiation_on()
rfreq = ring.revolution_frequency

variable_mpole = at.VariableThinMultipole('VMULT', at.ACMode.SINE,
                                          AmplitudeB=[1.0e-5, 0.0],
                                          FrequencyB=0.3*rfreq)
ring.append(variable_mpole)

p0 = np.zeros(6)
p1, *_ = ring.track(p0, nturns=128)
p1 = np.squeeze(p1)
fft_p1 = np.fft.fft(p1[0])

plt.subplot(211)
plt.plot(p1[0])
plt.xlabel('Turn #')
plt.ylabel('x [m]')
plt.subplot(212)
plt.plot(np.linspace(0, 1, len(fft_p1)), abs(fft_p1))
plt.xlabel('Ampl.')
plt.ylabel('Tune')
plt.tight_layout()
plt.show()

nbunch = 992
ring.set_fillpattern(nbunch)
beam = np.zeros((6, nbunch))
beam[5] = ring.bunch_spos
variable_mpole.FrequencyB=3*rfreq
ring.track(beam, nturns=1, in_place=True, unfold_beam=False)
plt.plot(beam[1])

plt.xlabel("x'")
plt.ylabel('Bunch #')
plt.tight_layout()
plt.show()
`
sine_multi_bunch

sine_single_bunch

swhite2401:
I have a few comments on the implementation:
-Would it make sense to put `FuncDeriv` in an arbitrary length array to provide more flexibility and more compact input?
-We need to have access to the polynom history for whitenoise as well, this is useful for instance to compute the beam transfer function-> in fact here you have to re-write in python what is done in C, something that was supposed to make things simpler is in fact complicating it. A history buffer is not that complicated to implement (we have it in other elements) and would cover all cases

swhite2401:
> Do you mean to any order ? Sorry, I don't understand. The length of every array has been fixed to match the length of the function. Maybe you refer to having arrays of different length ?

I mean the number of coefficient, presently you stop at 4, but it could go to `n` in principle using the same expansion just looping through an array

swhite2401:
> I also ignore translations (T1,T2) and rotations(R1,R2) are ignored. They might be implemented in a future PR if they are necessary.

I agree these are not need for the moment

swhite2401:
> I did not try to implement the buffer as @lfarv considered that this will slow-down the pass method. Should I revert this and do the buffer ?

I don't think it slows down the passmethod, it is just a bit more complicated as you have to handle memory allocation in C, however for me this is a non-issue since it is already implemented elsewhere

oscarxblanco:
Dear all,
this is PR is ready for review. I will summarize the modifications.

Since the introduction of the VariableMultipole pass method implemented in C, there were matlab AT and pyat elements that were set with incompatible parameter names and incomplete structures to be saved and exchangeable between matlab and python. This PR was opened to solve it, and on the way there were additional requirements:
- Change the name to VariableThinMultipole
- Add a buffer to store the random numbers used during tracking of a variable thin multipole WHITENOISE mode.
- Add the possibility to generate a half-sine function.
- Include custom functions to be represented by an arbitrary taylor expansion to k-th order.
- Add a function to inspect the variable multipole before using it in tracking.

All of them are implemented here.

There are two questions that are open because of this work.
- How variable elements will affect lattice parameters ? For the moment I have included in the documentation to set the pass method to IdentityPass when necessary, but, there might be a better solution.
- The implementation of the WHITENOISE mode buffer revealed a weird behaviour when tracking a few turns with a given seed. I expected the seed to fix the sequence and always obtain the same result, but, instead I get a cycle that is repeated every 2 or 3 tracks. I will open a separate issue on this, I might be missing something on the seed implementation.

PR #842, opened 2024-10-15
Fix LatticeObservable

lfarv:
A bug in `LatticeObservable` caused a crash for scalar `Element` parameters (like `Length` to `K`). This is fixed here.


swhite2401 (approved):
Ok for me

PR #843, opened 2024-10-15
Fix the "repr" representation of Elements (after numpy2.0)

lfarv:
Since the release of numpy 2.0, the"repr" representation of `Element` objects is polluted by "np.float64" strings. Example:
```
Quadrupole('QF1A', 0.311896, np.float64(2.5394599781303304))
```
Unlike the .m file output already fixed, this has no apparent consequence, the .repr format still works, so the fix is purely cosmetic. The new output is:
```
Quadrupole('QF1A', 0.311896, 2.5394599781303304)
```

Additionally, `elements.py` was reformatted…


lfarv:
Note: the tests require #842 to be applied !

swhite2401 (approved):
With all the formatting changes I can't really identify where the changes took place, but I trust you please go ahead with the merge

lfarv:
@swhite2401 : True, the formatting makes reviewing difficult. Hopefully, it's done once for ever. Here the modifications are ver simple:
```
    def K(self) -> float:
        """Focusing strength [mˆ-2]"""
        arr = self.PolynomB
        return 0.0 if len(arr) < 2 else arr[1]
```
is replaced be:
```
    def K(self) -> float:
        """Focusing strength [mˆ-2]"""
        arr = self.PolynomB
        return 0.0 if len(arr) < 2 else float(arr[1])
```
Just a conversion to a python float makes `repr` cleaner.

lfarv:
Tests are now OK!

PR #844, opened 2024-10-17
Loading of MAD and Elegant files

lfarv:
The processing of MAD files is improved:
#### 2. handling of deferred assignments:
deferred assignments are correctly handled when MAD parameters are interactively modified. Example:
assuming an input files containing:
```
QF:  QUADRUPOLE,  L=0.5,  K1:=KQF
KQF = 0.75
```
And the following PyAT commands:
```python
>>> parser = MadxParser()
>>> parser.parse_files(input_file)      # Read the input file and fill the database
>>> parser["KQF"]                       # Look at the initial value of KQF
0.75
>>> ring1 = parser.lattice(use="RING")  # Generate a 1st lattice with KQF == 0.75
>>> parser["KQF"] = 0.8                 # Modify the value of KQF
>>> parser["KQF"]                       # check the modified value of KQF
0.8
>>> ring2 = parser.lattice(use="RING")  # Generate another lattice with KQF == 0.8
```
One can now interactively modify any parameter, and the updated values will be taken into account when generating lattices.
With this, one can automatise the generation of AT lattices with variable conditions. Note however that this is much slower than using AT `Variable` objects when it is possible.

In the present state, MAD format may be used as a native input format for PyAT.


lfarv:
Parsing of Elegant files is also rebuilt and should now be fully compliant with the Elegant reference.

swhite2401:
@lfarv, what you propose is to have an object `parser` that contains all the lattice information including the so called deferred variables. 
However everytime you change a parameter in this parser object you need to instantiate a new ring so in effect it is the same as changing the MADX input file, is that correct?

lfarv:
@swhite2401 : you are right, the parser is just an image in memory of the MAD-X file, as it is already the case in the present version. So after modifying a variable, you need indeed to instantiate a new ring (like you need to call the "USE" command in MAD) to take new values into account. But:
1. This is much faster than parsing again a new MAD file: the interpretation of the MAD syntax is complicated and is done once for ever,
2. It's much easier to change programmatically a variable in the parser than to modify a text file.

The new point in this version is that the expansion of deferred variables was previously done while parsing the file, while it is now done when the variable is used (as in MAD). So in the previous version, modifying parameters had no interest.

Note that in simple cases, the `load_madx` function does everything in one command (parsing and instantiating the lattice). Accessing the parser gives more flexibility, like generating several lattices from different SEQUENCES (already possible in the present version), or modifying variables (now possible with this version), without reading the file again.

simoneliuzzo:
I may test the mad files, but I have no elegant files. May @carmignani has an elegant file we could use as test?

simoneliuzzo:
Dear @lfarv,

in this branch it is now possible to save mad and elegant files? This would bring a full omni-directional converter feature to pyAT!

I will try the features added in this branch.

best regards
Simone

lfarv:
@simoneliuzzo: You noticed the new feature even before I mentioned it ! Yes, you are right, both loading and saving are now available. I still need to document the limitations on saving (limited to elements available on both sides, tilt not yet taken into account…), but I tested on a few files that when loading a MAD (or elegant) file, and then saving it back to MAD, the resulting file is of course completely different from the initial one, but MAD gives identical results.

However, this needs to be tried on many files. I have about 10 MAD test files and ~5  elegant files. Every test you can perform is welcome, and is likely to show some problems, which will have to be fixed step by step. The code is simple and flexible enough so that should be easy.

simoneliuzzo:
Dear @lfarv,

I tested `load_madx`, `load_mad8`, `save_elegant` and `load_elegant` and they work as expected, producing identical phase space pictures. Assuming it is a complicated case, I used the FCC-ee lattice for test. 

I tested also on EBS lattices, that I thought simpler. In fact the lattice I chose has bpms names "BPM_C03-08" and I think those are source of an issue for `load_elegant` and generate the error:
"SyntaxError: leading zeros in decimal integer literals are not permitted; use an 0o prefix for octal integers"

load_madx and load_mad8 work as expected.

I understood from the output that the `save_madx`, `_mad8` and `_elegant` functions use of the `RingParam` element name for the lattice variable. This may be added in the documentation, or in a printed message. 

Some additional help in the load / save methods could be nice.

I am unable to test if the elegant files produced by AT are usable in elegant

lfarv:
@simoneliuzzo : Excellent, thanks a lot ! Can you post the lattice file with problems ?

> I understood from the output that the save_madx, _mad8 and _elegant functions use of the RingParam element name for the lattice variable

`RingParam` in python ? We use the `name` attribute of the `Lattice` object, which you can modify at any time. Does'n it make sense ? Anyway, I have to write the documentation…

simoneliuzzo:
Dear @lfarv,

if you have access to ESRF's machfs, those are the lines I am running: 
 
r_at = at.load_lattice('/machfs/liuzzo/EBS/S28F/LATTICE/AT/S28F.mat', mat_key='LOW_EMIT_RING_INJ')
    at.save_madx(r_at, filename='/machfs/liuzzo/accelerator_toolbox_tests/python/load_save_mad_elegant/S28F_save_madx.mdx')
    r_madx = at.load_madx('/machfs/liuzzo/accelerator_toolbox_tests/python/load_save_mad_elegant/S28F_save_madx.mdx', use='EBS', strict=False)
    at.save_mad8(r_at, filename='/machfs/liuzzo/accelerator_toolbox_tests/python/load_save_mad_elegant/S28F_save_madx.md8')
    r_mad8 = at.load_mad8('/machfs/liuzzo/accelerator_toolbox_tests/python/load_save_mad_elegant/S28F_save_madx.md8', use='EBS', strict=False)
    at.save_elegant(r_at, filename='/machfs/liuzzo/accelerator_toolbox_tests/python/load_save_mad_elegant/S28F_save_ele.lte')
    r_elegant = at.load_elegant('/machfs/liuzzo/accelerator_toolbox_tests/python/load_save_mad_elegant/S28F_save_ele.lte', use='EBS', strict=False)

simoneliuzzo:
Dear @lfarv,

I have repeated the tests for EBS. now the elegant converted file works.

The NumIntSteps information is apparently lost in the translation, I think this is due to elegant and mad not storing this information at element level. I fixed the values to 20 for all lattices.

All the translated lattices reloaded in AT show a common slight deviation from the original file:

Screenshot 2024-11-14 at 12 31 43

Screenshot 2024-11-14 at 12 31 37

I tried comparing the json files, but I could not spot the difference.

lfarv:
@swhite2401, @simoneliuzzo : The documentation is now improved: can you look [here](https://at.readthedocs.io/en/load-madx/p/api/at.load.mad8.html#module-at.load.mad8) and give your comments?
I think everything is now ready!

simoneliuzzo:
Dear @lfarv,

documentation looks very nice! Thank you!

the mad8 documentation refers to the MADX. This is ok for me.
For elegant and tracy there is not much detail instead.

best regards
simone

simoneliuzzo (approved):

lfarv:
@simoneliuzzo:
I corrected the documentation for Elegant, masked some private methods and classes, and added a link in the "How to" menu (in the left column).
But I have to ask you to approve again… Sorry.

Concerning tracy, I have not touched the existing code at all and I do not know if it's working, but to work on it, I need some tracy documentation and examples, which I could not find. If someone has a link, tell ne

simoneliuzzo (approved):

Issue #841, opened 2024-10-08
Possible Bug in random generators


swhite2401:
In #840 a bug with random generators was reported and fixed at the python level. 
The issue reported was that `ohmi_envelope()` would exist with a segmentation fault when an element with random number is present in the lattice, this issue appeared in #737.

This could be a sign of a profound bug in the random generators, could affect more function than just `ohmi_envelope()` and would need to be fixed at the C level.

lfarv:
Corrected in #845

PR #845, opened 2024-10-18
fix uninitialised random generators in elempass

lfarv:
A bug reported in #841 causes a segmentation fault in `ohmi_envelope` when called with  an element using random generators in the lattice. This was due to a non-initialised variable in the C function `elempass`.

This is corrected here.

There are very few functions calling `elempass` (`ohmi_envelope` is maybe the only one, and it's not supposed to be called with random elements), so this was not noticed until recently.


swhite2401 (approved):
Great!

PR #849, opened 2024-10-27
Fix "fastring"

lfarv:
This fixes a regression bug that appeared in #837, causing the chromaticies and detuning of fast rings to be wrong.

For reviewers, the only significant modification is `fasting.py` line 45:
```python
return all_rings, Lattice(ringm, energy=ring.energy, periodicity=1)
```
which forces the analysis to be done on a single cell. Everything else is formatting.

This bug was identified in the `matlab_test` series of python tests which cannot be run on GitHub because of a problem of Matlab license. So now, a new test is added in `pytests.m`, executed on Matlab on each commit.

### DIFF
--- a/atmat/attests/pytests.m
+++ b/atmat/attests/pytests.m
@@ -1,4 +1,6 @@
 classdef pytests < matlab.unittest.TestCase
+    % ">> run(pytests)" to run all tests
+    % ">> run(pytests, 'orbit4')" to run a the single orbit4 method
 
     properties(Constant)
         mlist=[...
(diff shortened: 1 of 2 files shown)


swhite2401:
@lfarv, modifications are ok for me, but could you update the doctring before merging?

lfarv:
@swhite2401 : nothing changes in the behaviour of the function. But here is the modified docstring:

```python
"""
    A fast ring consisting in:

    * a RF cavity per distinct frequency,
    * a 6x6 linear transfer map,
    * a detuning and chromaticity element,
    * a quantum diffusion element (for radiation ring).


    2 new lattices are returned, one with radiation and one without. These lattices
    keep the same attributes (energy, particle, circumference, periodicity,…)
    as the initial one.

    It is possible to split the original ring in multiple "fastrings"
    using the ``split_inds`` argument.
"""
```
Do you see more to be mentioned?

PR #851, opened 2024-11-03
Fix  a bug in load_madx

lfarv:
This fixes a problem reported by @elafmusa in #850.

The fixes concern the length of RBEND magnets and the determination of the harmonic number.

### DIFF
--- a/pyat/at/load/madx.py
+++ b/pyat/at/load/madx.py
@@ -216,7 +216,7 @@ class rbend(_MadElement):
     @staticmethod
     @set_tilt
     def convert(name, l, angle, e1=0.0, e2=0.0, **params):  # noqa: E741
-        hangle = 0.5 * angle
+        hangle = abs(0.5 * angle)
         arclength = l / sinc(hangle)
         return sbend.convert(
             name, arclength, angle, e1=hangle + e1, e2=hangle + e2, **params
--- a/pyat/at/load/madx.py
+++ b/pyat/at/load/madx.py
@@ -224,6 +224,7 @@ class rbend(_MadElement):
 
     @property
     def length(self):
+        """Element length"""
         hangle = 0.5 * self["angle"]
         return self["l"] / sinc(hangle)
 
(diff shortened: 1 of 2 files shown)


lfarv:
I found another problem with entrance/exit angles of bending magnets with negative bending angle… Working on it.

lfarv:
Problem fixed, ready for review.

pyat/at/load/madx.py
```diff
@@ -475,9 +477,9 @@ def flatten(self, parser, offset: float = 0.0, refer: float = 1.0, refpos=None):
 
     def expand(self, parser: MadxParser) -> Generator[elt.Element, None, None]:
         def insert_drift(dl, el):
-            if dl > 1.0e-10:
+            if dl > 1.0e-5:
```
swhite2401, line 480: Is there a reason to change this tolerance?

pyat/at/load/madx.py
```diff
@@ -707,13 +709,18 @@ def beta() -> float:
 
             params["_length"] = cell_length
             rev = beta() * clight / cell_length
+
+            # Set the frequency of cavities in which it is not specified
+            hn = 2147483647
```
swhite2401, line 714: Where does number come from?

swhite2401 (commented):
@lfarv, just a couple comments or rather questions in fact.

pyat/at/load/madx.py
lfarv: From the MADX manual:

_When the sequence is expanded in a [USE](https://mad.web.cern.ch/mad/webguide/manual.html#Ch4.S10) command, MAD-X generates the drift spaces between elements according to the following rules:_

_1. When the distance between the exit of the previous element and the entrance of the next element is positive and greater than a threshold of **1μm** an explicit drit is generated with its own name, unless an already existing drift space with the same length (within 10−12m tolerance) can be re-used.
2. When the absolute value of the distance between the exit of the previous element and the entrance of the next element is less than a given tolerance of **1μm**, no drift space is created and the elements are considered as contiguous.Note that in very specific cases this can cause very small errors to accumulate and the actual length of the sequence can vary slightly from the declared length. (see second example below)
3. When the distance between the exit of the previous element and the entrance of the next element is negative and less than **−1μm**, the elements are considered to be overlapping and MAD-X terminates with a “negative drift length” fatal error._

MADX sets the threshold at 1.E-6. The previous value causes the generation of many extremely small drifts. 1.E-6 still caused a few unwanted drifts absent in MADX, so I increased to 1.E-5. I can put back 1.E-6 if you prefer, a few μm long drifts are harmless.

pyat/at/load/madx.py
lfarv: 2^31 - 1
In python, here is no "maxint", the `int` type has no upper bound. I took the largest positive integer that can be coded in a 32bit integer… Any large value fits…

swhite2401 (approved):
Ok thanks for the precisions, all good for me

PR #853, opened 2024-11-07
Suppress a wrong warning when using linopt in beamline mode

lfarv:
When using `linopt*`/`get_optics` in beam line mode by providing the `twiss_in` entry, a warning _"Unstable beam"_ sometime appeared. This is fixed here.

In addition, the execution time of linopt is reduced by 16% !


lfarv:
In addition, warnings and errors are secured. Real warnings will always appear when relevant, instead of only on the first occurence.

swhite2401 (approved):
Thanks!

PR #854, opened 2024-11-07
Fix atplot for single-element lattices

lfarv:
This fixes a crash when using `atplot` on single-element lattices, as reported by @simoneliuzzo in #848

### DIFF
--- a/atmat/atplot/atplotsyn.m
+++ b/atmat/atplot/atplotsyn.m
@@ -36,60 +36,54 @@ labs=labs(sok);
 indx=indx(sok);
 [dipoles,qpoles,spoles,mpoles,bpms]=cellfun(@eltype,rok);
 
-ampl=0.05*tlim;     % Dipoles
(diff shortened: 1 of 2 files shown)


simoneliuzzo:
Dear @lfarv,

I have tested this branch but it does not seem to solve the issue. 

Screenshot 2024-11-12 at 13 02 18

and here a test with input twiss

Screenshot 2024-11-12 at 13 05 14

thank you for looking into this and my apologies for taking so long time to test.

best regards
Simone

lfarv:
@simoneliuzzo: sorry, I missed another problem.
You can try with this new version. Don't worry for delays, this branch is not blocking anything!

simoneliuzzo:
Dear @lfarv,

now it seems ok!

It still gives error for ringparam and marker elements. But I think this is not an issue.

thank you
best regards
Simone

simoneliuzzo (approved):
tested, ok

lfarv:
> It still gives error for ringparam and marker elements. But I think this is not an issue.

Yes, and it still gives error for empty lattices. But these cases have no real interest…

Issue #852, opened 2024-11-05
pyat acceptance doc string mistake


eliaserracarbonell:
Hi, my name is Èlia. I've been working with AT and I've seen that the function `at.get_momentum_acceptance` returns the survived and tracked particles in a different order than stated in the doc string:
https://github.com/atcollab/at/blob/09ec7bc9bfa3f700ee2d0011cef6144be59e11a9/pyat/at/acceptance/acceptance.py#L445-L447

https://github.com/atcollab/at/blob/09ec7bc9bfa3f700ee2d0011cef6144be59e11a9/pyat/at/acceptance/boundary.py#L372-L375

PR #855, opened 2024-11-07
Fix docstrings in the acceptance.py module

lfarv:
Fixes a documentation error reported in #852: the order of outputs is swapped


swhite2401 (approved):

swhite2401:
Same comment here could you black and use numpy as np?

swhite2401 (approved):

PR #859, opened 2024-11-26
Set "periodicity" to 1 by default

lfarv:
Following the discussion in https://github.com/atcollab/at/pull/837#issuecomment-2411445444, this PR modifies the behaviour of the Lattice creation:

- for a lattice initialised from a sequence of elements, the periodicity will be set to 1 by default. Up to now it was deduced from the total bending angle. This is also the case when creating lattices form "foreign" files (MAD and Elegant) which have no notion of periodicity,
- however, the periodicity is still saved and retrieved when saving/loading `.m`, `.mat` and  `.json` files. So nothing changes there,
- the previous behaviour (automatic determination) can now be obtained by specifying `periodicity=0`.

The idea is that the periodicity will be set to 1 unless it is **explicitly set** to something else. This should avoid part of the unexpected behaviour linked to `periodicity`, at the cost of a slight behaviour change/

### DIFF
--- a/pyat/at/load/matfile.py
+++ b/pyat/at/load/matfile.py
@@ -157,6 +157,7 @@ def ringparam_filter(
         else:
             yield elem
     params["_radiation"] = radiate
+    params.setdefault("periodicity", 0)
 
     if len(ringparams) > 1:
         warn(
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -98,12 +98,13 @@ class Lattice(list):
 
     # noinspection PyUnusedLocal
     def __init__(self, *args, iterator: Filter = None, scan: bool = False, **kwargs):
+        # noinspection PyUnresolvedReferences
         """
-        Lattice(elements, **params)
-        Lattice(filter, [filter, ...,] iterator=iter,**params)
+        Lattice(elements: Iterable[Element], **params)
+        Lattice(filter, [filter, ...,] iterator=iter, **params)
 
         Parameters:
-            elements: iterable of Element objects
+            elements: iterable of Element objects.
             iter: function called as :pycode:`iter(params, *args)`. It must
               return an iterable of :py:class:`.Element` objects for building
               the lattice. It must also fill the *params* dictionary providing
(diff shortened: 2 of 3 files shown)


lfarv:
In order to ease the reviewing, I postponed the "blackening" of `lattice_object.py` which will add many changes. When agreed, I will do it in a next step.

lfarv:
"black" and numpy -> np modified in` lattice_object.py`

swhite2401 (approved):

swhite2401:
Hello @lfarv , I missed it but periodicity is set to 0 in this PR not one, so in effect there no change at all....
If we set the default to 1, tests also have to adapted.
I restored the branch so we can fix it

lfarv:
Hello !

> periodicity is set to 0 in this PR not one

I don’t think so, because I had to modify test_lattice_object and set explicitly periodicity=0, otherwise the assessment on periodicity failed as it was now 1. 

Keep in mind that when loading à file, the file value is restored, it’s not computed again. It has to be so, because the file is expected to restore the lattice exactly as it was when it was saved, including periodicity.

Remains that as you noticed, setting periodicity=0 does not trigger the computation, it could be done.

I’m away, I’ll look at that next week !

pyat/at/load/matfile.py
```diff
@@ -157,6 +157,7 @@ def ringparam_filter(
         else:
             yield elem
     params["_radiation"] = radiate
+    params.setdefault("periodicity", 0)
```
swhite2401, line 160: The default is 0 not 1...

swhite2401: Or am I missing something...

swhite2401:
We may also load hmba.mat, and dba.mat with periodicity1 to avoid problems in the future in case we change files

lfarv:
Hello @swhite2401.

I still do not see the problem: the code you are showing refers to loading a `.mat` file, not creating a new lattice. As I said above, nothing changes when loading files: the periodicity is written in the file when saving and restored when loading. That's what a file is expected to do and that's similar for `.m`, `.repr `and `.json` files.

The line you show acts when reading a file created with an old Matlab version, without a RingParam element. So no lattice attribute (energy, periodicity, particle etc.) was saved in the file. In this particular case, we could set periodicity to 1. But I think that changing the behaviour of file reading could hit a lot of users. 

The change here deals with **creating** new lattices from a list of elements.
```python
d = at.Dipole(
    "d1",
    1,
    BendingAngle=np.pi / 16,
    Energy=5.0e6,
    PassMethod="BndMPoleSymplectic4RadPass",
)
lat = at.Lattice([d], name="lattice", energy=3.0e6)
lat.periodicity
```
Now you get
```
1
```
while before you got
```
32
```

swhite2401:
> Hello @swhite2401.
> 
> I still do not see the problem: the code you are showing refers to loading a `.mat` file, not creating a new lattice. As I said above, nothing changes when loading files: the periodicity is written in the file when saving and restored when loading. That's what a file is expected to do and that's similar for `.m`, `.repr `and `.json` files.
> 
> The line you show acts when reading a file created with an old Matlab version, without a RingParam element. So no lattice attribute (energy, periodicity, particle etc.) was saved in the file. In this particular case, we could set periodicity to 1. But I think that changing the behaviour of file reading could hit a lot of users.
> 
> The change here deals with **creating** new lattices from a list of elements.
> 
> ```python
> d = at.Dipole(
>     "d1",
>     1,
>     BendingAngle=np.pi / 16,
>     Energy=5.0e6,
>     PassMethod="BndMPoleSymplectic4RadPass",
> )
> lat = at.Lattice([d], name="lattice", energy=3.0e6)
> lat.periodicity
> ```
> 
> Now you get
> 
> ```
> 1
> ```
> 
> while before you got
> 
> ```
> 32
> ```

Ok, I think I read though a bit too fast, somehow I thought that files would be loaded with periodicity of 1.... but this is fine for me.

However, we need to handle the case where users set periodicity=0 for existing lattice, for the moment you end up with ZeroDivisionError

lfarv:
> we need to handle the case where users set periodicity=0 for existing lattice

Agreed. This implies changing the `periodicity` attribute to a property, which means significant modifications. I'll start from scratch for that.

Issue #862, opened 2024-12-02
difusion_matrix.c does not compile in Linux


ZeusMarti:
Hi all,
I'm having problems compiling the new release under Linux (Debian 11). Specifically the line 105 in `atmexall` fails:

`compile([alloptions, {passinclude}], fullfile(cdir,'diffusion_matrix.c'));`

Maybe it should be replaced by the following?

`compile([alloptions, {passinclude},LIBDL], fullfile(cdir,'diffusion_matrix.c'));`

Cheers,

Zeus

lfarv:
Fixed in #863

PR #863, opened 2024-12-03
Fix diffusion_matrix compilation on Linux

lfarv:
This fixes a compilation bug reported by @ZeusMarti in #862.


lfarv:
@ZeusMarti : thanks for the fix!

ZeusMarti (approved):
Exactly!

Issue #865, opened 2024-12-04
atx function fails with element EnergyLoss


oscarxblanco:
Dear all,
the matlab AT function `atx` fails when a ring contains an element `EnergyLoss`.
```matlab
Warning: Emittance computation failed:
'PassMethod' field must be a string
> In atx/go (line 208)
In atx (line 116) 
```
It seems to be solved if one includes the pass method in `atdiffmat`.
https://github.com/atcollab/at/blob/8aaae3e1f36bac5a5a2e47fc79088414b6dab10a/atmat/atphysics/Radiation/atdiffmat.m#L16-L18
Modified to
```matlab
newmethods = {'BndMPoleSymplectic4RadPass', ...
              'StrMPoleSymplectic4RadPass', ...
              'ExactMultipoleRadPass', ...
              'EnergyLossRadPass'};
```

lfarv:
Fixed in #870

PR #866, opened 2024-12-04
add EnergyLossRadPass to newmethods in atdiffmat.m

oscarxblanco:
Dear all,
this PR fixes an issue described in #865 , where the function atx fails to compute all parameters when there is an `EnergyLoss` in the ring.

The pass method `EnergyLossRadPass` is added to atdiffmat cell called `newmethods`.

### DIFF
--- a/atmat/atphysics/Radiation/atdiffmat.m
+++ b/atmat/atphysics/Radiation/atdiffmat.m
@@ -15,7 +15,8 @@ function [BCUM,Batbeg] = atdiffmat(ring,varargin)
 
 newmethods = {'BndMPoleSymplectic4RadPass', ...
               'StrMPoleSymplectic4RadPass', ...
-              'ExactMultipoleRadPass'};
+              'ExactMultipoleRadPass', ...
+              'EnergyLossRadPass'};
 
 NumElements=length(ring);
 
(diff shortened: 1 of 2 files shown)


oscarxblanco:
Dear @swhite2401 and @lfarv ,
who could review this PR ?

swhite2401:
Looks ok for me, but I am not very familiar with this

swhite2401 (approved):

swhite2401:
Hello @oscarxblanco , in fact the python version is broken as well.... I add @lfarv as reviewer

swhite2401:
> Hello @oscarxblanco , in fact the python version is broken as well.... I add @lfarv as reviewer

Same fix works in python too, we should integrate it in this PR

lfarv:
@oscarxblanco, @swhite2401

I added the fix for python.

The change is OK, and allows to use ohmi_envelope with EnergyLoss. But there will be no diffusion computed for this element, so it will contribute to damping without introducing excitation.

If this is OK for you, you can go on and merge. @swhite2401; back tp you for approval…

swhite2401 (approved):

lfarv:
@oscarxblanco: up to you for merging, if you agree with forgetting a possible diffusion which is impossible to compute without knowing the exact energy loss mechanism.

lfarv:
This is fixed in #870 where these changes are included, among similar ones

PR #867, opened 2024-12-09
Imprve errors in find_orbit function

swhite2401:
This PR improves the problems reported in #858:
-A `try/except` block protect the `linalg.solve()` function in the closed orbit search
-the energy loss calculation is also protected: tracking is used primarily, if it fails the calculation falls back to integral method, if it still fails an `AtError()` is raised.

Error message in all cases are improved to provide better insight on the problem for the user.

The behavior of the `find_orbit` function is unchanged

### DIFF
--- a/pyat/at/physics/orbit.py
+++ b/pyat/at/physics/orbit.py
@@ -56,7 +56,13 @@ def _orbit_dp(ring: Lattice, dp: float = None, guess: Orbit = None, **kwargs):
         j4 = (in_mat[:4, :4] - in_mat[:4, 4:]) / scaling
         a = j4 - id4  # f'(r_n) - 1
         b = ref_out[:4] - ref_in[:4]
-        b_over_a = numpy.linalg.solve(a, b)
+        try:
+            b_over_a = numpy.linalg.solve(a, b)
+        except numpy.linalg.LinAlgError:
+            msg = 'Particle lost during orbit search, returning NaN'
+            warnings.warn(AtWarning(msg))
+            b_over_a = numpy.zeros(4)
+            b_over_a[:] = numpy.nan
         r_next = ref_in - numpy.append(b_over_a, numpy.zeros((2,)))
         # determine if we are close enough
         change = numpy.linalg.norm(r_next - ref_in)
@@ -106,7 +112,13 @@ def _orbit_dct(ring: Lattice, dct: float = None, guess: Orbit = None, **kwargs):
         j5 = (in_mat[idx, :5] - in_mat[idx, 5:]) / scaling
         a = j5 - id5  # f'(r_n) - 1
         b = ref_out[idx] - numpy.append(ref_in[:4], 0.0) - theta5
-        b_over_a = numpy.linalg.solve(a, b)
+        try:
+            b_over_a = numpy.linalg.solve(a, b)
+        except numpy.linalg.LinAlgError:
+            msg = 'Particle lost during orbit search, returning NaN'
+            warnings.warn(AtWarning(msg))
+            b_over_a = numpy.zeros(5)
+            b_over_a[:] = numpy.nan
         r_next = ref_in - numpy.append(b_over_a, 0.0)
         # determine if we are close enough
         change = numpy.linalg.norm(r_next - ref_in)
@@ -321,7 +333,14 @@ def _orbit6(ring: Lattice, cavpts=None, guess=None, keep_lattice=False,
     harm_number = round(f_rf*l0/ring.beta/clight)
 
     if guess is None:
-        _, dt = get_timelag_fromU0(ring, method=method, cavpts=cavpts)
+        try:
+            _, dt = get_timelag_fromU0(ring, method=method, cavpts=cavpts)
+        except AtError:
+            try:
+                _, dt = get_timelag_fromU0(ring, method=ELossMethod.INTEGRAL, cavpts=cavpts)
+            except AtError:
+                msg = 'Energy loss per turn calculation failed, closed orbit not found'
+                raise AtError(msg)
         # Getting timelag by tracking uses a different lattice,
         # so we cannot now use the same one again.
         if method is ELossMethod.TRACKING:
@@ -353,7 +372,13 @@ def _orbit6(ring: Lattice, cavpts=None, guess=None, keep_lattice=False,
         a = j6 - id6  # f'(r_n) - 1
         b = ref_out[:] - ref_in[:] - theta
         # b_over_a, _, _, _ = numpy.linalg.lstsq(a, b, rcond=-1)
-        b_over_a = numpy.linalg.solve(a, b)
+        try:
+            b_over_a = numpy.linalg.solve(a, b)
+        except numpy.linalg.LinAlgError:
+            msg = 'Particle lost during orbit search, returning NaN'
+            warnings.warn(AtWarning(msg))
+            b_over_a = numpy.zeros(6)
+            b_over_a[:] = numpy.nan
         r_next = ref_in - b_over_a
         # determine if we are close enough
         change = numpy.linalg.norm(r_next - ref_in)


pyat/at/physics/orbit.py
```diff
@@ -56,44 +57,53 @@ def _orbit_dp(ring: Lattice, dp: float = None, guess: Orbit = None, **kwargs):
...
         b = ref_out[:4] - ref_in[:4]
-        b_over_a = numpy.linalg.solve(a, b)
-        r_next = ref_in - numpy.append(b_over_a, numpy.zeros((2,)))
+        try:
+            b_over_a = np.linalg.solve(a, b)
+        except np.linalg.LinAlgError:
+            msg = "Particle lost during orbit search, returning NaN"
+            warnings.warn(AtWarning(msg))
+            b_over_a = np.zeros(4)
+            b_over_a[:] = np.nan
```
lfarv, line 66: Here you are significantly changing the behaviour of `find_orbit`: by changing an error into a NaN orbit, you will break any existing code relying on error catching. I'm thinking of Simone's code for instance, but also of internal AT code like linear optics or matching (I did not check this)

- We agreed that we should never change the behaviour except for correcting a bug. This error is not a bug, it's justified. The only possible thing is to change the message to make it more clear (I do not think any code is checking the message!). Here your message is misleading: a particle lost in tracking gives a NaN solution, not a `LinAlgError`.
- The problem reported by Simone has nothing to do with `LinAlgError`: it's restricted to the computation of radiation losses: it's where we should act
- If we were allowed to change something, I would rather change the NaN orbit happening for tracking failure into an error: propagating NaNs delays the problem, but finally ends in an error. The error is cleaner and easier to detect. But again, **I would not change the existing behaviour**,
- I have the feeling that the `LinAlgError` cannot happen: a successful tracking leading to a singular matrix seems to violate Liouville's theorem, but I may be wrong. Did you ever encounter it? I did some tests with unstable machines, errors, lost particles… and could not get it.

So my preference here would be to leave it as it is. But I have nothing against another message like "Failure in closed orbit computation", since we don't know exactly the reason.

pyat/at/physics/orbit.py
```diff
@@ -281,65 +315,78 @@ def find_sync_orbit(ring: Lattice, dct: float = None, refpts: Refpts = None, *,
...
+        try:
+            _, dt = get_timelag_fromU0(ring, method=method, cavpts=cavpts)
+        except AtError:
+            try:
+                _, dt = get_timelag_fromU0(
+                    ring, method=ELossMethod.INTEGRAL, cavpts=cavpts
+                )
+            except AtError:
+                msg = "Energy loss per turn calculation failed, closed orbit not found"
+                raise AtError(msg)
```
lfarv, line 369: I think that the check, if necessary, should be be done earlier in `get_timelag_fromU0`: if the energy loss computation fails for missing RF voltage, no need to try INTEGRAL, the tracking is right because it includes elements ignored by INTEGRAL, and the error message is correct. The problem might be if the tracking gives a NaN loss, and then I agree, one could try INTEGRAL. For this we need to check for NaN before the comparison with RF voltage.

So here, the option would be to catch the "timelag" error, complement the message to mention that it happens during closed orbit guess, and re-raise it.

However, I still do not see how we can get a NaN loss… We would need a dump of a failing lattice to examine it

pyat/at/physics/orbit.py
swhite2401: Discussing with @simoneliuzzo I understood that a failure in closed orbit search would return NaNs and that this was the expected behavior in this case. @simoneliuzzo could you confirm?

swhite2401: > * I have the feeling that the `LinAlgError` cannot happen

I does happen when the particle is lost within one turn (or one tracking call if your prefer). This results in having NaNs in a and b and raises LinalgError.

simoneliuzzo: yes, I expect nan orbit if there is no closed orbit. no errors. code continues.

lfarv: > This results in having NaNs in a and b and raises LinalgError.

But that's not what I get: on the 2nd iteration, the one-turn tracking fails, and I do not get the LinAlgError, but a NaN orbit, as @simoneliuzzo expects. With additional prints in `_orbit6`:
```
---------- in_mat -----------
[[ 6.02651155e-03  6.02670706e-03  6.02651317e-03  6.02667349e-03
   6.02855442e-03  6.02654493e-03  6.02655181e-03]
 [-7.61424073e-04 -7.61353560e-04 -7.61408829e-04 -7.61370303e-04
  -7.61267285e-04 -7.61403325e-04 -7.61402515e-04]
 [-5.44880542e-03 -5.44881730e-03 -5.44877699e-03 -5.44885298e-03
  -5.44901320e-03 -5.44880371e-03 -5.44880209e-03]
 [-7.52769235e-04 -7.52722692e-04 -7.52720798e-04 -7.52736033e-04
  -7.53354257e-04 -7.52742340e-04 -7.52743806e-04]
 [-2.14422729e-06 -2.13921913e-06 -2.14379602e-06 -2.14166384e-06
   8.39597934e-07 -2.16121821e-06 -2.14281617e-06]
 [-7.92100070e-02 -7.92100391e-02 -7.92100196e-02 -7.92100234e-02
  -7.92095488e-02 -7.92070197e-02 -7.92100184e-02]]
---------- a -----------
[[-2.34227642e+00  5.17488254e+00 -1.28829548e+00  4.05570499e+00
   6.67536730e-01 -2.29591985e-03]
 [-7.18615811e-01  6.31825741e-01 -2.10488213e-01  1.07373606e+00
   4.50765897e-02 -2.69923696e-04]
 [-1.11162286e-01 -5.07042335e-01 -1.63335855e-01 -1.69639203e+00
  -7.03717319e-02 -5.40052338e-04]
 [-8.47649237e-01  7.03798664e-01  7.66936094e-01 -7.40913412e-01
  -2.03483764e-01  4.88581154e-04]
 [-4.70372931e-02  1.19901316e-01 -3.26615604e-02  3.84110411e-02
  -5.86196551e-03 -6.13401374e-03]
 [ 3.79189519e-01 -6.91308787e-01 -3.99132139e-02 -1.66529417e-01
   1.56546119e-01 -4.40069429e-04]]
---------- b -----------
[ 6.02655181e-03 -7.61402515e-04 -5.44880209e-03 -7.52743806e-04
 -2.14281617e-06  4.96409263e-04]
---------- b_over_a -----------
[ 0.00441301  0.00239309  0.0047167   0.00151365  0.00584008 -0.00793069]
---------- in_mat -----------
[[nan nan nan nan nan nan nan]
 [ 0.  0.  0.  0.  0.  0.  0.]
 [ 0.  0.  0.  0.  0.  0.  0.]
 [ 0.  0.  0.  0.  0.  0.  0.]
 [ 0.  0.  0.  0.  0.  0.  0.]
 [ 0.  0.  0.  0.  0.  0.  0.]]
---------- a -----------
[[nan nan nan nan nan nan]
 [ 0. -1.  0.  0.  0.  0.]
 [ 0.  0. -1.  0.  0.  0.]
 [ 0.  0.  0. -1.  0.  0.]
 [ 0.  0.  0.  0. -1.  0.]
 [ 0.  0.  0.  0.  0. -1.]]
---------- b -----------
[       nan 0.00239309 0.0047167  0.00151365 0.00584008 0.07177574]
---------- b_over_a -----------
[nan nan nan nan nan nan]
```
It has been like that for ever.

lfarv: @swhite2401 : can you post the example which gives a LinAlgError ?

lfarv: @simoneliuzzo : coming back to the problem you encountered: when you get the "Not enough RF Voltage" error, could you catch the error and dump the lattice in a file ? Again, there are only 3 possible cases which we need to investigate:

1. a bug in the energy loss computation: that's what we are looking for here, and I still don't see how it can happen,
2. the energy loss in you specific sample lattice is more than 5 times larger than the nominal one,
3. the RF Voltage is broken for some reason

Unless we can reproduce the problem, it's very difficult to find a solution

swhite2401: > @swhite2401 : can you post the example which gives a LinAlgError ?

I have sent you the script and output in your mail box, the lattice file is the same that is producing the RF error

swhite2401:
Threated in #877

PR #868, opened 2024-12-11
Fix Matlab PassMethod type

lfarv:
#865 revealed another bug in Matlab: The PassMethod name substituted for the "old" *RadPass methods must be a "char array" and not a "string". This explains the strange error message reported in #865 and prevents using methods like `BndMPoleSymplectic4E2RadPass`.


lfarv:
This is fixed in #870, I close this PR

PR #870, opened 2024-12-11
Fix ohmienvelope with wigglers

### LINKED ISSUE #869: ohmi_envelope() fails with wiggler element
Inserting a wiggler element with 6d enabled causes ohmi_envelope() to fail with the following error:

`AttributeError: 'Wiggler' object has no attribute 'PolynomA'`

Running the same code within a test_mode context works.

lfarv:
Fixes #869. ohmienvelope fails with wigglers because of a typo in passmethod names

This also includes similar bug fixes in the same files in #868 and #866

### DIFF
--- a/atmat/atphysics/Radiation/atdiffmat.m
+++ b/atmat/atphysics/Radiation/atdiffmat.m
@@ -15,7 +15,9 @@ function [BCUM,Batbeg] = atdiffmat(ring,varargin)
 
 newmethods = {'BndMPoleSymplectic4RadPass', ...
               'StrMPoleSymplectic4RadPass', ...
-              'ExactMultipoleRadPass'};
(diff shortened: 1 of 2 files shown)


swhite2401 (approved):
Very fast, thanks!

lfarv:
@swhite2401 : did you check your test file ? I don't have a test file immediately available…

swhite2401:
I haven't test this specific branch but I fixed the passmethod name in my test branch and it works so on my side it is fine

swhite2401:
> All this shows the drawback of having hard-coded passmethod names… Too bad that there is no other way!

We could make an enum somewhere that maps the passmethod names to centralizes in only one place the hard-coded names. This would mitigate this sort of typos

PR #875, opened 2024-12-12
Set the wiggler indivisible

lfarv:
This avoids a problem in plot_beta reported in #871 by setting a wiggler as "indivisible".

### DIFF
--- a/atmat/atplot/xplot.m
+++ b/atmat/atplot/xplot.m
@@ -76,7 +76,8 @@ if nargout>0, varargout={curve}; end
 
     function newelems=splitelem(elem)
         if isfield(elem,'Length') && elem.Length > 0 ...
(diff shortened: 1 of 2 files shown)


lfarv:
@swhite2401: Not sure it works since I don't have a test case. Can you check?

lfarv:
The fix in Matlab is much more delicate. I'll look at it…

lfarv:
By the way, a "long" RF (rarely seen!) cavity has the same problem. If inactive, no problem, it's a drift. But if active, we should either set it "indivisible", or split the voltage!

swhite2401:
Works on python!

lfarv:
Fixed on Matlab too, ready for merging!

swhite2401 (approved):
I have tested matlab, but python works nicely!

PR #882, opened 2025-01-02
Fix OpenMP compilation

lfarv:
The new integrators including the computation of diffusion matrices do not compile with openMP.
 The bug in the `#pragma omp` is fixed here.

### DIFF
--- a/atintegrators/BndMPoleSymplectic4RadPass.c
+++ b/atintegrators/BndMPoleSymplectic4RadPass.c
@@ -67,7 +67,7 @@ void BndMPoleSymplectic4RadPass(double *r, double le, double irho, double *A, do
         A[0] += sin(KickAngle[1])/le;
     }
     #pragma omp parallel for if (num_particles > OMP_PARTICLE_THRESHOLD) default(none) \
-    shared(r,num_particles,R1,T1,R2,T2,RApertures,EApertures,\
+    shared(r,num_particles,R1,T1,R2,T2,RApertures,EApertures,bdiff,\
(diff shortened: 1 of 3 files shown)


oscarxblanco:
Dear @lfarv ,
I have compiled it in AT matlab and pyat. Both seem to work.
However, I noticed that for a ring with an `EnergyLoss` element the emittance calculated by `radiation_parameters` and `ohmi_envelope` in pyat are different.
``` python
# ring with EnergyLoss element
>>> emit0['emitXYZ'] # from ohmi envelope in pyat
array([1.87729544e-10, 5.67160085e-38, 2.28673862e-06])
>>> pars.emittances # from radiation_parameters in pyat
array([2.00381041e-10,            nan, 3.05585928e-06])
```
This is not the case in AT matlab, `atsummary` and `atx` return the same emittance
```matlab
# [~,pa]=atx(THERING,0);
>> pa.modemittance(1)
   1.8778e-10
# atsummary(THERING);
Natural Emittance: 		 1.87429e-10 [mrad] ##  I think this should be rad not mrad !!!
```

And they all return about 200 pm.rad if the `EnergyLoss` element pass method is set to `IdentityPass`.

I believe the reason to have this energy loss pass method was to include the missing energy loss from IDs (maybe some other effects), and it seems to have a non-negligible effect on emittance. The `EnergyLossRadPass` was included in PR #870 for AT matlab. Maybe it should also be included in pyat `radiation_parameters` ?

oscarxblanco (approved):

lfarv:
@oscarxblanco: thanks for testing.
I created a new issue reporting your findings. Since the errors you mention are not correlated to the OpenMP compilation bug, they should be covered in another PR, and I'd like to merge this one first.

Issue #883, opened 2025-01-03
Physical constants updated in SciPy


lfarv:
scipy 1.15 has been released on January 3rd. It updates the physical constants to CODATA 2022, instead of CODATA 2018 used previously. As a consequence, the Matlab tests run on Github fail until the Matlab constants are updated.

Until this is done, ignore the following errors in Matlab tests:

- pytests/ringparameters(lat2=hmba)
- pytests/ringparameters(lat2=spear3)

This is due to a different relativistic gamma because of the updated electron mass.

The constants in C are already set to CODATA 2022 (in #742)

lfarv:
The problem is slightly worse than anticipated: ScyPy 1.15 requires python >= 3.10. So python 3.7, 3.8 and 3.9 are stuck with the old constants.

Consequently, AT results are slightly different depending on python version…

lfarv:
The Matlab values are updated in #884. With this, the values in python>=3.10, Matlab and hard-coded in C are now identical.

The values in python<3.10 are still form CODATA 2018

lfarv:
Partially solved in #884. But for python<3.10, we are stuck with old values

PR #884, opened 2025-01-05
Update of Matlab physical constants to CODADA 2022

lfarv:
As reported in #883, the python physical constants in `scipy.constants` have been updated to CODATA 2022. This brings the AT Matlab values on par with the python values.


oscarxblanco:
Dear @lfarv , some of the changes are just changing rows.
From the header I understand this is an automatically generated list. I see no issue on the calculations I launch on MATLAB R2023a.

oscarxblanco (approved):

lfarv:
@oscarxblanco : I generate the list from a text file published by [NIST](https://pml.nist.gov/cuu/Constants/Table/allascii.txt). Unfortunately, the order of this list has changed since the previous version, which makes the comparison more difficult. But there are indeed small modifications, in particular 3 more digits on the electron rest mass, inducing a (very small) modification ofAT results…

Issue #886, opened 2025-01-08
The EnergyLoss element is ignored in the radiation_parameters PyAT function


lfarv:
I noticed that for a ring with an `EnergyLoss` element the emittance calculated by `radiation_parameters` and `ohmi_envelope` in pyat are different.
``` python
# python
# ring with EnergyLoss element
>>> emit0['emitXYZ'] # from ohmi envelope in pyat
array([1.87729544e-10, 5.67160085e-38, 2.28673862e-06])
>>> pars.emittances # from radiation_parameters in pyat
array([2.00381041e-10,            nan, 3.05585928e-06])
```
This is not the case in AT matlab, `atsummary` and `atx` return the same emittance
```matlab
% matlab
% [~,pa]=atx(THERING,0);
>> pa.modemittance(1)
   1.8778e-10
% atsummary(THERING);
Natural Emittance: 		 1.87429e-10 [mrad] ##  I think this should be rad not mrad !!!
```

And they all return about 200 pm.rad if the `EnergyLoss` element pass method is set to `IdentityPass`.

I believe the reason to have this energy loss pass method was to include the missing energy loss from IDs (maybe some other effects), and it seems to have a non-negligible effect on emittance. The `EnergyLossRadPass` was included in PR #870 for AT matlab. Maybe it should also be included in pyat `radiation_parameters` ?

P.D. There seems to be an error in the units printed aside of the `atsummary` emittance.

_Originally posted by @oscarxblanco in https://github.com/atcollab/at/issues/882#issuecomment-2575920193_

swhite2401:
Solved in #888

PR #888, opened 2025-01-10
Improve radiation_parameters

lfarv:
This answers 2 issues reported by @oscarxblanco in #886:

- a unit bug in `atsummary` (Matlab)
- the `EnergyLoss` element is taken into account in the radiation integrals. Note that the energy loss element adds damping ($I_2$ integral), thus reducing the equilibrium emittance, but generates no diffusion. This makes sense assuming that the losses summarised in the element occur in non-dispersive locations.


swhite2401 (approved):

oscarxblanco (approved):

oscarxblanco:
Dear @lfarv ,
I would suggest to include your paragraph in the element's help message. Otherwise, I tested it and it works as expected.

- the EnergyLoss element is taken into account in the radiation integrals. Note that the energy loss element adds damping ( I 2 integral), thus reducing the equilibrium emittance, but generates no diffusion. This makes sense assuming that the losses summarised in the element occur in non-dispersive locations.

lfarv:
> I would suggest to include your paragraph in the element's help message

Your are right, @oscarxblanco. The help the element is modified in both Matlab and python. But I need you to approve again, sorry…

Issue #887, opened 2025-01-09
off energy rdts


oscarxblanco:
Dear all,
it would be nice to have the energy offset as optional input for the rdt calculation.
Right now `avlinopt` uses the default dp=0.

https://github.com/atcollab/at/blob/d2f764ec13980efd5f2c08837abfedf73cd43577/pyat/at/physics/rdt.py#L505

PR #890, opened 2025-01-14
Add dp to rdts

oscarxblanco:
This PR adds the energy offset as option to the calculation of the optics used in AT matlab `computeRDT` and pyat `get_rdts`.

### DIFF
--- a/atmat/atphysics/NonLinearDynamics/computeRDT.m
+++ b/atmat/atphysics/NonLinearDynamics/computeRDT.m
@@ -6,27 +6,35 @@ function RDT=computeRDT(ring, index, varargin)
 %   
 %   RDT=computeRDT(ring, index, varargin)
 %
+%   INPUTS
 %   ring is the AT lattice
 %   index is the vector of indexes where one wants to compute RDTs
 %   The additional arguments can be up to five strings:
 %   chromatic, coupling, geometric1, geometric2 and tuneshifts
+%
+%   OPTIONS (order does not matter)
+%   dp: Default 0. Energy offset to calculate the optics parameters.
 %   
-%   example:
+%   EXAMPLES
 %   RDT=computeRDT(ring, indexBPM, 'geometric1', 'tuneshifts');
 %   creates an array of structs (the length of the array is the number of 
 %   indexes where you want to compute driving terms) with first order
 %   geometric driving terms and tune shifts with amplitude.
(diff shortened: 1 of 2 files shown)


oscarxblanco:
Dear @swhite2401  and @lfarv , who could review this PR ? 
This addresses issue #887 .

swhite2401:
@oscarxblanco this looks fine to me, are the off-momentum RDTS the same as the on-momentum ones? High order terms appear no?

oscarxblanco:
@swhite2401 , @lfarv .
Before talking about the rdt I have a doubt. Does `avlinopt` requires ring.is_6d False ?
I'm getting this error
```python
>>> ring.avlinopt(refpts=at.All)
  File "/home/oblanco/Documents/public/progs/pyenv/pyenvat/lib/python3.11/site-packages/at/lattice/utils.py", line 279, in wrapper
    raise AtError('{0} needs "ring.is_6d" {1}'.format(
at.lattice.utils.AtError: avlinopt needs "ring.is_6d" False
```

swhite2401:
I thought I had fixed this by calling linopt6 in avlinopt, do you run the latest master version?

oscarxblanco:
> I thought I had fixed this by calling linopt6 in avlinopt, do you run the latest master version?

@swhite2401 , yes, it is the latest master
```
>>> at.__version__
'0.6.2.dev66+gbc8b345a'
```

oscarxblanco:
> By the way I discussed with @carmignani on that point and the implementation you propose is missing all the feed down effects from the off-energy orbit. It is possible to implement but would require more work.

oh, I see.
I guess I should
- calculate the closed orbit in all  reference points
- get the feed-down polynom B and A for those reference points in a magnet,
- extract the quad, sextupole and octupole components
- operate those components with the off-energy optics.

Am I missing something ?

swhite2401:
> > By the way I discussed with @carmignani on that point and the implementation you propose is missing all the feed down effects from the off-energy orbit. It is possible to implement but would require more work.
> 
> oh, I see. I guess I should
> 
>     * calculate the closed orbit in all  reference points
> 
>     * get the feed-down polynom B and A for those reference points in a magnet,
> 
>     * extract the quad, sextupole and octupole components
> 
>     * operate those components with the off-energy optics.
> 
> 
> Am I missing something ?

Yes I think this is about it

PR #891, opened 2025-01-15
Fix OpenMP compilation bug

lfarv:
This fixes an OpenMP compilation error in the specific case of macOS with OpenMP installed with the MacPorts package manager.


lfarv:
It's difficulty to test all possible cases, but I hope this is the last one.

oscarxblanco:
Dear @lfarv ,
I don't have a mac to test it. I will approve as @swhite2401 did in the previous one.

PR #892, opened 2025-01-15
Improved matching

lfarv:
This branch tries to follow the conclusions of the discussion #880 concerning improvement of matching process.

It's a work in progress as long as the discussion evolves.

This branch is fully compatible with the master branch, and you are welcome to test and check the improvements being discussed in #880.

Improvements:
- New LocalOptics codes `muf`, `mu2pi`, `mu2pif`
- Unnamed user-defined functions are accepted (`partial()`, lambda functions)
- The `plane` keyword is used the restrict the output of user-defined functions (in LocalOptics and GlobalOptics)


lfarv:
Hello @swhite2401 

The `name` keyword already exists for all observables. You can give an arbitrary name to any observable to identify it easily. By default, AT creates a name based on the observable definition. So you can already use the `partial` function, as shown in this example:

```python
def phase_advance(elemdata, index=Ellipsis):
    mu = elemdata.mu
    v = mu[-1] - mu[0]
    return v[index]

obs = ObservableList()

# Observe all planes: directly call the user-defined function
obs.append(
    LocalOpticsObservable([33, 101], phase_advance, all_points=True, summary=True)
)
# Observe the horizontal plane: use the index keyword
obs.append(
    LocalOpticsObservable([33, 101], partial(phase_advance, index=0), name="horizontal", all_points=True, summary=True)
)
# Observe the vertical plane
obs.append(
    LocalOpticsObservable([33, 101], partial(phase_advance, index=1), name="vertical", all_points=True, summary=True)
)

obs.evaluate(hmba_lattice)
[array([3.10650429, 2.99741974]),
 np.float64(3.1065042949785924),
 np.float64(2.9974197440054082)]

print(obs)
    location              Initial            Actual         Low bound        High bound         deviation 
phase_advance
                 [ 3.107     ...]  [ 3.107     ...]               -                 -    [ 0.0       ...] 
horizontal
                           3.1065            3.1065               -                 -                 0.0 
vertical
                          2.99742           2.99742               -                 -                 0.0 
```

lfarv:
However, we can think of other ways to answer the question:
- The `plane` keyword is ignored , because I thought that the user-defined function can return anything, possibly unrelated with any "plane" notion. But we can use it, to select a single item in the result of the user-defined function, as it is done for standard parameters. By default, as usual, it would return the whole result.
- The easiest way is probably to use the `statfun` keyword and specify the numpy `diff` function: no user-defined function, no need to give a name, the defaults are ok. One can use the `plane` keyword and all the `mu`, `m2pi`, `muf`, `mu2pif` variants:
  ```python
  obs = ObservableList()
  
  # Phase advance
  obs.append(LocalOpticsObservable([33, 101], "mu", all_points=True))
  # Differentiation of the phase advance
  obs.append(LocalOpticsObservable([33, 101], "mu", all_points=True, statfun=np.diff))
  
  obs.evaluate(hmba_lattice)
  [array([[ 2.83611057,  1.27674451],
          [12.22580017,  4.27416425]]),
   array([[9.3896896 , 2.99741974]])]
  
  print(obs)
    location              Initial            Actual         Low bound        High bound         deviation 
  mu
      BPM_03                    -    [ 2.836     ...]               -                 -    [ 0.0       ...] 
      BPM_09                    -    [ 12.23     ...]               -                 -    [ 0.0       ...] 
  diff(mu)
                                -    [ 9.39      ...]               -                 -    [ 0.0       ...]
  ```
  I think I will rewrite the example notebook to use this method. I have to check the behaviour if more that 2 refpoints are given: the shape of the output is then `nb_refs-1`

swhite2401:
> ```python
> obs.append(
>     LocalOpticsObservable([33, 101], partial(phase_advance, index=1), name="vertical", all_points=True, summary=True)
> )
> ```

Ah but I did exactly that and it still complained.... Let me try again then

swhite2401:
> > ```python
> > obs.append(
> >     LocalOpticsObservable([33, 101], partial(phase_advance, index=1), name="vertical", all_points=True, summary=True)
> > )
> > ```
> 
> Ah but I did exactly that and it still complained.... Let me try again then

By the way what is all_points doing? Is this a phase specific argument to get the correct integer phase advance?

lfarv:
> what is all_points doing?

It specifies that the optics should be computed at all points (see [LocalOpticsObservable documentation](https://at.readthedocs.io/en/latest/p/api/at.latticetools.observables.html#at.latticetools.observables.LocalOpticsObservable)). It's automatically set when "mu" is requested, but for a user-defined function, it cannot guess…

lfarv:
@swhite2401 : Following your remarks, here are 2 improvements:

1. A user-defined function without a name (`partial()` for instance) will not be refused anymore: the Observable name is set to "\<function\>". Similarly, a lambda function is named "\<lambda\>". However, it is still recommended to give the Observable a more meaningful name by using the `name` keyword,
2. The `plane` keyword is now used for user-defined functions: it selects the desired values in the output of the user-defined function. Integers, slice objects, Ellipsis, tuples of such items can be used for the `plane` keyword. By default, the whole output of the user-defined function is returned

swhite2401:
> > what is all_points doing?
> 
> It specifies that the optics should be computed at all points (see [LocalOpticsObservable documentation](https://at.readthedocs.io/en/latest/p/api/at.latticetools.observables.html#at.latticetools.observables.LocalOpticsObservable)). It's automatically set when "mu" is requested, but for a user-defined function, it cannot guess…

Sorry for my poor memory maybe we already had this discussion, but do we have an option to look only at the fractional par of mu so we do not have to compute optics at all refpts? This significantly speeds-up the matching

lfarv:
> Sorry for my poor memory maybe we already had this discussion, but do we have an option to look only at the fractional par of mu so we do not have to compute optics at all refpts? This significantly speeds-up the matching

Yes, sir ! This branch accepts the `muf`, `mu2pi`, `mu2pif` codes that you defined in #880. `all_points` is set for the `mu` and `mu2pi` options only,

swhite2401:
> > Sorry for my poor memory maybe we already had this discussion, but do we have an option to look only at the fractional par of mu so we do not have to compute optics at all refpts? This significantly speeds-up the matching
> 
> Yes, sir ! This branch accepts the `muf`, `mu2pi`, `mu2pif` codes that you defined in #880. `all_points` is set for the `mu` and `mu2pi` options only (or if explicitly specified).

Ok great thanks! I have been very busy with other things these days and did not have much time to concentrate on AT...

lfarv:
Improved documentation:
- more info on the shape of observable values and on broadcasting: useful for specifying the target for matching
- a new notebook shows how to use the new matching
- deprecation warning in the documentation of the old matching

swhite2401 (approved):

PR #898, opened 2025-01-29
Crab cavity passmethod

swhite2401:
The crab cavity pass method is add to atintegrators.
Matlab and python constructors need to be added.


swhite2401:
I created this branch from PR #839 , this is wrong sorry... let me restore it to the master branch

PR #905, opened 2025-03-05
Allow setting the seed of random generators

lfarv:
Following #879, we introduce an optional `seed` keyword argument in
- `lattice_track` (python)
- `linepass`, `ringpass` (Matlab)

When provided, this will reset the seed of the two C random generators. If omitted, the sequences of random numbers will continue as before: no change.

### DIFF
--- a/atintegrators/TestRandomPass.c
+++ b/atintegrators/TestRandomPass.c
@@ -1,4 +1,4 @@
-/* RandomPass.c
+ /* RandomPass.c
    Accelerator Toolbox
 
    Test of random generators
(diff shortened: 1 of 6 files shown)


carmignani:
Hello @lfar, thank you very much for the modification. I tested the input 'seed' in ringpass and linepass and it works well. I also sent many jobs in the cluster with radiation and quantum excitation, all identical except the seed in ringpass, and I get a nice equilibrium emittance.
I did not check the python version.

carmignani (approved):

PR #906, opened 2025-03-06
Update beamloading

lcarver:
Added the ability to specify an additional detuning angle in the beam loading element. When considering stability issues for active cavities, it can be desirable to apply a delta detuning angle from the optimum working point.

By specifying 'detune_angle' the feedback loop will feedback on the phase position of the optimum + detune_angle.

Below is a plot comparing the PyAT results with results based on the analytical formula (taken from MB2). There is some small disagreement, but the analytical formula contains a large amount of assumptions, whereas the numerical tracking of PyAT has fewer assumptions and also takes into account the bunch distribution 

unnamed

I also show here a plot showing that the detune_angle is implemented correctly, as the beam remains centered on 0 as a function of the detune_angle.

unnamed

### DIFF
--- a/atmat/atmexall.m
+++ b/atmat/atmexall.m
@@ -257,7 +257,7 @@ warning(oldwarns.state,oldwarns.identifier);
                 compile(compargs, pmeth);
             catch errcomp
                 if fail
-                    rethrow(err);
+                    rethrow(errcomp);
                 else
                     fprintf(2, 'Could not compile %s:\n%s\n', pmeth, errcomp.message);
                 end
--- a/atmat/atphysics/LongitudinalDynamics/atsetcavity.m
+++ b/atmat/atphysics/LongitudinalDynamics/atsetcavity.m
@@ -1,15 +1,16 @@
 function ring = atsetcavity(ring,varargin)
 %ATSECAVITY Set the parameters of RF cavities
 %
-%WARNING: This function modifies the time reference,
-%this should be avoided
-%
 %ATSETCAVITY may be used in two modes:
 %
 %Upgrade mode
 %===================================================
+% By default, ATSETCAVITY will act on the "main" cavities: they are defined by the
+% cavpts ring property, or if absent by cavities at the lowest frequency.
+%
 %NEWRING=ATSETCAVITY(RING,...,'Frequency',FREQUENCY,...)
-%   Set the cavity frequency [Hz].
+%   Set the cavity frequency [Hz]. FREQUENCY is a scalar or an array as
+%   long as the list of selected cavities
 %
 %NEWRING=ATSETCAVITY(RING,...,'Frequency','nominal',...)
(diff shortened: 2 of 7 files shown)


lfarv:
@lcarver
You are experiencing a known bug in atmexall which hides the compilation errors. This has been corrected in the "crab_cavity" branch, but it's still far from being merged…
The easiest way would be to correct it here, it's such an obvious modification. Or open another PR just for that, which I would immediately approve

lcarver:
Hello Laurent,
I have pushed the fix here. I found the bug but was trying to merge it separately. But if you think it's ok to mix the two together. I open this PR for review, which includes the beamloading update and the bugfix for atmexall

atintegrators/BeamLoadingCavityPass.c
```diff
@@ -103,7 +105,7 @@ void BeamLoadingCavityPass(double *r_in,int num_particles,int nbunch,
         psi = vgenk[1];
     }
     /*Track RF cavity is always done. */
-    trackRFCavity(r_in,le,vgen/energy,rffreq,harmn,tlag,-psi,nturn,circumference/C0,num_particles);
+    trackRFCavity(r_in,le,vgen/energy,rffreq,harmn,tlag,-psi+detune_angle,nturn,circumference/C0,num_particles);
```
swhite2401, line 108: Don't you want detune_angle to have the same sign as psi?

pyat/at/collective/beam_loading.py
```diff
@@ -154,6 +154,8 @@ def __init__(self, family_name: str, length: float, voltage: float,
             cavitymode (CavityMode):  Is cavity ACTIVE (default) or PASSIVE
             buffersize (int):  Size of the history buffer for vbeam, vgen, vbunch
                 (default 0)
+            detune_angle:      Fixed detuning from optimal tuning angle. [rad]
```
swhite2401, line 157: Should you describe the sign convention here?

swhite2401: You may want to add this in the docstring of add_beamloading function as well no?

atmat/atmexall.m
```diff
@@ -149,7 +149,7 @@ function atmexall(varargin)
                 compile(compargs, pmeth);
             catch errcomp
                 if fail
-                    rethrow(err);
+                    rethrow(errcomp);
```
swhite2401, line 152: Is this to be fixed in another PR then? Maybe you should start with this?

atintegrators/BeamLoadingCavityPass.c
lcarver: No. The Vgen[1] is psi, and for a main cavity in electron machines the optimum is at -ve psi, which is -ve detuning in frequency due to the -ve gradient. It is the opposite for harmonic cavities.

I prefer to keep it as an absolute delta, which is the most logical.

PR #907, opened 2025-03-06
Bugfix atmexall

lcarver:
While trying to fix the errors for BeamLoadingCavityPass, I found a small error in atmexall.

The error was "Line 152: err not found". Basically it caught errcomp, but then errorred trying to throw the message, so no useful information was given for debugging.


lcarver:
I am making a mess. Please wait.

PR #909, opened 2025-03-14
RDT observables

lfarv:
This PR implements RDT observables. This allows to use RDTs in matching, response matrices…

The `RDTObservable` class is similar to `LocalOpticsObservable`, but needs the RDT name (e.g. "h20000").


lfarv:
module `at/physics/rdt.py`:

On top of many docstring modifications to make Sphinx happy, there are 2 code modifications:

1. for the `rdt_type` argument, instead of the `Sequence` type, we allow the wider `Container` type: in addition to the previous types, this allows a `set` argument, perfectly adequate for this kind of input. No side effect is expected,
2. for the output, `rdt` is changed from `dict` to numpy record arrays (`numpy.recarray`):
   - accessing a value with `rdt["h20000"] `works exactly as before, like with a dictionary,
   - access is also possible with `rdt.h20000` (attribute syntax)
   - more useful (and required for the observable), `rdt[:10]` works, and means the array of the 10 first rdts. So one can write either `rdt["h20000"][1:10]` as before, or `rdt[:10]["h20000"]`.
   - this is the output type used in `get_optics`, `ohmi_envelope` and others;

These modifications don't change the usage, they just add new possibilities. However, they are not mandatory and I can easily revert them. The conversion from dict to recarray is a single line, I can easily remove it and make the conversion in the observable. The only side effects are the fact that the `recarray` does not implement the dict methods `dict.pop`, `dict.update`, adding new fields to the rdt result. I don't think these are of any interest. The only point I found is that, to get the list of existing keys, one must replace` rdt.keys()` by `rdt.dtype.names`.

So it's up to you to decide if we keep the record array or revert to the dict.

lfarv:
As this is the 1st complex observable we have, I added a new keyword called `kind`. It specifies a post-processing of the values to return the real part, imaginary part, module or angle of the result.

I think this could generalised to all observables, but I'm not satisfied with the `kind` name. Any better idea ? There is already a `statfun` argument, reserved for "statistic" processing (`mean`, `std`, `max`…). Unlike `kind`, this one modifies the shape of the result. I think both attributes are useful:  you may want to take the `max` of the `abs`  of a complex array.

Also, `statfun` is entered as a function (`np.mean` for instance), should it be similar for the "kind" keyword (`np.imag`)? At the moment, it's a string. What do you think?

lfarv:
The modules `observable.py` and `observablelist.py` have been significantly modified, but this is for a better behaviour when one of the evaluation fails: then the "evaluation" still succeeds, but accessing the "missing" value raises an exception. This PR fixes several bugs appearing when trying to display (`repr`) ot print (`str`) the observable list or a single observable in these degraded conditions.

The modification for RDTs are limited and mostly located in a new module `rdt_observable.py`.

lfarv:
> Also, statfun is entered as a function (np.mean for instance), should it be similar for the "kind" keyword (np.imag)? At the moment, it's a string. What do you think?

Now, `statfun` may be entered as a function (`np.mean`) as before, but also as a string in {"min", "max", "mean", "std", "var"}. It's is available for subclasses of `ElementObservable`

Similarly, the new `procfun` may be entered either as a function (`np.imag`), or as a string in {"real", "imag", "abs", "angle", "log", "exp", "sort"}. It's available for all `Observable`s.

The evaluation order is: `result=statfun(procfun(data))`. See the [documentation](https://at.readthedocs.io/en/rdt_observables/p/api/at.latticetools.observables.html#).

Any ideas for a better name for the arguments, instead of `procfun` ?

swhite2401 (approved):

lfarv:
@swhite2401 : I had to solve a conflict after merging the response matrices, and changed to "procfun" argument into "postfun". So I need you reapproval, sorry…

swhite2401 (approved):

Issue #910, opened 2025-03-14
PR #877 (Fix Energy Loss Computation) has broke the beamloading computation


lcarver:
There is a problem with the 'ts' setpoint I get from the function get_timelag_fromU0 and the problem originated with PR #877 . The knock-on effect is the beam loading compensation no longer works (it is inverted due to bad synchronous phase, the Vgen should be larger than the cavity setpoint for some positive current). I think there is a sign error possibly, originating from this PR. I am investigating.

```

import numpy as np
from at.constants import clight
import at
        
ring_dict = {
            'circumference': 843.977345496621,
            'harmonic_number': 992,
            'ac': 8.511728693902269e-05,
            'energy': 6e9,
            'sigma_e': 0.0009356632506279087,
            'U0': 2532505.857673822
            }

Vc = 6.5e6
Q0 = 37500
beta = 2.8
QL = Q0/(1+beta)
Rs = 145*13*Q0   
Rsl = 145*13*QL
   
phi_s = np.pi - np.arcsin((ring_dict['U0']/Vc))


Nbunches = 992
I0 = 50e-3
Nslice = 100

nparts_per_bunch = 1
Nparts = Nbunches

Nturns = 1000
sigma_matrix = at.sigma_matrix(betax=1, betay=1, alphax=0, alphay=0, emitx=100e-12, emity=10e-12, espread=ring_dict['sigma_e'], blength=11e-3)

simple_ring = at.simple_ring(ring_dict['energy'], ring_dict['circumference'], ring_dict['harmonic_number'], 0.1, 0.1, Vc, ring_dict['ac'], U0=ring_dict['U0'], tauz=3229.114299199956, espread=ring_dict['sigma_e'])
simple_ring.set_fillpattern(Nbunches)
simple_ring.set_beam_current(I0)

#Setup main cavity
mc_elem = simple_ring[0]
tl_main = clight*(np.pi - phi_s)/(2*np.pi*mc_elem.Frequency)
mc_elem.TimeLag = tl_main

# add beamloading
at.add_beamloading(simple_ring, QL, Rsl, cavpts=[0], Nslice=Nslice, blmode = at.BLMode.PHASOR, VoltGain=0.01, PhaseGain=0.01)

parts = at.beam(Nparts, sigma_matrix)

for iturn in np.arange(Nturns):
    if iturn%100==0:
        print('Tracking:', iturn)
        print(simple_ring[0].Vgen)
[...]

PR #911, opened 2025-03-14
Fix the energyloss computation for simple ring

lcarver:
This fixes issue #910 

The energy loss element has the name SimpleRadiationPass so was being ignored by the endswith('RadPass') when trying to compute the U0 for a lattice with a simple_ring.

### DIFF
--- a/atintegrators/SimpleRadiationRadPass.c
+++ b/atintegrators/SimpleRadiationRadPass.c
@@ -13,7 +13,7 @@ struct elem
   double EnergyLossFactor;
 };
 
-void SimpleRadiationPass(double *r_in,
+void SimpleRadiationRadPass(double *r_in,
            double *damp_mat_diag, double dispx, double dispxp,
            double dispy, double dispyp, double EnergyLossFactor, int num_particles)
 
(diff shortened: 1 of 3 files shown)


lfarv:
Hello @lcarver
I think that relying on the PassMethod is not the right way: the `EnergyLoss` element is also skipped when computing the energy loss. The correct test is:
```python
if isinstance(e, _Radiative) and e.longt_motion:
```
- The `e.longt_motion` property ensures that the element is "active": 6d motion is turned on,
- `isinstance(e, _Radiative)` takes out the cavities and collective effects.

Going back to the `Element` inheritance graph, I noticed another two problems:
1. `QuantumDiffusion` and `SimpleQuantDiff` are not considered as "radiating" (they don't inherit from `_Radiative`). They are useless in this computation, but more generally, I wonder if this is correct. For instance, a dipole with a `BndMPoleSymplectic4QuantPass` is considered as radiating. The question is open…
2. The distinction between `_Radiative` and `Radiative` is not used anywhere, and is even incorrect: `SimpleRadiation` and `EnergyLoss` are part of the elements turned on by default, though not inheriting from `Radiative`. Both should be grouped under the `Radiative` name. This would avoid accessing private classes. I'll open another PR for that.

lfarv:
On the other hand, in Matlab there is no other option than checking the passmethod. It's a pity that the passmethod was not called `SimpleRadPass` or `SimpleRadiationRadPass`… Would such a change be an option? Otherwise, you should set your test in `at/atmat/atphysics/Radiation/atgetU0.m`

lcarver:
I think changing the name to SimpleRadiationPass is the most straightforward and immediate solution. (I prefer this name to SimpleRadPass as the name is a little more instructive about what it is doing) . l get to it on Monday morning. 

As for your pt 1. We have run into issues in the past of trying to compute u0 with a quant diff element, the result is a u0 that changes each time you run it. Maybe some issues can start to arise when using energy loss computation with BndMpole..QuantPass. for example if you add beam loading on a ring with BndMpole...QuantPass, the energy loss is computed on the ring that you are setting it to and you will have a jittery synchronous phase.

lfarv:
> We have run into issues in the past of trying to compute u0 with a quant diff element, the result is a u0 that changes each time you run it. 

You get the correct result by tracking many particles and taking the average. That's clearly not the most efficient way, but I made this check, which revealed a [problem with random generators](https://github.com/atcollab/at/issues/592#issuecomment-1500188375) with OpenMP. That's why OpenMP is still disabled in QuantPass methods…

lcarver:
I changed the name from SimpleRadiationPass to SimpleRadiationRadPass. With my checks, it passes the logic and computes the U0 correctly. 

I think I am the only one who uses simple_ring, so for me there is no further checks to do.

PR #912, opened 2025-03-17
Add non-linear chromaticity to DeltaQPass

lcarver:
Hello,
I have added the feature to use non-linear chromaticity in DeltaQPass.

Basically now, instead of Qpx,Qpy as input, it takes an array of arbitrary length (chromx_arr, chromy_arr) which corresponds to [Q'x, Q''x, Q'''y...].

I also ensured that for the simple_ring, if you specify just Qpx and Qpy, it still works (it just makes an array of zeros and sets the Q' value). I tested it with simple_ring, with the script below which works well. (I also found a bug in at.chromaticity where the factorial was being divided instead of multiplied which was incorrect).

I have not done anything for the MatLab side. @carmignani can you take a look?

```
import at
import numpy as np
from scipy.special import factorial

simple_ring = at.simple_ring(3e9, 528, 176, 0.1, 0.1, 1e6, 0.000306, U0=0.0, tauz=1000, espread=0.000769,Qpy=1.0, chromx_arr=[0,1e3,1e5])

simple_ring.enable_6d()

fitOrder = 4
(fitH, fitV), dpa, qz = at.chromaticity(simple_ring, method='linopt', dpm=3e-3, npoints=20, order=fitOrder, dp=0)

fitH_pv = fitH/factorial(np.arange(fitOrder+1))
fitH_pv = fitH_pv[::-1]

fitV_pv = fitV/factorial(np.arange(fitOrder+1))
fitV_pv = fitV_pv[::-1]

labelH = r'$Q={:.1f}$'.format(fitH[0]) + '\n' +  \
         r"$Q'={:.2f}$".format(fitH[1]) + '\n' +   \
         r"$Q''={:.4f}$".format(fitH[2])

labelV = r'$Q={:.1f}$'.format(fitV[0]) + '\n' +  \
         r"$Q'={:.2f}$".format(fitV[1]) + '\n' +   \
         r"$Q''={:.4f}$".format(fitV[2])

fig, ax = plt.subplots(2,1)
ax[0].plot(dpa, np.polyval(fitH_pv, dpa), label=labelH)
ax[0].plot(dpa, qz[:,0], linestyle='None', marker='x')
ax[1].plot(dpa, np.polyval(fitV_pv, dpa), label=labelV)
ax[1].plot(dpa, qz[:,1], linestyle='None', marker='x')
ax[0].legend()
ax[1].legend()

plt.show()

```

image


lcarver:
It fails the tests because the output of at.chromaticity is different. Now it is a meaningful value, whereas before it was not. What should I do?

pyat/at/physics/fastring.py
```diff
@@ -142,6 +142,8 @@ def simple_ring(
     name: str = "",
     particle: str | Particle = "relativistic",
     TimeLag: float | Sequence[float] = 0.0,
+    chromx_arr: Sequence[float] = 0.0,
+    chromy_arr: Sequence[float] = 0.0
```
lfarv, line 146: Can't you just keep the `Qpx` and `Qpy` existing arguments?
```python
Qpx: float | Sequence[float],
Qpy: float | Sequence[float],
...
chromx_arr = np.ravel(Qpx)  # Forces 1d array
chromy_arr = np.ravel(Qpy)
chrom_maxorder = max(chromx_arr.size,chromy_arr.size)
```
This avoids introducing redundant new arguments.

lfarv:
> What should I do?

If the results were wrong and are now correct, you have to change the test, no other way…

pyat/at/physics/fastring.py
lcarver: I considered this but Qpx refers to Qprime which is linear chromaticity which may be a bit misleading.

lfarv: Right, but a logic like "ignored if…' is also misleading. Considering that Qpx is the 1st term of the series, that would be preferred solution!

lcarver:
Except for the comment about Qpx, Qpy (which remains open) , the python side is done.

lcarver:
I have made several changes. Firstly, I had forgotten to update the python fast_ring. This is now done and works. I only do the linear chromaticity for the fast_ring. 

I also had a go at updating all the matlab files. I made the implementation in atsimplering and atfastring. I updated the tests in the pytests.m as well as in test_matlab/test_cmp_physics.

Probably the var parsing in simplering and fastring for matlab can be improved. But I think the main work is done. Should be ready for review

pyat/at/physics/fastring.py
```diff
@@ -287,6 +293,21 @@ def simple_ring(
...
+    chromy_arr = np.ravel(Qpy)
+    chrom_maxorder = max(chromx_arr.size, chromy_arr.size)
+
+    if not np.any(chromx_arr):
+        chromx_arr = np.zeros(chrom_maxorder)
+        chromx_arr[0] = Qpx
+        
+    if not np.any(chromy_arr):
+        chromy_arr = np.zeros(chrom_maxorder)
+        chromy_arr[0] = Qpy
```
lfarv, line 306: Useless now: it's already using Qpx

lcarver: Thanks, this has been removed.

lcarver:
Hello @lfarv, I am not familiar with the syntax for the matlab tests. Can you advises on how to pass it? The problem is for the fast ring check in pytests.m

atmat/attests/pytests.m
```diff
@@ -268,8 +268,8 @@ function checkattr(rm, rp)
                 testCase.verifyEqual(rm{end}.Alphay, rp{3}.Alphay, AbsTol=1.e-10);
                 testCase.verifyEqual(rm{end}.Betax, rp{3}.Betax, RelTol=1.e-10);
                 testCase.verifyEqual(rm{end}.Betay, rp{3}.Betay, RelTol=1.e-10);
-                testCase.verifyEqual(rm{end}.Qpx, rp{3}.Qpx, RelTol=1.e-8);
-                testCase.verifyEqual(rm{end}.Qpy, rp{3}.Qpy, RelTol=1.e-8);
+                testCase.verifyEqual(rm{end}.chromx_arr, rp{3}.chromx_arr, RelTol=1.e-8);
+                testCase.verifyEqual(rm{end}.chromy_arr, rp{3}.chromy_arr, RelTol=1.e-8);
```
lfarv, line 272: I think it's missing a conversion from numpy array to double:

```Matlab
testCase.verifyEqual(rm{end}.chromx_arr, double(rp{3}.chromx_arr), RelTol=1.e-8);
testCase.verifyEqual(rm{end}.chromy_arr, double(rp{3}.chromy_arr), RelTol=1.e-8);
```
Try this !

lcarver: Thank you!

lcarver:
Ready for review

lfarv:
OK for me, but I let @swhite2401 and @carmignani comment !

atintegrators/DeltaQPass.c
```diff
@@ -111,14 +123,16 @@ ExportMode struct elem *trackFunction(const atElem *ElemData,struct elem *Elem,
         R2=atGetOptionalDoubleArray(ElemData,"R2"); check_error();
         T1=atGetOptionalDoubleArray(ElemData,"T1"); check_error();
         T2=atGetOptionalDoubleArray(ElemData,"T2"); check_error();
-        
+        chrom_maxorder=atGetOptionalLong(ElemData,"chrom_maxorder", 5); check_error();
```
carmignani, line 126: The default value should be 1, not 5. Otherwise if you give only Qpx and Qpy the pass method does a loop on chromx_arr and you can get a segmentation fault or similar error.

atmat/lattice/atsimplering.m
```diff
@@ -115,9 +128,9 @@
 nonlin_elem=atbaselem('NonLinear','DeltaQPass','Class','NonLinear',...
     'Betax',betax,'Betay',betay,...
     'Alphax',alphax,'Alphay',alphay,...
-    'Qpx',qpx,'Qpy',qpy,...
+    'chromx_arr',chromx_arr,'chromy_arr',chromy_arr,...
```
carmignani, line 131: You should add also the field chrom_maxorder in the nonlinear element.

atmat/lattice/atfastring.m
```diff
@@ -76,13 +76,13 @@
...
     'Alphax',lindata.alpha(1),'Alphay',lindata.alpha(2),...
-    'Qpx',xsi(1),'Qpy',xsi(2),...
+    'chromx_arr',[xsi(1)],'chromy_arr',[xsi(2)],...
     'A1',r1(1),'A2',r1(2),'A3',r1(4),...
     'T1',-orbit4{end},'T2',orbit4{end});
 nonlin_elemrad=atbaselem('NonLinear','DeltaQPass',...
     'Betax',lindata.beta(1),'Betay',lindata.beta(2),...
     'Alphax',lindata.alpha(1),'Alphay',lindata.alpha(2),...
-    'Qpx',xsi(1),'Qpy',xsi(2),...
+    'chromx_arr',[xsi(1)],'chromy_arr',[xsi(2)],...
```
carmignani, line 85: Same here: you should add chrom_maxorder=1 in the nonlinear element.

carmignani (changes requested):

atintegrators/DeltaQPass.c
```diff
@@ -161,11 +176,12 @@ void mexFunction(	int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
         R2=atGetOptionalDoubleArray(ElemData,"R2"); check_error();
         T1=atGetOptionalDoubleArray(ElemData,"T1"); check_error();
         T2=atGetOptionalDoubleArray(ElemData,"T2"); check_error();
+        chrom_maxorder=atGetOptionalLong(ElemData,"chrom_maxorder", 5); check_error();
```
carmignani, line 179: Same comment on the default value also here in the mexFunction. It should be 1, not 5.

lcarver: done

atmat/lattice/atfastring.m
lcarver: done

atmat/lattice/atsimplering.m
lcarver: done

atintegrators/DeltaQPass.c
lcarver: done

carmignani (approved):

PR #916, opened 2025-03-26
Main frequency as min frequency

ZeusMarti:
This solves an issue with lattices were the first cavity along the ring is not the main cavity.


lfarv:
Assuming that the 1st cavity was the main one was indeed a bug. But #917 gives a more general solution by scaling all cavities by the same amount.

PR #920, opened 2025-04-07
Gpu tracking

JeanLucPons:
Hello,
Here is the first release of AT GPU tracking.

Installation note can be found [here](https://github.com/atcollab/at/blob/gpu_tracking/docs/p/howto/multiprocessing.rst#gpu-tracking). In this very first release, GPU tracking is implemented only for acceptance calculations and requires `at.GridMode.CARTESIAN`. To get optimal performance, the particle number should be a multiple of gpu_core_number * 64.
GPU core number (this is not the number of cuda core) can be obtained using `at.tracking.gpu_info()` and `at.tracking.gpu_core_count()`. There is no constraint on the number of turns. Multi GPU tracking is not yet supported.

Example:

Momemtum aceptance:
```
b, *_ = ring.get_momentum_acceptance(1.0/10240.0,0.1,
refpts=[0,10,20],
grid_mode=at.GridMode.CARTESIAN,
nturns=1000,
use_gpu=True,
verbose=True)
```

For a dynamic aperture:
```
amplitudes=[-0.013, 0.005]
planes = ['x', 'y']
npoints = [32, 64]
nturns = 1000
b2, s2, g2 = ring.get_acceptance(planes,npoints,amplitudes,
grid_mode=at.GridMode.CARTESIAN,
nturns=nturns,
use_gpu=True,
verbose=True,
offset=np.array([1e-7,0,1e-7,0,0,0]))
```

Track function can also be called:
```
rout, *_ = ring.track(rin, nturns=nturns, use_gpu=True)
```

Thanks to test it.


lfarv:
> GPU tracking is implemented only for acceptance calculations

Simple tracking with `Lattice.track()` works as well !

lfarv:
@JeanLucPons 
> Multi GPU tracking is not yet supported.

Does this mean that the `gpu_pool` argument must contain only a single integer ?

JeanLucPons:
> Does this mean that the `gpu_pool` argument must contain only a single integer ?

For the time being, yes. If you try a sequence of GPU id, it will failed with an exception.
But it is planned to have multi-gpu support.

With Simon, we are also wondering how we can use an enum type for the parrelelisation method and keep backward compatibility. i.e `use_mp = True` and `mp_method = at.MPType.CPU` or `at.MPType.GPU` with a default to `at.MPTuype.CPU`.

JeanLucPons:
Just few words on the GPU accuracy and difference you can expect.
GPUs do not implement `IEEE 754 extended-precision` as CPUs generaly do.
So when tracking in non-linear area, after numerous number of turns, you can experience significant variation as shown below.
Lattice file (EBS) used can be found [here](https://www.esrf.eu/computing/cs/tango/data/betamodel.mat)

image


```python
import at
import numpy as np

nturn = 9999
ring = at.load_lattice("at_test/lattice/betamodel.mat",mat_key="betamodel")
rIdx = [i for i in range(0,len(ring))]
gpu_info = at.tracking.gpu_info()

# CPU tracking
rin = np.array([0.003,0,0.002,0,0,0])
ring.track(rin, nturns=nturn, in_place=True)
rout,*_ = ring.track(rin, nturns=1, refpts=rIdx)

# GPU tracking
rin2 = np.array([0.003,0,0.002,0,0,0])
ring.track(rin2, nturns=nturn, use_gpu=True, gpu_pool=[0], in_place=True, verbose=True)
rout2,*_ = ring.track(rin2, nturns=1, use_gpu=True, gpu_pool=[0], refpts=rIdx, verbose=True)

plt.ylabel("x [m] @turn " + str(nturn+1))
plt.xlabel("s [m]")
plt.plot(ring.get_s_pos(rIdx),rout[0,0,:,0],label="CPU")
plt.plot(ring.get_s_pos(rIdx),rout2[0,0,:,0],linestyle='dashed',label="GPU " + gpu_info[0][0])

plt.legend()
plt.show()
```

lfarv:
> `use_mp=True, mp_method=at.MPType.GPU `rather than` use_gpu=True`

I'm not sure it makes things easier: it requires to specify 2 keywords for GPU computing instead of a single one, and needs to memorise the enum type…

What about a single keyword `use_mp` with values;
- `False`
- `True` (for compatibility, equivalent to `at.MPType.CPU`)
- `at.MPType.CPU`
- `at.MPType.GPU`

But the present situation with 2 bool args is ok for me.

swhite2401:
> > `use_mp=True, mp_method=at.MPType.GPU `rather than` use_gpu=True`
> 
> I'm not sure it makes things easier: it requires to specify 2 keywords for GPU computing instead of a single one, and needs to memorise the enum type…
> 
> What about a single keyword `use_mp` with values;
> 
>     * `False`
> 
>     * `True` (for compatibility, equivalent to `at.MPType.CPU`)
> 
>     * `at.MPType.CPU`
> 
>     * `at.MPType.GPU`
> 
> 
> But the present situation with 2 bool args is ok for me.

Hello @lfarv, two booleans are not nice because it requires to implement exclusive conditions. Having a single keyword as you propose is fine for me

JeanLucPons:
Unfortunately it does not work simply using enum. An alternative could be to use simple integer constants.

```python
from enum import Enum

class MPMode(Enum):
    NONE = 0  #: No multiprocessing
    CPU = 1  #: CPU multiprocessing
    GPU = 2  #: GPU multiprocessing

val = True

if val == MPMode.CPU:
    print("ok")
else:
    print("not ok")

if val == 1:
    print("ok")
else:
    print("not ok")
```

```
not ok
ok
```

lfarv:
I does work simply with the proposed argument:

```python

class MPMode(Enum):
    CPU = 0  # CPU multiprocessing
    GPU = 1  # GPU multiprocessing

_mp_table = {
    False : (False, False),
    True: (True, False),
    MPMode.CPU: (True, False),
    MPMode.GPU: (False, True),
}

def lattice_track(..., use_mp=False,...):
   # Convert the single argument to the usual pair:
    use_mp, use_gpu = _mp_table[use_mp]
    # All the rest is unchanged

```

```python
use_mp, use_gpu = _mp_table[True]
print(use_mp, use_gpu)
True False
```

```python
use_mp, use_gpu = _mp_table[MPMode.GPU]
print(use_mp, use_gpu)
False True
```
And so on…

pyat/at/acceptance/acceptance.py
```diff
@@ -120,19 +121,22 @@ def get_acceptance(
...
 
+    # For backward compatibility (use_mp can be a boolean)
+    if isinstance(use_mp,bool) and use_mp:
+        use_mp = MPMode.CPU
+
     if verbose:
         nproc = multiprocessing.cpu_count()
         print(f"\n{nproc} cpu found for acceptance calculation")
-        if use_mp:
+        if use_mp==MPMode.CPU:
```
lfarv, line 131: better written as:
```python
if use_mp is MPMode.CPU:
```

pyat/at/tracking/track.py
```diff
@@ -314,13 +338,13 @@ def lattice_track(
...
-    use_mp = kwargs.pop("use_mp", False)
     start_method = kwargs.pop("start_method", None)
     pool_size = kwargs.pop("pool_size", None)
     if use_mp:
         kwargs.update({"pool_size": pool_size, "start_method": start_method})
         rout = _plattice_pass(lattice, r_in, nturns=nturns, refpts=refpts, **kwargs)
     else:
+        kwargs["use_gpu"] = use_gpu
         rout = _lattice_pass(
             lattice, r_in, nturns=nturns, refpts=refpts, no_varelem=False, **kwargs
```
lfarv, line 349: No need to modify `kwargs`:
```python
    else:
        rout = _lattice_pass(
            lattice, r_in, nturns=nturns, refpts=refpts, no_varelem=False, use_gpu=use_gpu, **kwargs
        )

File: pyat/at/acceptance/acceptance.py [hunk 3]
```diff
@@ -115,14 +121,23 @@ def get_acceptance(
     if start_method is not None:
         kwargs["start_method"] = start_method
 
+    # For backward compatibility (use_mp can be a boolean)
+    if isinstance(use_mp,bool) and use_mp:
+        use_mp = MPMode.CPU
```
lfarv, line 126: And a last one:
```python
if use_mp is True:
    use_mp = MPMode.CPU
```

Different from `if use_mp:` which would be `True` for all except `False`.

lfarv:
@JeanLucPons : That looks ok for me, ready for merging this 1st version of GPU tracking!

Finally, if you allow me, I'd like to run `acceptance.py` and `track.py` through the [black](https://black.readthedocs.io/en/stable/) code formatter: that's what we use to get a uniform code style (no code change).

JeanLucPons:
Hello @lfarv,
So ok for me. Feel free to run black.

lfarv (approved):
Great addition to AT, thanks @JeanLucPons !
Any advice, @swhite2401 and @oscarxblanco ?

lcarver:
Hello @JeanLucPons ,

The only question I have, is what have you NOT implemented in the GPU tracking? What lattices will not work?

Very impressive work!

oscarxblanco:
Dear @JeanLucPons ,
I get the following Warning. Is this a problem ?
```
$ python
Python 3.9.16 (main, Dec  8 2022, 00:00:00) 
[GCC 11.3.1 20221121 (Red Hat 11.3.1-4)] on linux
>>> import at
>>> at.tracking.gpu_info()
Warning, Platform AMD Accelerated Parallel Processing: clGetDeviceIDs: Device not found.
[['NVIDIA L4', '8.9', 58, 'NVIDIA CUDA OpenCL 3.0 CUDA 12.1.68'], ['NVIDIA L4', '8.9', 58, 'NVIDIA CUDA OpenCL 3.0 CUDA 12.1.68'], ['NVIDIA L4', '8.9', 58, 'NVIDIA CUDA OpenCL 3.0 CUDA 12.1.68'], ['NVIDIA L4', '8.9', 58, 'NVIDIA CUDA OpenCL 3.0 CUDA 12.1.68'], ['pthread-AMD EPYC 7502P 32-Core Processor', '0.0', 64, 'Portable Computing Language OpenCL 3.0 PoCL 3.1  Linux, Release, RELOC, SPIR, LLVM 14.0.6, SLEEF, DISTRO, POCL_DEBUG']]
>>> 

```

oscarxblanco:
Dear @JeanLucPons ,
`gpu_core_count()` seems to fail.
```python
>>> at.tracking.gpu_core_count()
TypeError: gpu_core_count() missing 1 required positional argument: 'gpuPool'

```

oscarxblanco:
Dear @JeanLucPons 
the `use_gpu`  you show above (and I copy here below) seems to be missing from the docstring message.
```python
rout, *_ = ring.track(rin, nturns=nturns, use_gpu=True)
```

oscarxblanco:
@JeanLucPons ,
I now have the following error in a lattice with apertures.
```python
>>> ring.track(z,nturns=1000,use_mp=at.MPMode.GPU, gpu_pool=[0])
AttributeError: 'Marker' object has no attribute 'RApertures'

The above exception was the direct cause of the following exception:

  File "/home/oblanco/Documents/public/progs/pyenv/pygpu2/lib64/python3.9/site-packages/at/tracking/track.py", line 349, in lattice_track
    rout = _lattice_pass(
  File "/home/oblanco/Documents/public/progs/pyenv/pygpu2/lib64/python3.9/site-packages/at/tracking/utils.py", line 106, in wrapper
    return func(lattice, r_in, *args, **kwargs)
  File "/home/oblanco/Documents/public/progs/pyenv/pygpu2/lib64/python3.9/site-packages/at/tracking/track.py", line 120, in _lattice_pass
    return _gpupass(lattice, r_in, nturns, refpts=refs, **kwargs)
SystemError: <built-in function gpupass> returned a result with an error set

```
Tracking the particle over the CPU gives no errors, and the particle survives all turns.
```
>>> _,_,lossesdata=ring.track(z,nturns=1000,losses=True)
>>> lossesdata['loss_map']['islost']
array([False])
```

JeanLucPons:
> @JeanLucPons , I now have the following error in a lattice with apertures.
> 
> ```python
> >>> ring.track(z,nturns=1000,use_mp=at.MPMode.GPU, gpu_pool=[0])
> AttributeError: 'Marker' object has no attribute 'RApertures'

Ah, this seems to be a bug (RApertures is supposed to be optional).
Could you send me your lattice, thanks ?

JeanLucPons:
> @JeanLucPons, I just sent the lattice to you by email.
Thanks i well received the lattice and I'm able to reproduce the issue.

JeanLucPons:
@oscarxblanco, I did few quick tests with this ALBA2 lattice and , at first glance, it seems good for me.

lfarv (approved):
Let's go on.

PR #921, opened 2025-04-10
Implementation of a new function to rotate an element

SebastienJoly:
Hello,

I tried to solve the issue with the dipole (or any element) rotation in pyAT. From what I have understood from the past issues on github and some tests I have done, the rotation transformation only changed the R1 and R2 attributes and left T1 and T2 untouched. This is true for a tilt/roll angle but not for a pitch or yaw angle.

I tried to reimplement the rotation function based on the Euler angles (https://en.wikipedia.org/wiki/Euler_angles) and this paper from ALS (https://www.sciencedirect.com/science/article/pii/S0168900222007793) on which is based the SC/pySC equivalent function. I put as many comments as possible to guide the user in understanding how the rotations are applied.

Now, the current function still returns non commutative rotations but agree perfectly with pySC. I used the MLS as an example where I rotated a quadrupole and observed the resulting closed-orbit.

Roll angle:
roll

Yaw angle:
yaw

Pitch angle:
pitch

We can clearly see the differences w.r.t the old implementation. I also tried it with a dipole and it agreed with the pySC curves. In that case, I also account for the bending angle in the R1, R2, T1 and T2 calculations.

I think it would be relevant to move the rotation functions outside of the utils.py module and create a rotation.py one. What do you think?

@TeresiaOlsson and I will compare these results with AT and SC next.

### DIFF
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -483,6 +483,15 @@ class Element:
     @dy.setter
     def dy(self, value: float) -> None:
         self._setshift(value, 2)
+        
+    @property
+    def dz(self) -> float:
+        """Longitudinal element shift"""
+        return -self._getshift(5)
+
+    @dz.setter
+    def dz(self, value: float) -> None:
+        self._setshift(-value, 5)
 
     @property
     def tilt(self) -> float:
--- a/pyat/at/lattice/elements.py
+++ b/pyat/at/lattice/elements.py
@@ -483,15 +483,6 @@ class Element:
     @dy.setter
     def dy(self, value: float) -> None:
         self._setshift(value, 2)
-        
-    @property
-    def dz(self) -> float:
-        """Longitudinal element shift"""
-        return -self._getshift(5)
-
-    @dz.setter
-    def dz(self, value: float) -> None:
-        self._setshift(-value, 5)
 
     @property
     def tilt(self) -> float:
--- a/pyat/at/lattice/transformation.py
+++ b/pyat/at/lattice/transformation.py
@@ -1,6 +1,5 @@
-import numpy
-#import re
-from typing import Union, Tuple, List, Type
+from __future__ import annotations
+import numpy as np
 from .elements import Element
 
 def _rotation(rotations):
@@ -17,19 +16,19 @@ def _rotation(rotations):
     gamma: Rotation about the Z-axis (roll/tilt).
     """
     alpha, beta, gamma = rotations  # ZYX intrinsic rotations (pitch, yaw, tilt)
-    R_x = numpy.array([
+    R_x = np.array([
         [1, 0, 0],
-        [0, numpy.cos(alpha), -numpy.sin(alpha)],
-        [0, numpy.sin(alpha), numpy.cos(alpha)]
+        [0, np.cos(alpha), -np.sin(alpha)],
+        [0, np.sin(alpha), np.cos(alpha)]
         ])
-    R_y = numpy.array([
-        [numpy.cos(beta), 0, numpy.sin(beta)],
+    R_y = np.array([
+        [np.cos(beta), 0, np.sin(beta)],
         [0, 1, 0],
-        [-numpy.sin(beta), 0, numpy.cos(beta)]
+        [-np.sin(beta), 0, np.cos(beta)]
         ])
-    R_z = numpy.array([
-        [numpy.cos(gamma), -numpy.sin(gamma), 0],
-        [numpy.sin(gamma), numpy.cos(gamma), 0],
+    R_z = np.array([
+        [np.cos(gamma), -np.sin(gamma), 0],
+        [np.sin(gamma), np.cos(gamma), 0],
         [0, 0, 1]
         ])
     return R_x @ R_y @ R_z
@@ -47,11 +46,11 @@ def _translation_vector(ld, r3d, offsets, X_axis, Y_axis):
     r3d: 3D rotation matrix
     offsets: 3D offsets [m]
     """
-    tD0 = numpy.array([
+    tD0 = np.array([
         -offsets @ X_axis, 0, -offsets @ Y_axis, 
          0, 0, 0
         ])
-    T0 = numpy.array([
+    T0 = np.array([
         ld * r3d[2, 0] / r3d[2, 2], r3d[2, 0], ld * r3d[2, 1] / r3d[2, 2],
         r3d[2, 1], 0, ld / r3d[2, 2]
     ])
@@ -69,7 +68,7 @@ def _r_matrix(ld, r3d):
     r3d: 3D rotation matrix
     Corresponds to Eq. (9)
     """
-    return numpy.array([
+    return np.array([
         [
             r3d[1, 1] / r3d[2, 2],
             ld * r3d[1, 1] / r3d[2, 2] ** 2,
@@ -174,11 +173,11 @@ def transform_elem(elem: Element, midpoint: str = "center",
     RB_half = _rotation([0, -elem_bending_angle / 2, 0]) # Eq. (12)
     
     # Define transverse offsets (element translation)
-    offsets = numpy.array([dx, dy, 0.])
+    offsets = np.array([dx, dy, 0.])
     
-    x_axis = numpy.array([1, 0, 0])
-    y_axis = numpy.array([0, 1, 0])
-    z_axis = numpy.array([0, 0, 1])
+    x_axis = np.array([1, 0, 0])
+    y_axis = np.array([0, 1, 0])
+    z_axis = np.array([0, 0, 1])
     
     # Extract current transformations if relative=True
     tilt0, pitch0, yaw0 = 0.0, 0.0, 0.0
@@ -193,9 +192,9 @@ def transform_elem(elem: Element, midpoint: str = "center",
                                  "'center' or 'entrance'.")
 
             # Reverse-engineer current angles from r3d
-            tilt0 = numpy.arctan2(-r3d[0, 1], r3d[0, 0])
-            yaw0 = numpy.arcsin(r3d[0,2])
-            pitch0 = numpy.arctan2(-r3d[1, 2], r3d[2, 2])
+            tilt0 = np.arctan2(-r3d[0, 1], r3d[0, 0])
+            yaw0 = np.arcsin(r3d[0,2])
+            pitch0 = np.arctan2(-r3d[1, 2], r3d[2, 2])
 
     # Apply new rotations (XYZ intrinsic order)
     tilt_total = tilt0 + tilt
@@ -209,9 +208,9 @@ def transform_elem(elem: Element, midpoint: str = "center",
         
         if elem_bending_angle:    
             Rc = elem_length / elem_bending_angle
-            OO0 =  Rc * numpy.sin(elem_bending_angle / 2) * \
+            OO0 =  Rc * np.sin(elem_bending_angle / 2) * \
                 RB_half @ z_axis # Eq. (34)
-            P0P = -Rc * numpy.sin(elem_bending_angle / 2) * \
+            P0P = -Rc * np.sin(elem_bending_angle / 2) * \
                 r3d_entrance @ RB_half @ z_axis # Eq. (36)
         else:
             OO0 =  elem_length / 2 * z_axis # Eq. (34)
@@ -236,7 +235,7 @@ def transform_elem(elem: Element, midpoint: str = "center",
     ld_entrance = Z_axis @ OP # Eq. (33)
             
     R1 = _r_matrix(ld_entrance, r3d_entrance)
-    T1 = numpy.linalg.inv(R1) @ _translation_vector(
+    T1 = np.linalg.inv(R1) @ _translation_vector(
         ld_entrance, r3d_entrance, OP, X_axis, Y_axis)
     
     # R2, T2
@@ -249,10 +248,10 @@ def transform_elem(elem: Element, midpoint: str = "center",
     
     if elem_bending_angle:
         Rc = elem_length / elem_bending_angle
-        OPp = numpy.array(
-            [Rc * (numpy.cos(elem_bending_angle) - 1), 
+        OPp = np.array(
+            [Rc * (np.cos(elem_bending_angle) - 1), 
              0, 
-             elem_length * numpy.sin(elem_bending_angle) / elem_bending_angle]
+             elem_length * np.sin(elem_bending_angle) / elem_bending_angle]
             ) # Eq. (24)
     else:
         OPp = elem_length * z_axis # Eq. (24)
--- a/pyat/at/lattice/transformation.py
+++ b/pyat/at/lattice/transformation.py
@@ -1,0 +1,280 @@
+import numpy
+#import re
+from typing import Union, Tuple, List, Type
+from .elements import Element
+
+def _rotation(rotations):
+    """
+    The implementation follows the one described in:
+    https://doi.org/10.1016/j.nima.2022.167487
+    All the comments featuring 'Eq' points to the paper's equations.
+    
+    3D rotation matrix (extrinsic rotation) using the Tait–Bryan angles convention
+    For more details, refer to https://en.wikipedia.org/wiki/Euler_angles
+    Corresponds to Eq. (3)
+    alpha: Rotation about the X-axis (pitch).
+    beta: Rotation about the Y-axis (yaw).
+    gamma: Rotation about the Z-axis (roll/tilt).
+    """
+    alpha, beta, gamma = rotations  # ZYX intrinsic rotations (pitch, yaw, tilt)
+    R_x = numpy.array([
+        [1, 0, 0],
+        [0, numpy.cos(alpha), -numpy.sin(alpha)],
+        [0, numpy.sin(alpha), numpy.cos(alpha)]
+        ])
+    R_y = numpy.array([
+        [numpy.cos(beta), 0, numpy.sin(beta)],
+        [0, 1, 0],
+        [-numpy.sin(beta), 0, numpy.cos(beta)]
+        ])
+    R_z = numpy.array([
+        [numpy.cos(gamma), -numpy.sin(gamma), 0],
+        [numpy.sin(gamma), numpy.cos(gamma), 0],
+        [0, 0, 1]
+        ])
+    return R_x @ R_y @ R_z
+
+def _translation_vector(ld, r3d, offsets):
+    """
+    The implementation follows the one described in:
+    https://doi.org/10.1016/j.nima.2022.167487
+    All the comments featuring 'Eq' points to the paper's equations.
+    
+    Translation vector resulting from the joint effect of a longitudinal
+    displacement (in the rotated frame), 3D offsets and the 3D rotation matrix.
+    Corresponds to Eqs. (8-11)
+    ld: Longitudinal displacement [m]
+    r3d: 3D rotation matrix
+    offsets: 3D offsets [m]
+    """
+    X_axis = numpy.dot(r3d, numpy.array([1, 0, 0]))
+    Y_axis = numpy.dot(r3d, numpy.array([0, 1, 0]))
+    tD0 = numpy.array([
+        -numpy.dot(offsets, X_axis), 0, -numpy.dot(offsets, Y_axis), 
+         0, 0, 0
+        ])
+    T0 = numpy.array([
+        ld * r3d[2, 0] / r3d[2, 2], r3d[2, 0], ld * r3d[2, 1] / r3d[2, 2],
+        r3d[2, 1], 0, ld / r3d[2, 2]
+    ])
+    return T0 + tD0
+
+def _r_matrix(ld, r3d):
+    """
+    The implementation follows the one described in:
+    https://doi.org/10.1016/j.nima.2022.167487
+    All the comments featuring 'Eq' points to the paper's equations.
+    
+    Rotation matrix operator (R1, R2).
+    Can take into account the effect of a longitudinal displacement.
+    ld: Longitudinal displacement [m]
+    r3d: 3D rotation matrix
+    Corresponds to Eq. (9)
+    """
+    return numpy.array([
+        [
+            r3d[1, 1] / r3d[2, 2],
+            ld * r3d[1, 1] / r3d[2, 2] ** 2,
+            -r3d[0, 1] / r3d[2, 2],
+            -ld * r3d[0, 1] / r3d[2, 2] ** 2,
+            0,
+            0,
+        ],
+        [
+            0,
+            r3d[0, 0],
+            0,
+            r3d[1, 0],
+            r3d[2, 0],
+            0,
+        ],
+        [
+            -r3d[1, 0] / r3d[2, 2],
+            -ld * r3d[1, 0] / r3d[2, 2] ** 2,
+            r3d[0, 0] / r3d[2, 2],
+            ld * r3d[0, 0] / r3d[2, 2] ** 2,
+            0,
+            0,
+        ],
+        [
+            0,
+            r3d[0, 1],
+            0,
+            r3d[1, 1],
+            r3d[2, 1],
+            0,
+        ],
+        [
+            0,
+            0,
+            0,
+            0,
+            1,
+            0,
+        ],
+        [
+            -r3d[0, 2] / r3d[2, 2],
+            -ld * r3d[0, 2] / r3d[2, 2] ** 2,
+            -r3d[1, 2] / r3d[2, 2],
+            -ld * r3d[1, 2] / r3d[2, 2] ** 2,
+            0,
+            1,
+        ],
+    ])
+    
+def transform_elem(elem: Element, midpoint: str = "center",
+                   dx: float = 0.0, dy: float = 0.0,                       
+                   tilt: float = 0.0, pitch: float = 0.0, yaw: float = 0.0,
+                   relative: bool = False) -> None:
+    r"""Set the tilt, pitch and yaw angle of an :py:class:`.Element`.
+    The tilt is a rotation around the *s*-axis, the pitch is a
+    rotation around the *x*-axis and the yaw is a rotation around
+    the *y*-axis.
+
+    A positive angle represents a clockwise rotation when
+    looking in the direction of the rotation axis.
+
+    The transformations are not all commmutative. The rotations are applied in 
+    the order *Z* -> *Y* -> *X* (tilt -> yaw -> pitch). The element is rotated 
+    around its mid-point. The mid-point can either be the element entrance 
+    (entry face of the downstream drift element) or center (axis joining the 
+    entry and exit points of the element).
+
+    If *relative* is :py:obj:`True`, the previous angles are rebuilt from the 
+    *r3d* matrix and incremented by the input arguments.
+    *relative* only allows to add the previous angles, not the transverse 
+    shifts.
+    The shift is always conserved regardless of the value of *relative*.
+    
+    pyAT describes the ultra-relativistic beam dynamics in 6D phase space 
+    coordinates, which differ from 3D spatial angles in an expansion with 
+    respect to the energy to first order by a factor (1 + $\\delta$) , where 
+    $\\delta$ is the relative energy offset. In general this introduces a small 
+    angle error, but could create an undesired effect for large energy offsets.
+    
+    The implementation follows the one described in:
+    https://doi.org/10.1016/j.nima.2022.167487
+    All the comments featuring 'Eq' points to the paper's equations.
+
+    Parameters:
+        elem:           Element to be tilted
+        midpoint:       Midpoint reference (entrance/center)
+        dx:             Horizontal shift [m]
+        dy:             Vertical shift [m]
+        tilt:           Tilt angle [rad]
+        pitch:          Pitch angle [rad]
+        yaw:            Yaw angle [rad]
+        relative:       If :py:obj:`True`, the rotation is added to the
+          previous one
+    """
+
+    elem_length = getattr(elem, "Length", 0)
+    elem_bending_angle = getattr(elem, 'BendingAngle', 0)
+    
+    # Rotated transfer matrix linked to a bending angle
+    RB = _rotation([0, -elem_bending_angle, 0]) # Eq. (12)
+    RB_half = _rotation([0, -elem_bending_angle / 2, 0]) # Eq. (12)
+    
+    # Define transverse offsets (element translation)
+    offsets = numpy.array([dx, dy, 0.])
+    
+    x_axis = numpy.array([1, 0, 0])
+    y_axis = numpy.array([0, 1, 0])
+    z_axis = numpy.array([0, 0, 1])
+    
+    # Extract current transformations if relative=True
+    tilt0, pitch0, yaw0 = 0.0, 0.0, 0.0
+    t10, t20 = numpy.zeros(6), numpy.zeros(6)
+    if relative:
+        if hasattr(elem, '_r3d'):
+            if midpoint == 'center':
+                r3d = numpy.dot(RB_half.T, numpy.dot(elem._r3d, RB_half))
+            elif midpoint == 'entrance':     
+                r3d = elem._r3d
+            else:
+                raise ValueError("Unsupported midpoint, please choose either "
+                                 "'center' or 'entrance'.")
+
+            # Reverse-engineer current angles from r3d
+            tilt0 = numpy.arctan2(-r3d[0, 1], r3d[0, 0])
+            yaw0 = numpy.arcsin(r3d[0,2])
+            pitch0 = numpy.arctan2(-r3d[1, 2], r3d[2, 2])
+
+    # Apply new rotations (XYZ intrinsic order)
+    tilt_total = tilt0 + tilt
+    pitch_total = pitch0 + pitch
+    yaw_total = yaw0 + yaw
+    rotations = [pitch_total, yaw_total, tilt_total]  # X, Y, Z convention
+    
+    if midpoint == 'center':
+        # Compute entrance rotation matrix in the rotated frame
+        r3d_entrance = numpy.dot(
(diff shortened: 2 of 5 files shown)


lfarv:
@SebastienJoly : very nice!
- I fully agree with the idea of moving rotation functions out of `utils.py`,
- I do not know what to do with the old implementation: you are keeping both side by side, and the name of the new implementation (`rotate_translate_elem`) is puzzling for me. And the global `set_rotation` function will still use the old implementation. Otherwise we could simply replace the old function by the new one. Or another possibility is to move the old implementation to a `deprecated` module or package: if necessary, it could be used with `from at.deprecated import rotate_elem`,
- I think that there is a problem related to AT's choice of coordinates (also present in the old implementation): T1[1] and T1[3] are used to model rotations around the X or Y axis. But these coordinates are $p_x= x' (1+\delta)$ and not $x'$: so the angular kick $x'$ depends on the momentum. No way to deal with that with a matrix… This should introduce an error on the dispersion generated by rotation errors. Since SC/pySC is based AT for the optics, I don't think comparing the dispersion will show anything. Is there another possible  benchmark?

oscarxblanco:
Dear @SebastienJoly , thank you very much. 
The documentation still misses the reference point around which the transformation is applied for curved reference systems.
Or, said in another way, does a dipole rotate with respect to the center ? or the start ? or the end point ? can this be implemented in a general way ? Which point do you currently use in your implementation ?
Also, wrt to this
```
the rotation transformation only changed the R1 and R2 attributes and left T1 and T2 untouched. This is true for a tilt/roll angle but not for a pitch or yaw angle
```
a dipole roll will also produce translations on the input and output reference systems T1 and T2.

SebastienJoly:
Thanks for the feedback @lfarv!

I will create a new file (transformation.py) to move all the new functions then. In addition, I have kept the old function as a quick way to benchmark the new implementation. Of course, the idea was to replace the old function and not to keep both in pyAT. We can also keep it in a deprecated library if needed.

Finally, thanks for bringing this to my attention. I did not realize this but you are absolutely right.

>     * I think that there is a problem related to AT's choice of coordinates (also present in the old implementation): T1[1] and T1[3] are used to model rotations around the X or Y axis. But these coordinates are px = x'(1+δ) and not x': so the angular kick x' depends on the momentum. No way to deal with that with a matrix… This should introduce an error on the dispersion generated by rotation errors. Since SC/pySC is based AT for the optics, I don't think comparing the dispersion will show anything. Is there another possible  benchmark?

Could we use Xsuite/elegant to cross check what we have and see the magnitude of the dispersion error due to the pyAT choice of coordinates?

SebastienJoly:
Thanks for the comments as well @oscarxblanco

> Dear @SebastienJoly , thank you very much. The documentation still misses the reference point around which the transformation is applied for curved reference systems. Or, said in another way, does a dipole rotate with respect to the center ? or the start ? or the end point ? can this be implemented in a general way ? Which point do you currently use in your implementation ?

The rotation is applied with respect to the center of the element. Based on my current understanding of the problem, while the rotation could be applied with respect to a face, we would need to change the implementation of the R1,R2 (and maybe T1,T2) elements. Is this a feature you would like to have? And if so, to study which problem?

> 
> ```
> the rotation transformation only changed the R1 and R2 attributes and left T1 and T2 untouched. This is true for a tilt/roll angle but not for a pitch or yaw angle
> ```
> 
> a dipole roll will also produce translations on the input and output reference systems T1 and T2.

I am not so sure. The way I picture this is, for a roll which is a rotation along the Z-axis, the entrance and exit faces of the element are simply rotated. By doing so, no additional displacement is added in any plane. Otherwise, it would mean that the use of skew elements also require T1 and T2 vectors.
On the contrary, a pitch or yaw will shift the entrance and exit faces; therefore, T1 and T2 are required.

lfarv:
@oscarxblanco, @SebastienJoly 

> The rotation is applied with respect to the center of the element

In my understanding, the rotation is around an axis joining the entry and exit points of the bending magnet. This seems to me the easiest description of bending magnet rotation.

SebastienJoly:
Yes, this is actually a better description of the rotation reference point in a dipole or any element.

oscarxblanco:
> > a dipole roll will also produce translations on the input and output reference systems T1 and T2.
> 
> I am not so sure. The way I picture this is, for a roll which is a rotation along the Z-axis, the entrance and exit faces of the element are simply rotated. By doing so, no additional displacement is added in any plane. Otherwise, it would mean that the use of skew elements also require T1 and T2 vectors. On the contrary, a pitch or yaw will shift the entrance and exit faces; therefore, T1 and T2 are required.

For a dipole, due to the curvature, there will be an angle between the magnet face and the plane perpendicular to the sagitta around which you and @lfarv propose to define the roll. You will need T1 and T2 to describe that angle.

lfarv:
@oscarxblanco:
Yes, rotating around the straight line joining entrance and exit requires a rotation at both faces around all 3 axis, which is handled by T1[1], T1[3] and rotation R1 (no translation), and similarly for exit. My point is that T1[1] and t1[3] are not true rotations, because of the dependence on $\delta$.

oscarxblanco:
> @oscarxblanco: Yes, rotating around the straight line joining entrance and exit requires a rotation at both faces around all 3 axis, which is handled by T1[1], T1[3] and rotation R1 (no translation), and similarly for exit. My point is that T1[1] and t1[3] are not true rotations, because of the dependence on δ .

Dear @lfarv , I have no easy solution or suggestion for this, and it is worth knowing. Could this be included as a note on the help message ? For example,
"AT describes the ultra-relativistic beam dynamics in 6D phase space coordinates, which differ from 3D spatial angles in an expansion w.r.t. energy to first order by a factor $(1+\delta)$,  where $\delta$ is the relative energy offset. In general this introduces a small angle error, but could create an undesired effect for large energy offsets."

Also, I think the function `rotate_traslate_element` should be renamed.

SebastienJoly:
I think I am beginning to understand now. Yet, I have no idea how to treat this problem with this implementation, apart from mentioning it in the documentation.
Do you have any comment/suggestion with the new function?
Otherwise, this is my plan:
1. Move the new function to transformation.py and rename it to rotate_elem()
2. Move the previous function to deprecated.py
3. Update the docstrings w.r.t the problem with T1[1] and T1[3] elements

Am I missing something?

oscarxblanco:
@SebastienJoly , there is one more thing on which I have to insist.
The dipole roll requires T1 and T3.

oscarxblanco:
I'm testing your function, @SebastienJoly, I have made the following python script
```python
import at
import numpy
qf = at.Quadrupole('qf',0.1,3)
at.rotate_translate_elem(qf,pitch=500e-6)
print(qf.T1)
print(qf.T2)
```
And I get the following
```python
>>> exec(open('test_sjoly_transformations.py').read())
[0.     0.     0.     0.0005 0.     0.    ]
[ 0.00000000e+00  0.00000000e+00 -5.00000042e-05 -4.99999979e-04
  0.00000000e+00  1.25000013e-08]
```
Which does not match your documentation. This is correct if the transformation is applied at the the element entrance, in which case I expect
```matlab
T1theo = [0 0 0 tan(apitch) 0 0]
T2theor = [0 0 -magLength*tan(apitch) -sin(apitch) 0 magLength*(sec(apitch)-1)]
```
The case you describe in your documentation is a rotation around the mid point, and the theoretical T1 and T2 are:
```matlab
T1theomidpoint=[   0 0 -magLength/2*(sin(apitch)) tan(apitch) 0 magLength/2*(1-cos(apitch))]
T2theomidpoint=[    0 0 -magLength/2*tan(apitch) -sin(apitch) 0 magLength/2*(sec(apitch)-1)]
```
You could do a numerical check in SC matlab using the following lines
```matlab
format long
dx = 0;
dy = 0;
dz = 0;
atilt = 0.0;
apitch = 0.5e-3;
ayaw = 0.0;
ax = apitch;
ay = ayaw;
az = atilt;
magTheta = 0;
magLength = 0.1;
[T1,T2,R1,R2] = SCgetTransformation(dx,dy,dz,ax,ay,az,magTheta,magLength,'refPoint','entrance');
%[T1,T2,R1,R2] = SCgetTransformation(dx,dy,dz,ax,ay,az,magTheta,magLength,'refPoint','center');
T1andT2fromSC=[T1';T2']
R1
R2
```
Could you please clarify ?

lfarv:
@SebastienJoly and @oscarxblanco 

I made another test on a tile of a bending magnet:
```
>>> dip=at.Dipole("ABCD", 1.0, 0.8)
>>> at.rotate_translate_elem(dip, tilt=0.01)
>>> print(dip)
Dipole:
       FamName: ABCD
        Length: 1.0
    PassMethod: BndMPoleSymplectic4Pass
  BendingAngle: 0.8
 EntranceAngle: 0.0
     ExitAngle: 0.0
      MaxOrder: 0
   NumIntSteps: 10
      PolynomA: [0. 0.]
      PolynomB: [0. 0.]
            R1: [[ 0.99995     0.          0.00999983  0.          0.          0.        ]
 [ 0.          0.99995     0.          0.00999983  0.          0.        ]
 [-0.00999983 -0.          0.99995     0.          0.          0.        ]
 [ 0.         -0.00999983  0.          0.99995     0.          0.        ]
 [ 0.          0.          0.          0.          1.          0.        ]
 [-0.         -0.         -0.         -0.          0.          1.        ]]
            R2: [[ 9.99975730e-01  1.35979871e-05 -6.96713024e-03 -9.47412466e-08
   0.00000000e+00  0.00000000e+00]
 [ 0.00000000e+00  9.99975730e-01  0.00000000e+00 -6.96695098e-03
   2.49891318e-05  0.00000000e+00]
 [ 6.96713024e-03  9.47412466e-08  1.00000146e+00  1.35983370e-05
   0.00000000e+00  0.00000000e+00]
 [ 0.00000000e+00  6.96695098e-03  0.00000000e+00  9.99950000e-01
  -7.17344135e-03  0.00000000e+00]
 [ 0.00000000e+00  0.00000000e+00  0.00000000e+00  0.00000000e+00
   1.00000000e+00  0.00000000e+00]
 [-2.49897748e-05 -3.39818883e-10 -7.17362593e-03 -9.75492405e-08
   0.00000000e+00  1.00000000e+00]]
            T1: [0. 0. 0. 0. 0. 0.]
            T2: [ 1.32068842e-05  2.49891318e-05 -3.79120049e-03 -7.17344135e-03
  0.00000000e+00  1.35983172e-05]
             K: 0.0
>>> 
```
Since T1 = 0, we are left with only an S rotation at entrance: this is clearly a **tilt applied to the entrance face**, not on the magnet centre. Consequently, the exit is out of the mid plane, and we get a y translation at the exit (`T2[2]= -3.79e-3`). This is possibly correct, but not what we expected. I think that the rotation around the centre (or middle of the sagitta for a bending magnet) would be much more convenient. Then for the tilt of a bending magnet, I expect something more symmetrical:

`T1[0] = T1[2] = T2[0] = T2[2] = 0`: no displacement
`T1[1] = ± T2[1]`: equal or opposite X rotations (I can't guess the sign)
`T1[3] = ± T2[3]`
`R1` and `R2` opposite rotations

SebastienJoly:
Thanks for pointing out the discrepancy between the midpoint in the docstrings and the one implemented. It turns out I did not fully understand the implementation. You were right, the midpoint was set on the entrance face.
I heavily modified the function to be closer to the one available in SC. We now have the possibility to choose where to set the midpoint (center or entrance) and also to apply transverse shifts (dx/dy).
We still have the possibility to apply relative rotations with both midpoint options. I just needed to store the rotation matrix in the element attributes, is it ok?
However, the transverse shifts cannot be preserved with the relative argument. It becomes too complex when rotations and shifts are present at the same time as the transformations are not commutative.
I spent some time reworking the docstrings and adding many comments pointing out to the equations I used from the paper from Marco Venturini and Thorsten Hellert.
I moved all the rotation functions in transformation.py and called the main function transform_elem.
I reshuffled the utils.py by moving the old rotation function to deprecated.py and changed the set_tilt, set_rotation, tilt_elem to call transform_elem.

The new function passed all the tests you listed. Unfortunately, I do not have access to matlab on my laptop so I cannot make more tests using SC. I will try to install matlab next week.

Is there something else that needs to be fixed or is it ready for review?

lfarv:
```
>>> at.transform_elem(dip, tilt=0.01)
>>> print(dip)
Dipole:
       FamName: ABCD
        Length: 1.0
    PassMethod: BndMPoleSymplectic4Pass
  BendingAngle: 0.8
 EntranceAngle: 0.0
     ExitAngle: 0.0
      MaxOrder: 0
   NumIntSteps: 10
      PolynomA: [0. 0.]
      PolynomB: [0. 0.]
            R1: [[ 9.99957582e-01  3.63825078e-22  9.21052627e-03  3.35116259e-24
   0.00000000e+00  0.00000000e+00]
 [ 0.00000000e+00  9.99957583e-01  0.00000000e+00  9.21045643e-03
  -1.79337528e-05  0.00000000e+00]
 [-9.21052627e-03 -3.35116259e-24  9.99965165e-01  3.63827837e-22
   0.00000000e+00  0.00000000e+00]
 [ 0.00000000e+00 -9.21045643e-03  0.00000000e+00  9.99950000e-01
  -3.89411852e-03  0.00000000e+00]
 [ 0.00000000e+00  0.00000000e+00  0.00000000e+00  0.00000000e+00
   1.00000000e+00  0.00000000e+00]
 [ 1.79338888e-05  6.52507528e-27 -3.89414805e-03 -1.41684882e-24
   0.00000000e+00  1.00000000e+00]]
            R2: [[ 9.99957582e-01  0.00000000e+00 -9.21052627e-03 -0.00000000e+00
   0.00000000e+00  0.00000000e+00]
 [ 0.00000000e+00  9.99957583e-01  0.00000000e+00 -9.21045643e-03
   1.79337528e-05  0.00000000e+00]
 [ 9.21052627e-03  0.00000000e+00  9.99965165e-01  0.00000000e+00
   0.00000000e+00  0.00000000e+00]
 [ 0.00000000e+00  9.21045643e-03  0.00000000e+00  9.99950000e-01
  -3.89411852e-03  0.00000000e+00]
 [ 0.00000000e+00  0.00000000e+00  0.00000000e+00  0.00000000e+00
   1.00000000e+00  0.00000000e+00]
 [-1.79338888e-05 -0.00000000e+00 -3.89414805e-03 -0.00000000e+00
   0.00000000e+00  1.00000000e+00]]
            T1: [-2.77555756e-17  1.79338888e-05 -2.21255173e-19 -3.89414805e-03
  0.00000000e+00 -1.41059322e-37]
            T2: [ 4.80618242e-21  1.79337528e-05 -5.21791960e-19 -3.89411852e-03
  0.00000000e+00  0.00000000e+00]
          _r3d: [[ 9.99957583e-01 -9.21045643e-03 -1.79337528e-05]
 [ 9.21045643e-03  9.99950000e-01  3.89411852e-03]
 [-1.79337528e-05 -3.89411852e-03  9.99992418e-01]]
             K: 0.0
```
This is now conform to my expectations:

`T1[0] ≈ T1[2] ≈ T2[0] ≈ T2[2] ≈ 0`: no displacement
`T1[1] = T2[1]`: equal X rotations (I can't guess the sign)
`T1[3] = T2[3]`; equal Y rotations
`R1` and `R2` opposite rotations

lfarv:
> I just needed to store the rotation matrix in the element attributes, is it ok?

If I understand well, it's just used to revert to the initial angles, for the relative option? No problem for me.

pyat/at/lattice/transformation.py
```diff
@@ -0,0 +1,280 @@
+import numpy
+#import re
+from typing import Union, Tuple, List, Type
```
lfarv, line 3: Adding this line on top will ensure compatibility with old python version (3.7 for us):
```python
```
Also, `Union, Tuple, List, Type` are not used, you can remove this line.

pyat/at/lattice/transformation.py
```diff
@@ -0,0 +1,280 @@
...
+    y_axis = numpy.array([0, 1, 0])
+    z_axis = numpy.array([0, 0, 1])
+    
+    # Extract current transformations if relative=True
+    tilt0, pitch0, yaw0 = 0.0, 0.0, 0.0
+    t10, t20 = numpy.zeros(6), numpy.zeros(6)
+    if relative:
+        if hasattr(elem, '_r3d'):
+            if midpoint == 'center':
+                r3d = numpy.dot(RB_half.T, numpy.dot(elem._r3d, RB_half))
```
lfarv, line 191: What about:
```python
r3d = RB_half.T @ elem._r3d @ RB_half
```
It looks more readable for me…

pyat/at/lattice/transformation.py
```diff
@@ -0,0 +1,280 @@
...
+            
+    R1 = _r_matrix(ld_entrance, r3d_entrance)
+    T1 = numpy.dot(numpy.linalg.inv(R1), _translation_vector(
+        ld_entrance, r3d_entrance, OP)
+        )
+    
+    # R2, T2
+    # XYZ - axes unit - vectors expressed in the xyz coordinate system
+    X_axis = numpy.dot(RB, x_axis)
+    Y_axis = numpy.dot(RB, y_axis)
```
lfarv, line 252: Same thing here

pyat/at/lattice/transformation.py
```diff
@@ -0,0 +1,280 @@
...
+    *r3d* matrix and incremented by the input arguments.
+    *relative* only allows to add the previous angles, not the transverse 
+    shifts.
+    The shift is always conserved regardless of the value of *relative*.
+    
+    pyAT describes the ultra-relativistic beam dynamics in 6D phase space 
+    coordinates, which differ from 3D spatial angles in an expansion with 
+    respect to the energy to first order by a factor (1 + $\\delta$) , where 
+    $\\delta$ is the relative energy offset. In general this introduces a small 
+    angle error, but could create an undesired effect for large energy offsets.
```
lfarv, line 153: It introduces a (hopefully small)  **spurious dispersion**: it's an angle proportional to the off-momentum. I would rephrase it that way…

lfarv:
@SebastienJoly : I get many PEP8 errors. I suggest that you run [flake8](https://flake8.pycqa.org/en/latest/#) to list the errors and [black](https://black.readthedocs.io/en/stable/) to reformat the code,

pyat/at/lattice/transformation.py
```diff
@@ -0,0 +1,280 @@
...
+    the order *Z* -> *Y* -> *X* (tilt -> yaw -> pitch). The element is rotated 
+    around its mid-point. The mid-point can either be the element entrance 
+    (entry face of the downstream drift element) or center (axis joining the 
+    entry and exit points of the element).
+
+    If *relative* is :py:obj:`True`, the previous angles are rebuilt from the 
+    *r3d* matrix and incremented by the input arguments.
+    *relative* only allows to add the previous angles, not the transverse 
+    shifts.
+    The shift is always conserved regardless of the value of *relative*.
```
lfarv, line 147: > The shift is always conserved regardless of the value of *relative*.

This sentence is not clear. Could it be "The shift is always absolute regardless of the value of *relative*"?

lfarv:
@SebastienJoly, @oscarxblanco, @swhite2401 
```
    Parameters:
        ...
        midpoint:       Midpoint reference (entrance/center)
        ...
```
I have a few remarks on this argument, and I would like to have the opinion of other people on that:
- the choice of `midpoint` for the argument name: it could be `reference`, or `origin`,
- rather the strings, we are usually using enumerations for that: this avoids misspellings. See [cavity_access.py](https://at.readthedocs.io/en/latest/p/api/at.physics.energy_loss.html#at.physics.energy_loss.ELossMethod)
- I prefer "centre" to "center" (U.K. English). See in MAD-X sequences "..., REFER=CENTRE,...". However, it seems that there are already some diverging options in AT…

lfarv:
@SebastienJoly: It seems that using R1 and R2 you can also apply longitudinal positioning errors. Isn't it worth adding a `ds` or `dz` argument to allow that?

pyat/at/lattice/transformation.py
```diff
@@ -0,0 +1,280 @@
...
+    # Define transverse offsets (element translation)
+    offsets = numpy.array([dx, dy, 0.])
+    
+    x_axis = numpy.array([1, 0, 0])
+    y_axis = numpy.array([0, 1, 0])
+    z_axis = numpy.array([0, 0, 1])
+    
+    # Extract current transformations if relative=True
+    tilt0, pitch0, yaw0 = 0.0, 0.0, 0.0
+    t10, t20 = numpy.zeros(6), numpy.zeros(6)
```
lfarv, line 187: `t10`, `t20` are never used. Can this be removed?

SebastienJoly: Yes, I forgot to remove them. It was when I tried to allow both relative rotations and translations.

SebastienJoly:
I have modified the functions to account for all your comments @lfarv 
Thanks a lot for taking the time to go through all of it!
I am not so familiar with the Enum library in python, feel free to correct what I have done. And @oscarxblanco @swhite2401, on what convention do we agree for the reference on which we apply the transformation? centre/entrance?

pyat/at/lattice/transformation.py
```diff
@@ -0,0 +1,315 @@
...
+                -r3d[1, 2] / r3d[2, 2],
+                -ld * r3d[1, 2] / r3d[2, 2] ** 2,
+                0,
+                1,
+            ],
+        ]
+    )
+
+
+class reference_point(Enum):
```
lfarv, line 139: Following PEP8, class names should follow the CapWords convention. So here it could be:
```python
class Reference(Enum):
# or
class AtRefer(Enum):
```

oscarxblanco:
> Hello @SebastienJoly
> 
> > it is not possible in pyAT to longitudinally shift an element. I do not know what would be the different to apply such a shift with R1, R2 matrices instead of T1, T2.
> 
> You cannot shift an element longitudinally with T1, T2. But R1 and R2 can represent any transfer matrix. For a straight element, if you put in R1 a small drift matrix and in R2 the same negative drift, it represents a longitudinal translation of the element.
> 
> I made a test with your last version:
> 
> ```
> >>> import at
> >>> qd=at.Quadrupole("QP", 1.0, 0.5)
> >>> at.transform_elem(qd, dz=0.02)
> >>> print(qd)
> Quadrupole:
>        FamName: QP
>         Length: 1.0
>     PassMethod: StrMPoleSymplectic4Pass
>       MaxOrder: 1
>    NumIntSteps: 10
>       PolynomA: [0. 0.]
>       PolynomB: [0.  0.5]
>             R1: [[ 1.    0.02 -0.   -0.    0.    0.  ]
>  [ 0.    1.    0.    0.    0.    0.  ]
>  [-0.   -0.    1.    0.02  0.    0.  ]
>  [ 0.    0.    0.    1.    0.    0.  ]
>  [ 0.    0.    0.    0.    1.    0.  ]
>  [-0.   -0.   -0.   -0.    0.    1.  ]]
>             R2: [[ 1.   -0.02 -0.    0.    0.    0.  ]
>  [ 0.    1.    0.    0.    0.    0.  ]
>  [-0.    0.    1.   -0.02  0.    0.  ]
>  [ 0.    0.    0.    1.    0.    0.  ]
>  [ 0.    0.    0.    0.    1.    0.  ]
>  [-0.    0.   -0.    0.    0.    1.  ]]
>             T1: [0.   0.   0.   0.   0.   0.02]
>             T2: [ 0.    0.    0.    0.    0.   -0.02]
>           _r3d: [[1. 0. 0.]
>  [0. 1. 0.]
>  [0. 0. 1.]]
>              K: 0.5
> ```
> 
> This looks almost correct: R1 and R2 are indeed drift matrices. But I do not understand T1[5] and T2[5]: the 6th coordinate is the path **lengthening with respect to the reference particle**. A additional drift should not contribute to that. It's as if this coordinate was the absolute path length and not the lengthening with respect to the reference. Here it does not matter too much because T1 and T2 cancel out, but have a doubt…

@lfarv , I consider T1[5] and T2[5] correct as they are implemented now by @SebastienJoly. In a straight magnet they will simply cancel out. For an RF cavity it will create a small phase offset when traversing the cavity. For a bending magnet the longitudinal displacement produces a small transverse offset and a shortened longitudinal displacement.

The sign also is correct as dz points in  the opposite direction of the longitudinal coordinate $c\tau$.

None of these will add to the total path lengthening of the particle trajectory.

oscarxblanco:
> I have modified the functions to account for all your comments @lfarv Thanks a lot for taking the time to go through all of it! I am not so familiar with the Enum library in python, feel free to correct what I have done. And @oscarxblanco @swhite2401, on what convention do we agree for the reference on which we apply the transformation? centre/entrance?

@SebastienJoly , centre and entrance seem OK to me. But you are free to choose.

oscarxblanco:
@SebastienJoly ,
I want comment about the implementation of the 'relative' argument.

I think the code could be more sustainable and understandable if is divided in pieces.
1) Transform from dx,dy,dz,ax,ay,az to T1,T2,R1,R2
2) Do the inverse, i.e. transform T1,T2,R1,R2 to dx,dy,dz,ax,ay,az (or in the current case just the angles from your implementation)
3) Now you could use either directly function 1) when doing absolute transformations; alternatively go through function 2) and add the displacements to be input to function 1) for relative displacements.

You gain a bit of flexibility by being able to recover the current state, and modified, instead of adding some values calculated inside the function that you would need to carefully inspect in case of problems.

SebastienJoly:
@oscarxblanco 
Thank you for the idea. I implemented your suggestion, and now I can use the relative argument without the _r3d attribute, both with shifts and rotations! I tested with a reference point set at the entrance or centre of the element and it is working in both scenarios.
By comparing the R1, R2, T1 and T2 of a full rotation/shift or two half rotations/shifts, I retrieve exactly the same results.
Now, the results are almost equivalent when I am comparing a full rotation+shift or two half rotation+shift. I get an error of the order of 1e-6. I suspect a loss of information when computing the matrix inverse.

As the relative argument works with shifts and rotations, I modified the shift_elem() function in utils.py and its tests as the new function allows longitudinal shifts and the effect of a bending angle on R1, R2, T1 and T2. Is it fine with @lfarv @swhite2401 ?

pyat/at/lattice/utils.py
```diff
@@ -1094,25 +1012,28 @@ def set_tilt(ring: Sequence[Element], tilts, relative=False) -> None:
     """
     tilts = numpy.broadcast_to(tilts, (len(ring),))
     for el, tilt in zip(ring, tilts):
-        tilt_elem(el, tilt, relative=relative)
+        transform_elem(el, tilt=tilt, relative=relative)
 
 
-def set_shift(ring: Sequence[Element], dxs, dzs, relative=False) -> None:
+def set_shift(ring: Sequence[Element], dxs, dys, dzs, relative=False) -> None:
```
lfarv, line 1018: This should be
```python
def set_shift(ring: Sequence[Element], dxs, dys, dzs=0.0, *, relative=False) -> None:
```
You have to make `dzs` a keyword argument with default 0.0, otherwise you break the compatibility for cases where it is called with only three positional arguments (`set_shift(ring, dxs, dys)`).

In addition, making `relative` keyword only avoids confusion when using 4 positional arguments.

pyat/test/test_lattice_utils.py
```diff
@@ -199,7 +199,7 @@ def test_tilt_elem():
 def test_shift_elem():
     elem = elt.Drift("Drift", 1.0)
     # Test shift_elem function
-    shift_elem(elem, 1.0, 0.5)
+    shift_elem(elem, 1.0, 0.5, 0.0)
```
lfarv, line 202: `shift_elem` has no entry for  longitudinal displacement. So the 4th argument is interpreted here as `relative=0.0`.

To ensure the backward compatibility, you have to make sure that `test_lattice_utils` still runs unchanged. The only thing you can do is to add new tests for the new features. But the old ones should not be modified.

pyat/at/lattice/transformation.py
```diff
@@ -0,0 +1,390 @@
...
+            # Reverse-engineer current angles from r3d
+            tilt0 = np.arctan2(-r3d[0, 1], r3d[0, 0])
+            yaw0 = np.arcsin(r3d[0, 2])
+            pitch0 = np.arctan2(-r3d[1, 2], r3d[2, 2])
+
+        else:
+            print(
+                "Cannot apply a relative transformation without a prior"
+                "transformation, assume an absolute transformation instead."
+            )
```
lfarv, line 307: Can't we assume that if T1, R1 are absent, it just means that they are 0 ?

For instance:
```python
try:
    oldR1 = elem.R1
except AttributeError:
    oldR1 = np.identity(6)
try:
    oldT1 = elem.T1
except AttributeError:
    oldT1 = np.zero(6)
```
This fixes the case where you have one and not the other: in you implementation, you assume that both are zero .

SebastienJoly: I do not think my implementation allows to have either T1 or R1. It sets both quantities the first time the function is called, even if the transformation results in a zero vector for T1 and an identity matrix for R1. Same thing for T2 and R2.

pyat/test/test_lattice_utils.py
SebastienJoly: The backward compatibility is now preserved with the new functions

SebastienJoly (commented):
I applied the discussed changes, you can find them in my different answers to your comments

lfarv:
Could you put the stars in the argument lists:
```python
def shift_elem(elem: Element, dx: float = 0.0, dy: float = 0.0, dz: float = 0.0, *, relative: bool = False) -> None:
...
def set_shift(ring: Sequence[Element], dxs, dys, dzs=0.0, *, relative=False) -> None:
```
This makes `relative` a keyword-only argument, which avoids confusion with a possible `dz` positional argument.

pyat/at/lattice/transformation.py
lfarv: Now, I don't think the `print` statement is useful: no attributes means the initial state is zero:
You could just remove the` else:` clause.

If you find the information really useful, then it should be a `warning` rather than a `print`.

SebastienJoly: The print was more of a sanity check but I agree it is not crucial
I removed it

lfarv:
Hello @SebastienJoly:

I found another problem: Using `set_rotation` in absolute mode (`relative=False`, the default) resets all the translations to zero. And similarly, `set_shift` resets all rotations to zero. 

Practically, it means that only the "relative" mode should be used to be able to set both translations and rotations. Or one should use directly the `transform_elem` function.

Ideally, all elementary functions like `shift_elem`, `tilt_elem` …, and the `Element` properties like `Element.dx`,… should go through `transform_elem`, which should have new default values: `dx=None,…` instead of `dx=0.0,…`, meaning "keep the value unchanged". Whatever `relative` is, one needs to reverse-engineer the current angles and translations, and then apply the explicitly specified values as relative or absolute modifications. Since this is a new function, we can still modify its default arguments.

This is still a big modification, it turns out that this PR goes much further than the implementation of the right formulas… What is your advice, @SebastienJoly, @swhite2401 and @oscarxblanco ?

Otherwise, we could temporarily take this PR as it is, with a word of caution, and leave the reorganisation for a second step?

pyat/at/lattice/elements.py
```diff
@@ -483,6 +483,15 @@ def dy(self) -> float:
...
         self._setshift(value, 2)
+        
+    @property
+    def dz(self) -> float:
+        """Longitudinal element shift"""
+        return -self._getshift(5)
+
+    @dz.setter
+    def dz(self, value: float) -> None:
+        self._setshift(-value, 5)
```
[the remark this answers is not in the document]
SebastienJoly: You are absolutely right, my mistake.

SebastienJoly:
> Hello @SebastienJoly:
> 
> I found another problem: Using `set_rotation` in absolute mode (`relative=False`, the default) resets all the translations to zero. And similarly, `set_shift` resets all rotations to zero.
> 
> Practically, it means that only the "relative" mode should be used to be able to set both translations and rotations. Or one should use directly the `transform_elem` function.
> 
> Ideally, all elementary functions like `shift_elem`, `tilt_elem` …, and the `Element` properties like `Element.dx`,… should go through `transform_elem`, which should have new default values: `dx=None,…` instead of `dx=0.0,…`, meaning "keep the value unchanged". Whatever `relative` is, one needs to reverse-engineer the current angles and translations, and then apply the explicitly specified values as relative or absolute modifications. Since this is a new function, we can still modify its default arguments.
> 
> This is still a big modification, it turns out that this PR goes much further than the implementation of the right formulas… What is your advice, @SebastienJoly, @swhite2401 and @oscarxblanco ?
> 
> Otherwise, we could temporarily take this PR as it is, with a word of caution, and leave the reorganisation for a second step?

I did not consider this, but it makes sense. Now that the rotations and shifts are handled together, it complicates the overall picture. 
I agree that in the best case scenario, the properties `Element.dx`, `Element.dy`, and `Element.dz` should be handled by `transform_elem`. In addition, there is now the possibility to reverse engineer the offsets and angles with the function. However, it feels like a considerable amount of work... Unfortunately, I will not be able to dedicate much more time on the function until the beginning of summer.
I vote for taking the PR as it is and investing more time in its reorganization later.

lfarv:
@SebastienJoly and @oscarxblanco : I agree with both of you, let's merge this PR and acknowledge the very nice work done be @SebastienJoly. All the rest is a rather simple reorganisation which I can handle in another PR (no need touch all the computations).

lfarv:
@SebastienJoly : the master branch has been updated and there is now a conflict in `utils.py`. I'll solve this and let @oscarxblanco or you approve the final state.

lfarv (approved):

lfarv:
@SebastienJoly, @oscarxblanco : the conflicts in `utils.py` are now solved. You can check in the "modified files" above that your changes to `utils.py` are intact. Feel free to approve (I can't do it) and merge.

oscarxblanco (approved):

T-Nicholls:
[like]  Nicholls, Tobyn (DLSLtd,RAL,LSCI) reacted to your message:
________________________________

PR #923, opened 2025-04-11
Setting the nominal frequency on 6d lattices

lfarv:
This modification follows what was done in #917: in 6d lattice, a momentum change is done by changing the RF frequency. #917 fixes a problem in chromaticity correction, we propose to do the same when setting the off-momentum for all optics computations: `find_orbit`, `linopt`, `get_tune`, `get_chrom`…

`frequency_control` (both Matlab and python) is responsible for setting the nominal frequency and will now act on all cavities for setting the off-momentum.

For this, `atsetcavity` (Matlab) and `set_cavity` (python) are modified for setting the nominal frequency. Their behaviour for other settings is unchanged. The modification is:

**Present behaviour:**
For all selected cavities:
1. Identify the "main" cavities and compute their new frequency for the required off momentum,
2. Set the main cavities to the new frequency and record the ratio new/old frequency,
3. Scale all other selected cavities by the same ratio.

**New behaviour:**
For all selected cavities (no change):
1. get the harmonic number of all cavities: take the cavity's HarmNumber attribute if existing, otherwise round the ratio cavity_frequency/revolution_frequency,
2. Apply h*frev on all selected cavities.

**Benefits:**
- no need to guess which are the "main" cavities (usually taken as the lowest frequency unless specified in Lattice attributes)
- the new setting totally ignores any previous one. If all cavity.HarmNumber are specified, one can start from unspecified (or 0)  frequencies, as in MAD-X

Apart from the new way of computing the nominal frequencies, the behaviour of  `atsetcavity`  and `set_cavity` is unchanged.

### DIFF
--- a/pyat/at/lattice/cavity_access.py
+++ b/pyat/at/lattice/cavity_access.py
@@ -1,19 +1,34 @@
-"""Access to properties of RF cavities"""
-from enum import Enum
-import numpy
-from typing import Optional
-import functools
-from .elements import RFCavity
-from .utils import AtError, make_copy, Refpts, get_bool_index
-from .lattice_object import Lattice
+"""Access to properties of RF cavities
 
-__all__ = ['frequency_control',
-           'get_rf_frequency', 'get_rf_voltage', 'set_rf_voltage',
-           'get_rf_timelag', 'set_rf_timelag', 'set_cavity', 'Frf']
+See :doc:`/p/howto/CavityControl` for a guide on using these functions.
+"""
+
+from __future__ import annotations
+
+import functools
+from enum import Enum
+
+import numpy as np
+
+from .elements import RFCavity
+from .lattice_object import Lattice
+from .utils import AtError, make_copy, Refpts
+
+__all__ = [
+    "frequency_control",
+    "get_rf_frequency",
+    "set_rf_frequency",
+    "get_rf_voltage",
+    "set_rf_voltage",
+    "get_rf_timelag",
+    "set_rf_timelag",
+    "set_cavity",
+    "Frf",
+]
 
 
 def frequency_control(func):
-    r""" Function to be used as decorator for
+    r"""Function to be used as decorator for
     :pycode:`func(ring, *args, **kwargs)`
 
     If :pycode:`ring.is_6d` is :py:obj:`True` **and** *dp*, *dct* or *df*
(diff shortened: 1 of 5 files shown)


pyat/at/lattice/cavity_access.py
```diff
@@ -286,9 +349,18 @@ def apply(ring: Lattice, cavpts, modif):
 Lattice.get_rf_frequency = get_rf_frequency
 Lattice.get_rf_timelag = get_rf_timelag
 Lattice.set_rf_voltage = set_rf_voltage
+Lattice.set_rf_frequency = set_rf_frequency
```
swhite2401, line 352: You don't run into this import issue anymore? How did you fix it? I see that the physics module that you need to compute the revolution frequency is imported later....

pyat/at/lattice/cavity_access.py
```diff
@@ -147,137 +162,185 @@ def get_rf_timelag(ring: Lattice, **kwargs) -> float:
...
-        array:      If False, the value is applied as described for
-                      set_rf_voltage, set_rf_timelag and set_rf_frequency
-                      If True, directly apply the value to the selected
-                      cavities. The value must be broadcastable to the number
-                      of cavities.
-        copy:       If True, returns a shallow copy of ring with new
-                      cavity elements. Otherwise, modify ring in-place
+        ring:           lattice description
+        Frequency:      RF frequency [Hz]
+        Voltage:        Total RF voltage [V]
```
swhite2401, line 267: A word on Frf.NOMINAL would be useful here

pyat/at/lattice/cavity_access.py
```diff
@@ -147,137 +162,185 @@ def get_rf_timelag(ring: Lattice, **kwargs) -> float:
...
+
+
+def set_rf_frequency(
+    ring: Lattice, frequency: float | Frf = Frf.NOMINAL, **kwargs
+) -> Lattice | None:
+    """Set the RF frequency
+
+    Parameters:
+        ring:           Lattice description
+        frequency:      RF frequency [Hz]. Default: nominal frequency.
```
swhite2401, line 186: Same comment as in set_cavity

swhite2401:
This all looks good to me, and it behaves as expected from the documentation. It would gain from a better API documentation.
It it possible to put link to the howto guide in the API doc?

pyat/at/lattice/cavity_access.py
lfarv: I don't import anything from the physics package:
```python
from .elements import RFCavity
from .lattice_object import Lattice
from .utils import AtError, make_copy, Refpts
```
All this is within the lattice package and previously imported in `lattice/__init__.py`.  The physics module is involved at run time (`cell_f = ring.get_revolution_frequency(**kwargs) * ring.periodicity`), as a `Lattice` method, after everything has been imported.

The point is that there must not be an import loop, which is not the case here.

pyat/at/lattice/cavity_access.py
lfarv: done

pyat/at/lattice/cavity_access.py
lfarv: done

lfarv:
> It it possible to put link to the howto guide in the API doc?

The link is on top of the module documentation.

swhite2401:
Ok for me, do you want to rebase before I approve?

lfarv:
> do you want to rebase before I approve?

No need, there is no conflict.

swhite2401 (approved):

PR #899, opened 2025-01-29
add crab cavity passmethod

swhite2401:
The crab cavity passmethod is added to atintegrators.
Constructors in python and matlab remain to be implemented

PR #926, opened 2025-04-14
New python and Matlab releases

lfarv:
It's more than time to issue new releases of AT and PyAT.

For this, we should look at pending pull requests:
#### 3 are waiting approval:
- #893
- #909
- #923

It would be nice to get approval for these

#### 2 are "active" drafts:
- #921
- #892

Should we wait for these or postpone them to the next release?

#### 3 are currently asleep:
- #906
- #899
- #890

### DIFF
--- a/atmat/Contents.m
+++ b/atmat/Contents.m
@@ -1,5 +1,5 @@
 % Accelerator Toolbox
-% Version 2.6.0 (atcollab) 20-Jun-2024
+% Version 2.7 (atcollab) 14-Apr-2025
 % 
 %   Contents file for ATMAT and its subfolders.
 %   
--- a/atmat/at.m
+++ b/atmat/at.m
@@ -1,5 +1,5 @@
 % Accelerator Toolbox
-% Version 2.6.0 (atcollab) 20-Jun-2024
+% Version 2.7 (atcollab) 14-Apr-2025
 %
 % The Accelerator Toolbox was originally created by Andrei Terebilo.
 % Development is now continued by a multi-laboratory collaboration, <a href="matlab:web('https://github.com/atcollab')">atcollab</a>
@@ -140,4 +140,4 @@
 %
 %       <a href="matlab:help symplectify">symplectify</a>              - Makes a matrix more symplectic
 %
-%<a href="matlab:web('/Applications/MATLAB_R2024a.app/help/3ptoolbox/atacceleratortoolbox/doc/AT_page.html')">See documentation for AT</a>
+%<a href="matlab:web('/Applications/MATLAB_R2024b.app/help/3ptoolbox/atacceleratortoolbox/doc/AT_page.html')">See documentation for AT</a>


lcarver:
I will finish off the beam loading branch to try and get it merged (but i have easter holidays coming soon so will try to get it done before then).

swhite2401:
@lfarv, I will free some time to review the pending PRs
- #923 and #909 are fine for me but I want to run some tests to make sure.
- #893 is an improvement but would require additional features in my opinion to be really useful. This is nevertheless non blocking and it could be ok to merge as is it and improve later.

Concerning the active drafts:
- I have been using the matching extensively and we can merge that version, I am not excluding further suggestions. On e question though, where do we stand with this parameters class, I am now running into a case where I have to create a lot of observables to match a lattice which is rather painful.... the parameters class that was proposed at some point would be extremely useful.
- For the rotate element function I have no strong opinion, up to you.

Finally for stale branches, I think they either need more discussion or significant development, no need to rush on these.

SebastienJoly:
For the element rotation branch, I think it does not require a lot of work before being ready. I will try to finish it as soon as possible.

lfarv:
About versioning for this new release:

Should we switch to the **new matching method**: if so, we should change the major version number to report some changes breaking the compatibility. Therefore the python version would be 1.0.0. This also needs some reorganisation of modules, moving the old matching to a "deprecated" module (not imported by default), and moving the new matching into standard.

Otherwise, we can postpone the new matching to the next release and keep the version number at 0.7.0.

I would push for jumping to 1.0.0.

swhite2401:
@lfarv, can we keep the new matching for future release? I think I am the only one that did the transition. At least let's first check with @simoneliuzzo if this is ok

lfarv:
> can we keep the new matching for future release?

Concerning the matching:
- we keep the old one by default, but
- I added a line in the release notes stating that the present matching will be deprecated in a future release, and suggesting to switch from now on to the new one.
- I added a line in the documentation of `at/matching/matching.py` (old one) saying that it is now obsolete, with links to the new variables, observables and matching,
- I added an example notebook showing the use of the new matching. The example is taken from the test modules, it's not optimum. A better example is welcome!

All this can be checked [here](https://at.readthedocs.io/en/improved_matching/p/api/at.matching.matching.html). It should help the transition.

lfarv:
@swhite2401, @SebastienJoly, @oscarxblanco, @simoneliuzzo
I think that we are now ready for new releases of AT and PyAT. Unless you suggest another modification… Please have a look at the [release notes](https://github.com/atcollab/at/releases/tag/untagged-a8b31dc6f08d020606c4) and decide.

oscarxblanco:
@lfarv I think this is OK to release.

SebastienJoly:
@lfarv Good for me as well

oscarxblanco (approved):

SebastienJoly (approved):

swhite2401 (approved):

Issue #925, opened 2025-04-14
set_tilt does not take refpts


lcarver:
For `set_tilt `you have to pass an array of length ring, but being able to provide a smaller array and using `refpts `would be much more convenient and more in line with other functions.

PR #932, opened 2025-05-08
Reorganise the rotation and translation functions

lfarv:
This is a follow-up of #921 and is based on @SebastienJoly's work.
As [mentioned previously](https://github.com/atcollab/at/pull/921#issuecomment-2853699215), the fact that element translations and rotations are handled simultaneously induced some problems, which are solved here:

#### Reorganisation
- the function getting back the translations and rotations from R1 and T1 is extracted and brought to the top level so that it can be used by element properties like `Element.tilt` (or anyone else),
- all functions (`shift_elem`,…) and properties (`Element.dx`,…) now use `transform_elem` and work consistently,
- all functions (`shift_elem`,…) and properties (`Element.dx`,…) have been moved from `utils.py` to `transformation.py.` This minimises inter-module references,
- The default for the arguments of `transform_elem` is set to `None`, indicating the previous value must be kept. This allows to set or increment some parameters while keeping the others unchanged, rather that setting them to zero,
- nothing is changed in the computations.

#### New features
- The value `None` can be used wherever a displacement or rotation is expected and means "Keep the value unchanged", including where the default is 0.0 like in `set_shift`,
- The lattice functions `set_tilt`, `set_shift`, `set_rotation` have now a `refpts` keyword argument. This was proposed in #925 and allows to set errors of specific elements of a lattice. The default is `All`, to be compatible with the present behaviour.

#### Conclusion
With this PR, multiple function calls for displacing a given element can be used to combine transformations, with the following precautions:
1. The order in which the transformations are applied is fixed, and is independent of the order of function calls,
2. All function calls must use the same reference point. That's why I did not propagate the _reference_ argument to the high level functions `set_shift`, `set_tilt`, `set_rotation`: I think it's safer there to stick to the default of `ReferencePoint.CENTRE`.

These precautions are mentioned in the module documentation.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,9 +7,9 @@ version: 2
 
 # Set the version of Python and other tools you might need
 build:
-  os: ubuntu-20.04
+  os: ubuntu-24.04
   tools:
-    python: "3.10"
+    python: "3.11"
     # You can also specify other tool versions:
     # nodejs: "16"
     # rust: "1.55"
--- a/pyat/at/lattice/__init__.py
+++ b/pyat/at/lattice/__init__.py
@@ -14,6 +14,7 @@ from .elements import *
 from .rectangular_bend import *
 from .idtable_element import InsertionDeviceKickMap
 from .utils import *
+from .transformation import *
 from .lattice_object import *
 # from .lattice_variables import *
 from .cavity_access import *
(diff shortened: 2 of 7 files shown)


lfarv:
Please check the updated documentation [here](https://at.readthedocs.io/en/reorganise-rotations-translations/p/api/at.lattice.transformation.html)

oscarxblanco:
Perfect, thank you @lfarv .

PR #940, opened 2025-06-11
cast lattice_object.nbunch to int()

swhite2401:
Problems appeared in the JSON encoder and lattice string representation due to the np.int64 type. This is related to the numpy version 2.3.0 where count_nonzero return np.int64 instead of int.
This function is used in lattice_object.nbunch the return value is now converted to int to allow string representation without errors

### DIFF
--- a/pyat/at/lattice/lattice_object.py
+++ b/pyat/at/lattice/lattice_object.py
@@ -854,7 +854,8 @@ class Lattice(list):
     @property
     def nbunch(self) -> int:
         """Number of bunches"""
-        return np.count_nonzero(self._fillpattern)
+        # cast to int required from numpy version 2.3.0
+        return int(np.count_nonzero(self._fillpattern))
 
     @property
     def harmonic_number(self) -> int:
--- a/pyat/at/load/json.py
+++ b/pyat/at/load/json.py
@@ -27,6 +27,10 @@ class _AtEncoder(json.JSONEncoder):
             return obj.tolist()
         elif isinstance(obj, Particle):
             return obj.to_dict()
+        elif isinstance(obj, np.integer):
+            return int(obj)
+        elif isinstance(obj, np.floating):
+            return float(obj)       
         else:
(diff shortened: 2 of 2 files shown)


lfarv:
@swhite2401, @lcarver: the problem appears with numpy 2.3.0 released on June 7th. It was not there with numpy 2.2.6. Another numpy compatibility problem. This needs a careful investigation…

lcarver:
Can we put an issue here?
https://github.com/numpy/numpy/issues

lfarv:
Consequences: in terms of computation, this should not change anything (I hope). It only affects text representations, so practically `save_m`, `save_repr` and `save_json`.

My preferred solution would be to modify the nbunch property:
```python
    @property
    def nbunch(self) -> int:
        """Number of bunches"""
       # Cast to int required starting from numpy 2.3.0
        return int(np.count_nonzero(self._fillpattern))
```
This solves all the problems and is cleaner than introducing particular cases.

swhite2401:
> This solves all the problems and is cleaner than introducing particular cases.

Agreed, this is what I was suggesting in my previous comment, I make a new commit now.

lfarv:
But on top of that, one could make the json encoder more robust (for the future), with a variant of the proposed solution:
```python
class _AtEncoder(json.JSONEncoder):
    """JSON encoder for specific AT types"""

    def default(self, obj):
        if isinstance(obj, Element):
            return obj.to_dict()
        elif isinstance(obj, np.ndarray):
            return obj.tolist()
        elif isinstance(obj, Particle):
            return obj.to_dict()
        elif isinstance(obj, np.integer):
            return int(obj)
        elif isinstance(obj, numpy.floating):
            return float(obj)
        else:
            return super().default(obj)
```
This deals with all kinds of signed and unsigned numpy integers (byte, short, long, longlong…) and floats (half, single, double…)

swhite2401:
All done

Issue #941, opened 2025-06-12
CorrectorPass is not implemented on at.GPU version.


gubaidulinvadim:
I've tried using GPUs to find momentum aperture and compute Touschek lifetime. If the AT lattice contains corrector elements, the GPU tracking will throw an error:

`RuntimeError: at_gpupass() build lattice failed: Not implemented PassMethod: CorrectorPass in element #29 (from #0) `

At SOLEIL, most of the lattices currently used for SOLEIL II have at.Corrector elements. We could replace it with StrMPoleSymplectic4Pass later, as the lattice is finalised now. However, for now, at SOLEIL, we would need to continue using at.Corrector.

swhite2401:
Solver in #942

PR #942, opened 2025-06-12
Atgpu add correctorpass

JeanLucPons:
Following issue #941 , i added the pass method.
I tested very quicly on EBS by replacing a SH by a CorrectorPass and it seems ok.


gubaidulinvadim (approved):
I've checked that the tracking is possible with the correctors included in the lattice and that the individual corrector seems to change particle coordinates correctly. However, I'm not sure how should I change the tracking mode to GPU for a single element_track() call. In the lattice_track there's a use_gpu option.

Issue #935, opened 2025-05-16
Lattice parameters when no cavity in lattice


TeresiaOlsson:
Hi,

In the pyAT documentation for radiation_parameters it says "Valid for uncoupled lattices with no RF cavity or radiating element." but if I try to use it for a lattice without a cavity inside I get the error "AtError: No cavity found in the lattice" because of the line "voltage = ring.rf_voltage".

Why is there a requirement to have a cavity in the lattice if the function should be valid without RF cavity? And if radiation_parameters can't be used, what is the correct way to calculate the lattice parameters for a lattice without a cavity? For example, during the early stages of lattice design.

lfarv:
@TeresiaOlsson :I see the problem, I'll look at that…

lfarv:
fixed in #943

PR #943, opened 2025-06-12
Improve radiation_parameters

lfarv:
This answers the question raised by @TeresiaOlsson in #935.

`radiation_parameters` now accepts rings without cavities, but of course then all RF parameters ($\varphi_s, \nu_s$,…) are set to `nan`.

The doc is modified to avoid the ambiguity noticed by @TeresiaOlsson: the function works with or without longitudinal motion (4D or 6D).


TeresiaOlsson (approved):
I have tested radiation_parameters for a lattice without cavities and now it works giving nan for the numbers which require a voltage.

PR #947, opened 2025-06-23
Matlab documentation

lfarv:
The full Matlab documentation is now handled by Sphinx and is available on the AT web site.

This results from two actions:
1. Recent Matlab versions now allow saving notebooks (Matlab so-called "live functions") as markdown files. This makes them suitable for Sphinx processing. 
2. The help text for all Matlab functions (docstrings) is now gathered and processed by a python script which converts the result to .rst files. Also suitable for Sphinx processing.

As a result, the already existing Matlab documentation on the Web site is now formatted as the rest of the site, but more important, the documentation of all functions is available, indexed and searchable.

Have a look to [the result](https://at.readthedocs.io/en/matlab_documentation/m/index.html) !

At the same time, the internal Matlab documentation had to be reorganised to work with new Matlab versions

Note: the help for Matlab functions is now available on line, but is unchanged. This reveals a number of small problems which should be corrected in other PRs.


lfarv:
See #946 for the explanation of test failures.

swhite2401 (approved):

Issue #695, opened 2023-11-12
pyAT 'Multipole has no attribute PolynomB' exception for loading multipoles from a .mat file with both PolynomB and K value defined


space-budgie:
### Issue
While trying to load a `.mat` lattice file using `at.load.matfile.load_mat`, multipole magnets that have both a `PolynomB` and a `K` value defined raise an error:

```Python
AttributeError: In element QFEND, parameter K: 'Multipole' object has no attribute 'PolynomB'
```
This was tested for Python accelerator toolbox version 0.4.0.

Here, QFEND is defined with with the Matlab code
```matlab
QFEND 	= atmultipole('QFEND' , 0.250000, [0 0 0 0], [0  3.653849 0 0], 'StrMPoleSymplectic4Pass','MaxOrder',2);  
QFEND.K   = QFEND.PolynomB(2);
```
### Potential cause
I believe that the issue lies in the function call

https://github.com/atcollab/at/blob/1e32037159aef2122d0fc0952c245acf129fe7cc/pyat/at/lattice/elements.py#L628

where passing in `K` as part of `**kwargs` causes a call to

https://github.com/atcollab/at/blob/1e32037159aef2122d0fc0952c245acf129fe7cc/pyat/at/lattice/elements.py#L704-L706

but `PolynomB` is only defined later in

https://github.com/atcollab/at/blob/1e32037159aef2122d0fc0952c245acf129fe7cc/pyat/at/lattice/elements.py#L634

### (Temporary) fix
I managed to bypass the issue by changing the lines
https://github.com/atcollab/at/blob/1e32037159aef2122d0fc0952c245acf129fe7cc/pyat/at/lattice/elements.py#L628-L634
to
```Python
        # Set MaxOrder while PolynomA and PolynomB are not set yet
        super(ThinMultipole, self).__setattr__('MaxOrder', maxorder)
        # Adjust polynom lengths and set them
        len_ab = max(self.MaxOrder + 1, len_a, len_b)
        self.PolynomA = lengthen(poly_a, len_ab - len_a)
        self.PolynomB = lengthen(poly_b, len_ab - len_b)
        super(ThinMultipole, self).__init__(family_name, **kwargs)
```
so that `PolynomB` is defined prior to the potential call to the setter. 

However, I have not tested if this breaks anything else in the code.

T-Nicholls:
Hi @TeresiaOlsson,
Yes, I got distracted by other things and forgot to circle back to this.
[...]

PR #950, opened 2025-07-03
Fix for #695 - Multipole Strength Prioritisation

T-Nicholls:
This makes `ThinMultipole`, `Multipole`, `Dipole`, `Quadrupole`, `Sextupole`, and `Octupole` all prioritise `PolynomB` over `K` and `H` in a consistent manner (according to the logic agreed in #895 ), and adds warnings for when we ignore `K` or `H`


lfarv:
2nd problem:

```
Python 3.11.13 (main, Jun  3 2025, 18:38:25) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
>>> import at
>>> 
>>> qp = at.Quadrupole("qp", 1.0, 0.1, PolynomB=[0., 0.])
>>> print(qp)
Quadrupole:
       FamName: qp
        Length: 1.0
      MaxOrder: 1
   NumIntSteps: 10
      PolynomA: [0. 0.]
      PolynomB: [0.  0.1]
             K: 0.1
>>> 
```
The `PolynomB` keyword should have priority over the 3rd argument, so `PolynomB` should be `[0.0, 0.0]`

lfarv:
Same problem when saving a .repr file:
```
Python 3.11.13 (main, Jun  3 2025, 18:38:25) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
>>> import at
>>> 
>>> mp=at.Multipole("MP1", 0.5, [], [0.001, 0.5])
>>> lat1 = at.Lattice([mp], energy=6e9)
>>> 
>>> at.save_repr(lat1)
{'name': '', 'energy': 6000000000.0, 'particle': Particle('relativistic'), 'periodicity': 1, 'beam_current': 0.0, 'nbunch': 1}
Multipole('MP1', 0.5, array([np.float64(0.0), np.float64(0.0)]), array([np.float64(0.001), np.float64(0.5)]), 0.5, 0.0)
````

In fact all this comes from the `repr` output of the `Multipole`:
```
>>> mp=at.Multipole("MP1", 0.5, [], [0.001, 0.5])
>>> mp
Multipole('MP1', 0.5, array([0., 0.]), array([0.001, 0.5  ]), 0.5, 0.0)
```
It's difficult to anticipate all side effects and compatibility problems coming from this change

**Comment by lfarv** [comment]
I think that this solution is unnecessarily complicated. A simple solution is:
```python
class Quadrupole(Radiative, Multipole):

...

    def __init__(
        self, family_name: str, length: float, k: float = 0.0, **kwargs
    ):
        k2 = kwargs.pop("K", k)
        super().__init__(family_name, length, [], [0.0, k2], **kwargs)
        if k != 0.0 and float(k) != self.K:
            warnings.warn(
                AtWarning(
                    f"In element {family_name!r}: conflicting values "
                    f"for the focusing strength.\nKeeping {self.K}"
                ),
                stacklevel=2,
            )
```
And similarly for the dipole and sextuple. A single line added to handle the `K` keyword, and a simple test for the warning. And no change for the `Multipole`!

Advantage of the solution:
- the logics of the Quadrupole and Multipole are clearly separated: quadrupole-specific quantities like K or H **are not forwarded** to the parent `Multipole`,
- the behaviour of the `Multipole` is unchanged,
- the change is so small that the risk of side effects is null.

The behaviour with this simple solution satisfies the priority rules:
```
Python 3.11.13 (main, Jun  3 2025, 18:38:25) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
>>> import at
>>> 
>>> # Default value
>>> qp = at.Quadrupole("qp", 1.0)
>>> print(qp.K)
0.0
>>> 
>>> # Positional parameter
>>> qp = at.Quadrupole("qp", 1.0, 0.1)
>>> print(qp.K)
0.1
>>> 
>>> # K keyword
>>> qp = at.Quadrupole("qp", 1.0, 0.1, K=0.2)
<stdin>:1: AtWarning: In element 'qp': conflicting values for the focusing strength.
Keeping 0.2
>>> print(qp.K)
0.2
>>> 
>>> # PolynomB keyword
>>> qp = at.Quadrupole("qp", 1.0, 0.1, K=0.2, PolynomB=[0., 0.3])
<stdin>:1: AtWarning: In element 'qp': conflicting values for the focusing strength.
Keeping 0.3
>>> print(qp.K)
0.3
>>> 
>>> # No warning when overloading the default value:
>>> qp = at.Quadrupole("qp", 1.0, K=0.2)
>>> print(qp.K)
0.2
>>> 
>>> qp = at.Quadrupole("qp", 1.0, PolynomB=[0., 0.3])
>>> print(qp.K)
0.3
>>> 
```
The problem with @TeresiaOlsson's file is different, I'll comment on that later…

T-Nicholls:
Hi @lfarv 
Re the points you've raised:

1. Why would this ever happen? Big `K` is in the quadrupole's build attributes and so will always be popped and converted to little `k` when creating from a lattice file. So the only way this can happen is if a user is manually creating a quadrupole and specifies it, which I'm not sure why they would do that.
2. I intended this - if `PolynomB[1]` is `0.0` (the default value) but `k` is not, then I made `k` have priority; since it doesn't make sense to prioritise a default value when a non-default value has been manually specified. `PolynomB` still has priority over `k` in all other situations.
3. I also intended this, I thought since `K` & `H` were on the `repr`s of `Quadrupole` & `Sextupole`, they should also go on the `Multipole` `repr`, however, since this would break loading new lattice files to Matlab I will reverse this.

lfarv:
@T-Nicholls: I understand your points, but here is how I see different use cases:

1. You are right, it may only occurs if  user explicitly specifies a K keyword. No good reason to do that, but a user is allowed to specify any attribute he wants to be set on the quad (predefined or custom), so at least it should not crash. A definition like `qp = Quadrupole("qf1", 0.2, K=0.5)` does not look so strange…

2. Here, there are good reasons to explicitly specify PolynomB, for instance to create a combined function magnet. For instance to add a sextupole component, one would write `PolynomB=[0.0, k, h]`. It would be misleading that PolynomB is partially modified in the particular case of k=0. If `PolynomB` has priority it should always be such,

3. Indeed, I cannot guess all the consequences of a change of the `Multipole` `repr`, so I really think it's safer to keep it unchanged.

lfarv:
@TeresiaOlsson : This `K` attribute was unexpected in a `Multipole` creation, but now it seems that @T-Nicholls' modification handles it correctly !

lfarv:
@TeresiaOlsson 

Historically, `K` was used in the strongly discouraged `QuadLinearPass` passmethod. This was modified long ago so that  `PolynomB[1] `is used if existing, with `K` being a fallback solution if `PolynomB` does not exist.

I don't know if MML runs with recent AT versions (though we did everything to ensure compatibility), and it possibly still relies on `QuadLinearPass`, which is a very bad solution…

T-Nicholls:
@lfarv I've committed something that should now behave as you describe; to summarise the expected behaviour:

- `PolynomA` is always prioritised over `poly_a` and an appropriate warning is given when that happens
- `PolynomB` is always prioritised over `K`, `H`, and `poly_b` and an appropriate warning is given when that happens
- `K` and `H` are always prioritised over `poly_b` and an appropriate warning is given when that happens
- `k` and `h` are not positional arguments for `Multipole` or `ThinMultipole` and so their `repr`s won't change
- `K` and `H` passed in kwargs are handled correctly by `Multipole` or `ThinMultipole` now

If you are happy with that, then on Thursday I will update the tests and I may also simplify the warning logic - as I think my current method of dynamically generating the warning messages is probably overkill (let me know your thoughts).

pyat/test/test_basic_elements.py
```diff
@@ -139,6 +140,162 @@ def test_multipole():
...
+            elem = element_type("Dipole", 1.0, 0.0, k=0.0, PolynomB=[0.0, 1.0])
+    assert elem.PolynomB[1] == 1.0
+    # No warnings or changes when K is None
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        if element_type is elements.Quadrupole:
+            elem = element_type("Quad", 1.0, k=None, PolynomB=[0.0, 1.0])
+        else:
+            elem = element_type("Dipole", 1.0, 0.0, k=None, PolynomB=[0.0, 1.0])
+    assert elem.PolynomB[1] == 1.0
```
lfarv, line 237: `None` is not an allowed value for the strength. If you don't specify PolynomB you get a nan in the element: not very useful.

You can remove this test, and feel free to change the type hint in the Quadrupole definition and suppress the None

lfarv:
@T-Nicholls  Follow-up about the .mat-file:

Since K and H are not part of a multipole definition, my 1st idea would be to simply ignore them. And that's what Matlab is doing.

However your processing is logically correct, except for the warning message which should not appear since here the values are identical.

My only concern is the unlikely case of a multipole defined with a K keyword different from poly_b[1]: the same file will give different results when read form Matlab (K ignored) and python (K has priority).

What is your opinion ?

TeresiaOlsson:
I have something I'm wondering about. What happens when adding field errors to a magnet? Are they added both to the H/K and the polynomials?

If so, I guess it's always a mistake when K/H don't have the same values as the polynomials and it should raise an exception so the user get the information and is forced to fix it?

T-Nicholls:
Furthermore, @TeresiaOlsson raises a good point; if we have contradictory values for focusing strength, then something is definitely wrong and we should probably raise an exception so that the user doesn't keep misusing the element constructor. For this reason, I propose that we do the following:

1. If the ignored strength value (either from `K`, `H`, or `poly_b`) is non-zero and is different to the strength value being prioritised, then we raise an error
2. If the ignored strength value is zero or the same as the strength value being prioritised, then we give a warning

The reason I think we should do this for each case is:

1. If we don't raise an error in this case, then the user can experience unintended behaviour, and there is no valid reason for a user to want to do this; for that reason, I think a user should be prevented from doing that, rather than just warned
2. In this case, there isn't a risk of unintended behaviour, but the user is still misusing the element constructor, which could lead to unintended behaviour in future; so, I think that a warning is sufficient here

T-Nicholls:
Ok, I have updated the logic now; however, I've run out of time today, so I'll update the tests tomorrow morning

swhite2401:
@T-Nicholls @lfarv, there is interest to add  `IntegratedPolynomA/B` properties (`elem.PolynomA/B*elem.Length`). Since this is rather simple development do you think we could integrate it to this PR?

swhite2401:
Ok I'll do a separate PR then

T-Nicholls:
Ready for review again, once you're happy with it, I'll clean up the commit history before we merge it

lfarv:
Hi @T-Nicholls
This look all good. Thanks for the reorganisation: it will be very easy to merge your last version in #950 
I think that #950 still needs a lot of checks, so I propose to merge this one right now, update #950 accordingly and leave time for more tests.

I usually use the "squash and merge" option which groups all commits in a single one, so if you agree, I can merge it as it is. No need for cleanup, unless you prefer to do it. Let me know.

lfarv (approved):
All good

PR #955, opened 2025-07-10
IBS module for pyat

Ritu-Bharadwaj:
These are the files to model the Intrabeam Scattering effect written purely in python but we intend to convert the passmethod in C in future.

### DIFF
--- a/pyat/at/collective/__init__.py
+++ b/pyat/at/collective/__init__.py
@@ -1,7 +1,9 @@
 """
 Collective effects
 """
+
 from .wake_elements import *
 from .wake_functions import *
 from .wake_object import *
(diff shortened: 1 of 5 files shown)


lcarver:
Hello @Ritu-Bharadwaj , thank you for this work!

Can you make an example script showing how to use it? I recommend that you put it in `pyat/examples/CollectiveEffects/`. Typically for the examples you need to use a lattice from the `machine_data` sub directory. We usually go for the old ESRF lattice.

This will make it much easier to understand and provide feedback.

swhite2401 (commented):
Very nice work! I added a few comments with possible simplifications.

PR #957, opened 2025-07-11
Elements package

lfarv:
This turns the `elements.py` module into a new `elements` package.

`elements.py` was growing to a size making it difficult to maintain. It is now split into 10 modules, which should reduce the possibility of conflicts during parallel development.

The `elements` package interface is identical to the old `elements` module's one, so no other modification is required, except for one point:

modules defining new `Element` derivatives which cannot be moved into this new package (because need the `physics` package for instance) may need to access the "conversion" functions (like `_array`). These are no longer publicly accessible accessible from the `lattice` package, they need to be imported from` lattice.elements.conversions`.


lfarv:
This has to wait for #950 to be merged, to introduce the last modifications from @T-Nicholls.

swhite2401:
Is this ready for review?

lfarv:
Just waiting for the last modifications in #950 to be included in magnet_elements.py. Everything else is ready for review. Le 16 juil. 2025 à 10:24, Simon White ***@***.***> a écrit :﻿swhite2401 left a comment (atcollab/at#957)
Is this ready for review?

lfarv:
Ready for review

swhite2401 (approved):
From what I could see this is just a reorganization / reformating of the present implementation, so this is fine for me. This makes it in fact easier to maintain.

PR #963, opened 2025-08-11
Fix nbyn idtable dat

oscarxblanco:
This PR takes back the work started in #683 .


swhite2401:
@oscarxblanco thanks for the update, I haven't followed this development at all and would need time to dive into it.
Probably the best persons to review would be @lfarv @lnadolski and @simoneliuzzo as they were involved in the discussion in #683 and are matlab users.

oscarxblanco:
Dear all, this PR is ready for review.

In addition to the comments above, now pyat accepts a dictionary to create an element. This should be enough to support non-text IDs, for example, IDs in .mat or .hdf5 formats because the user can create a function that reads the non-text file and returns a dictionary with the needed keys.

I did not implement directly the reading of .mat or .hdf5 files because .mat files vary in format. Old ones seem to be a matlab format compatible with scipy.io while others are .hdf5 in disguise. One could test whether it is one or the other, but it seemed to me unreasonable to implement a file format check and version check only to feed data to this element.

lfarv:
@oscarxblanco:
This looks fine for me. But I cannot do any independent check of the behaviour of ID kick maps. So my proposal is that I'll approve the PR (I trust you for the processing of kick maps),  we try to get a review from @lnadolski, and if not, you are free to merge when you like. This say, we are not locked…

lfarv (approved):
OK

PR #961, opened 2025-08-06
Add `CHROMATIC2` to pyat `get_rdts`

oscarxblanco:
Dear all,
I have implemented in pyat the chromatic2 rdts: second order wrt sextupoles and first order wrt to octupoles.

Who could review this pull request ?

PR #964, opened 2025-08-12
Adding the  new CHROMATIC2 RDTs to the RDT Observables

lfarv:
This adds the CHROMATIC2 RDTs implemented in #961 to the list of RDT Observables;

### DIFF
--- a/pyat/at/latticetools/rdt_observable.py
+++ b/pyat/at/latticetools/rdt_observable.py
@@ -4,8 +4,6 @@ __all__ = ["RDTObservable"]
 
 from functools import partial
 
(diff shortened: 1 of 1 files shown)


lfarv:
@oscarxblanco:

The docstring of `get_rdts` mentions a "h00103" rdt which is not computed. Should this mention be removed ?

oscarxblanco:
> @oscarxblanco:
> 
> The docstring of `get_rdts` mentions a "h00103" rdt which is not computed. Should this mention be removed ?

@lfarv , yes. It should not be there. Thank you for noticing it.

PR #970, opened 2025-08-13
Include Strength and IntegratedStrength in magnet elements

swhite2401:
This PR proposes a minor enhancement.

The properties strength and integrated strength pointing to `PolynomB[DefaultOrder]` and `PolynomB[DefaultOrder]*Length` are added. A complete octupole element with default strength propertie `O` is also implemented.
The `Strength` properties could eventually replace `K, H, O` to simplify the code.

Few open questions remain:
- Element without `DefaultOrder` are not handled -> raise an error when trying to access `Strength` or `IntegratedStrength`, here I considered taking into account the lowest non-zero element in `PolynomB` but I think it is dangerous, what do you think?
- ThinMultipoles are not handled as they do not have `DefaultOrder`. Concerning this point I am considering merging the 2 passmethods, ThinMultipole would then become a special case of multipole with length=0.0, this would allow to create thin element simply by setting their length to zero.
For example:
`thinquadrupole = at.Quadrupole('QUAD', 0.0, 1.0)`
which would be a much more natural user interface.

If considered useful, this last point should be treated in a separate PR.

### DIFF
--- a/pyat/at/physics/rdt.py
+++ b/pyat/at/physics/rdt.py
@@ -27,7 +27,9 @@ class RDTType(Enum):
     #: Amplitude detuning coefficients
     #: optionally includes the second order contribution of sextupoles
     TUNESHIFT = 6
-    CHROMATIC2 = 7  #:Chromatic RDTs from sextupoles
+    #: Chromatic RDTs from octupoles
+    #: optionally includes the second order contribution of sextupoles
+    CHROMATIC2 = 7
 
 
 def _get_polynom(elem, attr, index):
@@ -644,7 +646,7 @@ def get_rdts(
 
     for :py:obj:`~RDTType.CHROMATIC2`:
         `h21001`, `h30001`, `h10021`, `h10111`, `h10201`, `h11002`
-        `h20002`, `h00112`, `h00202`, `h10003`, `h00103`, `h00004`
+        `h20002`, `h00112`, `h00202`, `h10003`, `h00004`
 
     for :py:obj:`~RDTType.TUNESHIFT`:
         `dnux_dJx`, `dnux_dJy`, `dnuy_dJy`


pyat/at/lattice/elements/magnet_elements.py
```diff
@@ -168,29 +176,80 @@ def __setattr__(self, key, value):
...
+        if order is None:
+            msg = (f"Default Order not defined for class "
+                   f"{self.__class__.__name__}, please access the strength "
+                   f"with PolynomA/B or use predefined magnet class")
+            raise AtError(msg)
+        return order
+    
+    def _get_strength(self, order, integrated=False):
+        try:
+            k = self.PolynomB[order]
```
lfarv, line 190: better:
```
k = float(self.PolynomB[order])
```
which converts from array scalar to float, as it was done before.
The result is the same, but the printout is nicer (avoids `np.float64(0.51)`), and allows to add a type hit to the function: `-> float`.

For the future: it also triggers the evaluation of parameters.

pyat/at/lattice/elements/magnet_elements.py
```diff
@@ -436,12 +495,38 @@ def items(self) -> Generator[tuple[str, Any], None, None]:
...
+    """Octupole element"""
 
-    _BUILD_ATTRIBUTES = Multipole._BUILD_ATTRIBUTES
+    _BUILD_ATTRIBUTES = LongElement._BUILD_ATTRIBUTES + ["O"]
+    _stacklevel = 7  # Stacklevel for warnings
 
     DefaultOrder = 3
 
+    def __init__(self, family_name: str, length: float, o: float = 0.0, **kwargs):
+        """
```
lfarv, line 506: Problem here: you change the signature of the constructor which was
```
Octupole(name, length, polya, poly)
```
which is not backward compatible. This is illustrated by the change you had to do in other files.

You could try 
```python
def __init__(self, name, length, *args, **kwargs):

    if isinstance(args[0], Iterable):
        polya, polyb = args[:2]
    else:
        polya = []
        polyb = [0.0, 0.0, 0.0, args[0]]

    super().__init__(family_name, length, poly, poly, **kwargs)
```
to keep compatibility

lfarv:
> To be noted that the notation K, H, O are not the most intuitive. K normally stands for the strength and K1, K2, K3 would be the quadrupole, sextupole and octupole strengths... Similarly, KL, K1L, K2L, etc... are integrated strengths and KS, K1S, K2S, etc... are skew strengths
> 
> Changing this would probably mean breaking backward compatibility but would allow for generalization with compact code.

Nothing prevents from adding new properties if it looks useful.

I would consider the ongoing [PALS](https://pals-project.readthedocs.io/en/latest/index.html) project which tries to set a standard for the [definition of magnetic multipoles](https://pals-project.readthedocs.io/en/latest/element-parameters.html#magneticmultipolep-magnetic-multipole-parameters): they chose **Kn1, Kn2…** for normal components and **Ks1, Ks2…** for skew components. And **Kn1L, Kn2L**… for integrated strengths.

Issue #966, opened 2025-08-13
pyat `save_m` and `load_m` bug with long numpy arrays


oscarxblanco:
Dear all,
while checking PR #963 , I realized that there is a bug in pyat when long arrays, for example the ones in ID tables, are written or read to/from text .m files.

The issue comes from numpy.array2string default behaviour
https://github.com/atcollab/at/blob/3b172e653a54499857e037d5d339b45c34d3d6c7/pyat/at/load/matfile.py#L429-L435
where long arrays are replaced by ellipsis "..."

oscarxblanco:
Solved in #971

PR #971, opened 2025-08-13
Fix pyat and matlab arrays read/write to/from m files

oscarxblanco:
Dear all,
this is a possible solution to bug #966 , where long arrays were mistakenly written with ellipsis in .m files.

The function `make_array` in pyat, used to read arrays, is rewritten to be compatible with pyat `save_m` output array format [[...];[...]] for .m files, and matlab `atwrite` [...;...;...]. All changes are therefore backwards compatible.

The function `convert_array` is modified to consider long arrays, which could happen when including ID elements.

All other changes are included to fix warnings from flake8 and follow the black format. As a note, the import order given by isort is different to the one prefered by black. I chose to leave the file `matfile.py` as it is after running black.

### DIFF
--- a/pyat/test/test_load_save.py
+++ b/pyat/test/test_load_save.py
@@ -1,5 +1,7 @@
 """pyat test."""
 
+from __future__ import annotations
+
 import os
 import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


swhite2401:
@oscarxblanco , @lfarv is certainly more knowledgeable that me in this topic, but you can get inspiration on how lattice file are handled. 
They are located in the machine data folder and loaded in `conftest.py`

lfarv (approved):
ok for me

PR #972, opened 2025-08-14
Reorgnize magnet elements and passmethods

swhite2401:
The default passmethods for thick multipole now accept zero length elements in thin lens approximation.
`ThinMpolePass` is not needed anymore.
The python interface to magnet elements is re-organize to threat thin elements.

### DIFF
--- a/atintegrators/StrMPoleSymplectic4Pass.c
+++ b/atintegrators/StrMPoleSymplectic4Pass.c
@@ -45,11 +45,22 @@ void StrMPoleSymplectic4Pass(double *r, double le, double *A, double *B,
     bool useLinFrEleExit = (fringeIntM0 != NULL && fringeIntP0 != NULL  && FringeQuadExit==2);
     double B0 = B[0];
     double A0 = A[0];
-
-    if (KickAngle) {   /* Convert corrector component to polynomial coefficients */
+    
+    /* Convert corrector component to polynomial coefficients */
+    if (KickAngle != NULL && le>0) {
         B[0] -= sin(KickAngle[0])/le;
         A[0] += sin(KickAngle[1])/le;
+    } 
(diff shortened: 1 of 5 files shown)


lfarv:
Hello @swhite2401: I see the point in accepting 0 length to define thin magnets. And I understand that this is a work in progress, still far from being ready, but I have right now some comments:

- Changing the inheritance diagram is a **very bad idea**. Inheritance has some logic, breaking it may lead to absurd results:
  ```
  >>> issubclass(ThinMultipole, LongElement)
  True
  >>>
  ```
  may lead to stupid conclusions. More generally, if in the future another property is added to `LongElement`, all thin elements will immediately inherit it, and another patch will be needed in the `_ThinElement` class to cancel that. `_ThinElement` itself looks like a nonsense, since it tries to revert back long elements into thin ones. A "ThinElement" is an `Element`, non divisible, non mergeable.

  For instance the new `ThinQuadrupole` class should inherit from `ThinMultipole`. Not from `Quadrupole`: most `Quadrupole` class attributes are irrelevant: `Radiative` ancestor, _BUILD_ATTRIBUTES, `FringeQuad`* attributes, `_stacklevel`, … The only common attribute is DefaultOrder.

  If the inheritance diagram is not absolutely clean, lot of errors will appear. Here inheriting from both `_ThinElement` and `LongElement` clearly indicated that something is wrong.

- Concerning C integrators, I think that the safest way, by far, to combine thin and thick elements is to put in the same C file (StrMPoleSymplectic4Pass.c) both functions `ThinMPolePass()` extracted from ThinMPolePass.c and `StrMPoleSymplectic4Pass()` (already there). And to have a single test in trackFunction calling one or the other depending on length. Otherwise, you'll need a lot of tests, within loops: it's inefficient and error prone.

Issue #981, opened 2025-09-01
pyat get_rdts h00112 gives a wrong result


oscarxblanco:
Dear all,
I have made a mistake on the implementation of the h00112 rdt.
In line 444, the term (pym2[i] - cpym2) should be multiplying.
https://github.com/atcollab/at/blob/fb8fd4f24e86ccf9b6f4fba799dc5c4e51b447f4/pyat/at/physics/rdt.py#L442-L446
and
<img width="1101" height="154" alt="Image" src="https://github.com/user-attachments/assets/244b2c17-4fcc-401c-8c1e-6780f4248a6e" />
from Chun-Xi Wang's  "Explicit formulas for 2nd-order driving terms due to sextupoles and chromatic effects of quadrupoles" ANL/APS/LS-330 March 10, 2012.

I'll open a pull request to fix it.

PR #982, opened 2025-09-01
fix bug pyat get_rdts h00112

oscarxblanco:
Dear all,
this is a fix to #981 where a typo in the implementation of h00112 was identified. The minus sign is replaced by a multiplication sign which is the correct operation of the exponential values. Parenthesis around have been removed as they are not longer needed because of the precedence of the multiplication sign.

Who could review this ?


oscarxblanco:

swhite2401 (approved):

swhite2401:
Good catch!

Issue #985, opened 2025-09-03
Diffusion matrix calculation problem for lattice with linear coupling


SebastienJoly:
Hello,

I am trying to generate a quantum diffusion element to run macroparticle tracking simulations in the presence of linear coupling. While I can use the envelope_parameters() function without any problem, I cannot compute the equivalent diffusion matrix for my ring. 
I used the following command:
`quantum_matrix = at.physics.radiation.gen_quantdiff_elem(ring)
`
And obtained this Numpy error:
`LinAlgError: Matrix is not positive definite`
From what I understand, the problem originates from the at.physics.radiation._lmat() function computing the Cholesky decomposition of the diffusion matrix. In my case, the 'except' clause in the try except block is being used. The problem is that the line `cmat = np.reshape(dmat[nz], (4, 4))` causes an error as the cmat matrix should be (6,6) in my example.
I calculated the diffusion matrix this way:
`
dmat=at.physics.radiation.quantdiffmat(ring)
`
And here is what it returned:
`[[ 3.22597462e-15  1.61254870e-16  4.63784182e-16  7.65692610e-17
   2.45454931e-14 -1.32686730e-17]
 [ 1.61254870e-16  7.96076281e-16  4.89655080e-17  8.97149100e-18
  -1.10542571e-14 -1.36570700e-18]
 [ 4.63784182e-16  4.89655080e-17  4.92090297e-16  8.01910010e-17
  -5.75678241e-14 -8.64565910e-17]
 [ 7.65692610e-17  8.97149100e-18  8.01910010e-17  1.30478310e-17
  -9.40095068e-15 -1.40729960e-17]
 [ 2.45454931e-14 -1.10542571e-14 -5.75678241e-14 -9.40095068e-15
   9.43297990e-12  1.18633604e-14]
 [-1.32686730e-17 -1.36570700e-18 -8.64565910e-17 -1.40729960e-17
   1.18633604e-14  2.18307480e-17]]
`
Could I use a different matrix decomposition method to overcome this issue? Or should I assume a quantum diffusion element using the uncoupled ring and use it in my tracking simulations with coupling?

swhite2401:
> [@SebastienJoly](https://github.com/SebastienJoly) you can certainly generate the element without coupling and add coupling at a later stage, emittances will be correctly re-distributed. However, I am curious too loo at you issue, could you send me your lattice file?

I am having second thoughts, this statement is only valid for certain conditions, however I never had this problem... let's see with your script and lattice

[...]

PR #990, opened 2025-09-11
Quantum diffusin matrix numerical noise issue

swhite2401:
This PR solves issue #985 where it was observed the the choleski decomposition was failing in some cases  due to non-positive defined diffusion matrix.


swhite2401:
> > I think in this case we may enter the new values by the new algorithm and keep the tolerance as it is
> 
> That's exactly what I was suggesting just above!

Oh yes, sorry, I need to read more carefully!

swhite2401:
Yes all good to me, I cannot approve though since I created the PR.

lfarv:
@SebastienJoly : nothing new for 1 week, so I merge the PR. Thank for your contribution !

Issue #989, opened 2025-09-10
3D geometry/survey for python AT


simoneliuzzo:
matlab AT has a 3D geometry computation feature.

It would be nice to have the same feature in python AT.

lfarv:
Implemented in #993

PR #993, opened 2025-09-20
3d geometry

lfarv:
This answers #989 by adding a function to get the 3d coordinates of lattice elements.
A test function is also added.

### DIFF
--- a/pyat/at/plot/standalone.py
+++ b/pyat/at/plot/standalone.py
@@ -111,45 +111,71 @@ def plot_acceptance(ring: Lattice, planes, *args, **kwargs):
 def plot_geometry(
     ring: Lattice,
     start_coordinates: tuple[float, float, float] = (0, 0, 0),
-    centered: bool = False,
+    centred: bool = False,
+    h_angle: float = 0,
+    v_angle: float = 0,
     ax: Axes = None,
+    plot3d: bool = False,
     **kwargs,
 ):
-    """Compute and plot the 2D ring geometry in cartesian coordinates.
+    """Compute and plot the 2D or 3D ring geometry in cartesian coordinates.
 
     Parameters:
-        ring: Lattice description
+        ring:           Lattice description
         start_coordinates: x,y,angle at starting point
-        centered: it True the coordinates origin is the center of the ring
+        centred:        it True the coordinates origin is the center of the ring
+        h_angle:        initial horizontal angle.
+        v_angle:        initial vertical angle.
         ax: axes for the plot. If not given, new axes are created
+        plot3d:         Draw a 3d line plot
 
     Keyword arguments are forwarded to the underlying
     :py:func:`~matplotlib.pyplot.plot` function
 
     Returns:
-        geomdata: recarray containing, x, y, angle
+        geomdata: :py:class:`~numpy.recarray` containing the fields *x*, *y*,
+          *z*, *angle*, *v_angle*.
         radius: machine radius
         ax: plot axis
 
     Example:
         >>> ring.plot_geometry()
     """
-    if not ax:
-        fig, ax = plt.subplots()
+    # Accept the US wording "centered" for compatibility
+    centred = kwargs.pop("centered", centred)
     geom, radius = ring.get_geometry(
-        start_coordinates=start_coordinates, centered=centered
+        start_coordinates=start_coordinates,
+        h_angle=h_angle,
+        v_angle=v_angle,
+        centred=centred,
     )
-    ax.plot(
-        geom["x"],
-        geom["y"],
-        "o:",
-        linewidth=kwargs.pop("linewidth", 0.5),
-        markersize=kwargs.pop("markersize", 2),
-        **kwargs,
-    )
+    if plot3d:
+        if not ax:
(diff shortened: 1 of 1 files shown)


lfarv:
This new feature is implemented as a new function `get_geometry3`, similar to the existing `get_geometry`. Ideally, it could replace the existing function, since adding a new "x" field to the output recarray does not break the compatibility. However, `get_geometry` has additional arguments whose usefulness can be discussed:
- a `radius` output, which tries to guess an estimate of the radius of the ring. This is useful only if the lattice represents a cell of a periodic ring. Practically, this result is meaningful only if the cell is symmetrical, and it's anyway different from the "classical" radius taken as $C/2\pi$.
- a `centred` input keyword, offsetting the output coordinates so that the origin is at the centre of the ring. With the same ambiguities mentioned above concerning the determination of the centre.

So we have 3 alternatives:
1. Keep 2 different functions, the old one kept for compatibility (2D only), the new one without the confusing arguments,
2. Replacing the old function with the new one, but breaking the compatibility. The `centred` keyword can still be accepted with a warning saying it's now ignored. But a single output instead of two breaks the compatibility,
3. We implement in the new function the same arguments as previously, acting on the x/y projection of the ring, knowing that  the determination of the centre of the ring is doubtful.

What is your opinion ?

simoneliuzzo:
Dear @lfarv,

to check that the 2D features are still available I resumed my test script on geometry 2D. 

this line failed:

data, radius, ax = ring.plot_geometry(centered=True)

with error
`Traceback (most recent call last):
  File "/mnt/multipath-shares/machfs/liuzzo/accelerator_toolbox_tests/python/geometry3D/test_geom_2d.py", line 27, in <module>
    data, radius, ax = ring.plot_geometry()
                       ^^^^^^^^^^^^^^^^^^^^
  File "/mnt/multipath-shares/machfs/liuzzo/accelerator_toolbox_tests/python/geometry3D/venvgeom/lib/python3.12/site-packages/at/plot/standalone.py", line 139, in plot_geometry
    geom, radius = ring.get_geometry()`

I think radius is used by the centered option

simoneliuzzo:
Dear Laurent,

I run this code:
```

import at
import numpy as np
import at.plot
import copy
from os.path import exists
from scipy.io import savemat, loadmat
import time
import sys

folder_data = '.'
show_figures = False


lattice_file = '/machfs/liuzzo/EBS/beamdyn/matlab/optics/sr/S28F_all_BM_27Mar2022/betamodel.mat'
ebs = at.load_lattice(lattice_file, mat_key='betamodel')

ring = at.Lattice(ebs)

data, radius, ax = ring.plot_geometry()


```
and now I get this message:

```
  File "/mnt/multipath-shares/machfs/liuzzo/accelerator_toolbox_tests/python/geometry3D/test_geom_2d.py", line 27, in <module>
    data, radius, ax = ring.plot_geometry()
                       ^^^^^^^^^^^^^^^^^^^^
  File "/mnt/multipath-shares/machfs/liuzzo/accelerator_toolbox_tests/python/geometry3D/venvgeom/lib/python3.12/site-packages/at/plot/standalone.py", line 139, in plot_geometry
    geom, radius = ring.get_geometry(
                   ^^^^^^^^^^^^^^^^^^
  File "/mnt/multipath-shares/machfs/liuzzo/accelerator_toolbox_tests/python/geometry3D/venvgeom/lib/python3.12/site-packages/at/lattice/geometry.py", line 222, in get_geometry
    coords = [coord0] + [increment(xyzc, convi, el) for el in ring]
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/multipath-shares/machfs/liuzzo/accelerator_toolbox_tests/python/geometry3D/venvgeom/lib/python3.12/site-packages/at/lattice/geometry.py", line 206, in increment
    xyz += conv @ np.array([length, 0.0, 0.0])
numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'

```

thank you for your help
best regards
Simone

lfarv:
@simoneliuzzo

> `arc.plot_geometry(offset=(100, -50))`

This line is wrong: `offset` is not accepted neither by `get_geometry` nor `plot_geometry`. So, as indicated in the doc, it's forwarded to the `plot` command, and is rejected there. It fails with the same message on the `master` branch.

Looking back in [history](https://github.com/atcollab/at/pull/460#issuecomment-1216813296), I see that this keyword appeared during development but was removed before merging (with your approval!). So this test looks like a survival of the development process.

`start_coordinates` is a correct keyword and works correctly on both the `master` and `geometry` branches.

When testing with your lattice file, I got a warning from `scipy` while loading the file. I don't know why, and I cannot fo anything since it's internal to `scipy`. Apparently this has no consequence.

simoneliuzzo (approved):

lfarv:
> may be plot_geometry could be modified to provide a 3D plot in case of 3D coordinates?

That's done. There is a new plot3d bool keyword triggering a 3d plot, whatever the lattice. Here is an example for a small test lattice:

Figure 11

I need a new approval, sorry…

simoneliuzzo (approved):

Issue #686, opened 2023-10-31
pyAT MAT file V7.3


lnadolski:
Dear all

the function at.load_lattice does not work for a V7.3 mat lattice file.

Per default, the default matlab is still configured with -V7 (if not changed in the user's preference)

Example to generate a 7.3 version matlab file
The matlab format has changed significantly

`save("newLattice.mat", "-v7.3")`

Best regards,

Laurent.

lfarv:
Hello @lnadolski,

In Matlab, `atloadlattice` works as expected with V7.3 lattice files: `load` (and consequently `atloadlattice`) automatically detects the format.

However in python, we rely on SciPy which state: _"v4 (Level 1.0), v6 and v7 to 7.2 matfiles are supported."_. In addition, SciPy mentions:

_"You will need an HDF5 Python library to read MATLAB 7.3 format mat files. Because SciPy does not supply one, we do not implement the HDF5 / 7.3 interface here."_

Decoding the private Matlab structure starting with a plain HDF5 reader looks difficult. So for file exchange with python, you must stick to the V7 format.

lnadolski:
Dear @lfarv 

This is good to know.
Maybe we should add a note in pyAT.

lfarv:
I found the [hdf5storage](https://github.com/frejanordsiek/hdf5storage) project which claims to support V7.3 .mat files. The project does not seem to be alive any more, but it may work. I would not like to have PyAT depend on this package, but if it works (still to be checked), we could have PyAT check that if the package is available (the user explicitly installed it), it will use it, otherwise it will stick to SciPy.

lfarv:
Forget [hdf5storage](https://github.com/frejanordsiek/hdf5storage), I could not make it work: it hangs forever…

T-Nicholls:
Hello both,

When I try and load a v7.3 `.mat` lattice file using `at.load_lattice` I get the error: `NotImplementedError: Please use HDF reader for matlab v7.3 files, e.g. h5py`. To me that's clearly telling the user that the v7.3 file is the issue; however, if we wanted to be more specific we could catch that error from `scipy` and reraise it with a message telling the user that they need to save the lattice file in Matlab using an earlier version.

Alternatively, if there is significant demand, we could write new functions specifically to support v7.3 files. I briefly tried loading a v7.3 file using `h5py` and it seemed feasible as it, at least partially, decodes the data but it's significantly more convoluted than using `scipy`:
```
>>> import h5py
[...]

PR #995, opened 2025-09-23
pyat load_mat for matlab files v7.3 hdf5

oscarxblanco:
Dear all,
this PR proposes to update the pyat `load_mat` method for matlab files.
This is done in order to support .mat files with version 7.3, see issue #686. I believe they are useful when a large amount of data is stored.
**h5py would need to be added to the pyat dependencies.**

I have tested these modifications in a local installation without problems, but, for sure it would need to be confirmed by others.
This is what I do in matlab
```matlab
save("newLattice.mat","THERING", "-v7.3")
```
while in python there is no change
```python
ring = at.load_mat('newLattice.mat')
```

`save_mat` is not modified, therefore the output .mat file from pyat is still the same.
Please, let me know if this is needed or useful, or if it is preferred to leave matlab file loading as it is.

### DIFF
--- a/pyat/at/load/matfile.py
+++ b/pyat/at/load/matfile.py
@@ -114,13 +114,16 @@ def _matfile_generator(
 
     def mcleanhdf5(data: any) -> any:
         matlab_class = data.attrs["MATLAB_class"]
-        if matlab_class == b"char":
+        if matlab_class == b"struct":
+            # Return a dict from recursion
+            dataout = {f: mcleanhdf5(data[f]) for f in data.keys()}
+        elif matlab_class == b"char":
             # Convert to string
             dataout = "".join(chr(i) for i in np.asarray(data).flatten())
-        if matlab_class == b"double":
+        else:
+            # e.g. matlab_class == b"double":
+            # Remove any surplus dimensions in arrays.
             dataout = np.squeeze(np.asarray(data))
-        if matlab_class == b"struct":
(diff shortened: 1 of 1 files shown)


lfarv:
@oscarxblanco : After rebasing on the new master branch, the tests succeed. However I have a problem reading the following .mat file:
[test73.mat.txt](https://github.com/user-attachments/files/22618226/test73.mat.txt)

This is the `hmba.mat` test file saved in Matlab as a v7.3 file.
It seems that in `mcleanhdf5` you have to handle the `b'struct'` Matlab class.

pyat/at/load/matfile.py
```diff
@@ -84,41 +85,79 @@ def _mat_encoder(v):
...
+    def mcleanhdf5(data: any) -> any:
+        matlab_class = data.attrs["MATLAB_class"]
+        if matlab_class == b"char":
+            # Convert to string
+            dataout = "".join(chr(i) for i in np.asarray(data).flatten())
+        if matlab_class == b"double":
+            dataout = np.squeeze(np.asarray(data))
+        if matlab_class == b"struct":
+            dataout = {f: mcleanhdf5(data[f]) for f in data.keys()}
+        return dataout
```
lfarv, line 124: If for some custom attribute the Matlab class in not one of the expected ones, this crashes (`dataout` is undefined). It would be wise to add a default behaviour, returning `np.squeeze(np.asarray(data))`.

Issue #997, opened 2025-09-25
Mad8 parser does not parse multipole definitions


kandre2:
Hello,

There is an issue when converting a lattice from mad8 to AT.

Multipoles in mad8 are defined as:

OC3FL: MULTIPOLE, K3L=KOC3FL
DEC1R: MULTIPOLE, K4L=KDEC1R

whereas KNL and KSL are used in madx.

So the conversion from mad8 to AT does not parse multipole elements correctly whereas it is correctly done from madx to AT.

lfarv:
@kandre2 : Sorry,I missed that in the MAD8 documentation. It's corrected in #999

PR #999, opened 2025-09-27
Fix multipole in MAD8

lfarv:
This corrects a bug in reading MAD8 files reported in #997


simoneliuzzo (approved):
looks correct to me.
I have tested on a lattice with octupoles and decapoles and the gradients appear correctly in AT

PR #1001, opened 2025-10-20
Define integrated strengths in multipoles

lfarv:
This PR defined 2 new properties of `Multipoles` and derivatives:
- `IntegratedPolynomA = Length * PolynomA`: skew strength 
- `IntegratedPolynomB = Length * PolynomB`: straight strength

### DIFF
--- a/pyat/at/lattice/elements/magnet_elements.py
+++ b/pyat/at/lattice/elements/magnet_elements.py
@@ -75,7 +75,7 @@ def _warn(doc: str) -> str:
     )
 
 
-def _k_property(attrname: str, order: int, doc: str):
+def _k_property(attrname: str, order: int, doc: str) -> property:
     """Return a property for a field component in MAD convention."""
 
     def kget(mult: ThinMultipole) -> float:
@@ -90,7 +90,7 @@ def _k_property(attrname: str, order: int, doc: str):
     return property(kget, kset, doc=doc)
 
 
-def _s_property(attrname: str, doc: str):
+def _s_property(attrname: str, doc: str) -> property:
     """Return a property for the strength of the main field component."""
 
     def kget(mult: ThinMultipole) -> float:


lfarv:
@JeanLucPons:
The real field definition is `PolynomA/B`, there is no change here. All the new properties introduced here are just views into this polynomial. So the "active" functions of a magnet are the ones for which `PolynomA/B[order]` is non-zero. A way to identify them could be:
```python
[n for n, v in enumerate(magnet.PolynomB[:magnet.MaxOrder+1]) if v != 0.0]
```
which gives a list of "active" orders.

lfarv:
Oh, are you talking of the way the magnet is printed ? That's something else. At the moment, properties are normally not printed (K is an exception).

JeanLucPons:
Yes, If property are well printed, then i can find what functions are available a bit as for KickAngle.

lfarv:
@JeanLucPons : Ah, hardware ! Well, up to now, AT totally ignores hardware, but you are free to add any attribute to elements, concerning anything you like. Additional attributes (other than the ones documented) are totally ignored and harmless.

In addition, it would be possible to decide that some attributes become part of AT, even if they are not used, just to freeze their names and avoid name conflicts.

lfarv:
@JeanLucPons : to illustrate what I was writing above, here is an example:
```
>>> from at import Sextupole
>>> 
>>> # Create a sextupole
>>> sx = Sextupole("SX1", 0.5, 100.0)
>>> 
>>> # Add some hardware-specific attributes
>>> sx.ML_Kn0 = "SR/PS/COR-H/SX1"   # Horizontal corrector
>>> sx.ML_Ks0 = "SR/PS/COR-V/SX1   # Vertical corrector
>>> sx.ML_Ks1 = "SR/PS/COR-SKEW/SX1"   # Skew quadrupole
>>> 
>>> print(sx)
Sextupole:
       FamName: SX1
        Length: 0.5
        ML_Kn0: SR/PS/COR-H/SX1
        ML_Ks0: SR/PS/COR-V/SX1
        ML_Ks1: SR/PS/COR-SKEW/SX1
      MaxOrder: 2
   NumIntSteps: 10
      PolynomA: [0. 0. 0.]
      PolynomB: [  0.   0. 100.]
>>> 
```
There you can see the hardware capabilities.
These "custom" attributes are totally ignored, and instead of a string, you can put any object like a device proxy

JeanLucPons:
I have a qestion concerning this `tan()`. Today:
```
    if (KickAngle) {   /* Convert corrector component to polynomial coefficients */
        B[0] -= sin(KickAngle[0])/le;
        A[0] += sin(KickAngle[1])/le;
    }
```
It uses a sin() so if you consider the pathlength for the hypotenuse, this should be a sin().
I know that it does not make big difference but what is correct ?

lfarv:
@JeanLucPons 

> overriding `def items(self):`

I strongly discourage using `items()` for that. Among other things, `items()` is used in file storage of elements, and defines what needs to be stored, so that the element can be rebuilt identically. Normally, this is exactly the list of attributes (but not properties, which are derived from the attributes). The only exception is `K`, because for historical reasons in Matlab `K` and `PolynomB` are different attributes. So `K` is added to the list and both are stored.

So overriding `items` may cause storage problems…

But there are many other ways to answer your question

Issue #1006, opened 2025-10-28
`atlinopt4` with twiss_in ignores dp


oscarxblanco:
Dear all,

in AT matlab I want to calculate the twiss functions of a transport line at different energies, but, using the parameter `dp` in `atlinopt4` combined with `twiss_in` doesn't seem to work, it always returns the same twiss output.

Best regards,
o

swhite2401:
@oscarxblanco , have you tried instead to change initial orbit condition? Does this work?

PR #1007, opened 2025-10-28
atlinopt4 with twiss_in and dp

oscarxblanco:
Dear all,

this is a possible work around issue #1006 where the twiss functions of a transport line at different energies did not change when changing the energy using `dp`.

The issue is the variable `orbitin`, which is overwritten several times along the calculation.

- The parameter orbit was read twice (by dpargs and orbitin). This duplication is now removed.
- For a circular machine (no twiss_in) the orbit is now explicitly passed to findorbit4, and all other modifications are just to limit the maximum length of the line.
- For a transfer line (passing twiss_in) the orbit is unmodified. An auxiliary orbit is created to calculate the effect of dispersion. 

With this logic the call to findM44 is unchanged for circular machines or transfer lines.
https://github.com/atcollab/at/blob/a22c9b19acff5b0a6311bcbe2057186b0d9ef003/atmat/atphysics/LinearOptics/atlinopt4.m#L141

### DIFF
--- a/atmat/atphysics/LinearOptics/atlinopt4.m
+++ b/atmat/atphysics/LinearOptics/atlinopt4.m
@@ -99,7 +99,7 @@ NE = length(ring);
 [get_w,varargs]=getflag(varargs,'get_w');
 [twiss_in,varargs]=getoption(varargs,'twiss_in',[]);
 [dp,varargs]=getoption(varargs,'dp',0.0);
-[dpargs,varargs]=getoption(varargs,{'orbit','dct','df'});
+[dpargs,varargs]=getoption(varargs,{'dct','df'});
 [orbitin,varargs]=getoption(varargs,'orbit',[]);
 [coupled,varargs]=getoption(varargs,'coupled',true);
 [mkey,varargs]=getoption(varargs,'mkey','M');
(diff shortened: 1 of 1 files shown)


swhite2401 (approved):
This looks ok to me, @lfarv any comments?

lfarv (approved):
Ok for me also.

wei0852:
Dear all,
A new bug was introduced in this pull request  causing an error in the dispersion calculation.

```
  -  [orbitP,o1P]=findorbit4(ring,dp+0.5*DPStep,refpts,orbitin,'XYStep',XYStep, 'strict', -1);
  -  [orbitM,o1M]=findorbit4(ring,dp-0.5*DPStep,refpts,orbitin,'XYStep',XYStep, 'strict', -1);
 +   [orbitP,o1P] = findorbit4(ring,dp+0.5*DPStep,refpts, ...
 +                             'orbit',orbitin,'XYStep',XYStep,'strict',-1);
 +   [orbitM,o1M] = findorbit4(ring,dp-0.5*DPStep,refpts, ...
 +                             'orbit',orbitin,'XYStep',XYStep,'strict',-1);
```
In the previous version, `orbitin` was not correctly passed into `findorbit4`, so the function used the default value `[]`:

```
[orbitin,varargs]=getoption(varargs,'orbit',[]);
```
which happened to allow the calculation at line 83 to proceed:
```
 orbitin=xorbit_dp(ring,dp,varargs{:});
```
However, after the modification, the `orbitin` inside `findorbit4` becomes `[0 0 0 0 0 0]`, and the calculated orbit is always 0:

```
if isempty(refpts)
    % return only the fixed point at the entrance of RING{1}
    orb4=orbitin(1:4,1);
else
    orb6 = linepass(ring,orbitin,refpts,args{:});
    orb4 = orb6(1:4,:);
```

The logic inside findorbit4 should be further modified, since the previous correct behavior was accidental.

oscarxblanco:
Dear @wei0852 , I'll have a look.

oscarxblanco:
Dear @wei0852 , you are right.
My mistake was to assume that findorbit4 behaviour wouldn't change after fixing the error on the orbitin variable.
I did a test and it seems enough to remove the orbitin parameter from the call to findorbit when twiss_in is not given.

Could you create an issue ? otherwise I will do it.
o

Issue #1012, opened 2025-11-05
Zero peak field wiggler not behaving as drift


SebastienJoly:
Hello,

I have found a strange behaviour when dealing with undulators. I naively thought that an undulator with a zero peak wiggler field would behave like a drift.
In some cases, they do, and in others, not. I could not pinpoint exactly which parameter pair (length, wiggle_period) was responsible for that. But I could see that it was related to the rounding error (e.g. if length / wiggle_period is not exactly equal to an integer, then the bug occurs)

I have created a simple example below where I generate a FODO cell. Then, I replace the last drift by an undulator with zero peak field. You can see that the resulting tunes will be different from the previous ones.

Does anyone know what could cause this issue? Is the at.Wiggler element trustworthy?

```
import numpy as np
import at

QF = at.Multipole("QF", 0.5, poly_b=np.array([0,1.2]), poly_a=np.zeros(2))

drift_length = 0.5
Dr = at.Drift("Dr", drift_length)
HalfDr = at.Drift("Dr2", drift_length/2)
Dr3 = at.Drift("Dr3", 0.075)
Dr4 = at.Drift("Dr4", 0.425)

QD = at.Quadrupole("QD", 0.5, -1.2)
Bend = at.Dipole("Bend", 1, 2 * np.pi / 20)

FODOcell = at.Lattice(
    [HalfDr, Bend, Dr, QF, Dr, Bend, Dr, QD, Dr3, Dr4],
    name="Simple FODO cell",
    energy=1e9,
)

print(FODOcell.radiation_parameters().tunes)
print(FODOcell.circumference)

und = at.Wiggler('und', length=Dr4.Length, wiggle_period=Dr4.Length/100, b_max=0)
FODOcell[-1] = und

print(FODOcell.radiation_parameters().tunes)
print(FODOcell.circumference)
```

oscarxblanco:
Dear @SebastienJoly ,

the length of `und` is wrong. You are replacing a half drift of 0.25 m with an undulator of 0.5 m.
Fixing the length makes the difference in tune disappear. 
``` python
...
und = at.Wiggler('und', length=0.25, wiggle_period=0.005, b_max=0)
...
```
```bash
>>> exec(open('example.py').read())
[0.26099678 0.2458889         nan]
[0.26099678 0.2458889         nan]

```

oscarxblanco:
@SebastienJoly , I finally could reproduce what you were saying, I got a little confused because you edited the code inside the initial message.

[...]

Issue #894, opened 2025-01-21
Wiggler Energy attribute error


T-Nicholls:
### Cause

If we create even the most basic Wiggler in Matlab it has the `Energy` attribute:
```
>> w = mywig("MPW15")

w = 

  struct with fields:

       FamName: "MPW15"
    PassMethod: 'GWigSymplecticPass'
        Length: 0
         Class: 'Wiggler'
            Lw: 0
          Bmax: 0
        Energy: 0
         Nstep: 5
         Nmeth: 4
            Bx: [6×0 double]
            By: [6×1 double]
        NHharm: 1
        NVharm: 0
```
So, if we then try to load a `.mat` lattice file containing a Wiggler in Python we get this error:
```
TypeError: at.lattice.elements.LongElement.__init__() got multiple values for keyword argument 'Energy'
```
This is because `Energy` was removed from the build attributes of the Wiggler element in #822 and so `Energy` doesn't get popped here anymore:
https://github.com/atcollab/at/blob/c8c58446715cf263fa94afc7a8990381e707138b/pyat/at/load/utils.py#L366
and so it isn't passed as the lowercase `energy` argument anymore. Instead, it's passed in as `Energy` with the other miscellaneous attributes in `kwargs`.
However, lowercase `energy` is also passed to `super().__init__()` as `Energy`:
https://github.com/atcollab/at/blob/c8c58446715cf263fa94afc7a8990381e707138b/pyat/at/lattice/elements.py#L1416-L1427
so we end up with an error because we are trying to pass two `Energy` keyword arguments to the constructor.

### Replication

Rather than having to boot-up Matlab and save a new lattice file, we can more easily replicate this directly from Python by doing:
```
>>> Wiggler("MPW15", 1.44, 0.06, 0.0, Energy=0.0) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jzs47954/git/at/pyat/at/lattice/elements.py", line 1424, in __init__
    super().__init__(
TypeError: at.lattice.elements.LongElement.__init__() got multiple values for keyword argument 'Energy'
```

### Solution

The first question is why was `Energy` removed from the build attributes of the Wiggler element in #822 and a default value of `energy=0.0` added?
Assuming there was a good reason for doing that and so it shouldn't be reverted, then a simple solution would be to just pop `Energy` from the `kwargs` dict before that is passed to `super().__init__()`.
[...]

PR #1013, opened 2025-11-07
Fix bug wiggler rounding

oscarxblanco:
Dear all,
this PR solves issue #1012 where there was in inconsistent behaviour of a wiggler set to Bmax=0 when compared to a drift of the same length.

The number of iterations is now calculated using `round`. The user must take into account that the ratio between total wiggler length and the wiggle_period should be an integer.

In addition, wiggler elements cannot be tracked alone. I include a check on the definition of gamma that is only set by the lattice, according to pull request #894 .

### DIFF
--- a/atintegrators/GWigSymplecticPass.c
+++ b/atintegrators/GWigSymplecticPass.c
@@ -25,6 +25,7 @@
 #include "atlalib.c"
 #include "gwig.c"
 
+
 struct elem {
     double Energy;
     double Length;
@@ -119,6 +120,7 @@ ExportMode struct elem *trackFunction(const atElem *ElemData,struct elem *Elem,
 
 {
     double gamma;
+    double wig_l_ratio, len_tolerance;
 
     if (!Elem) {
         double *R1, *R2, *T1, *T2;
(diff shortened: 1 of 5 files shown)


SebastienJoly:
I am confused, I confirm that the tune difference still appear in your example. Moreover, I get a difference of the order of 1e-8 and 1e-13 for the first and last element of the list obtained from:
`
print(zout2.T-np.ravel(zout).T)
`
The difference drops to 1e-20 when I use an undulator length which does not lead to a tune difference.

lfarv (approved):
OK

oscarxblanco:
Hi @SebastienJoly , there is still something we are doing differently. I get the correct tune, and a single particle tracking difference of 10^{-22}. See here below
```python
tune with drift          [0.32536863 0.2625984         nan]
tune with wiggler Bmax=0 [0.32536863 0.2625984         nan]
[[4.250e-06]
 [1.000e-05]
 [0.000e+00]
 [0.000e+00]
 [0.000e+00]
 [2.125e-11]]
Particle tracking difference : [[-8.47032947e-22  0.00000000e+00  0.00000000e+00  0.00000000e+00
   0.00000000e+00 -2.94036858e-25]]
```

With respect to the output you show, the last line says there is an error of 10^{-8} in the single particle tracking. I copy it here below for clarity
```
[[4.250e-08 0.000e+00 0.000e+00 0.000e+00 0.000e+00 2.125e-13]]
```
I can reproduce your result if I use the master branch, where there is a mistake on the number of step during the integration. So, my suspicions are still about the installation or compilation of the pass method. Could you try to add one line to the pass method and check the output ? It is the line I wrote in a previous message.
```c
printf("Nstep=%d, dl=%.6g,PN=%d,Nw=%d,Lw=%.3g\n",Nstep,dl,pWig->PN,pWig->Nw,pWig->Lw);
```
In that way we will know that the code you are editing is  the one you execute, and we could see some of the integrator parameters. Also, you could also check that the `round` function is implemented in the gwig.c file.

Other than that, I have tested the installation with pip gpu flags, pip mpi flags, just pip inside my copy of the repo, and all of them give me the right tune.

SebastienJoly (approved):

PR #1014, opened 2025-11-07
bugfix: switch to BndMPoleSymplectic4RadPass in ohmi envelope

swhite2401:
Presently the pass methods not implemented for ohmi envelope (exact bends) switch to StrMPoleSymplectic4RadPass, this is not correct as it transforms dipoles into multipoles resulting in wrong radiation parameters returned by ohmi_envelope()
Now the passmethods are correctly switched to BndMPoleSymplectic4RadPass.


lfarv (approved):
Absolutely right ! This was indeed a bug, thanks for pointing it out  and correcting.

Issue #1015, opened 2025-11-07
Wiggler tracking


oscarxblanco:
Dear all,
while debuggin issue #1012 , I noticed that tracking a single wiggler is not possible in pyat. For example,
```python
>>> und = at.Wiggler('und', length=0.425, wiggle_period=0.425/100, b_max=0,Nmeth=2,Nstep=10,energy=1e9)
>>> zin = np.zeros((6,1))
>>> zin[1] = 10e-6
>>> und.track(zin)
array([[nan],
       [nan],
       [nan],
       [nan],
       [ 0.],
       [nan]])
```
When the energy is removed from the wiggler parameters, gamma has a wrong value and produces nan from an undefined division in the GWigSymplectic pass method implementation. See the following lines.
https://github.com/atcollab/at/blob/780657c6111176673321e2c75623d54670619947/atintegrators/gwig.c#L157
where `pWig->Po` has been previously set to:
https://github.com/atcollab/at/blob/780657c6111176673321e2c75623d54670619947/atintegrators/gwig.c#L40

This could be worked around by including the undulator as a single element in a ring. For example,
```python
ring0 = at.Lattice([und],energy=1e9)
zin = np.zeros((6,1))
zin[1] = 1e-5
zout,*_ = ring0.track(zin)
```

However, it is a strange behavior for a pyat element, and one could easily get distracted by the thought of having missed something important on the wiggler parameters. 

Is there a way to raise an error when tracking a single wiggler ? or could the energy parameter be restored, which I know could be confusing when having different energy values for the lattice and the wiggler ?

PR #1016, opened 2025-11-10
Check elem track needs energy

oscarxblanco:
Dear all,
this PR addresses issue #1015 , where tracking a well defined wiggler element returns nans because the energy is not  defined.

As explained by @lfarv in issue #1013 , the energy could be pass in the track method, therefore, this PR checks that the energy parameter is used when any of the element has any of the two wiggler pass methods, GWigSymplecticPass or GWigSymplecticPass.

Additionally I changed numpy  to np, I run black, isort and solved some flake8 messages.

### DIFF
--- a/atintegrators/atelem.c
+++ b/atintegrators/atelem.c
@@ -83,7 +83,6 @@ double atEnergy(double ringenergy, double elemenergy)
             return elemenergy;
         else {
             atError("Energy not defined.");
-            check_error();
             return 0.0;   /* Never reached but makes the compiler happy */
         }
 }


atmat/attrack/elempass.m
```diff
@@ -19,13 +19,22 @@
...
 [particle,varargs]=getoption(varargs,'Particle',[]);
 [methodname,varargs]=getoption(varargs,'PassMethod',elem.PassMethod); %#ok<ASGLU>
 if ~isempty(particle)
     props.Particle=particle;
 end
 
+% check pass methods that require energy
+needs_energy = {'GWigSymplecticPass', 'GWigSymplecticRadPass'};
+if any(strcmp(needs_energy,methodname)) && varenergy == 0
+    error('Energy needs to be defined');
```
lfarv: I think that all the *RadPass methods need energy. Plus the wiggler,  because of the absolute field definition in T… And also RFCavityPass.

Unfortunately, the pass method names must be hard-coded. So It has to be revised when new pass methods are introduced.

But I guess it's better than nothing. So OK for me.

atintegrators/GWigSymplecticPass.c
```diff
@@ -121,6 +121,11 @@ ExportMode struct elem *trackFunction(const atElem *ElemData,struct elem *Elem,
 {
     double gamma;
 
+    if (Param->energy == 0){
+            atError("Energy needs to be defined. Check lattice parameters or pass method options.\n");
+            check_error();
+    }
+
```
lfarv: A comment before you replicate this in all the passmethods.

In Matlab, the energy may be given by an element attribute. So you have to check the result of `atEnergy`/`atGamma`, and it has to be checked before the memory allocation:
```C

    if (!Elem) {
        double *R1, *R2, *T1, *T2;
        double *By, *Bx;
        double Ltot, Lw, Bmax, Energy;
        int Nstep, Nmeth;
        int NHharm, NVharm;

        Ltot = atGetDouble(ElemData, "Length"); check_error();
        ...
        /* Optional fields */
        Energy=atGetOptionalDouble(ElemData,"Energy",Param->energy); check_error();
        ...

        gamma = atGamma(Param->energy, Energy, Param->rest_energy);
        if (gamma == 0) {
            atError("Energy needs to be defined. Check lattice parameters or pass method options.\n");
            check_error();
        }

        Elem = (struct elem*)atMalloc(sizeof(struct elem));
        ...
    }
    else {
        gamma = atGamma(Param->energy, Elem->Energy, Param->rest_energy);
    }
```
Note that the 2nd gamma computation is done because `Param->energy` may be used for energy ramping, and may change on each call. But there I don't think it needs to be check again.

lfarv:
@oscarxblanco : this looks correct.

Another option is to move the test inside the python `atEnergy` (similarly to the Matlab version) and follow the call with a `check_error`. This would make Matlab and python versions more similar and avoid a double test in Matlab. But `atEnergy`/`atGamma` still have to be moved to where you just put them. Up to you !

lfarv:
@oscarxblanco : after checking `RFCavityPass`, I notice that unlike all other elements, `RFCavityPass`, as it is in the master branch, does use the element `Energy` field if necessary.  Do we keep a special case for it?

PR #1017, opened 2025-11-12
Plot observables

lfarv:
This PR introduces a new plot function: `plot_observables`. It plots the values of Observable objects along a lattice. It generalises `plot_linear` by giving access to any ElementObservable (geometry, RDT, trajectories, transfer matrix elements…) and offer a very simple input. It has a dual y-axis capability, takes care of plot and axis labels and is highly configurable for line properties and labels.

It can create a new figure or plot in predefined axes, possibly as part of a global panel.

It plots a smooth curve if `refpts` is `at.All` or plots markers if `refpts` selects element positions, with the possibility of superimposing both (see the example).

Example 1: plot trajectories on the left axis and module of RDTs on the right axis. Don't ask me how useful this plot can be, it's just a demo!.

```python
# Define the trajectory starting point
orbit=np.zeros(6)
orbit[0] = 0.001
orbit[2] = 0.001

# Trajectory observables
obstraj = ObservableList(
    [
        # Plot curves
        TrajectoryObservable(at.All, axis="x"),
        TrajectoryObservable(at.All, axis="y"),
        # Plot markers at specified locations
        TrajectoryObservable(at.Quadrupole, axis="x", label="_"),
        TrajectoryObservable(at.Quadrupole, axis="y", label="_"),
    ],
    r_in=orbit,
)

# RDT observables
obsrdt = ObservableList(
    [
        # Take the module of RDTs
        RDTObservable(at.All, "h20000", postfun=np.abs),
        RDTObservable(at.All, "h00200", postfun=np.abs),
    ]
)

# Plot
axl, axr = plot_observables(
    ring, obstraj, obsrdt, title="Test", s_range=[0.0, 10.0]
)
```
traj_rdt

### DIFF
--- a/pyat/at/plot/__init__.py
+++ b/pyat/at/plot/__init__.py
@@ -1,11 +1,11 @@
-"""AT plotting functions"""
+"""AT plotting functions."""
 
 try:
     # noinspection PyPackageRequirements
     import matplotlib
-except (ImportError, RuntimeError) as exc:
+except (ImportError, RuntimeError):
     print("matplotlib is unavailable => plotting is disabled.")
-    print("To enable plotting functions, run \"pip install matplotlib\".")
+    print('To enable plotting functions, run "pip install matplotlib".')
 else:
     from .synopt import *
     from .generic import *
--- a/pyat/at/plot/generic.py
+++ b/pyat/at/plot/generic.py
@@ -1,19 +1,22 @@
-"""AT generic plotting function"""
+"""AT generic plotting function."""
 from __future__ import annotations
+
+__all__ = ["baseplot"]
+
+from collections.abc import Callable
 from itertools import chain, repeat
+
 # noinspection PyPackageRequirements
 import matplotlib.pyplot as plt
-from typing import Callable
+
 from .synopt import plot_synopt
 from ..lattice import Lattice
 
 SLICES = 400
 
-__all__ = ['baseplot']
-
 
 def baseplot(ring: Lattice, plot_function: Callable, *args, **kwargs):
-    """Generic lattice plot
+    """Generic lattice plot.
 
     :py:func:`baseplot` divides the region of interest of ring into small
     elements, calls the specified function to get the plot data and calls
@@ -76,7 +79,7 @@ def baseplot(ring: Lattice, plot_function: Callable, *args, **kwargs):
 
     def plot1(ax, yaxis_label, x, y, labels=()):
         lines = []
-        for y1, prop, label in zip(y, props, chain(labels, repeat(None))):
+        for y1, prop, label in zip(y, props, chain(labels, repeat(None)), strict=False):
             ll = ax.plot(x, y1, **prop)
             if label is not None:
                 ll[0].set_label(label)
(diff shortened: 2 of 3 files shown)


lfarv:
Exemple 2

This demonstrates the use of the `postfun` post-processing attribute of observables to plot the beam envelopes for arbitrary emitttance values:

```python
# Define the emittances
emit_x = 130.0e-12
emit_y = 10.0e-12

# beam size computation
sigma_x = lambda x: 1.0e6 * np.sqrt(x * emit_x)  # result in um
sigma_y = lambda y: 1.0e6 * np.sqrt(y * emit_y)  # result in um

# Observables
obsenv = ObservableList(
    [
        LocalOpticsObservable(
            at.All, "beta", plane="x", postfun=sigma_x, label=r"$\sigma_x$"
        ),
        LocalOpticsObservable(
            at.All, "beta", plane="y", postfun=sigma_y, label=r"$\sigma_y$"
        ),
    ]
)

# Plot
(ax5,) = plot_observables(ring, obsenv, title="Beam envelopes")
ax5.set_ylabel(r"Beam size [${\mu}m$]")
```
Here we fix the curve labels since the default value is "<lambda>($\beta_x$)".

envelope

oscarxblanco:
Dear @lfarv , thank you. The example works for me. In my opinion, this solves several custom plots I would normally do in a custom script. For me this could be approved to see how it behaves with more complicated studies.

Please, consider the text here below as an opinion:

I see you made a function called `evaluate` which seems quite useful and that I would think one should have access to.

Suppose that after one does a plot, one needs the result to do something else. The most recent example I have in mind is scanning the tune vs energy, make a plot, and then from the raw data get the energy at which the tune crosses a given resonance, for example 0.5. The plot is a first exploratory step when trying to understand, and later I would include the routine to check the resonance crossing automatically for as many lattices as needed without the plot. The plot, however, is still needed to further exploring the effect of errors in some lattices or changes in it.

I wonder if one could somehow get the data and the plot separately. Maybe it is better to expose both plot and evaluate to the user, so the user can use what is more adequate.

oscarxblanco (approved):

lfarv:
@oscarxblanco : I think that with all the features of the `Observable` and `ObservableList` you have what you are looking for: the plot functions (both `plot_response` and `plot_observables`) do nothing more than evaluating the observables and plotting the data. For instance with `plot_observables`, taking the 1st example above in this discussion:

```python
# Define the trajectory starting point
orbit=np.zeros(6)
orbit[0] = 0.001
orbit[2] = 0.001

# Trajectory observables
obstraj = ObservableList(
    [
        # Plot curves
        TrajectoryObservable(at.All, axis="x"),
        TrajectoryObservable(at.All, axis="y"),
    ],
    r_in=orbit,
)

# RDT observables
obsrdt = ObservableList(
    [
        # Take the module of RDTs
        RDTObservable(at.All, "h20000", postfun=np.abs),
        RDTObservable(at.All, "h00200", postfun=np.abs),
    ]
)

# Replace the plot by a simple evaluation
datatraj = obstraj.evaluate(ring=ring)

# get the x position at the entrance of all elements
xtraj = datatraj[0]
# you can get the same with
xtraj = obstraj[0].value

# get the y position at the entrance of all elements
ytraj = datatraj[1]
```
which is exactly what is plotted.

Similarly:
```
datardt = obsrdt.evaluate(ring=ring)

# get the module of h20000 at the entrance of all elements
h20000 = datardt[0]

# get the module of h00200 at the entrance of all elements
h00200 = datardt[1]
```
which are the data plotted on the right axis.

Look at the entry for [observables](https://at.readthedocs.io/en/latest/p/api/at.latticetools.observables.html#at.latticetools.observables.Observable)  and [ObservableList](https://at.readthedocs.io/en/latest/p/api/at.latticetools.observablelist.html#at.latticetools.observablelist.ObservableList) in the documentation and at the [example notebook](https://at.readthedocs.io/en/latest/p/notebooks/observables.html#observables). Of course we can add any additional feature that you would consider useful.

PR #1004, opened 2025-10-26
Plot any observable as a function of a variable

lfarv:
This PR introduces a new plot function. It plots the value of an `ObservableList` (one or several observables) as a function of a variable.

This PR uses the modifications introduced in #1003. It also take profit of the new magnet properties introduced in #1001.

PR #1019, opened 2025-11-18
Fix matching

lfarv:
This fixes a problem in matching revealed in an [example](https://github.com/atcollab/at/pull/1017#issuecomment-3543284001) suggested by @oscarxblanco: matching $\delta$ to fulfil a tune constraint.

The present implementation of `match` freezes `dp`, `dct`, and `df`, preventing from using them as `EvaluationVariable` (introduced in #1004).

A constraint on `dp`, `dct` or `df` can be set on the ObservableList input instead of as a `match` keyword.

### DIFF
--- a/pyat/at/lattice/variables.py
+++ b/pyat/at/lattice/variables.py
@@ -508,6 +508,8 @@ class VariableBase(abc.ABC):
         then restored when exiting the context.
 
         Keyword Args:
+            initial:    If :py:obj:`True`, clear the history and set the variable
+              initial value
             \*\*setkw:  Keyword arguments to be passed to the *setfun* function.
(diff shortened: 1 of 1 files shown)


lfarv:
Ready for reviewing. However I have a question:

After the modifications, we end up with a private function `_match` which is a lattice-independent matching function. It seems useful to make it public by renaming it. The main difference with `match` is that the present `match` makes an optional copy of the lattice before calling the more general `_match` function.

If we want to keep `match` unchanged for compatibility, what could be the name of the general lattice-independent matching function ?

Or are we allowed to change the name of the `match` function (still in `at.future`) ?

swhite2401:
This looks ok to me but I let @oscarxblanco approve

PR #1023, opened 2025-12-04
Change output order of simple ring lattice

lcarver:
Hello,

I am doing some housekeeping before I make a PR for a major upgrade of the beam loading element.

In the simple_ring lattice, the output order of the elements is

`simple_ring = Lattice([RF_Cavity, Linear_Element, NonLinear_Element, SimpleRadiation, QuantumDiffusion])`

This is not so ideal, as if you take some beam diagnostics at the end of the turn, (either by tracking in a for loop or by appending a beam monitor), then you are sort of taking the diagnostics half way through the turn, which results in an offset of the `delta` by `-U0/energy`. NOTE: It is NOT wrong, it is just a little misleading.

This PR reorders the output of simple_ring, such that all linear transformations (linear element and non linear element) occur at the end of the turn. 

I also had to update the tests to pass this.

@lfarv @swhite2401  i had thought about a similar modification to the fastring outputs, but there it is trickier because the radiation damping and energy loss is also wrapped up in the linear element with the M66 matrix, and on top of that you can slice a ring into several sections. So for now I leave it be, as I think a rewrite of fast_ring is both planned and long overdue.

### DIFF
--- a/pyat/test/test_physics.py
+++ b/pyat/test/test_physics.py
@@ -430,10 +430,10 @@ def test_quantdiff(hmba_lattice):
 def test_simple_ring():
     ring = physics.simple_ring(6e9, 844, 992, 0.1, 0.2, 6e6, 8.5e-5)
     assert_equal(len(ring), 5)
-    
-    assert_equal(ring[at.SimpleQuantDiff][0].PassMethod, "SimpleQuantDiffPass")
+    num_cavities = len(ring[at.RFCavity])
+    assert_equal(ring[num_cavities+1].PassMethod, "SimpleQuantDiffPass")
     ring.disable_6d()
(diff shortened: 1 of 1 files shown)


pyat/test/test_physics.py
```diff
@@ -430,9 +430,9 @@ def test_quantdiff(hmba_lattice):
 def test_simple_ring():
     ring = physics.simple_ring(6e9, 844, 992, 0.1, 0.2, 6e6, 8.5e-5)
     assert_equal(len(ring), 5)
-    assert_equal(ring[-1].PassMethod, "SimpleQuantDiffPass")
+    assert_equal(ring[2].PassMethod, "SimpleQuantDiffPass")
```
swhite2401: Shouldn't the index here be number of cavities + 1?

swhite2401 (approved):

PR #1025, opened 2025-12-05
Upgrade and major bug fix to the beam loading module

lcarver:
There has been a major bug present in the beam loading module since it was first written. To summarise, the beam induced voltage used in the cavity feedback was taken at the wrong point. For a single harmonic system, (like the work published in IPAC : https://inspirehep.net/files/cc5691837fadff23276bd1022058a71e ) the passmethod was working as expected and this bug was not visible, however when a harmonic cavity was present, the phase of the beam induced voltage became incorrect, causing unstable results. This was not the correct behaviour.

This PR completely remodels the computation to compute the beam induced voltage identically to mbtrack. 

This PR is a work in progress. Still to do:

1. ~~In the PassMethod I need the full fill pattern and the total number of buckets for the main RF as I also want to record the vbeam for the empty buckets. At the moment all I can pass is bunch_spos and bunch_currents which have length of nbunch. For now I do a dirty reconstruction of the fill pattern and I compute the ring harmonic number from the RF frequency, but needs to be cleaned up.~~

2. ~~Benchmark uniform, 16b and asymmetric fill pattern with MBTRACK. Check with multiple slices per bunch and 1 slice per bunch.~~

3. ~~Check the passive cavity cases including the passive cavity feedback.~~

4. ~~Check buffers and windowing.~~

5. ~~Documentation.~~

6. ~~Update examples.~~

Other relevant major changes:

BLMode.WAKE has been removed in its entirety, it was not useful, slow and not needed. The tests have been updated and the default and now only method is PHASOR.

Vgen was an array of two [Vgen, psi], now it is an array of four [Vgen, thetag, psi, Vgr]. This has been updated everywhere, but breaks a bit some old scripts.

The **ring** harmonic number and the filling pattern are now passed to the passmethod through the ring param.


lcarver:
The main functions are ready for review. I am still testing, doing documention and examples. But the major components are ready.

atintegrators/atimplib.c
```diff
@@ -300,176 +300,130 @@ static void wakefunc_long_resonator(double ds, double freqres, double qfactor, d
...
     double kloss = rshunt*omr/(2*qfactor);
     double bc = beta*C0;
     double *vbr = vbunch;
     double *vbi = vbunch+nbunch;
-    double totalW=0.0;
-    double *totalWb = atMalloc(nbunch*sizeof(double));
+    int ibunch, islice, total_slice_counter;
+    int bunch_counter = 0;
+    double bucket_curr = 0.0;
+    double main_bucket = circumference / (double) M;
```
swhite2401: M as variable name is not great, can you put a more meaningful name?

docs/p/howto/Collective.rst
```diff
@@ -323,22 +323,48 @@ A final note of importants, when parallelising, Nslice refers to the number of s
...
+
+An important thing to consider for the functions found in the **BeamLoadingCavityPass.c** and all related functions is the slightly different definition compared to other codes like **mbtrack**. In PyAT, the RF cavity definition is :math:`V_{rf}=-sin(k*(z-z_{lag}))` as opposed to :math:`V_{rf}=cos(k*(z-z_{lag}))`. This means that when computing the real and imaginary parts of the beam induced voltage from the amplitude and phase of :math:`V_{b}`, the following conversion must be used: 
+
+.. math:: V_{b,real}=-|\tilde{V}_{b}|sin(\theta_{b})
+.. math:: V_{b,imag}=|\tilde{V}_{b}|cos(\theta_{b})
+
+When the beam loading element is first initialise, the expected beam location is computed using the function **get_timelag_fromU0**. This function returns the parameter **ts**, which is the expected equalibrium **z** coordinate of the beam. If you set your timelag correctly, ts is equal to 0. Considering this parameter allows the usual behaviour of the TimeLag of the main cavity to be retained i.e. a shift in TimeLag of the main cavity is simply a shift in the reference of the full system.
```
swhite2401: This part is not very well described.
What does setting "timelag correctly" mean? it can be anything this is just a change of referential.

get_timelag_fromU0 return the synchronous phase of the reference particule define by the energy loss per turn due to synchrotron radiation U0, if the TimeLag attribute of the cavities is set to minus this value the particle ct coordinate will therefore oscillate around zero.

docs/p/howto/Collective.rst
```diff
@@ -323,22 +323,48 @@ A final note of importants, when parallelising, Nslice refers to the number of s
...
+
+The phasor model is updated with each slice. At the end of each bucket (filled or empty) the beam induced voltage is back tracked to the expected beam location in the bucket (:math:`t=ts/clight` for that bucket). This is why the ts is important, as it allows the feedback to ensure the cavity setpoints are fulfilled at the expected beam position. The final value of :math:`V_{b}` at the end of the turn is the mean of all the beam induced voltage from all buckets, measured at :math:`t=0` after the bunch has been tracked. The :math:`V_{b}` used for the compuation of the :math:`V_{g}` can either be taken from the most recent turn, or it can be the mean of a buffer of a specified length.
```
swhite2401: Again this is confusing: the beamloading element will ensure that the energy loss per turn is restored by the cavities at the theoretical synchronous phase computed using get_timelag_fromU0

pyat/at/collective/beam_loading.py
```diff
@@ -324,25 +307,39 @@ def __init__(
...
         super(BeamLoadingElement, self).__init__(
             family_name, length, voltage, frequency, harmonic_number,
             energy, **kwargs)
+            
         if ts is None:
-            _, ts = get_timelag_fromU0(ring)
+            _, ts = get_timelag_fromU0(ring)            
         self._ts = ts
+        
         self._phis = (2 * numpy.pi * self.Frequency *
```
swhite2401: Chamge all occurence to np setting import to : import numpy as np

Also please run black for formatting

atintegrators/atimplib.c
lcarver: `M` has been changed to `ring_harmn `to distinguish it from `harmn `which is from the cavity element.

lcarver:
I fixed the TBL case and made the changes suggest by @swhite2401. Please read the docs again, if you think it is not clear, feel free to make changes yourself.

Given that I found a bug in the TBL case, I plan to check a quick 16b case too to make sure I haven't missed anything.

Issue #1034, opened 2025-12-17
False warning in atfastring AT matlab


oscarxblanco:
Dear all,
the function `atfastring` strips out the ring parameters too early creating an unnecessary warning when calling findm66.

```matlab
>> [ra,rb]=atfastring(THERING);
Warning: Slow access to properties because there is no RingParam element.
Consider adding it with the command: ">> ring=atSetRingProperties(ring)". 
> In atGetRingProperties (line 79)
In frequency_control (line 10)
In findm66 (line 52)
In atfastring/rebuild (line 108)
In atfastring (line 74) 
Warning: Slow access to properties because there is no RingParam element.
Consider adding it with the command: ">> ring=atSetRingProperties(ring)". 
> In atGetRingProperties (line 79)
In frequency_control (line 10)
In findm66 (line 52)
In atfastring/rebuild (line 114)
In atfastring (line 74) 
```

PR #1035, opened 2025-12-17
keep RingParam in atfastring slice

oscarxblanco:
Dear all,
this PR proposes a solution to issue #1034 where a false warning is thrown because the ring parameters are stripout of the ring slice starting with xbeg and ending with xend.

In this modification the ring parameters are kept.

Also, I have a doubt about the variable `global GLOBVAL.E0`.
https://github.com/atcollab/at/blob/a6840ddc194f2ff4a7d6cf07df7369f2c10fa43e/atmat/lattice/atfastring.m#L30
https://github.com/atcollab/at/blob/a6840ddc194f2ff4a7d6cf07df7369f2c10fa43e/atmat/lattice/atfastring.m#L57
 Is it still necesary ? If so I will mark it as `global GLOBVAL %#ok<GVMIS>`, otherwise I would remove it.

### DIFF
--- a/atmat/lattice/atfastring.m
+++ b/atmat/lattice/atfastring.m
@@ -8,17 +8,17 @@ function [newring,newringrad] = atfastring(ring0,varargin)
 %
 %   [FASTRING,FASTRINGRAD]=ATFASTRING(RING)
 %
-%RING:          original AT structure, with no RF and no radiation.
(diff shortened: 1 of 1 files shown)


oscarxblanco:
Oh, I got it, there is an extra diffusion element in rb4 and rb6.

Wrt to the help message I will update it.

lfarv (approved):
OK. The documentation is much better!

Issue #1028, opened 2025-12-12
Different tracking results in versions 0.7.1 and 0.7.2


alberdibenat:
Dear all, 

I have been using the attached script to calculate the dynamic aperture of a ring with some kind of Flood Fill algorithm that relies on the track function of pyAT. I have the development version 0.7.2 of pyAT that I have cloned from git, and I get a result of 14.50mm^2 as an output. However, when I tried to cross-check the results in another system with the 0.7.1 version of pyAT, I got 13.93mm^2. This mismatch happened for several different lattices. 

I understand that both versions should give the same result when tracking and registering particle losses. 

[at_track_issue.zip](https://github.com/user-attachments/files/24125138/at_track_issue.zip)

alberdibenat:
I have just tested with an ideal lattice, and the results for both versions are much more similar than for the lattice with errors, but there are still small differences in the boundary region. 

<img width="1000" height="800" alt="Image" src="https://github.com/user-attachments/assets/2277ec9a-1109-4a7d-bef2-699fcba9c0a5" />

<img width="1000" height="800" alt="Image" src="https://github.com/user-attachments/assets/69c940fd-4257-40cd-a159-825884f56c57" />

oscarxblanco:
I had a quick look on the git modifications of 'StrMPoleSymplectic4RadPass.c', 'BndMPoleSymplectic4RadPass.c', 'RFCavityPass.c' using the blame tab and I see no modifications between June the 3rd (release 7.1) and yours.

I guess it must be something on the track function, but, one would go through the commit history bisecting until we could find the difference. There are 55 commits from Jun3/2025 up to now, so, you will need to repeat 6 times the bisecting to get there.

TeresiaOlsson:
These are the results I get.

**Version 0.7.1**

<img width="1000" height="800" alt="Image" src="https://github.com/user-attachments/assets/0974a9d8-291b-4469-a7bd-e3943128a7e4" />

**17 Aug a6fd964**

<img width="1000" height="800" alt="Image" src="https://github.com/user-attachments/assets/6662a422-8793-47e6-951d-335f0bd1ebbb" />

**20 Aug e053163** 

<img width="1000" height="800" alt="Image" src="https://github.com/user-attachments/assets/06dd9913-8b3b-4360-967e-48459101aef7" />

21 Aug 84657e2

[...]

PR #1036, opened 2025-12-20
Tracking consistency

lfarv:
This answers two questions raised in #1028:

1. The computation of the relativistic $\beta$ and $\gamma$ was done before the particle mass was defined (python only). Now the right order is restored,
2. As long as non-relativistic tracking is not implemented, for consistency the revolution period must assume $\beta == 1$ (both Matlab and python).

This cleaning does not change the tracking results for relativistic particles. Non-relativistic particles are ignored and tracked as if the were relativistic, as mentioned in the doc. In Matlab the results were inconsistent.


JeanLucPons:
Personally, I prefer avoid to rely on specific associative rules and add parenthesis.

PR #1057, opened 2026-02-18
Make ExactQuantPass methods from ExactRadPass methods and SymplecticQuantPass methods

kandre2:
Pass Methods needed for FCC studies specifically in the final focus quadrupoles around the interaction points as the lumped element does not reproduce this local effect on e.g. DA/MA.

v106a_quantum_2000turns_special


swhite2401:
You have a problem in the ExactRectangularBendPass for the ds calculation. See the transformation on line 159, you may need to subtract SL+refdz to ds....

Also you did not implement it for ExactSectorBend?

PR #1059, opened 2026-03-12
Add parallel `get_fmap` to pyat

oscarxblanco:
Dear all,
I have made an implementation of the frequency maps called `get_fmap`.
```python
        >>> fmapdata = get_fmap(
                ring,
                window=(-10e-3,10e-3,-10e-3,10e-3),
                offset=np.array([0,0,0,0,eoffset,0]),
                grid_size=(100,100),
                axes=('x','y'),
                n_moving_slices=10,
                shift_zero=1e-9,
                verbose=False,
                use_mp=at.MPMode.CPU,
                pool_size=pool_size,
                nturns=1024,
                max_mem=2048,
                mem_margin=0,
                )
```
It speeds up the calculation by a factor 2 in the cases I tried with at.MPMode.CPU. Instead of tracking particles in blocks that could only be reused when all particles are done, it uses every CPU to track an individual particle, and extract the frequency and frequency variation, signaling when it is done in order to immediately track the next.

Memory handling is very conservative as there is no need to keep all information from tracking until the frequency is analyzed, but, if the users prefers to set limits on max memory or minimum free memory in the system it could be set. This memory handling is done through `psutil` which would need to be installed but it is not a general requirement because the method works well without it.

It also extends the functionality to do other planes, e.g. $\delta$-x, and the output is now a dictionary containing the axis names. Axes indexes (0 to 5) are also store in the dictionary.

There is also the possibility to calculate the tune variation with a moving window which should highlight smaller differences. See for example D. Shatilov, E. Levichev, E. Simonov and M. Zobov  https://journals.aps.org/prab/abstract/10.1103/PhysRevSTAB.14.014001. 

In addition, we keep the number of turns the particle survived which is a complementary information of particle motion in the area beyond the boundary of stable motion. It could show if unstable particles are lost rapidly or slowly.

Any (6,n) array of particles could be provided using the argument 'particles', otherwise, the arguments 'window', 'axes' and 'grid_size' will create a rectangular 2D array for the user.

Unfortunately at.MPMode.GPU is not yet implemented.


oscarxblanco:
Dear all, who could review this P.R. ?

oscarxblanco:
Dear @TeresiaOlsson ,
this addresses the limitations in the frequency map that you mentioned here
https://github.com/atcollab/at/discussions/755#discussion-6514775

could you have a look on this function to get the frequency map ?
You could use `axes=('x','dp')`.

Dear @swhite2401  and @lfarv , this implementation uses `psutil` in the same way that `matplotlib` is used, i.e. it is not a requirement but it suggest to have it everytime the function is used. Would this be OK ?

PR #1067, opened 2026-03-26
Quad curvature correction in bending magnets

lfarv:
This adds another term to the bending magnets to try and cancel a difference of tracking of combined bending-focusing magnets between AT and Xsuite.

The so-called "K1h" term of the Hamiltonian is included in:
- BndMPoleSymplectic4Pass, BndMPoleSymplectic4RadPass,
- ExactSectorBendPass, ExactSectorBendRadPass.

A successful comparison is necessary before accepting this term. @swhite2401 and @kandre2, can you look at that ?


swhite2401:
> @swhite2401:
> 
> Bug fix : I swapped the contributions to px and py. This is now corrected:
> 
> ```python
>    r6[1] -= L * (ReSum + irho*B[1]*(x*x-0.5*y*y));
>    r6[3] += L * (ImSum + irho*B[1]*x*y);
> ```
> 
> Maybe it's better? however the contribution is still very small…

Ah very good, yes it is better now! In fact I did a few tests and when I set the face angles to zero, Fringe*=0, and in XSuite edge_model = suppressed I get perfect agreement for the SBEND (the RBEND still is different but I will look at it seperately taking into account your previous remarks)

So, this tells me that the body integrators are now identical even with K!=0. There is still some issue with the Fringe/edge model . However, looking at the XSuite physics manual it look like we have the same implementation for the ExactSectorBend, did you take a look, could you confirm this?

lfarv:
The face angles you are mentioning are with the AT definition (same convention for sector and rectangular)?

swhite2401:
> > Still does not work, (conversion)
> 
> It works for Quadrupoles, but not for Multipoles. The problem seems to be in Xsuite, when creating the element form a dictionary (using the `from_dict` class method)

Ok yes it does, and for sextupole as well. Fringe field behaves the same as for the multipole for these elements as well.
I will report the problem with multipoles to the CERN colleagues.

lfarv:
> StrMPoleSymplectic4Pass does not have the multipole fringe model, this is why it does not match.

With fringe=1, `StrMPoleSymplectic4Pass` has the same fringe model as `ExactMultipolePass`, but restricted to a single order (the quadrupole). As there is no other order in your magnet, this should be correct. So another question here. I'll check again the code…

lfarv:
Quadrupolar wedge correction added in both `ExactSectorBendPass` and `ExactRectangularBendPass`! Up to you !

swhite2401:
> It seems that in Xsuite it is conditioned by the "full" fringe field model.

I confirm this, and do not understand either why it should depend on the fringe field model...

swhite2401:
@lfarv, I went through everything again and all looks fine except for the RBEND.
I would need to talk to CERN colleagues to understand some things that don't look correct as we discussed, the main developer is also on vacation for the moment, we can resume when you both are back.

swhite2401:
@lfarv, line 531 to 541 in xsuite.py are causing problems. k1s does not seem to be allowed in a dipole so a skew component in PolynomA will be ignored in the conversion. We probably need to revise the logic here and make a specific case for dipoles.

Also looking at these things, we have seen  adding either:
-a horizontal offset in a bend
-a B0, A0 component in AT
Gives different results between AT and XSuite

kandre2:
@lfarv line 542 breaks when the element does not have NumIntSteps parameter.
```
if match_model:
            self["num_multipole_kicks"] = 3 * atparams["NumIntSteps"]
```
```
Cell In[81], line 162, in Multipole._set_xs_poly(self, atparams, match_model)
    160             self["ksl"] = list(pola)
    161         if match_model:
--> 162             self["num_multipole_kicks"] = 3 * atparams["NumIntSteps"]
    163 #         if (numintsteps := atparams.get("NumIntSteps")) is not None:
    164 #             self["num_multipole_kicks"] = 3 * numintsteps
    165         if (scaling := atparams.get("FieldScaling")) is not None:

KeyError: 'NumIntSteps'
```

Replacing by another condition fixed it.
```
if (numintsteps := atparams.get("NumIntSteps")) is not None:
            self["num_multipole_kicks"] = 3 * numintsteps
```

lfarv:
@swhite2401 and @kandre2:

- For bending magnets: `K0`, `K1`, `K2` are possibly set, but not their skew components,
- `NumIntSteps` may be absent (for thin magnets)

No idea yet from the problem with dx, A0, B0…

lfarv:
The AT exact methods are now correctly using the `FullGap`, `FringeInt1`, `FringeInt2` attributes previously used only in `BndMPoleSymplectic2Pass`.

Can you check with non-zero gaps and fringe field integrals?

swhite2401:
> The AT exact methods are now correctly using the `FullGap`, `FringeInt1`, `FringeInt2` attributes previously used only in `BndMPoleSymplectic2Pass`.
> 
> Can you check with non-zero gaps and fringe field integrals?

@lfarv, I have just tested it for the ExactSectorBend and it does not match in vertical for face angles=0.0, for face angles !=0.0 I also see some differences in the horizontal plane.

Any other tests you would like me to run?

swhite2401:
> * or bending magnets: `K0`, `K1`, `K2` are possibly set, but not their skew components,

The latest implementation correctly handles magnet order and does fix this issue with K1, K1s reported previously.

swhite2401:
> No idea yet from the problem with dx, A0, B0…

This remains an issue but I haven't heard any back form CERN, I would almost be tempted to merge this branch soon, we could list the remaining discrepancies in the documentation, what do you think?

lfarv:
I found a bug in the multipole fringe field in Xsuite: the multipole fringe field in Xsuite makes use of the particle charge `q`, which makes no sense for magnetic fields. As a consequence, the tracking with multipole fringe field (edge_*_model= "full") differs for electrons and positrons. There is a perfect agreement with AT when forcing positrons in Xsuite.

lfarv:
@swhite2401 : I reproduced your test setup and here are my findings, taking into account these 2 problems in Xsuite:
1. The quadrupole wedge is not taken into account in Xsuite when edge_*_model == "dipole_only", while it is in AT (unless we decide to change that)
2. The multipole fringe field is wrong in Xsuite for electrons, while it should be independent of particle.

#### For sector magnets (ExactSectorBendPass) with K1 == 0.0
- the magnet core agrees within 1.0e-13,
- the face angle agrees within 1.0e-13,
- the dipole fringe ("dipole-only") agrees within 1.0e-13, including with fringe field integrals,
- the multipole rings field agrees within 1.0e-13,
- the dipole fringe field ("linear") also agrees within 1.0e-13. This is not normally available in AT, I made a specific branch to test that.

#### For sector magnets (ExactSectorBendPass) with K1 != 0.0
- For the magnet core, the agreement goes down to 1.0e-7  instead of 1.0e-13, but with a much better large central area.
- with face angles, the agreement falls to 1.0e-5, with the same square central area

#### For rectangular magnets (ExactRectangularBendPass)
Agreement only with K1 == 0.0 **and** face_angle == 0.5*bending_angle. There is clearly a problem there.

#### For sector magnets (BndMPoleSymplectic4Pass)
No result yet, I have to compare with the Xsuite "drift-kick-drift-expanded" model.

swhite2401:
@lfarv, thanks for the summary, sorry I had forgotten about this charge issue... I agree with your findings!
To complement, the problem with the rectangular bend faces is most likely from XSuite, to probe this I have compared XSuite sector and rectangular bends with different face angles and only the case with 0.5*angle matches...

swhite2401:
@lfarv see issue #1076 from @gubaidulinvadim maybe it could be integrated in there?

swhite2401:
@lfarv we had a discussion with CERN colleagues yesterday and made some prgress see below.


> @swhite2401 : I reproduced your test setup and here are my findings, taking into account these 2 problems in Xsuite:
> 
> 1. The quadrupole wedge is not taken into account in Xsuite when edge_*_model == "dipole_only", while it is in AT (unless we decide to change that)
The quadrupole wedge effect was recently implemented and this is a way for them to turn it off, they admit this is not very clean. A dedicated switch should be better, shall we implement it as well?

> 2. The multipole fringe field is wrong in Xsuite for electrons, while it should be independent of particle.
> 
They agree on that and will fix this

> #### For sector magnets (ExactSectorBendPass) with K1 == 0.0
> * the magnet core agrees within 1.0e-13,
> * the face angle agrees within 1.0e-13,
> * the dipole fringe ("dipole-only") agrees within 1.0e-13, including with fringe field integrals,
> * the multipole rings field agrees within 1.0e-13,
> * the dipole fringe field ("linear") also agrees within 1.0e-13. This is not normally available in AT, I made a specific branch to test that.
> 
> #### For sector magnets (ExactSectorBendPass) with K1 != 0.0
> * For the magnet core, the agreement goes down to 1.0e-7  instead of 1.0e-13, but with a much better large central area.
> * with face angles, the agreement falls to 1.0e-5, with the same square central area
> 
We could not reproduce this, for us the sector bend agrees well for all face angles

> #### For rectangular magnets (ExactRectangularBendPass)
> Agreement only with K1 == 0.0 **and** face_angle == 0.5*bending_angle. There is clearly a problem there.
> 
Here there is some subtelties related to X0ref, to match XSuite and AT.

Without running rbendtune (X0ref=0): xs_sbend[0].rbend_shift += xs_sbend[0]._x0_in
With this we have agreement for all face angle and K1=0

With running rbendtune (X0ref!=0): xs_sbend[0].rbend_shift += xs_sbend[0]._x0_in - at_sbend[0].X0ref
With this we have agreement for parallel faces and K1!=0


> #### For sector magnets (BndMPoleSymplectic4Pass)
> No result yet, I have to compare with the Xsuite "drift-kick-drift-expanded" model.

swhite2401:
Remaining issue:

B0, A0: They include these components in the multipole fringe field calculation, while we have the skip_b0 option activated. When using dipole fringe only I get agreement with A0 but not B0 (as expected)

dx: this is strange we get agreement only with paralell faces, there is something fishy there either for us or for them...do we neglect the change in dipole length when applying a horizontal offset in a sector bend?

X0ref: we apply the shift only on the body of the magnet while it seem that rbendshift applies an offset on everything including edges (I checked this by using parallel faces and turning off multipole fringe, in this case the 2 codes agree)

I think this is all... there are so many cases I am getting confused myself with all these tests.... but we are slowly converging!

lfarv:
@swhite2401 : very nice, we are progressing. In the meantime, I made some additional tests:

#### For sector magnets (BndMPoleSymplectic4Pass)
Perfect agreement with the `drift-kick-drift-expanded` model, in all conditions of K1, face angles and "linear" fringe field.

#### For quadrupoles (StrMPoleSymplectic4Pass)
Perfect agreement with the `drift-kick-drift-expanded model`. A very small deviation in the fringe field model can be easily corrected. `dx` and `dy` are also perfectly reproduced


#### For quadrupoles (ExactMultipolePass)
Perfect agreement with the `bend-kick-bend model`.

So I found no additional problems compared to your list above.

swhite2401:
> Perfect agreement with the `bend-kick-bend model`.

Great! Just for my understanding, I would have expected this to match with kick-drift-kick-exact, why do you need to bend in a multipole?

lfarv:
> Great! Just for my understanding, I would have expected this to match with kick-drift-kick-exact, why do you need to bend in a multipole?

Sorry, it's a typo, you are right!

lfarv:
> X0ref: we apply the shift only on the body of the magnet

You are right. it should be applied to the whole magnet. I just pushed a correction for that. Let me know what you find.

swhite2401:
> > X0ref: we apply the shift only on the body of the magnet
> 
> You are right. it should be applied to the whole magnet. I just pushed a correction for that. Let me know what you find.

You introduced a bug in xsuite.py, the keyword for fringe in xsuite is "suppressed" and not "suppress"

swhite2401:
Other problem, in ExactRectangular bend the case where FringeBend*=1 and QuandBend*=0 now translates to 'linear' instead of 'dipole-only'.

swhite2401:
So it is strange now... some things that used to work don't anymore.... are you sure X0ref is correctly handle? Don't you need to change rbendtune() as well?

lfarv:
> You introduced a bug in xsuite.py, the keyword for fringe in xsuite is "suppressed" and not "suppress"

corrected

> in ExactRectangular bend the case where FringeBend*=1 and QuandBend*=0 now translates to 'linear' instead of 'dipole-only'.

Yes, here is what is happening at the moment: according to the doc: 
```
    /*     method 0 no fringe field
     *     method 1 legacy version Brown First Order
     *     method 2 SOLEIL close to second order of Brown
     *     method 3 THOMX
     */
```
So 1 should indeed translate to "linear". But at the moment, **the AT exact methods ignore the value of FringeBend*** and always use another method (Forest), even when specifying 0. In the future, this will be method 4, and 1 will be indeed the "linear" method (this is how I tested the agreement with Xsuite). For the time being, don't specify `FringeBend*` on exact methods: by default it is translated to "dipole-only".

I'm presently working to implement all 4 methods on all the dipole passmethods.

lfarv:
> are you sure X0ref is correctly handle? Don't you need to change rbendtune() as well?

`rbendtune` just iterates over `X0ref` until the reference particle exits at 0 angle, whatever `x0ref` does. So no need to change it. The resulting value may be different from what it was before, but it's correct.

swhite2401:
> > are you sure X0ref is correctly handle? Don't you need to change rbendtune() as well?
> 
> `rbendtune` just iterates over `X0ref` until the reference particle exits at 0 angle, whatever `x0ref` does. So no need to change it. The resulting value may be different from what it was before, but it's correct.

Still I still get strange results:
```
nsteps = 10000
energy = 10e9

at_sbend = at.Lattice([at.Dipole('D', 1.0, 1.0, k=0.0, PassMethod='ExactRectangularBendPass', 
                              NumIntSteps=nsteps, EntranceAngle=0.5, ExitAngle=0.5,
                              PolynomB=[0.0, 0.0], PolynomA=[0.0, 0.0], dx = 0.0)],
                              energy = energy)

fbend = 1
fquad = 1
at_sbend[0].FringeBendEntrance = fbend
at_sbend[0].FringeBendExit = fbend
at_sbend[0].FringeQuadEntrance = fquad
at_sbend[0].FringeQuadExit = fquad

p0, *_ = at_sbend.track(np.zeros(6))
print(np.squeeze(p0))
at_sbend[0].rbendtune()
p0, *_ = at_sbend.track(np.zeros(6))
print(np.squeeze(p0))
```

For me the reference particle is np.zeros(6) before rbendtune() it already exists at zero, so why is rbendtune() moving it to some other coordinates? ... maybe I miss someting?

In any case not running rbendtune in this case I get perfect agreement using:

```
xs_sbend = at.line_from_lattice(at_sbend, match_model=True)
xs_sbend[0].rbend_shift += xs_sbend[0]._x0_in  - getattr(at_sbend[0], 'X0ref', 0.0)
```

I would have think that running rbendtune and shifting the xsuite element by the same amount... so something is strange....and clearly the problem lies there

lfarv:
@swhite: you magnet has K1 == 0. So `x0Ref` has no effect: you can shift a uniform magnetic field by any amount, the result is the same. if K1 == 0, the result should be the same with or without running `rbendtune`.

Similarly, if K1 == 0, the Xsuite result should be independent of `rbend_shift`. If not, there is a problem there !

lfarv:
@swhite2401: I checked that before iterating, `rbendtune` shifts the magnet by half the sagitta, so that the trajectory spans equally on each side of the magnetic axis. This gives a good approximation. If K1 == 0, it stops there.

swhite2401:
> @swhite: you magnet has K1 == 0. So `x0Ref` has no effect: you can shift a uniform magnetic field by any amount, the result is the same. if K1 == 0, the result should be the same with or without running `rbendtune`.
> 
> Similarly, if K1 == 0, the Xsuite result should be independent of `rbend_shift`. If not, there is a problem there !

I agree, I did not expect this! This is why I am surprised, but maybe I am doing something wrong.... let me check and get back to you...

lfarv:
This new release makes 3 changes:

1. The point where the `x0ref` is applied is moved: after the axis rotation and before the fringe field. With this, the agreement with Xsuite for a rectangular magnet with K1 != 0 is much improved,
2. I suppressed the quadrupole wedge when FringeQuad* is 0: this allows a direct comparison with Xsuite for the "dipole-only" model. We can restore it later when everything is understood,
3. I rewrote the "exact bend fringe", bringing a small improvement in the comparison

lfarv:
A comment on `dx` processing: a dipole with parallel faces (face_angle == 0.5 * bending_angle) has no effect in the horizontal plane: it's a drift. So its horizontal displacement has no effect. This is true in Xsuite, but not in AT, whatever the passmethod (BndMPoleSymplectic4Pass, ExactSectorBendPass, ExactRectangularBendPass).

This looks like a problem in transformations.

dx works perfectly on straight magnets !

swhite2401:
@lfarv I am having a look back and here is a summary of the present situation:

**Multipole:** all looks good

**SBEND** is ok for any value of FringeQuad*, face angles or K1, dy as long as dx, B0 and A0 are equal to zero
-with dx!=0 I get 1.0e-7 disagreement (instead of 1.,0e-14) that can degrade to 1.0e-6 depending on the value of K1 or the face angles
-with A0!=0 and FringeQuad*!=1 I get 1.0e-7 disagreement that can go up to 1.0e-5 depending of face angles and K1. For FringeQuad*=0 I get good agreement in all cases
-with B0!=0 strong disagreement in all cases

**RBEND**
-with K1=0 everything looks fine
-with K1!=0 and FringeQuad*=0 good agreement is obtained if rbendtune() is not called
-with K1!=0 and FringeQuad*=1 strong disagreement except for the case where face angles=0.5*bendingangle
-same observation as SBEND for dx, dy, A0, B0

Few questions / remarks come:
-Did you look into the possibility of integrating B0 into the fringe field calculation as it is done in XSuite?
-There is a problem with dx as you pointed out
-X0ref is still problematic for face angle != 0.5*bendingangle

lfarv:
@swhite2401 
> **RBEND**
> -with K1=0 everything looks fine

I don't see that: even in that case, I see a strong disagreement as soon as the entrance/exit angles are not half the bending angle

lfarv:
@swhite2401 : after another review, my conclusions on RBEND are different from yours:

### With faceangles == 0.5*bendingangle
**Everything works**, including
- K1 != 0,
- `rbendtune`() and the corresponding `rbend_shift`,
- dipole fringe field with non-zero fringe field integrals,
- multipole fringe field.

Here is my test dipole:
```python
elem = at.Dipole('D', 2.0, 1.0, k=1.0, PassMethod='ExactRectangularBendPass',
                  EntranceAngle=0.5, ExitAngle=0.5,
                  FullGap=0.1, FringeInt1=0.65, FringeInt2=0.65,
                  FringeQuadEntrance=1, FringeQuadExit=1,
                  NumIntSteps=1000)
```
Here is the translated Xsuite element:
```
View of RBend(length_straight=1.92, angle=1, length=2, k0='from_h', k1=1, h=0.5, k0_from_h=True, model='drift-kick-drift-exact', knl=[0., 0.], ksl=[0., 0.], knl_rel=[0.], ksl_rel=[0.], edge_entry_active=np.int64(1), edge_exit_active=np.int64(1), edge_entry_model='full', edge_exit_model='full', edge_entry_angle=0, edge_exit_angle=0, edge_entry_angle_fdown=0, edge_exit_angle_fdown=0, edge_entry_fint=0.65, edge_exit_fint=0.65, edge_entry_hgap=0.05, edge_exit_hgap=0.05, shift_x=0, shift_y=0, rot_s_rad=0)
```
And the result:
Figure 3-2

### With faceangles != 0.5*bendingangle
**nothing works**

swhite2401:
Strange let me go case by case in details then.

First here is the script I use:

```python

nsteps = 10000
energy = 10e9

at_sbend = at.Lattice([at.Dipole('D', 1.0, 1.0, k=0.0, PassMethod='ExactRectangularBendPass', 
                              NumIntSteps=nsteps, EntranceAngle=0.5, ExitAngle=0.5,
                              PolynomB=[0.0, 0.0], PolynomA=[0.0, 0.0], dx = 0.0)],
                              energy = energy)

#fquad = 1
at_sbend[0].FringeQuadEntrance = fquad
at_sbend[0].FringeQuadExit = fquad
at_sbend[0].EntranceAngle = 0.0
at_sbend[0].ExitAngle = 0.0
at_sbend[0].K = 0.0
at_sbend[0].dx = 0.0
at_sbend[0].dy = 0.0
at_sbend[0].PolynomB[0] = 0.0
at_sbend[0].PolynomA[0] = 0.0

#at_sbend[0].rbendtune()

xs_sbend = at.line_from_lattice(at_sbend, match_model=True)
xs_sbend[0].rbend_shift += xs_sbend[0]._x0_in - getattr(at_sbend[0], 'X0ref', 0.0)

nax, _, _, diffx = at_xs_track(at_sbend, xs_sbend, axes= [0, 1])
nay, _, _, diffy = at_xs_track(at_sbend, xs_sbend, axes= [2, 3])

plt.subplot(211)
plt.contourf(nax, nax, diffx)
plt.xlabel('x [m]')
plt.ylabel('px')
plt.colorbar()

plt.subplot(212)
plt.contourf(nay, nay, diffy)
plt.xlabel('y [m]')
plt.ylabel('py')
plt.colorbar()

plt.show()

```

swhite2401:
**With K1=0 and rbendtune() not called**: work for all face angle and model

**With K1=0 and rbendtune() called**:
- FringeQuad = 0, face angles = 0.0: not ok
- FringeQuad = 0, face angles = 0.5: ok
- FringeQuad = 1, face angles = 0.0: not ok
- FringeQuad = 1, face angles = 0.5: ok

**With K1=1 and rbendtune() not called** :
- FringeQuad = 0, face angles = 0.0: ok
- FringeQuad = 0, face angles = 0.5: ok
- FringeQuad = 1, face angles = 0.0: not ok
- FringeQuad = 1, face angles = 0.5: ok

**With K1=1 and rbendtune() called** :
- FringeQuad = 0, face angles = 0.0:  not ok
- FringeQuad = 0, face angles = 0.5: ok
- FringeQuad = 1, face angles = 0.0: not ok
- FringeQuad = 1, face angles = 0.5: ok


For me there is still a problem with X0ref the agreement should not depend on whether we call rbendtune or not as long as rbend_shift is set correctly (am I doing something wrong?)

The the special case with K1!=0, FringeQuad=1 and face angle !=0.5 never work, so there is a problem there too unrelated to X0ref

lfarv:
@swhite2401: 
For both our tests, there is at least a common conclusion: everything is ok when the face angles are 0.5.

If not, my tests always fail, while you get agreement for your 1st case: K1 == 0 and no `rbend_tune()`. The only difference is that even in that case, you tune `rbend_shift` while I let it to the default value. Apparently this makes the difference.

So I wonder if the problem is not just a question of transverse positioning : one should always check that the reference particle comes out at 0.0 angle. If not, `x0Ref` or `rbend_shift` is wrong. Could it be that when face_angles == 0, your tuning function `rbend_shift += xs_sbend[0]._x0_in - getattr(at_sbend[0], 'X0ref', 0.0)` would not work ?

if k1 == 0, normally there is no reason to call r`bend_tune()`. I told you that however, translating a uniform magnetic field does not change anything: that's true only if the faces are parallel, otherwise the magnet length changes. So I will change `rbend_tune()` to set `X0Ref` to 0 if K1 == 0. With this your test case 2 will be identical to test case 1.

Remain the problems for K1 != 0.

lfarv:
Now `rbend_tune` will always keep `X0Ref` at 0 if K1 ==0. This ensures a correct reference trajectory whatever the face angles are.

However I found a problem: whatever K1 and face angles, after calling `rbendtune()`, the output orbit in AT is in the 1.e-15 range. Applying the conversion formula for `rbend_shift`, I find that:

- if face angles are 0.5 x bending angle, the output orbit in Xsuite is in the 1.e-12 range: significantly larger, but good enough to get a reasonably good agreement on the plots.
- if the face angles are 0, the output orbit in Xsuite is in the 0.02 range, far from zero: the conversion formula does not work. So disagreement between AT and Xsuite comes from the tuning of `rbend_shift`. Apart from building a loop to cancel the output orbit, as it is done in AT, I do not know what to do…

swhite2401:
> * if the face angles are 0, the output orbit in Xsuite is in the 0.02 range, far from zero: the conversion formula does not work. So disagreement between AT and Xsuite comes from the tuning of `rbend_shift`. Apart from building a loop to cancel the output orbit, as it is done in AT, I do not know what to do…

We are progressing!

I did write a loop to cancel the output orbit in XSuite but could not reach agreement either.... however, you have made a lot of changes since then. Let me try to run these tests again.

swhite2401:
@lfarv, I confirm that even with a numerical loop on the xsuite bend agreement is still not good for K1!=0 and face angles = 0.0

swhite2401:
@lfarv, I may have some news concerning the issue with `dx` . Looking at the reference:

https://www.sciencedirect.com/science/article/pii/S0168900222007793?ref=cra_js_challenge&fr=RR-1

we are implementing the linearized map and what we observe is rather good agreement up to a certain amplitude which could indicate that the exact map would be needed for large amplitude.
The paper mentions issue with the exact map relating to sqrt of negative numbers in case of "extreme" misalignment, not sure what extreme means in this case.

I think it would be worth implementing the exact map, this could be done in separate PR, @SebastienJoly when you implemented this transformations, did you by any chance take a look at this exact map?

swhite2401:
For reference this is the kind of error we get for dx=1mm

image

lfarv:
@swhite2401 : I came to the same conclusion. Implementing the exact map for dx, dy, dz is trivial: it is what is done to apply `X0Ref` ! It does not use the R1/2 and T1/2 matrices. But I do not know what to do with rotations.

However from your plots, it looks like for position within 5 cm (which is huge !), the error is within 1.e-7, so it's not so critical.

swhite2401:
@lfarv, some more observation with B0 this time. I have tried to simplify the problem as much as possible.
For this I use exact sector bend with face angles=0.0, K1=0 and Fringe*=0.0 so basically only the body is considered, in XSuite this results in edge mode=suppressed.

In this case I still don't get agreement between the 2 codes, which seems very strange to me because I thought the implementation of the body was solid and well benchmarked....to be followed, I will do the same test with the small angle approximation passmethod just to be sure.

SebastienJoly:
@swhite2401 I did not try implementing the exact map. I chose to implement the linearized map to be able to compare the results with SC/pySC, which also rely on it.
Up to +- 5 cm, what is the order of magnitude of the error compared to Xsuite?

swhite2401:
> @swhite2401 I did not try implementing the exact map. I chose to implement the linearized map to be able to compare the results with SC/pySC, which also rely on it. Up to +- 5 cm, what is the order of magnitude of the error compared to Xsuite?

It is very small, we are talking 1.0e-7 difference on the norm of the phase space coordinates, but since we are going for this one to one agreement, we are going through every possible differences. As @lfarv said this is not a problem for now, it is already quite good. But when on of us as time it should be rather simple to implement the exact map.

PR #1071, opened 2026-03-31
add Matrix66RadPass passmethod

swhite2401:
This PR adds a Matrix66RadPass passmethod to allow switching radition on or off for an M66 element using the standard enable_6d/disable_6d function.

This is the first step to a complete refurbishment of the fastring module.

This is for the comment a proposal, any comments or suggestions are more than welcome.

### DIFF
--- a/atintegrators/Matrix66RadPass.c
+++ b/atintegrators/Matrix66RadPass.c
@@ -1,0 +1,97 @@
+#include "atelem.c"
+#include "atlalib.c"
+
+struct elem
+{
+    double *M66;
+    /* optional fields */
+    double *R1;
+    double *R2;
+    double *T1;
+    double *T2;
+};
+
+void Matrix66RadPass(double *r, const double *M,
+        const double *T1, const double *T2,
+        const double *R1, const double *R2, int num_particles)
+{
+    double *r6;
+    int c;
+    for (c = 0; c<num_particles; c++) {	/*Loop over particles  */
+        r6 = r+c*6;
+        if (!atIsNaN(r6[0])) {
+            if (T1 != NULL) ATaddvv(r6, T1);
+            if (R1 != NULL) ATmultmv(r6, R1);
+            ATmultmv(r6, M);
+            if (R2 != NULL) ATmultmv(r6, R2);
+            if (T2 != NULL) ATaddvv(r6, T2);
+        }
+    }
+}
+
+#if defined(MATLAB_MEX_FILE) || defined(PYAT)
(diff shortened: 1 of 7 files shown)


swhite2401:
@lfarv @lcarver I believe this is ready for review. I will start the new fastring development based on this in a new branch.

pyat/at/lattice/elements/basic_elements.py
```diff
@@ -443,25 +443,30 @@ def set_longt_motion(self, enable, new_pass=None, **kwargs):
...
+            m66rad:         Transfer matrix including radiation. Default: Identity matrix.
+
 
         Default PassMethod: ``Matrix66Pass``
         """
         if m66 is None:
-            m66 = np.identity(6)
+            m66 = np.asfortranarray(np.identity(6))
+        if m66rad is None:
+            m66rad = np.asfortranarray(np.identity(6))
```
lfarv: If `m66rad` is not provided, would not it be better to set it to `m66`, rather that to identity ?

atintegrators/Matrix66RadPass.c
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#if defined(MATLAB_MEX_FILE) || defined(PYAT)
+ExportMode struct elem *trackFunction(const atElem *ElemData,struct elem *Elem,
+        double *r_in, int num_particles, struct parameters *Param)
+{
+    if (!Elem) {
+        double *M66;
+        double *R1, *R2, *T1, *T2;
+        M66=atGetDoubleArray(ElemData,"M66Rad"); check_error();
```
lfarv: This raises a small compatibility problem: now, when turning radiation on with a M66 element, the element is processed as non-radiating. With this branch, turning radiation on with an old lattice (no m66rad attribute), it will instead raise an exception. One solution could be to revert to `M66` if `M66Rad` is absent:
```C
M66=atGetOptionalDoubleArray(ElemData,"M66Rad");
if (M66 == NULL) {
    M66=atGetDoubleArray(ElemData,"M66");
}
```

pyat/at/lattice/elements/basic_elements.py
swhite2401: Done

atintegrators/Matrix66RadPass.c
swhite2401: Done

swhite2401:
@lfarv, I have implemented the suggested changes, thanks for implementing the matlab!

lfarv (approved):
Ready for the new `fastring`

Issue #465, opened 2022-08-23
Octupole element creator not as Sextupole


simoneliuzzo:
it is not possible to create an octupole as 

`Octupole('OCT',length_oct,K_oct)`

while it is possible to create a Sextupole in this way

`Sextupole('OCT',length_sext,K_sext)`

lfarv:
Exact. `Octupole` needs `PolynomA` and `PolynomB`. It should be possible to additionally allow the syntax you propose by checking the number of inout arguments.

TeresiaOlsson:
Is there a reason for not changing the syntax of the octupole? We are using the octupole element in pyAT and it is strange for the users that this element doesn't follow the same syntax as other elements.

PR #1077, opened 2026-05-04
Add atoctupole to Matlab version

TeresiaOlsson:
When saving a pyAT lattice to a m-file using `at.save_m` it can currently not be loaded in the Matlab version if the lattice contains octupoles since there is no atoctupole element in the Matlab version.

This is the output of `at.save_m`:

```
atoctupole('OMK1RP', 0.1, [0. 0. 0. 0.], [ 0.   0.   0. -50.]);
```

I have added an atoctupole element to the Matlab version based on the same syntax as atmultipole.

### DIFF
--- a/atmat/lattice/element_creation/atoctupole.m
+++ b/atmat/lattice/element_creation/atoctupole.m
@@ -1,44 +1,43 @@
 function elem=atoctupole(fname,varargin)
-%ATOCTUPOLE Creates an octupole element with class 'Octupole'
+%ATOCTUPOLE Creates an octupole element
 %
-%  ATOCTUPOLE(FAMNAME,LENGTH,S,PASSMETHOD)
+%  ATOCTUPOLE(FAMNAME,LENGTH,POLYNOMA,POLYNOMB,PASSMETHOD)
 %	
 %  INPUTS
-%	 1. FNAME        	family name 
-%    2. LENGTH			length [m]
-%    3. O				strength [m-3]
-%    4. PASSMETHOD     tracking function, defaults to 'StrMPoleSymplectic4Pass'
+%  1. FNAME      - Family name 
+%  2. LENGTH     - Length[m]
+%  3. POLYNOMA   - Skew [dipole quad sext oct];	 
+%  4. POLYNOMB   - Normal [dipole quad sext oct]; 
+%  5. PASSMETHOD - Tracking function. Defaults to 'StrMPoleSymplectic4Pass'
 %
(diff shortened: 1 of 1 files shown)


TeresiaOlsson:
However, the PR is related to issue #465. Is there a reason why the octupole hasn't been implemented similarly to the sextupole? To change that both in the Matlab and Python versions would make the octupole nicer to use.

lfarv:
@TeresiaOlsson : I agree that the Octupole should have been implemented similarly to the Sextupole. Here is a python implementation which does this while keeping the compatibility with the previous syntax:
```python
class Octupole(Multipole):
    """Octupole element, with no changes from multipole at present."""

    _BUILD_ATTRIBUTES = [*LongElement._BUILD_ATTRIBUTES, "_O"]
    _stacklevel = 7  # Stacklevel for warnings

    DefaultOrder = 3

    def __init__(self, family_name: str, length: float, *args, **kwargs):
        """
        Args:
            family_name:    Name of the element
            length:         Element length [m]
            strength:       Octupolar strength [m^-4].

        Keyword arguments:
            PolynomB:           normal multipoles
            PolynomA:           skew multipoles
            MaxOrder:           Number of desired multipoles
            NumIntSteps=10:     Number of integration steps
            KickAngle:          Correction deviation angles (H, V)
            FieldScaling:       Scaling factor applied to the magnetic field
              (*PolynomA* and *PolynomB*)

        Default PassMethod: ``StrMPoleSymplectic4Pass``
        """
        match args:
            case [poly_a, poly_b]:
                pass
            case [strength]:
                poly_a = kwargs.pop("PolynomA", [])
                poly_b = [0.0, 0.0, 0.0, strength]
            case []:
                poly_a = kwargs.pop("PolynomA", [])
                poly_b = kwargs.pop("PolynomB", [0.0, 0.0, 0.0, 0.0])
            case _:
                msg = "Octupole strength should be a float."
                raise ValueError(msg)
        super().__init__(family_name, length, poly_a, poly_b, **kwargs)

    # noinspection PyPep8Naming
    @property
    def _O(self) -> float:
        """Octupolar strength [m^-4]."""
        arr = self.PolynomB
        return 0.0 if len(arr) < 4 else float(arr[3])

    # noinspection PyPep8Naming
    @_O.setter
    def _O(self, strength):
        self.PolynomB[3] = strength
```


Now `save_m` outputs:
```
atoctupole('OC', 1.0, 50.0);...
```
With this you can implement the Matlab octupole in a nicer way.

PR #1078, opened 2026-05-12
fix(xsuite): remove duplicate keyword args in Corrector.from_at

AmarildoTopalli:
atelem.to_file() serializes all element attributes including PolynomA, PolynomB and MaxOrder into atparams. These were then passed twice to Multipole.from_at() — once explicitly and once via **atparams — causing a TypeError.

Fix by popping PolynomA, PolynomB and MaxOrder from atparams before the call, so only the values recomputed from KickAngle are used.

### DIFF
--- a/pyat/at/load/xsuite.py
+++ b/pyat/at/load/xsuite.py
@@ -848,6 +848,13 @@ class Corrector(_XsFactory):
         if (length := atparams["Length"]) != 0.0:
             pola /= length
             polb /= length
+
+      ####################################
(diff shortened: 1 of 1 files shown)


lfarv:
@AmarildoTopalli : in AT, a Corrector does not have PolynomA/B nor MaxOrder attributes. And if explicitly added to the element, they are ignored. But your correction adds some safety for that, it's OK.

lfarv (approved):

PR #1082, opened 2026-05-19
New fastring

swhite2401:
This PR proposes a new implementation for the fast ring allowing to switch radiations on and off with standard lattice function.
Only one lattice object is then produced.
It is also possible to model higher order chromaticity terms with manual input. The automated generation considers only linear terms. This is safer since fit parameters have to be adapted depending on the order the the user wants to consider.


swhite2401:
@lfarv, @lcarver, @oscarxblanco this is still WIP and more testing is needed but I believe all the functionalities requested are now implemented. Please take a first look and let me know if you see something that needs to be modified/improved.

swhite2401:
For the moment the new function is in `physics/new_fastring.py` to allow for easy comparison and testing. Once validated I will move everything to `fastring.py` and replace the old implementation.

swhite2401:
@lcarver still waiting for your review....

lcarver:
```
import at
from at.physics.new_fastring import *

ring = at.load_mat('/machfs/carver/Repo/EBSlattices/AT/S28F.mat', mat_key = 'LOW_EMIT_RING_INJ')
ring.enable_6d()
print(ring.radiation, ring.energy_loss)


fring = fast_ring_new(ring)
print(fring.radiation, fring.energy_loss)

fring.enable_6d()
print(fring.radiation, fring.energy_loss)

```
output:
```
True 2532505.842292285
True 0.0
True 0.0
```

if I try:
`at.get_energy_loss(fring, method=at.ELossMethod.TRACKING)`
I get:
`2375654.915012846`

lcarver:
Hello, the problem is here:
```

    @check_radiation(True)
    def tracking(ring):
        """Losses from tracking"""
        energy = ring.energy
        particle = ring.particle
        delta = 0.0
        try:
            o6, *_ = ring.find_orbit(method=ELossMethod.INTEGRAL)
            o6l, *_ = ring.disable_6d(RFCavity, copy=True).track(o6)
            delta = np.squeeze(o6l)[4] - o6[4]
        except:
            for e in ring:
                if e.PassMethod == 'SimpleRadiationRadPass':
                    delta -= e.U0 / energy #Needed to prevent mixing with rad. damping
                elif e.PassMethod.endswith('RadPass'):
                    ot = e.track(np.zeros(6), energy=energy, particle=particle)
                    delta += ot[4]
        return -delta*energy
```

For the fast ring, the code does not raise an error when computing find_orbit with the INTEGRAL method, and it does not exclude the Quantum Diffusion, so it never reaches the tracking part in the second section which does exclude it.

swhite2401:
Variable elements are now excluded and a warning is issued when the closed orbit calculation fails and energy loss calculation falls back to the element-by-element method that ignores closed orbit effects

oscarxblanco:
Dear @swhite2401, I tried to use it but got this error. Seems to miss the repr.
```python
fring = fast_ring_new(ring)
```
```python
>>> fring
  File "<python-input-4>", line 1, in <module>
    fring
  File "/home/oblanco/Documents/public/progs/pyenv/pyenv313at/lib/python3.13/site-packages/at/lattice/lattice_object.py", line 296, in __repr__
    return f"Lattice({super().__repr__()}, {ks})"
                      ~~~~~~~~~~~~~~~~^^
  File "/home/oblanco/Documents/public/progs/pyenv/pyenv313at/lib/python3.13/site-packages/at/lattice/elements/element_object.py", line 119, in __repr__
    clsname, args, kwargs = self.definition
                            ^^^^^^^^^^^^^^^
  File "/home/oblanco/Documents/public/progs/pyenv/pyenv313at/lib/python3.13/site-packages/at/lattice/elements/element_object.py", line 227, in definition
    defelem = self.__class__(*arguments)
TypeError: DeltaQ.__init__() takes 7 positional arguments but 11 were given
```

oscarxblanco:
Maybe only DeltaQ is missing  the representation ?
image

swhite2401:
@oscarxblanco did you rebuild AT?

swhite2401:
Ah sorry I had forgotten to add the new passmethod

oscarxblanco:
@swhite2401 , DeltaQ is still missing the representation.
image

oscarxblanco:
@swhite2401 , it may be a problem related to the difference in arguments when the element is built for the first time and when is read from memory.

swhite2401:
@oscarxblanco I cannot reproduce this issue, could you confirm that you have rebuilt AT? DeltaQPass method was changed and now requires more input arguments.
Looking at the error message you sent me it seems you are running on the old passmethod.

oscarxblanco:
@swhite2401 , yes, it is the last commit at.__version__ = 0.8.1.dev51+g85a33dc6
In the python terminal I examine one element at the time and when I reach DeltaQ the error appears
image

oscarxblanco:
@swhite, I don't think it is the PassMethod. It is either a missing repr or a default behaviour of the init which is receiving less arguments than expected.
Also, I'm constructing DeltaQ, I just take the ring returned by new_fastring
image

swhite2401:
Ok! Now I can reproduce it. The problem appears only in a python command line prompt when you type `fring[2]`, typing `print(fring[2])`  or running a script works.... this is is strange but at least now I can look into it! thanks

oscarxblanco:
@swhite2401 , the element is read again as a dictionary when you give it as an entry. Perhaps your __init__ only supports the construction from user input, but not from the dictionary it is translated into

oscarxblanco:
Here is a possible solution @lfarv gave me 5 years ago for the InsertionDeviceKickMap.
The number of arguments changes when the element is first defined and when it is read from memory. Check the following lines :

https://github.com/atcollab/at/blob/0b11096aa7efa2f23a638d434ad3e965ed264cf8/pyat/at/lattice/elements/idtable_element.py#L88-L108

swhite2401:
> @swhite2401 , the element is read again as a dictionary when you give it as an entry. Perhaps your **init** only supports the construction from user input, but not from the dictionary it is translated into

Yes you were right thanks! I have applied a simpler workaround that is to move all argument to keyword arguments with default and remove them from `_BUILD_ATTRIBUTES`.

It seems to work on my side now, could you confirm?

swhite2401:
By the way, with vacation and so on this PR was left in an intermediate state, @lcarver and myself still need to run tests and I am not excluding further bugs on the new features.

If you prefer I can let you know when we are though with these tests, otherwise your review is very useful and more than welcome!

oscarxblanco:
> > @swhite2401 , the element is read again as a dictionary when you give it as an entry. Perhaps your **init** only supports the construction from user input, but not from the dictionary it is translated into
> 
> Yes you were right thanks! I have applied a simpler workaround that is to move all argument to keyword arguments with default and remove them from `_BUILD_ATTRIBUTES`.
> 
> It seems to work on my side now, could you confirm?

Yes, it works !

oscarxblanco:
> By the way, with vacation and so on this PR was left in an intermediate state, @lcarver and myself still need to run tests and I am not excluding further bugs on the new features.
> 
> If you prefer I can let you know when we are though with these tests, otherwise your review is very useful and more than welcome!

Yes, I can try it out. 
I have a question wrt the LinearMatrix. It has two additional fields T1Rad and T2Rad. What is their purpose ?
I don't see any documentation about them, and it is the first time I encounter them.
image

oscarxblanco:
Understood, thank you @swhite2401 .

I have others questions,
Do T1Rad and T2Rad need to be again in DeltaQ ?
Are they the same T1Rad and T2Rad in Matrix66RadPass ?
Is the chromaticity of the longitudinal plane ignored ?
image

oscarxblanco:
@swhite, I see that beta and alpha in DeltaQ are calculated for a ring with is_6d=False, i.e. this are 4D.
Do alpha and beta are recalculated when switching on and off radiation ?

oscarxblanco:
Is there a function in at that calculates the chromatic coefficients qpx and qpy ? If not, could I provide one ?
image

swhite2401:
Do T1Rad and T2Rad need to be again in DeltaQ ?
Yes otherwise you get artificial tune shift because the beam is on the wrong 6D orbit when entering the element

Are they the same T1Rad and T2Rad in Matrix66RadPass ?
Not necessarily. DeltaQ is always lumped for the whole ring so T1/T2 are computed at the end of the ring, for M66 it is different because the fast ring can be split into several matrices in order to insert elements (we have found that for EBS we needed to distribute the quadrupole impedance to get the correct dynamics with collective effects) in which case they would be different

Is the chromaticity of the longitudinal plane ignored ?
Yes for the moment it is, shall it be added?

swhite2401:
> @swhite, I see that beta and alpha in DeltaQ are calculated for a ring with is_6d=False, i.e. this are 4D. Do alpha and beta are recalculated when switching on and off radiation ?

You are correct, and no for the moment they are not changed with radiation which is obviously wrong in case you want to model an off-momentum fastring. I assumed this would never be the case.... but I can certainly add the corresponding `*Rad` attributes if you think this would be useful

swhite2401:
> Is there a function in at that calculates the chromatic coefficients qpx and qpy ? If not, could I provide one ? image

Yes there is already one in `physics.nonlinear.chromaticity`, I will mention it in the documentation

oscarxblanco:
>Is the chromaticity of the longitudinal plane ignored ?
>Yes for the moment it is, shall it be added?

I don't see why it should be ignored. The longitudinal chromaticity will change the phase of off-momentum particles which I think could have an impact on the particle distribution,  but I could be wrong.

oscarxblanco:
> > Is there a function in at that calculates the chromatic coefficients qpx and qpy ? If not, could I provide one ? 
> Yes there is already one in `physics.nonlinear.chromaticity`, I will mention it in the documentation

Thank you, it is good to know.

oscarxblanco:
> > @swhite, I see that beta and alpha in DeltaQ are calculated for a ring with is_6d=False, i.e. this are 4D. Do alpha and beta are recalculated when switching on and off radiation ?
> 
> You are correct, and no for the moment they are not changed with radiation which is obviously wrong in case you want to model an off-momentum fastring. I assumed this would never be the case.... but I can certainly add the corresponding `*Rad` attributes if you think this would be useful

I want to track the effect of variable skew quadrupole on several lattices with errors and corrections, they are all off-momentum by some small amount. For the moment I have been using the whole ring because the previous fastring ignored non-linear chromaticity.

 So, in this case the beta functions of the 6D model are important otherwise you will miscalculate the effect of the skew quad on the transverse beam size.

oscarxblanco:
Also, the default calculation of the amplitude detuning ignores radiation. I think this should be stated somewhere because one might want to use other detuning factors that better match the behaviour with radiation.

swhite2401:
> > Is the chromaticity of the longitudinal plane ignored ?
> > Yes for the moment it is, shall it be added?
> 
> I don't see why it should be ignored. The longitudinal chromaticity will change the phase of off-momentum particles which I think could have an impact on the particle distribution, but I could be wrong.

My understanding is that amplitude dependent synchrotron tune is handled by the cavity elements, same as for the full ring. The only thing missing that I can see is higher order momentum compaction. The first order is integrated in the M66... higher order terms could be added to DeltaQ, do you see cases where this could be useful?

swhite2401:
> So, in this case the beta functions of the 6D model are important otherwise you will miscalculate the effect of the skew quad on the transverse beam size.

I have added *Rad attributes for optics function in the DeltaQ element

oscarxblanco:
> My understanding is that amplitude dependent synchrotron tune is handled by the cavity elements, same as for the full ring. The only thing missing that I can see is higher order momentum compaction. The first order is integrated in the M66... higher order terms could be added to DeltaQ, do you see cases where this could be useful?

Ok, understood. I don't have an immediate case but I also don't see why should higher order momentum compaction factor should be excluded.

oscarxblanco:
> I have added *Rad attributes for optics function in the DeltaQ element

Would it be possible to store the dispersion ? It may not be implemented on the pass method but could be useful as information about the ring.

swhite2401:
> > I have added *Rad attributes for optics function in the DeltaQ element
> 
> Would it be possible to store the dispersion ? It may not be implemented on the pass method but could be useful as information about the ring.

I am not so much in favor or storing data in elements attributes, if you want the optics at the beginning of the ring can simply run: ` l0, *_ = fast_ring.get_optics()`

oscarxblanco:
> l0, *_ = fast_ring.get_optics()

They seem to be slightly different, but, it may not be a problem. OK.
image

swhite2401:
@oscarxblanco I think I broke something with the latest pushes... I will let you know when this is ready for review again

swhite2401:
So I have been running some tests with non-linear chromaticity. In 4D (1st picture) all looks fine but in 6D (2nd picture) I can see some descrepencies in the horizontal plane.
It is not very clear yet what is the problem...

nl_chrom_4d
nl_chrom_6d

oscarxblanco:
@swhite2401 , would you like me to check it ? Or should I wait ?

swhite2401:
> @swhite2401 , would you like me to check it ? Or should I wait ?

I think it is the dispersion at the observation point that is causing issues, the contribution of dispersion to the beam coordinates as to be removed before apply the tune shift form Q' etc... (rotation in phase space)

Now I have to figure out a smart way to handle this...

swhite2401:
Ok so it was very stupid, the non-linear chromaticity fit was failing due to bad points at very large momentum offset, for the mooment I increased the order of the fit and it is all good.

See the test code:

```python
import at
from at.physics import new_fastring as nf
from at.physics import fastring as of
import numpy as np
from scipy.special import factorial

ring = at.load_lattice("../lattice/hmba.mat", periodicity=1)

cav = ring.get_elements(at.RFCavity)[0]
cav.HarmNumber=992
for i in ring.get_uint32_index(at.RFCavity):
    name = ("DR_" + ring[i].FamName)
    ring[i] = at.Drift(name, ring[i].Length) if ring[i].Length > 0 else at.Marker(name)

ring = ring.repeat(32)
ring.append(cav)
ring.harmonic_number=992
ring.rf_voltage = 5.5e6
ring.set_rf_frequency()

at.save_lattice(ring, "./ring_full_1cav.mat")

ring = at.load_lattice("./ring_full_1cav.mat", periodicity=1)
ring.disable_6d()
ring.enable_6d(at.RFCavity)

dpmax = 0.05
order = 15
chrom_arr, dpo, qo  = at.chromaticity(ring, order=order, npoints=31, dpm=dpmax+0.01)
plt.plot(dpo, qo[:, 0], '.', color="C01")
plt.plot(dpo, qo[:, 1], '.', color="C02")
px = np.poly1d((chrom_arr[0, :]/ factorial(np.arange(order + 1)))[::-1])
py = np.poly1d((chrom_arr[1, :]/ factorial(np.arange(order + 1)))[::-1])
plt.plot(dpo, px(dpo), color="C01")
plt.plot(dpo, py(dpo), color="C02")
plt.show()
              
fring = nf.fast_ring_new(ring, qpx=chrom_arr[0, 1:], qpy=chrom_arr[1, 1:])
fring.disable_6d()
fring.enable_6d(at.RFCavity)

f, dp, q = at.chromaticity(ring, npoints=21, order=5, dpm=dpmax)
ff, dpf, qf = at.chromaticity(fring, npoints=21, order=5, dpm=dpmax)

plt.plot(dp, q[:, 0], color='C01', label="Full ring - linopt - X")
plt.plot(dpf,qf[:, 0], '+', color='C01', label="Fast ring - linopt- X")

plt.plot(dp, q[:, 1], color='C02', label="Full ring - linopt - Y")
plt.plot(dpf,qf[:, 1], '+', color='C02', label="Fast ring - linopt - Y")

f, dp, q = at.chromaticity(ring, npoints=21, order=5, dpm=dpmax, method='laskar', nturns=512, fmin=0.01, fmax=0.99)
ff, dpf, qf = at.chromaticity(fring, npoints=21, order=5, dpm=dpmax, method='laskar', nturns=512, fmin=0.01, fmax=0.99)

plt.plot(dp, q[:, 0], color='C03', linestyle='dashed', label="Full ring - tracking - X")
plt.plot(dpf,qf[:, 0], 'x', color='C03', label="Fast ring - tracking - X")

plt.plot(dp, q[:, 1], color='C04', linestyle='dashed', label="Full ring - tracking - Y")
plt.plot(dpf,qf[:, 1], 'x', color='C04', label="Fast ring - tracking - Y")

plt.xlabel('dp')
plt.ylabel('Q')
plt.legend()
```

and figure in 6D:
nl_chrom_6d

swhite2401:
Now I have also corrected for dispersion in DeltaQ element and it helps tracking data to fit better in the presence of large momentum offsets, I am cleaning this up now and should be able to propose a new version for review today

oscarxblanco:
@swhite2401 , great !
On another topic, I would rename the function new_something is never a good name. Could it be something like non-linear fast ring ? nlfastring ?

swhite2401:
> On another topic, I would rename the function new_something is never a good name. Could it be something like non-linear fast ring ? nlfastring ?

Oh yes absolutely! I you look at it I keep both versions alive for the moment because it is covenient for testing within the same python env.
Once validated I will replace the present fastring with the new one, so new will be removed.

swhite2401:
This is the effect of dispersion compensation in the DetlaQ element, slightly better agreement with the full ring for large momentum offsets.

dispersion

swhite2401:
Effect of non-linear momentum compaction factor:
alphac

swhite2401:
@oscarxblanco I believe this is now ready for review

oscarxblanco:
@swhite2401 , there is still something missing to make it work with variable thin multipoles.

I am checking the coupling induced by a variable skew quad. The first set of three plots comes from tracking the whole ring and it behaves as expected, while the second set of three plots comes from tracking with a fast ring with the same skew quad and looks wrong

image

image

oscarxblanco:
@swhite2401 , it seems that I don't know how to model the longitudinal motion with the fast ring element. Here is the same example but with fewer turns. The third column is the longitudinal phase space in AT coordinates. On top is the behaviour with the whole ring while below is the fast ring.
image
I will have a look into the RF cavities.

oscarxblanco:
@swhite2401 , why is T1Rad and T2Rad fifth coordinate not exactly the same magnitud with opposite sign ? 
image

oscarxblanco:
@swhite2401 , the help message says that the non-linear momentum compaction factor could be obtained with `physics.mcf` but I get the following error
image

oscarxblanco:
The ring mcf seems to only provide the first order.
image

swhite2401:
@oscarxblanco , few answers:

1- Sorry there is a typo in the docstring, `ring.mcf()` uses default kw argument, instead it should be `ring.get_mcf()`, here is an example usage: 

```python
ring = at.load_lattice("./ring_full_1cav.mat", periodicity=1)
ring.disable_6d()
mcf = ring.get_mcf(fit_order=4, n_step=11, DPStep=0.04)
ring.enable_6d(at.RFCavity)

chrom_arr, *_ = at.chromaticity(ring, order=15, npoints=31, dpm=0.04)             
fring = nf.fast_ring_new(ring, qpx=chrom_arr[0, 1:], qpy=chrom_arr[1, 1:], alphac=mcf)
fring.disable_6d()
fring.enable_6d(at.RFCavity)
```

2- Note that here I enable only the cavity element, this is to turn off quantum diffusion in the fastring. The reason is that it introduces a random part in the tracking that prevents comparison with the full ring where quantum diffusion is not integrated by default: could it be that this is the reason why you see discrepancies in the longitudinal motion?

3-The difference between T1 and T2 (on the M66?) is to account for the energy loss introduced by this fraction of the ring

oscarxblanco:
@swhite2401 , I just recall that there is a bug in the VariableThinMultipole because the Polynoms are not set back to zero after the tracking. I will fix it again.

Apart from that bug, it seems that the fast ring makes the beam have a large energy excursion even if the beam starts in the closed orbit. Here below, the tracking of 100 turns, in blue the whole ring and in orange the fast_ring.
image


The closed orbit for both rings is the same. Although, the difference in T1Rad and T2Rad in the linear matrix is is still there
image

swhite2401:
> Apart from that bug, it seems that the fast ring makes the beam have a large energy excursion even if the beam starts in the closed orbit. Here below, the tracking of 100 turns, in blue the whole ring and in orange the fast_ring.

Yes as I said in my previous comment I think this is due to the quantum diffusion element of the fastring, could you try turning it off?

swhite2401:
Here is a script with a variable skew: turning off quatum diffusion in the fastring seem to provide agreement in the transverse planes. The full ring (blue) shows a small longitudinal motion that is not seen in the fastring (orange)

Results:

image


```python
import at
from at.physics import new_fastring as nf
from at.physics import fastring as of
import numpy as np


ring = at.load_lattice("./ring_full_1cav.mat", periodicity=1)
ring.disable_6d()
mcf = ring.get_mcf(fit_order=4, n_step=11, DPStep=0.04)
ring.enable_6d()

chrom_arr, *_ = at.chromaticity(ring, order=15, npoints=31, dpm=0.04)             
fring = nf.fast_ring_new(ring, qpx=chrom_arr[0, 1:], qpy=chrom_arr[1, 1:], alphac=mcf)
fring.enable_6d()
fring.disable_6d(at.QuantumDiffusion)

o6, *_ = ring.find_orbit()
o6[0] = 100.0e-6

vmult = at.VariableMultipole("VM", AmplitudeA=0.1, FrequencyA=0.21*at.clight/ring.circumference, MaxOrder=1)

ring.append(vmult)
fring.append(vmult)


nturns = 100

pout, *_ = ring.track(o6, nturns=nturns)
pout = np.squeeze(pout)
vmult.PolynomB*=0.0
vmult.PolynomA*=0.0

pout2, *_ = fring.track(o6, nturns=nturns)
pout2 = np.squeeze(pout2)
vmult.PolynomB*=0.0
vmult.PolynomA*=0.0

plt.subplot(131)
plt.plot(pout[0, :], pout[1, :], "x")
plt.plot(pout2[0, :], pout2[1, :], "+")
plt.xlabel("x")
plt.xlabel("px")
plt.subplot(132)
plt.plot(pout[2, :], pout[3, :], "x")
plt.plot(pout2[2, :], pout2[3, :], "+")
plt.xlabel("y")
plt.xlabel("py")
plt.subplot(133)
plt.plot(pout[5, :], pout[4, :], "x")
plt.plot(pout2[5, :], pout2[4, :], "+")
plt.xlabel("ct")
plt.xlabel("dp")
plt.tight_layout()
plt.show()
```

oscarxblanco:
Thank you @swhite2401 .  I did remove the quantum diffusion element 
```python
del fast_ring[fast_ring.get_refpts(at.QuantumDiffusion)]
```
but your solution looks less dangerous 
```python
fring.disable_6d(at.QuantumDiffusion)
```
The longitudinal motion now seems OK, but, I see some dissagreement on the transverse plane and it seems to come from detuning.
image

swhite2401:
> Thank you @swhite2401 . I did remove the quantum diffusion element
> 
> ```python
> del fast_ring[fast_ring.get_refpts(at.QuantumDiffusion)]
> ```
> 
> but your solution looks less dangerous
> 
> ```python
> fring.disable_6d(at.QuantumDiffusion)
> ```
> 
> The longitudinal motion now seems OK, but, I see some dissagreement on the transverse plane and it seems to come from detuning. image

Could you send your test script?

oscarxblanco:
@swhite, here it is 
```python
"""This script tests the fast ring implementation."""

# oblanco 2026jul07

import at
from at.physics.new_fastring import *
import numpy as np


print(f'at.__version__ = {at.__version__}')
ring = at.load_mat('ring.mat')
idx = ring.get_refpts('LSS')
ring = ring.rotate(idx[1])
#ring.disable_6d()
#ring.enable_6d(at.RFCavity,at.Dipole)

fast_ring = fast_ring_new(ring)
# remove QuantumDiffusion
#del fast_ring[4]
fast_ring.disable_6d(at.QuantumDiffusion)


fast_ring[fast_ring.get_refpts(at.DeltaQ)][0].Detuning = fast_ring[fast_ring.get_refpts(at.DeltaQ)][0].Detuning

#li, *_ = ring.get_optics()
#lf, *_ = fast_ring.get_optics()

ringlist = [ring,fast_ring]


for ringtest in ringlist:

#ringtest.disable_6d()

# closed orbit¶
    co,_ = ringtest.find_orbit()

    revfreq = ringtest.revolution_frequency
    tune = ringtest.get_tune()
    coup_freq = abs(tune[0]-tune[1])*revfreq
    stripa1 = 0e-4
    stripa2 = 0.3e-3
    nturns = 30000
    nturns =  100
    elecoup = at.VariableMultipole('ACSKEW',AmplitudeA=[stripa1,stripa2],FrequencyA=coup_freq)
    elecoup.PassMethod="VariableThinMPolePass"
    idx = len(ringtest)
    ringtest.insert(idx,elecoup,copy_elements=True)
    ringtest[-1].PassMethod="VariableThinMPolePass"

# tracking¶
zin = np.zeros((6,1))
zin = zin.T + co
zin = zin.T
#xoffset =  22e-6
xoffset =  1e-6
#xoffset =  5e-3¶
#xoffset =  4.05e-3¶
zin[0] = zin[0] + xoffset
zin[4] = zin[4] + 0e-3
print(f'nturns={nturns}')
print(f'xoffset={xoffset}')

fig, axs = plt.subplots(1,3)
for ringtest in ringlist:
    zout, _, _ = ringtest.track(zin,nturns=nturns)
    xscale1 = 1e3
    yscale1 = 1e3
    xscale2 = 1e3
    yscale2 = 1e2
    axs[0].plot(xscale1*zout[0,0,0,:],yscale1*zout[1,0,0,:],'o')
    axs[1].plot(xscale1*zout[2,0,0,:],yscale1*zout[3,0,0,:],'o')
    axs[2].plot(xscale2*zout[5,0,0,:],yscale2*zout[4,0,0,:],'o')
    axs[0].set_xlabel('x [mm]')
    axs[0].set_ylabel('$p_x$ [mrad]')
    axs[1].set_xlabel('y [mm]')
    axs[1].set_ylabel('$p_y$ [mrad]')
    axs[2].set_xlabel(r'c$\tau$ [mm]')
    axs[2].set_ylabel(r'$\delta$ [%]')

plt.show(block=True)
```

oscarxblanco:
Here is what I get even with a very small initial amplitude
image

I'll send you the ring by email.

oscarxblanco:
@swhite2401 , I have discarded the variable thin multipole.

Tracking the horizontal plane with a very small amplitude shows a difference in particle coordinates on the horizontal phase space. Here is the result over 60 turns and without acskewquad
image

swhite2401:
I am now using your test script, the problem shows only in 6D.

4D tracking
image

6D tracking
image

swhite2401:
@oscarxblanco so I have found the problem.

I had done all my test on a ring with a single cavity at the end. However your ring has cavities distributed.
The first ting fast-ring does is to merge cavities and place then at the end, so it is generated with a different configuration and 6D orbit. This somehow results in a different tune and results as you have shown.

I have changed you lattice to have all cavities at the end here I now get good agreement even in 6D. Now I need to find a solution for this....

image

swhite2401:
And with variable skew just for completeness:
image

swhite2401:
@oscarxblanco, I need you opinion.

There is no real way around the problem above unless the fastring is split into subsets ending/starting at cavities. So here is what I would like to propose:
1-Use as default the present implementation. It is an approximation but provides the "fastest" ring
2-Implement an option "split_at_cavities" that would automatically segment the ring at cavities, each segment ending with a cavity

What do you think? Do you have any other ideas?

oscarxblanco:
@swhite2401 , the fast tracking already matches the ring behaviour and I think that is enough to make it useful.

If I understood correctly the main risk is having distributed rf cavities, so, I would suggest to warn the user. The simplest behavior would be to check for the RFCavities indexes, launch a warning saying that rf will be reorganized and then proceed with a modified ring as needed inside `new_fast_ring`. 

Whenever distributed rf cavities are supported the warn could be removed. In this scenario there would not be a flag `split_at_cavities`, as it will follow the ring model with unique or distributed rf.

swhite2401:
@oscarxblanco , there is already a splitting feature implemented, I can work something out.

swhite2401:
@oscarxblanco , I have now included the keep cavity option, everything is working fine on my side.
I send you back your modified example, for the moment I also return the `split_ring`, I will remove it later on once the implementation is validated.

The is still something not working: the manual emittance input. I will fix it this afternoon, but meanwhile you can rerun your previous tests to confirm that the parts we discussed are ok.

```python
"""This script tests the fast ring implementation."""

# oblanco 2026jul07

import at
from at.physics.new_fastring import *
import numpy as np
import warnings


print(f'at.__version__ = {at.__version__}')
with warnings.catch_warnings():
    warnings.filterwarnings("ignore")
    ring = at.load_mat('ring_oscar.mat')

idx = ring.get_refpts('LSS')
ring = ring.rotate(idx[1])
ring.set_rf_frequency()

chrom_arr, *_ = at.chromaticity(ring, order=15, npoints=31, dpm=0.02, keep_cavities=True)   
fast_ring, split_ring = fast_ring_new(ring)
# remove QuantumDiffusion
fast_ring.enable_6d()
fast_ring.disable_6d(at.QuantumDiffusion)
#fast_ring.disable_6d()

split_ring.enable_6d()

for e in fast_ring:
    print(e.FamName)
    if hasattr(e, "T1Rad"):
        print(e.T1Rad)
        print(e.T2Rad)    

print(ring.get_tune(), split_ring.get_tune(), fast_ring.get_tune())
print(ring.get_tune(method="laskar"), split_ring.get_tune(method="laskar"), fast_ring.get_tune(method="laskar"))
print(ring.get_energy_loss(method=at.ELossMethod.TRACKING),
      split_ring.get_energy_loss(method=at.ELossMethod.TRACKING),
      fast_ring.get_energy_loss(method=at.ELossMethod.TRACKING))


ringlist = [ring, split_ring, fast_ring]
mlist = ['x', '+', '1']
cor,_ = ring.find_orbit()
cosr,_ = split_ring.find_orbit()
cofr,_ = fast_ring.find_orbit()
colist = [cor, cosr, cofr]

print(colist[0])
print(colist[1])
print(colist[2])


for ringtest in ringlist:
# closed orbit¶
    tune = ringtest.get_tune()
    revfreq = ringtest.revolution_frequency
    coup_freq = abs(tune[0]-tune[1])*revfreq
    stripa1 = 0e-4
    stripa2 = 0.3e-3
    nturns = 30000
    nturns =  100
    elecoup = at.VariableMultipole('ACSKEW',AmplitudeA=[stripa1,stripa2],FrequencyA=coup_freq)
    elecoup.PassMethod="VariableThinMPolePass"
    idx = len(ringtest)
    ringtest.insert(idx,elecoup,copy_elements=True)
    ringtest[-1].PassMethod="VariableThinMPolePass"

# tracking¶
fig, axs = plt.subplots(1,3)
for ringtest, mkr, co in zip(ringlist, mlist, colist):
    zin = np.zeros((6,1))
    zin = zin.T + co
    zin = zin.T
    xoffset =  100e-6
    zin[0] = zin[0] + xoffset
    zin[4] = zin[4] + 0e-3
    print(f'nturns={nturns}')
    print(f'xoffset={xoffset}')
    zout, _, _ = ringtest.track(zin,nturns=nturns)
    xscale1 = 1e3
    yscale1 = 1e3
    xscale2 = 1e3
    yscale2 = 1e2
    axs[0].plot(xscale1*zout[0,0,0,:],yscale1*zout[1,0,0,:],mkr)
    axs[1].plot(xscale1*zout[2,0,0,:],yscale1*zout[3,0,0,:],mkr)
    axs[2].plot(xscale2*zout[5,0,0,:],yscale2*zout[4,0,0,:],mkr)
    axs[0].set_xlabel('x [mm]')
    axs[0].set_ylabel('$p_x$ [mrad]')
    axs[1].set_xlabel('y [mm]')
    axs[1].set_ylabel('$p_y$ [mrad]')
    axs[2].set_xlabel(r'c$\tau$ [mm]')
    axs[2].set_ylabel(r'$\delta$ [%]')
plt.legend(['Ring', 'Re-arranged ring', 'Fast ring'])
plt.show(block=True)
```

swhite2401:
Sorry there is a typo in the script I sent you.
It should be:
```python
chrom_arr, *_ = at.chromaticity(ring, order=15, npoints=31, dpm=0.02)   
fast_ring, split_ring = fast_ring_new(ring, keep_cavities=True)
```

and not 
```python
chrom_arr, *_ = at.chromaticity(ring, order=15, npoints=31, dpm=0.02, keep_cavities=True)   
fast_ring, split_ring = fast_ring_new(ring)
```

oscarxblanco:
@swhite2401 
for our lattice the amplitude detuning is not negligible, I had to add it, and the new fast ring matches pretty well the element by element tracking.

```python
chrom_arr, *_ = at.chromaticity(ring, order=15, npoints=31, dpm=0.02)
_, dnuwa, *_ = at.detuning(ring)
fast_ring, split_ring = fast_ring_new(ring, keep_cavities=True, detuning_coeff=[dnuwa[0,0],dnuwa[0,1],dnuwa[1,1]])
```

image


I will now test tracking an off-energy particle and adding the non-linear chromaticiy to the new fast ring.

oscarxblanco:
@swhite2401 , I saw the coefficients returned by `at.chromaticity` and I have a doubt. Are they the result of a Taylor expansion ? i.e. are the coefficients multiplied by $k!$ ? 
It is not explicitly writen in the docstring of at.chromaticity but  I need to divide each component by its factorial in order to get what I expect.

swhite2401:
> @swhite2401 , I saw the coefficients returned by `at.chromaticity` and I have a doubt. Are they the result of a Taylor expansion ? i.e. are the coefficients multiplied by k ! ? It is not explicitly writen in the docstring of at.chromaticity but I need to divide each component by its factorial in order to get what I expect.

I had the same issue and yes the coefficients are multiplied by `k!`.
I do the inverse transformation in the fastring so that it is directly compatible with the output of `chromaticity()`.

I will add a word in the documentation of chromaticity.

swhite2401:
> for our lattice the amplitude detuning is not negligible, I had to add it, and the new fast ring matches pretty well the element by element tracking.

I am surprised that you had to manually add them, if not specified they should be automatically computed form the input lattice.

oscarxblanco:
@swhite2401 , I still don't manage to get an off-energy particle tracking with a good result.
image
This is what I have modified:
```python
chrom_arr, *_ = at.chromaticity(ring, order=3, npoints=31, dpm=0.02)
_, dnuwa, *_ = at.detuning(ring)

fast_ring, split_ring = fast_ring_new(ring, keep_cavities=True,
                                      detuning_coeff=[dnuwa[0,0],dnuwa[0,1],dnuwa[1,1]],
                                      qpx = chrom_arr[0,:],
                                      qpy = chrom_arr[1,:],
                                      )
...
    zin[4] = zin[4] + 10e-3
```

swhite2401:
> @swhite2401 , I still don't manage to get an off-energy particle tracking with a good result. image This is what I have modified:
> 
> ```python
> chrom_arr, *_ = at.chromaticity(ring, order=3, npoints=31, dpm=0.02)
> _, dnuwa, *_ = at.detuning(ring)
> 
> fast_ring, split_ring = fast_ring_new(ring, keep_cavities=True,
>                                       detuning_coeff=[dnuwa[0,0],dnuwa[0,1],dnuwa[1,1]],
>                                       qpx = chrom_arr[0,:],
>                                       qpy = chrom_arr[1,:],
>                                       )
> ...
>     zin[4] = zin[4] + 10e-3
> ```

Be carefull the first element of chrom_arr is the tune, you fastring instanciation should be:

```python
fast_ring, split_ring = fast_ring_new(ring, keep_cavities=True,
                                       detuning_coeff=[dnuwa[0,0],dnuwa[0,1],dnuwa[1,1]],
                                       qpx = chrom_arr[0,1:],
                                       qpy = chrom_arr[1,1:],
                                       )
```

swhite2401:
With these 2 lines:
```python
chrom_arr, *_ = at.chromaticity(ring, order=15, npoints=31, dpm=0.02)   
fast_ring, split_ring = fast_ring_new(ring, qpx=chrom_arr[0, 1:], qpy=chrom_arr[1, 1:], keep_cavities=True)
```

I get:
image

oscarxblanco:
@swhite2401 , all OK for off-energy particles up to 0.1%. 
When tracking a particle 1% off-energy it seems that I need the non-linear momentum compaction factor.
```python
zin[4] = zin[4] + 10e-3
```
image
I will try the non-linear momentum compaction factor.

swhite2401:
> I will try the non-linear momentum compaction factor.

Sure you can but I suspect it would change much, the transfer matrices and other quantities are computed around the on-momentum closed orbit so as you deviate form this it is not a surprise that it starts diverging.

If you need to use the fast ring off-momentum I can add a dp argument to compute everything around the dispersive orbit, would you find it useful?

oscarxblanco:
I did a test with 2% energy offset, and +100 um in  the horizontal plane.
With the following lines in the script the longitudinal plane looks perfect.
```python
mcf = at.physics.get_mcf(ring.disable_6d(copy=True),fit_order=3,n_step=10)
fast_ring, split_ring = fast_ring_new(ring,
                                      qpx=chrom_arr[0, 1:],
                                      qpy=chrom_arr[1, 1:],
                                      keep_cavities=True,
                                      alphac=mcf,
                                      )
```

The transverse, however, behaves really non-linear, while the fast ring shows an ellypse which I think is expected.
image

I think this difference is not a problem, but, adding a `dp`  argument is a good idea.

oscarxblanco:
Here is the result with horizontal offset of 20 um, and energy offset of 0.2% which is closer to what the beam would be. It seems all is OK.
image

swhite2401:
> I think this difference is not a problem, but, adding a `dp` argument is a good idea.

Ok I will add it

oscarxblanco:
@swhite2401 , could the amplitude detuning be a polynom ?

swhite2401:
> @swhite2401 , could the amplitude detuning be a polynom ?

Yes I could add this

swhite2401:
@oscarxblanco , for the amplitude detuning, everything is new done w.r.t. action which is cumbersome when going to higher orders, I would like to change this to simplify the implementation.
Before that, I will wait that @lcarver is back because I need to break backward compatibility to move to amplitude.

Is the rest of the code ok for you? Can I start cleaning up this PR and prepare for the merge?

oscarxblanco:
@swhite2401 , all the rest looks OK for me.

The current detuning as a function of action already works, and the polynomial amplitude detuning might not be necessary.

swhite2401:
> The current detuning as a function of action already works, and the polynomial amplitude detuning might not be necessary.

Ok I start cleaning up and wait for @lcarver approval.

But let's see it is a good occasion to improve the detuning function, maybe in a separate PR. Let's see.
Thanks a lot for the review!

swhite2401:
@lfarv , @lcarver  I believe this is getting ready to merge. Any further comments?

Concerning the non-linear amplitude detuning requested by @oscarxblanco , I think I will keep it for another PR as this is one is already quite large.

lcarver:
```
import at

ring = at.load_mat('/machfs/carver/Repo/EBSlattices/AT/S28F.mat', mat_key = 'LOW_EMIT_RING_INJ')
ring.enable_6d()
ring.set_cavity_phase()
print('ring integral', ring.is_6d, at.get_energy_loss(ring))
print('ring tracking', ring.is_6d, at.get_energy_loss(ring, method=at.ELossMethod.TRACKING))


fring = at.fast_ring(ring)
fring.pop(-1)
print('fring.energy_loss', fring.is_6d, fring.energy_loss)
print('fring.energy_loss with tracking', fring.is_6d, at.get_energy_loss(fring, method=at.ELossMethod.TRACKING))
```

gives as output

```
ring integral True 2532505.842292285
ring tracking True 2532456.7605992286
fring.energy_loss True 0.0
fring.energy_loss with tracking True 2532506.462107789

```

so u0 is not the same

swhite2401:
> ```
> import at
> 
> ring = at.load_mat('/machfs/carver/Repo/EBSlattices/AT/S28F.mat', mat_key = 'LOW_EMIT_RING_INJ')
> ring.enable_6d()
> ring.set_cavity_phase()
> print('ring integral', ring.is_6d, at.get_energy_loss(ring))
> print('ring tracking', ring.is_6d, at.get_energy_loss(ring, method=at.ELossMethod.TRACKING))
> 
> 
> fring = at.fast_ring(ring)
> fring.pop(-1)
> print('fring.energy_loss', fring.is_6d, fring.energy_loss)
> print('fring.energy_loss with tracking', fring.is_6d, at.get_energy_loss(fring, method=at.ELossMethod.TRACKING))
> ```
> 
> gives as output
> 
> ```
> ring integral True 2532505.842292285
> ring tracking True 2532456.7605992286
> fring.energy_loss True 0.0
> fring.energy_loss with tracking True 2532506.462107789
> ```
> 
> so u0 is not the same

Fast ring merges cavities and puts then at the end, the calculation of energy loss per turn cannot compare if 6D orbit is different.

The are two options for your benchmark test:
1- first format your lattice to move the cavites at the end

Here is the code:
```python
import at

ring = at.load_mat('/machfs/carver/Repo/EBSlattices/AT/S28F.mat', use = 'LOW_EMIT_RING_INJ', periodicity=1)
cav = ring.get_elements(at.RFCavity)[0]

for i in ring.get_uint32_index(at.RFCavity):
    name = ("DR_" + ring[i].FamName)
    ring[i] = at.Drift(name, ring[i].Length) if ring[i].Length > 0 else at.Marker(name)

ring.append(cav)
ring.append(at.Marker('M'))
ring.harmonic_number=992
cav.HarmNumber = 992
ring.rf_voltage = 5.5e6
ring.set_rf_frequency()
ring.enable_6d()

print(ring.radiation, ring.get_energy_loss(method=at.ELossMethod.TRACKING))

fring = at.fast_ring(ring)
fring.enable_6d()
print(fring.radiation, fring.get_energy_loss(method=at.ELossMethod.TRACKING))
```

Output:
```
True 2532506.462107546
True 2532506.462107463
```

2- Use the new option `keep_cavities=True`. This creates one segment/cavity with the cavity placed at the end.

Here is the code:

```python
import at

ring = at.load_mat('/machfs/carver/Repo/EBSlattices/AT/S28F.mat', use = 'LOW_EMIT_RING_INJ', periodicity=1)
ring.enable_6d()
print(ring.radiation, ring.get_energy_loss(method=at.ELossMethod.TRACKING))

fring = at.fast_ring(ring, keep_cavities=True)
fring.enable_6d()
print(fring.radiation, fring.get_energy_loss(method=at.ELossMethod.TRACKING))
```

Output:
```
True 2532456.7605991713
True 2532456.5768542183
```

This by the way illustrates how the 6D orbit affect energy loos per turn calculations.

lcarver:
Makes sense!

lcarver (approved):

PR #1088, opened 2026-06-15
Allow for Thin multipoles in RDT calculation

JonasKallestrup:
Previously, the integrated multipole strength of magnets were calculated as k*L, but for thin multipoles this would result in zero integrated strength.  Now, to get the integrated strength calculation correct, the zero-length elements are treated to have length of 1.0.

### DIFF
--- a/pyat/at/physics/rdt.py
+++ b/pyat/at/physics/rdt.py
@@ -682,14 +682,23 @@ def get_rdts(
     beta = avebeta[idx_mag]
     etax = avedisp[idx_mag, 0]
     phi = avemu[idx_mag]
+    
+    Lengths = [e.Length for e in ring[idx_mag]]
+    Lengths = [1 if L == 0 else L for L in Lengths] 
+    
+    pola1 = [_get_polynom(e, "PolynomA", 1) for e in ring[idx_mag]]
+    polb1 = [_get_polynom(e, "PolynomB", 1) for e in ring[idx_mag]]
+    polb2 = [_get_polynom(e, "PolynomB", 2) for e in ring[idx_mag]]
+    polb3 = [_get_polynom(e, "PolynomB", 3) for e in ring[idx_mag]]
+    
     pols = [
         [
-            _get_polynom(e, "PolynomA", 1) * e.Length,
-            _get_polynom(e, "PolynomB", 1) * e.Length,
-            _get_polynom(e, "PolynomB", 2) * e.Length,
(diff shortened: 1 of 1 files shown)


oscarxblanco:
Dear @JonasKallestrup , just a warning.
In the current implementation of the geometric RDTs of second order wrt to sextupoles, each sextupole needs a different position along the 's' coordinate. This is always true for thick elements but could be a problem if one defines two thin sextupoles one after the other.


This comes from the summation term over pairs of sextupoles inherited from C, used in matlab and implemented as
https://github.com/atcollab/at/blob/7414e40dfda373e5295b3a3039c92cd339e81cce/pyat/at/physics/rdt.py#L263-L264, which is basically a matrix with zeros in the diagonal, and plus 1 or minus 1 elsewhere, then multiplied by the sextupole strength. See the term`sm[i]-sm` .

During the implementation by @swhite2401  in python I did a review and ask to consider using indexes because I thought it was risky to substract two float values in order to get exactly zero to define the sign, but at that time @swhite2401 preferred to leave the equation exactly as in C for matlab.

Instead, for the chromatic second order RDTs I wrote the same sumation sign as 
https://github.com/atcollab/at/blob/7414e40dfda373e5295b3a3039c92cd339e81cce/pyat/at/physics/rdt.py#L362-L365
which produces the same matrix without the need of float comparison
```python
>>> nelem = 3
>>> np.tri(nelem, nelem, -1) - 1 + np.tri(nelem)
array([[ 0., -1., -1.],
       [ 1.,  0., -1.],
       [ 1.,  1.,  0.]])

```

oscarxblanco:
Here is the equation I refer to
(taken from https://www.aps.anl.gov/files/APS-sync/lsnotes/files/APS_1429490.pdf )
image
where $i,j$ refer to sextupole index $i$ and $j$, and in order to exclude the case of $i=j$ the C implementation checks for $sign(s_i - s_j)$.

swhite2401:
@JonasKallestrup  to get the integrated strength you may simply use the newly added attributes `IntegratedPolynomA` and `IntegratedPolynomB` this should correctly handle all cases. 

@oscarxblanco , we may reconsider in view of this failure scenario where we have successive thin lens elements, at the time I preferred to have a conservative approach and remain as close as possible to the matlab implementation. Now that we have more experience we could move away from this... up to you!

oscarxblanco:
@JonasKallestrup , the sign implementation is solved in the master branch. You could continue with this PR.

PR #1092, opened 2026-07-01
Load corrupt files

swhite2401:
The multipole element instantiation includes sanity checks that are sometimes returning errors. In particular for Quadrupole and Sextupole when K and H are not consistent with PolynomB.

This caused issues when trying to load old lattice files saved when this was not enforced. This PR proposes a work around to load this corrupt files, ignoring K and H.

This is just a proposal, the alternative would be to always issue warnings instead of errors and prioritize PolynomB in case of conflict.

### DIFF
--- a/pyat/at/load/utils.py
+++ b/pyat/at/load/utils.py
@@ -18,6 +18,8 @@ __all__ = [
 import collections
 import re
 import sysconfig
+import warnings
+import copy
 from typing import ClassVar
 from pathlib import Path
 from warnings import warn, filterwarnings
@@ -28,7 +30,7 @@ import numpy as np
 from at import integrators
 from at.lattice import AtWarning
 from at.lattice import elements as elt
-from at.lattice import Lattice, Particle, Element, Marker
+from at.lattice import Lattice, Particle, Element, Marker, AtError, AtWarning
 
 _ext_suffix = sysconfig.get_config_var("EXT_SUFFIX")
 _plh = "placeholder"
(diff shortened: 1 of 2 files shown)


swhite2401:
It is not clear why the legacy matching test fails

lfarv:
@swhite2401 : I can reproduce the test problem on macOS : I get success with python 3.10 and failure with python 3.14. Numpy and scipy versions are different between both cases, but I see that the ubuntu test succeeds even with that latest numpty…

The master branch now fails with python 3.14 in the same way. It succeeded 2 weeks ago in the following configuration:
```
Darwin (macOS) 24.6.0
python 3.14.5
numpy 2.4.6
scipy 1.17.1

```
I now fails with
```
Darwin (macOS) 25.4.0
python 3.14.6
numpy 2.5.0
scipy 1.18.1

```

Everything changed, difficult to identify what causing the problem.

This can be solved by increasing the test threshold from 1.e-8 to 2.e-8, as we did in other occasions, but it's still worrying to see results changing without any intervention.

pyat/at/load/utils.py
```diff
@@ -28,7 +30,7 @@
 from at import integrators
 from at.lattice import AtWarning
 from at.lattice import elements as elt
-from at.lattice import Lattice, Particle, Element, Marker
+from at.lattice import Lattice, Particle, Element, Marker, AtError, AtWarning
```
lfarv: `AtWarning` is imported twice !

pyat/at/load/utils.py
```diff
@@ -331,7 +333,17 @@ def sanitise_class(index, cls, elem_dict):
     cls = find_class(elem_dict, quiet=quiet, index=index)
     if check:
         sanitise_class(index, cls, elem_dict)
-    return cls.from_file(elem_dict)
+    try:
+        elem = cls.from_file(copy.deepcopy(elem_dict))
```
lfarv:
```python
elem = cls.from_file(elem_dict.copy())
```
is enough. But it's a detail…

lfarv (commented):
Setting up all the protections we have now on element creation took a long time, and we see now that it creates problems…

So this simple solution looks perfect to me.

pyat/at/load/utils.py
swhite2401: done

pyat/at/load/utils.py
swhite2401: Corrected

swhite2401:
All suggestions implement and legacy matching test modified

lfarv (approved):
Ok !

Issue #1089, opened 2026-06-15
Setting a misalignment properties (e.g. yaw) modifies the reference point


kparasch:
Setting a misalignment property (e.g. yaw) modifies the reference point back to at.ReferencePoint.CENTRE if the ReferencePoint was set to "ENTRANCE" before. Is this the expected behavior or should it instead use the ReferencePoint of the element?

It seems to happen because the setter of yaw (and the other properties) calls `transform_elem` (L160 in at/pyat/at/lattice/transformation.py) with `reference=None`. Then in L229, it assigns reference to "centre".

Perhaps reference should be passed in the property setters?

Using version 0.8.0.

Example to reproduce it:

```
import at

el = at.Quadrupole('q', 1)

print(f"{el.ReferencePoint=}, {el.yaw=}")

print('Setting reference point to entrance')
el.ReferencePoint = at.ReferencePoint.ENTRANCE
print(f"{el.ReferencePoint=}, {el.yaw=}")

print('Setting yaw to 1e-6')
el.yaw = 1e-6

print(f"{el.ReferencePoint=}, {el.yaw=}")
```

outputs:
```
el.ReferencePoint=<ReferencePoint.CENTRE: 0>, el.yaw=0.0
Setting reference point to entrance
el.ReferencePoint=<ReferencePoint.ENTRANCE: 1>, el.yaw=0.0
Setting yaw to 1e-6
el.ReferencePoint=<ReferencePoint.CENTRE: 0>, el.yaw=1e-06
```

swhite2401:
@kparasch yes, it was decided at the time that if the reference point is not specified the default CENTRE is used. We thought this was safer behavior: the user always has to specify the reference point if he wants something other than the default.

It can nevertheless be change in case you see issues with this.

kparasch:
But I do specify the reference point in the el.ReferencePoint field, and it gets changed back to the default.

swhite2401:
Ah yes I see the problem when you set the properties it uses default values.... to change the reference point you need to explicitly use the function `transform_elem()` 

This is in fact very annoying. 

@lfarv I believe this is a good argument to change the logic and use the element reference point instead of restoring the default... what do you think?

kparasch:
Another issue I just realized is what should happen if I change the reference point after I have already transform with respect to another reference point?

swhite2401:
If you apply absolute changes all transformations are recalculated using the new reference point. If you are applying relative changes, you will get an error if you have inconsistent reference points

[...]

PR #1093, opened 2026-07-02
make reference point persistent for transformations

swhite2401:
This PR solves #1089. The transformation reference point  is now kept in case it was previously set.

### DIFF
--- a/pyat/at/lattice/transformation.py
+++ b/pyat/at/lattice/transformation.py
@@ -226,7 +226,7 @@ def transform_elem(
     """
 
     if reference is None:
-        reference = transform_options.referencepoint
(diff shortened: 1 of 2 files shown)


lfarv (approved):
If @kparash agrees, this is OK for me

lfarv:
@KParash: can you confirm that this branch solves your problem ?

lfarv:
@KParash: sorry for the confusion, I mistyped the intended recipient…
@kparasch: can you confirm that this branch solves your problem ?

kparasch:
Hi @lfarv , sorry for the delay from my part. The branch does solve my problem.

Issue #1098, opened 2026-07-08
AT matlab atguessclass cannot guess atvariablethinmultipole


oscarxblanco:
The function AT matlab atguessclass cannot guess atvariablethinmultipole because it is made of a generic element with no requirements.

PR #1099, opened 2026-07-08
Require Mode and set ModeName in atvariablethinmultipole

oscarxblanco:
This PR  proposes to resolve the class determination at AT matlab and pyat by requiring Mode to be always defined for the atvariablethinmultipole. See issue #1098 .

### DIFF
--- a/pyat/at/lattice/elements/variable_elements.py
+++ b/pyat/at/lattice/elements/variable_elements.py
@@ -43,13 +43,13 @@ class VariableThinMultipole(Element):
     )
 
     def __init__(
-        self, family_name, mode, AmplitudeA=None, AmplitudeB=None, **kwargs
+        self, family_name, modename, AmplitudeA=None, AmplitudeB=None, **kwargs
     ):
         # noinspection PyUnresolvedReferences,SpellCheckingInspection
         r"""
         Parameters:
             family_name(str):    Element name
-            mode:  one of the following:
+            modename:  one of the following:
 
               * :py:attr:`SINE`: sine function
               * :py:attr:`WHITENOISE`: gaussian white noise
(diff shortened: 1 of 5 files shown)


oscarxblanco:
Dear all,
the elements atvariablethinmultipole in AT matlab and VariableThinMultipole in pyat had diferences in the mode setting. In matlab the mode name was used to create the element, while in pyat the custom class was used. This is an incompatibility. In addition, once the mode was set it was impossible to know the modename unless the docstring or help message was at hand.

I propose to bring both elements a step closer requiring always the mode name (and not the mode index) to create the element in both matlab and pyat. And also to add an extra field to store separately mode and modename.

Here is an example in matlab. Please remark Mode and ModeName.
```matlab
eleskew = atvariablethinmultipole('VARSKEW','SINE','FrequencyA',40e3,'AmplitudeA',[0.001]);
>> eleskew

eleskew = 

  struct with fields:

       FamName: 'VARSKEW'
        Length: 0
          Mode: 0
      ModeName: 'SINE'
      PolynomA: 0
      PolynomB: 0
    FrequencyA: 40000
    AmplitudeA: 1.0000e-03
      MaxOrder: 0
``` 
And python
```python
elskew = at.VariableThinMultipole('VARSKEW','SINE',FrequencyA=40e3,AmplitudeA=0.001);
>>> print(elskew)
VariableThinMultipole:
       FamName: VARSKEW
        Length: 0.0
    AmplitudeA: [0.001]
    FrequencyA: 40000.0
      MaxOrder: 0
          Mode: 0
      ModeName: SINE
      Periodic: True
        PhaseA: 0.0
      PolynomA: [0.]
      PolynomB: [0.]

```

oscarxblanco:

lfarv (changes requested):
@oscarxblanco:

I agree with the new `ModeName` field and the new Matlab constructor. However, in python, I think that using a string as input argument rather than an enumeration is a regression (need to check the value), independently of the backward compatibility issue. I would keep the enumeration `mode` input argument, and store it as
```python
self.Mode = mode.value
self.ModeName = mode.name
```

oscarxblanco:
Dear @lfarv , ok, understood. I'll make the changes.

oscarxblanco:
Dear @lfarv , I think this is ready again for review.

pyat/at/lattice/elements/variable_elements.py
```diff
@@ -42,24 +43,24 @@ class VariableThinMultipole(Element):
         Periodic=bool,
     )
 
-    def __init__(
-        self, family_name, AmplitudeA=None, AmplitudeB=None, mode=ACMode.SINE, **kwargs
-    ):
+    def __init__(self, family_name, mode, AmplitudeA=None, AmplitudeB=None, **kwargs):
```
lfarv: Is there a reason to remove the default value for mode ?

oscarxblanco: Yes, to make mode an explicit requirement in both pyat and AT matlab.
Would you prefer to have SINE as default in pyat ? Also in AT matlab ?

lfarv: I see. Previously, as far as I can see, SINE was the default in both Matlab and python. For compatibility, I would keep this, since anyway the correct `Mode` and `ModeName` fields will now always be set in the element. This is enough to get the correct class determination in Matlab.

However if you prefer to have the mode explicitly declared at instantiation, I don't think that it will create problems. Up to you, let me know.

oscarxblanco:
Dear @lfarv , the SINE mode is again the default.

lfarv (approved):

PR #1085, opened 2026-06-04
Add atError when kz=0 in gwig.c

SebastienJoly:
I added some safeguards in the gwig.c integrator following the discussion in: https://github.com/atcollab/at/issues/1075
It raises an atError whenever a wiggler with kz=0 (in By or Bx) is called.

I am not sure whether I should call check_error() after atError(). What would you recommend?

Issue #1075, opened 2026-04-29
at.Wiggler does not handle kz=0


SebastienJoly:
I am currently trying to implement undulators with different H and V harmonics in pyAT. For that, I use the By and By arguments in `at.Wiggler`.
I have a list of H/V harmonics, each with a different amplitude, wavenumbers, and phase. The By harmonics respect the condition ky^2 = kx^2 + kz^2, and the Bx ones kx^2 = ky^2 + kz^2. 
pyAT consistently returns an unstable ring as soon as kz=0, whether with By or Bx harmonics. 
From the integrator gwig.c, there are several kw/kz when calculating the vector potentials. Yet, this special case is not documented nor safely handled in GWigSymplecticPass.c and GWigSymplecticRadPass.c.

Should we add some safeguards in gwig.c or in a different integrator?

You can find a reproducible example below. Note that it is working if I am using By=(1, 1, 0, 1, 1, 0) (kx=0, ky=1, kz=1) instead.
```
import numpy as np
import at

QF = at.Multipole("QF", 0.5, poly_b=np.array([0,1.2]), poly_a=np.zeros(2))
Dr = at.Drift("Dr", 0.5)
HalfDr = at.Drift("Dr2", 0.25)
Dr3 = at.Drift("Dr3", 0.075)
QD = at.Quadrupole("QD", 0.5, -1.2)
Bend = at.Dipole("Bend", 1, 2 * np.pi / 20)
Und = at.Wiggler('und', length=0.425, wiggle_period=0.425/50, b_max=1,
                 #By=(1, 1, 0, 1, 1, 0))
                 By=(1, 1, 1, 1, 0, 0))

FODOcell = at.Lattice(
    [HalfDr, Bend, Dr, QF, Dr, Bend, Dr, QD, Dr3, Und],
    name="Simple FODO cell",
    energy=1e9,
)

print(FODOcell[-1:].track(r_in=np.zeros(6)))

print(FODOcell.radiation_parameters())
```
I use the following AT and numpy versions:
`print(at.__version__)
print(np.__version__)`

0.8.1.dev6+gfd9eed9e
1.26.4

swhite2401:
@SebastienJoly , I can see that there is already some sanity check in the wiggler class of the python interface, maybe you could simply handle this there in the same way? @lfarv what do you think?

SebastienJoly:
What about the matlab users? Should I only add a warning in the python interface?

lfarv:
[...]

PR #1101, opened 2026-07-09
Add kz check in wiggler integrators

SebastienJoly:
This PR is related to the discussion in: https://github.com/atcollab/at/issues/1075
And the PR: https://github.com/atcollab/at/pull/1085
I have now moved the kz checks inside `trackFunction` as requested by @lfarv 

I created a check function which I placed into gwig.c and called in GWigSymplecticPass.c and GWigSymplecticRadPass.c. Let me know if this is better now.

### DIFF
--- a/atintegrators/GWigSymplecticPass.c
+++ b/atintegrators/GWigSymplecticPass.c
@@ -137,6 +137,7 @@ ExportMode struct elem *trackFunction(const atElem *ElemData,struct elem *Elem,
         NVharm = atGetLong(ElemData, "NVharm"); check_error();
         By = atGetDoubleArray(ElemData, "By"); check_error();
         Bx = atGetDoubleArray(ElemData, "Bx"); check_error();
+        GWigCheckKz(NHharm, NVharm, By, Bx); check_error();
         /* Optional fields */
         Energy=atGetOptionalDouble(ElemData,"Energy",Param->energy); check_error();
         R1 = atGetOptionalDoubleArray(ElemData, "R1"); check_error();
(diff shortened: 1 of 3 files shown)


lfarv:
Hello @SebastienJoly. This now looks correct, but did you make tests in both Matlab and python with correct and incorrect wiggler definitions to be sure that the behaviour is as expected ?

SebastienJoly:
@lfarv I could try with the python version and I do get a ValueError from the track function, as expected. Sadly, I do not have access to matlab so I could not test it.

oscarxblanco:
Dear @SebastienJoly , could you put an example in python ? I could translate it and run it in matlab .

lfarv (approved):
> I could try with the python version and I do get a ValueError from the track function, as expected. Sadly, I do not have access to matlab so I could not test it.

Ok, let's take it !

SebastienJoly:
Dear @oscarxblanco 
I have added a simple example below
`Und = at.Wiggler('und', length=0.425, wiggle_period=0.425/50, b_max=1,
                 By=(1, 1, 1, 1, 0, 0))
                 #Bx=(1, 1, 1, 1, 0, 0))
print(Und.track(r_in=np.zeros(6), energy=1e9))`
You can also comment the By line and uncomment the Bx one. The same error should pop up.

oscarxblanco:
Here is the matlab code
```matlab
%%
Und = atwiggler('und',0.425,0.425/50,1,'By',[1,1,1,1,0,0]);
elempass(Und,zeros(6,1),'energy',1e9);
%%
Und = atwiggler('und',0.425,0.425/50,1,'Bx',[1,1,1,1,0,0]);
elempass(Und,zeros(6,1),'energy',1e9);
```
and the result, the same message in both elempass calls:
image

SebastienJoly:
@oscarxblanco Thanks a lot for taking the time to check it in matlab. Did you get exactly the same error message with both options? Or did you first get 'GWig error: By harmonic...' and then 'GWig error: Bx harmonic...'?
I want to make sure it points to the right problem, namely either the By or the Bx array.

oscarxblanco:
Here are the first and second calls
image
image

SebastienJoly:
Excellent then everything is working as expected :)

Issue #1102, opened 2026-07-17
RADIA kickmap reading not working for files with both horizontal and vertical kicks


gubaidulinvadim:
There might be an issue with reading of RADIA kickmaps into at.latticel.elements.idtable_element. I'm trying to simply check that the kickmaps are read correctly from our typical SOLEIL file. 

```
import at 
import matplotlib.pyplot as plt

hu44 = at.idtable_element.InsertionDeviceKickMap('HU44', 10, 'HUS2_44_II10_g11500_Jaw0_kicks.txt', 2.75)
plt.plot(hu44.xkick[:, 57])
plt.plot(hu44.ykick[:, 57], ls='--')

```
This yields identical kickmaps in the horizontal and vertical planes, which are NOT identical in the file. I've also checked that the same issue persists with the kickmaps attached by @simoneliuzzo in https://github.com/atcollab/at/pull/558 for W150. The .ykick and .xkick properties are also identical.

@oscarxblanco @simoneliuzzo Could you check on your side if the kickmaps are read correctly for horizontal AND vertical kicks? If not, I can open a pull request and fix the issue.

oscarxblanco:
HI @gubaidulinvadim ,
it is quite strange. I'll have a look with one of the ALBA kickmaps right now.

oscarxblanco:
@gubaidulinvadim , you are right. 

<img width="337" height="91" alt="Image" src="https://github.com/user-attachments/assets/8354a317-c978-4a81-98a6-40edbf43dec6" />
I'll have a look.

gubaidulinvadim:
@oscarxblanco I can also do it, but if you want to, please proceed. Since you know your code better.

PR #1103, opened 2026-07-17
pyat ID table text file read H and V fix

oscarxblanco:
This PR solves issue #1102 where the horizontal and vertical kickmap had the same values.

This bug was introduced when extending from 2 to 4 blocks the functionality of the idtable read from text file  in the pull request #963 on August 25, 2025. I hope this has not created too much damage. 

The solution is to make a copy of the block before inserting it into a list, i.e. there was a missing np.copy call.

### DIFF
--- a/pyat/at/lattice/elements/idtable_element.py
+++ b/pyat/at/lattice/elements/idtable_element.py
@@ -339,8 +339,8 @@ class InsertionDeviceKickMap(Element):
                     if block_lines > v_points:
                         block_lines = 0
                         kick_block_list.append(np.copy(kick_block))
-                        kick_haxes_list.append(haxis)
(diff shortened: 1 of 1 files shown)


oscarxblanco:
Dear @gubaidulinvadim ,
I think I found the error. Could you please check it works for you ?

gubaidulinvadim:
@oscarxblanco Yes, now it reads correctly two different kickmaps.

oscarxblanco:
@gubaidulinvadim Would you approve to merge ? or would you prefer to wait for others to review it ?

gubaidulinvadim (approved):
I can approve (before the tests were still running). The bug is fixed. But I would suggest adding a check in the automated tests that a small kickmap is read correctly to avoid this kind of bug in the future.

PR #1104, opened 2026-07-17
add InsertionDevice block values H and V check in pyat

oscarxblanco:
This Pull Request adds a test to the insertion device creation in pyat in order to check that the first elemnt of the horizontal and vertical blocks are read correctly.

### DIFF
--- a/pyat/test/test_load_save.py
+++ b/pyat/test/test_load_save.py
@@ -4,18 +4,17 @@ from __future__ import annotations
 
 import os
 from importlib.resources import files
-from typing import Any
 from pathlib import Path
 from tempfile import mkstemp
-
-import pytest
-from numpy.testing import assert_allclose, assert_equal
+from typing import Any
 
 import machine_data
+import pytest
 from at.lattice import Lattice
 from at.lattice import elements as elt
 from at.lattice.elements.idtable_element import InsertionDeviceKickMap
-from at.load import save_xsuite, load_xsuite
+from at.load import load_xsuite, save_xsuite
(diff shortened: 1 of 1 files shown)


oscarxblanco:
This test was suggested by @gubaidulinvadim during the review of the pull request #1104 .

pyat/test/test_load_save.py
```diff
@@ -4,18 +4,17 @@
...
+from typing import Any
 
 import machine_data
+import pytest
 from at.lattice import Lattice
 from at.lattice import elements as elt
 from at.lattice.elements.idtable_element import InsertionDeviceKickMap
-from at.load import save_xsuite, load_xsuite
+from at.load import load_xsuite, save_xsuite
+from numpy.testing import assert_allclose, assert_equal
```
gubaidulinvadim, line 17: I don't understand why the order of imports changed. Did pyAT change the tool that it checking the imports (ruff, isort, etc)?

oscarxblanco: This is effect of running isort.

gubaidulinvadim: Yes, but why isort was not run on the original file before?

oscarxblanco: Dear @gubaidulinvadim , I believe it was missing.
Looking at the history of the file the 15th of August last year, I ran isort as described in the commits. However, other merges added imports. It may as well be that it was not isorted.
image

gubaidulinvadim (approved):

Issue #1113, opened 2026-07-29
atvariablethinelment wrong initialization


oscarxblanco:
Dear all,
PolynomA, PolynomB and MaxOrder for the atvariablethinelement are wrong when Amplitude is zero.

<img width="508" height="297" alt="Image" src="https://github.com/user-attachments/assets/cb3fa6ec-f8d2-46c5-bd1c-b61a80e59e3c" />

PR #1114, opened 2026-07-29
fix MaxOrd and Amplitude for atvariablethinmultipole

oscarxblanco:
Dear all,
this is a possible solution to issue #1113 where  the PolynomA, PolynomB and MaxOrder are wrongly initialized if the initial amplitude is zero.

The function setparam is simplified, and the function setmaxorder has been modified to initialize correctly the polynoms and MaxOrder fields.

For example,
image


oscarxblanco:
Dear @lfarv or @swhite2401 , who could review this PR ?

lfarv (approved):

Issue #1116, opened 2026-07-31
pyat VariableThinMultipole print error inside python terminal


oscarxblanco:
Dear all,
the VariableThinMultipole element in pyat cannot be printed in the terminal. The error message asks for AmplitudeA or AmplitudeB to be defined.

<img width="1248" height="196" alt="Image" src="https://github.com/user-attachments/assets/95d3e594-ea98-436c-900a-0ac3c5f579f0" />

lfarv:
Fixed in #1117.

PR #1117, opened 2026-07-31
fix pyat variablethinmultipole print

oscarxblanco:
Dear all,
this is a possible solution to issue #1116 where the pyat VariableThinMultipole element throw an error when printed in a python terminal.

The number of kwargs is used to determine if the element is created from the user input or if it already exists.
Here is the output:
image

### DIFF
--- a/pyat/at/lattice/elements/variable_elements.py
+++ b/pyat/at/lattice/elements/variable_elements.py
@@ -77,10 +77,10 @@ class VariableThinMultipole(Element):
                          of the excitation
 
               * ``t<t0``: excitation amplitude is zero
-              * ``t0<t<t1``: exciation amplitude is linearly ramped up
-              * ``t1<t<t2``: exciation amplitude is constant
-              * ``t2<t<t3``: exciation amplitude is linearly ramped down
-              * ``t3<t``: exciation amplitude is zero
+              * ``t0<t<t1``: excitation amplitude is linearly ramped up
+              * ``t1<t<t2``: excitation amplitude is constant
+              * ``t2<t<t3``: excitation amplitude is linearly ramped down
+              * ``t3<t``: excitation amplitude is zero
 
         Examples:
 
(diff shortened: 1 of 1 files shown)


oscarxblanco:
Dear @lfarv or @swhite2401 , who could review this PR ?

lfarv (approved):
With this solution, we can get:
```
>>> vmp=at.VariableThinMultipole("VMP")
>>> print(vmp)
VariableThinMultipole:
       FamName: VMP
        Length: 0.0
    PassMethod: IdentityPass
>>> 
```
Not very useful, but correct.

Apparently another option would be to have the python constructor similar to the Matlab one, giving default values for PolynomB, AmplitudeB, FrequencyB, PhaseB. The default would be a sine wave with 0 frequency and 0 amplitude: not more useful.

So I approve this simple solution, up to you to further refine if you think it's worth.

oscarxblanco:
Thank you @lfarv ! Right, `print` is a solution but sometimes it is worth inspecting the element without the fancy format.

I would prefer to merge and move to the next modification of this element.

oscarxblanco:
On a second thought @lfarv , it is better to have both pyat and AT matlab behaving in a similar way. I will implement the default right now.

lfarv:
> Right, print is a solution but sometimes it is worth inspecting the element without the fancy format

Sure. But I was using a `print` to show the full contents of the element

`print` and `repr` have different goals:
- `print(elem)` should give full details of the element in a human-readable form,
- `repr(elem)` or `elem`, should print a string which, when executed, will generate an identical element. So default values are usually not displayed with `repr`. But it's still correct if they do…

oscarxblanco:
> > Right, print is a solution but sometimes it is worth inspecting the element without the fancy format
> 
> Sure. But I was using a `print` to show the full contents of the element
> 
> `print` and `repr` have different goals:
> 
>     * `print(elem)` should give full details of the element in a human-readable form,
> 
>     * `repr(elem)` or `elem`, should print a string which, when executed, will generate an identical element. So default values are usually not displayed with `repr`. But it's still correct if they do…

oh, alright, understood. Thank you for the explanation @lfarv

PR #1122, opened 2026-08-06
Touschek lifetime update

AmarildoTopalli:
The calculation of Touschek lifetime has been updated following the work of Fu et al doi/10.1103/PhysRevAccelBeams.27.124401 (2024). In addition, the polarization dependency of Touschek lifetime has been added. The function can calculate the scattering C(s) and F(s) functions, as well as the average analyzing power, which is an important parameter used to fit the polarization of the beam.

The formula was tested agains the existing Piwinski formula and gives similar results for the tested lattices.

### DIFF
--- a/pyat/at/acceptance/touschek_polarization.py
+++ b/pyat/at/acceptance/touschek_polarization.py
@@ -1,0 +1,316 @@
+"""
+Implementation of the calculation of Touschek lifetime including polarization
+Reference: Fu et al. doi/10.1103/PhysRevAccelBeams.27.124401.
+
+The Touschek lifetime can be given as 1/t = <C> + <F>P**2,
+where t is the Touschek lifetime, P is the polarization, and <..> is the average.
+C and F are a function of the longitudinal position.
+"""
+
+import numpy
+from scipy import integrate
+from scipy.special import iv
+
+import at
+
+from .touschek import (
+    get_beam_sizes,
+    _init_ma_rp,
+)
+from ..constants import qe, clight, _e_radius
+from ..lattice import Lattice
+
+
+def int_C(k, km, um, B1, B2):
+    r"""
+    Integrand of C(s) function.
+    The infinite integral has been turned into a finite integral with the transformation
+    t = tan(k)**2
+    """
+    t = numpy.tan(k) ** 2
+    tm = numpy.tan(km) ** 2
+    
+    one_pl_t5 = (1+t) * (1+t) * (1+t) * (1+t) * (1+t)
+    t3 = t * t * t
+   
+    beta_ma2 = um
+    
+    fact =(
+    numpy.sqrt(one_pl_t5 / t3 )  *
+    (
+    t / (beta_ma2 *(1 + t))
+    - 1
+    + 0.5 * numpy.log(beta_ma2 *(1 + t)/ t)
+    )
+    )
+    
+    if B2 * t < 500:
+        intc = fact * numpy.exp(-B1 * t) * iv(0, B2 * t)  * 2 *  numpy.sqrt( t) * (1+t)
+    else:
+        intc = (
+            fact
+            * numpy.exp(B2 * t - B1 * t)
+            / numpy.sqrt(2 * numpy.pi * B2 * t)
+           * 2 *  numpy.sqrt( t) * (1+t)
+        )
+    return intc
+
+
+def int_F(k, km, um, B1, B2):
+    r"""
+    Integrand of F(s) function.
+    The infinite integral has been turned into a finite integral with the transformation
+    t = tan(k)**2
+    """
+    t = numpy.tan(k) ** 2
+    tm = numpy.tan(km) ** 2
+    
+    one_pl_t5 = (1+t) * (1+t) * (1+t) * (1+t) * (1+t)
+    t3 = t * t * t
+    beta_ma2 = um
+    fact =(
+    0.5 *    numpy.sqrt(one_pl_t5 / t3 )     
+    * numpy.log(beta_ma2 *(1 + t)/ t)
+     )
+    
+    if B2 * t < 500:
+        intf = fact * numpy.exp(-B1 * t) * iv(0, B2 * t)  * 2 *  numpy.sqrt( t) * (1+t)
+    else:
+        intf = (
+            fact
+            * numpy.exp(B2 * t - B1 * t)
+            / numpy.sqrt(2 * numpy.pi * B2 * t)
+           * 2 *  numpy.sqrt( t) * (1+t)
+        )
+    return intf
+
+def _get_vals(
+    ring,
+    rp,
+    ma,
+    emity,
+    bunch_curr,
+    emitx=None,
+    sigs=None,
+    sigp=None,
+    zn=None,
+    epsabs=1.0e-16,
+    epsrel=1.0e-12,
+):
+    r"""
+    Calculate the integrals to obtain C(s) and F(s)
+    """
+    emitx, sigs, sigp = get_beam_sizes(
+        ring, bunch_curr, zn=zn, emitx=emitx, sigs=sigs, sigp=sigp
+    )
+
+    nc = bunch_curr / ring.revolution_frequency / qe
+    beta2 = ring.beta * ring.beta
+    gamma2 = ring.gamma * ring.gamma
+
+    emit = numpy.array([emitx, emity])
+    _, _, ld = ring.get_optics(refpts=rp)
+
+    bxy = ld.beta
+    bxy2 = bxy * bxy
+    axy = ld.alpha
+    dxy = ld.dispersion[:, [0, 2]]
+    dxy2 = dxy * dxy
+    dpxy = ld.dispersion[:, [1, 3]]
+
+    sigb = numpy.sqrt(bxy * emit)
+    sigb2 = sigb * sigb
+    sigp2 = sigp * sigp
+
+    dt = dxy * axy + dpxy * bxy
+    dt2 = dt * dt
+
+    sigh2 = 1 / (1 / sigp2 + numpy.sum((dxy2 + dt2) / sigb2, axis=1))
+
+    bs = bxy2 / sigb2 * (1 - (sigh2 * (dt2 / sigb2).T).T)
+
+    bg2i = 1 / (2 * beta2 * gamma2)
+
+    B1 = bg2i * numpy.sum(bs, axis=1)
+
+    B2sq = (
+        bg2i**2
+        * (
+            numpy.diff(bs, axis=1).T**2
+            + 4
+            * sigh2**2
+            * numpy.prod(bxy2 * dt2, axis=1)
+            / numpy.prod(sigb2**2, axis=1)
+        )
+    )
+
+    B2 = numpy.squeeze(numpy.sqrt(B2sq))
+
+    val_C = numpy.zeros((2, len(rp)))
+    val_F = numpy.zeros((2, len(rp)))
+
+    for i in range(2):
+
+        dpp = ma[:, i]
+        um = beta2 * dpp * dpp
+        km = numpy.arctan(numpy.sqrt(um / (1 - um)))
+
+        for ii in range(len(rp)):
+
+            args = (km[ii], um[ii], B1[ii], B2[ii])
+
+            val_C[i, ii], *_ = integrate.quad(
+                int_C,
+                args[0],
+                numpy.pi / 2,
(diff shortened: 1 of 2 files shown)


gubaidulinvadim:
> The formula was tested agains the existing Piwinski formula and gives similar results for the tested lattices.

You should maybe show your tests :)

oscarxblanco:
Dear @AmarildoTopalli ,
I did a test of your function and I notice this implementation is much slower than the current one. Is there any way you could increase the speed ?
The result of the Touschek lifetime for this test seems to change very little:
```python
current implementation: Touschek lifetime w/o harmon = 9.34 h
Amarildo implementation: Touschek lifetime w/o harmon = 9.39 h
```

AmarildoTopalli:
Dear @oscarxblanco ,
I added a polarization parameter that will calculate the polarization-dependent terms only when polarization = True. By default I set polarization = False, and in this case it takes the same amout of time to run the code as the original get_lifetime function.
If the polarization-depent term is calculated, it takes twice as long to run the code, because of the additional F(s) integral.

gubaidulinvadim:
@AmarildoTopalli Shouldn't polarisation parameter be a float? To allow calculating lifetime at a different polarisation degree?

AmarildoTopalli:
@gubaidulinvadim What is useful for the experimentalists is the average analysing power A_avg used to fit the polarization time and equilibrium beam polarization. With that value they can calculate the lifetime of the beam with X polarization degree, without needing to run a heavy script.

oscarxblanco:
> Dear @oscarxblanco , I added a polarization parameter that will calculate the polarization-dependent terms only when polarization = True. By default I set polarization = False, and in this case it takes the same amout of time to run the code as the original get_lifetime function. If the polarization-depent term is calculated, it takes twice as long to run the code, because of the additional F(s) integral.

@AmarildoTopalli , I confirm the calculation time is now very similar.

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
+"""
+Implementation of the calculation of Touschek lifetime including polarization
+Reference: Fu et al. doi/10.1103/PhysRevAccelBeams.27.124401.
+
+The Touschek lifetime can be given as 1/t = <C> + <F>P**2,
+where t is the Touschek lifetime, P is the polarization, and <..> is the average.
+C and F are a function of the longitudinal position.
+"""
+
+import numpy
```
oscarxblanco: use
import numpy as np

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+from scipy import integrate
```
oscarxblanco: missing isort

gubaidulinvadim:
> @gubaidulinvadim What is useful for the experimentalists is the average analysing power A_avg used to fit the polarization time and equilibrium beam polarization. With that value they can calculate the lifetime of the beam with X polarization degree, without needing to run a heavy script.

I don't think it's correct to assume that the user knows how to calculate the lifetime of a polarised beam and doesn't want to just do `get_lifetime(a, b, c, polarisation=0.72)'

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+    """
+    t = numpy.tan(k) ** 2
+    tm = numpy.tan(km) ** 2
+
+    one_pl_t5 = (1 + t) * (1 + t) * (1 + t) * (1 + t) * (1 + t)
+    t3 = t * t * t
+
+    beta_ma2 = um
+
+    fact = (
```
oscarxblanco: Typically black formatting opens up the equations in a way that is more difficult to read.
You could use `fmt: off` and then `fmt: on` after the equation

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+The Touschek lifetime can be given as 1/t = <C> + <F>P**2,
+where t is the Touschek lifetime, P is the polarization, and <..> is the average.
+C and F are a function of the longitudinal position.
+"""
+
+import numpy
+from scipy import integrate
+from scipy.special import iv
+
+import at
```
oscarxblanco: ruff tells me it is never used

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+def int_C(k, km, um, B1, B2):
+    r"""
+    Integrand of C(s) function.
+    The infinite integral has been turned into a finite integral with the
+    transformation t = tan(k)**2.
+    """
+    t = numpy.tan(k) ** 2
+    tm = numpy.tan(km) ** 2
```
oscarxblanco: `tm` is never used according to ruff. Is this correct ?

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+    val_C *= prefactor
+
+    if polarization:
+        val_F *= prefactor
+        return val_C, val_F
+
+    return val_C
+
+
+def get_touschek_complete(
```
oscarxblanco: the word complete in `get_touschek_complete` is too generic, is there a better name for this function ?

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+import at
+
+from .touschek import (
+    get_beam_sizes,
+    _init_ma_rp,
+)
+from ..constants import qe, clight, _e_radius
+from ..lattice import Lattice
+
+
```
oscarxblanco: ``__all__`` is missing

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+    sigs=None,
+    sigp=None,
+    zn=None,
+    momap=None,
+    refpts=None,
+    offset=None,
+    polarization=False,
+    **kwargs,
+):
+    """Touschek lifetime calculation.
```
oscarxblanco: Could you use google style DocStrings?

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+
+    Computes the Touschek scattering C(s) function.
+
+    If polarization=True, also computes the polarization-dependent
+    F(s) function and A_avg.
+
+    If polarization=False, F(s) and A_avg are not calculated.
```
oscarxblanco: An example would be of help.

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+        offset=offset,
+        momap=momap,
+        interpolate=interpolate,
+        check_zero=True,
+        **kwargs,
+    )
+
+    length_all = numpy.array([e.Length for e in ring[rp]])
+
+    if polarization:
```
oscarxblanco: Wouldn't it be enough passing `polarization = polarization` and remove the `if` ?

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+    emity,
+    bunch_curr,
+    emitx=None,
+    sigs=None,
+    sigp=None,
+    zn=None,
+    momap=None,
+    refpts=None,
+    offset=None,
+    polarization=False,
```
oscarxblanco: Shouldn't it be `polarization = True` by default, and leave the flag to do a faster but less precise calculation ?

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+    """Touschek lifetime calculation.
+
+    Computes the Touschek scattering C(s) function.
+
+    If polarization=True, also computes the polarization-dependent
+    F(s) function and A_avg.
+
+    If polarization=False, F(s) and A_avg are not calculated.
+
+    Returns:
```
oscarxblanco: This docstring is incomplete

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+    fact = (
+        numpy.sqrt(one_pl_t5 / t3)
+        * (
+            t / (beta_ma2 * (1 + t))
+            - 1
+            + 0.5 * numpy.log(beta_ma2 * (1 + t) / t)
+        )
+    )
+
+    if B2 * t < 500:
```
oscarxblanco: The reason to split this integral based on the criteria B2 *t < 500 seems to be inherited from 
https://github.com/atcollab/at/blob/44a0e5c9c9230bec3fdf408db116abdaecce42f9/pyat/at/acceptance/touschek.py#L85-L110
Could you please include your comments? Otherwise it is difficult to know if you follow the same criteria or created a new one that is not written in the article

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
+"""
+Implementation of the calculation of Touschek lifetime including polarization
+Reference: Fu et al. doi/10.1103/PhysRevAccelBeams.27.124401.
```
oscarxblanco: Please mention the article in the main function DocString.

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+    zn=None,
+    momap=None,
+    refpts=None,
+    offset=None,
+    polarization=False,
+    **kwargs,
+):
+    """Touschek lifetime calculation.
+
+    Computes the Touschek scattering C(s) function.
```
oscarxblanco: Please, include a brief description of the approximation followed by this implementation.

I see that several terms of the Moller cross section are omitted to allow the inclusion of polarization.

For the few lattices I have at hand the ommission of those terms in the Moller cross section have little effect, but, there might be rings at higher energy where this incomplete treatment leads to less precise Touschek lifetime estimation.

Therefore, this is not a more precise Touschek lifetime implementation, it is an implementation of the Touschek lifetime including the effect of polarization valid for non-relativistic electron pair interaction in the center or mass.

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+def int_F(k, km, um, B1, B2):
+    r"""
+    Integrand of F(s) function.
+    The infinite integral has been turned into a finite integral with the
+    transformation t = tan(k)**2.
+    """
+    t = numpy.tan(k) ** 2
+    tm = numpy.tan(km) ** 2
```
oscarxblanco: `tm` seems to be never used.

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+
+    beta_ma2 = um
+
+    fact = (
+        0.5
+        * numpy.sqrt(one_pl_t5 / t3)
+        * numpy.log(beta_ma2 * (1 + t) / t)
+    )
+
+    if B2 * t < 500:
```
oscarxblanco: These lines of code are repeated, and the criteria to split the integral in two is undocumented.

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+    dpxy = ld.dispersion[:, [1, 3]]
+
+    sigb = numpy.sqrt(bxy * emit)
+    sigb2 = sigb * sigb
+    sigp2 = sigp * sigp
+
+    dt = dxy * axy + dpxy * bxy
+    dt2 = dt * dt
+
+    sigh2 = 1 / (
```
oscarxblanco: Are you sure about the signs in the term sigh2 ?

Image

pyat/at/acceptance/touschek_polarization.py
```diff
@@ -0,0 +1,358 @@
...
+    get_beam_sizes,
+    _init_ma_rp,
+)
+from ..constants import qe, clight, _e_radius
+from ..lattice import Lattice
+
+
+def int_C(k, km, um, B1, B2):
+    r"""
+    Integrand of C(s) function.
```
oscarxblanco: Does this DocString  really belong to this function ?
The integrand seems to be written as in Eq (27), but, the integration limits are not.

PR #1124, opened 2026-08-12
Add multi-kickmap store to InsertionDeviceKickMap

### LINKED ISSUE #1105: Insertion device kickmap swap
Dear all, 

I propose to have a utility method for the InsertionDeviceKickMap object to swap the kickmap file and update xtable, ytable, xkick, and ykick. I've tried to run InsertionDeviceKickMapInstance.from_user() in the hope of updating the kickmap, but it did not work. Is there a way to just replace the kickmap without making a new InsertionDeviceKickMap and swapping it with the old one in the lattice?

At the moment, the workaround is
- Create a new InsertionDeviceKickMap for the new mode
- Remove the old kickmap from the lattice
- Insert a new kickmap into the same index

gubaidulinvadim:
Adds an interface for storing and selecting between multiple named kickmaps on a single InsertionDeviceKickMap element. This supports workflows where the same ID element must represent different operating modes (e.g. different gap settings or modes (LV/LH/LT/Circular/Elliptical) without recreating the element.

New public API
--------------
- add_kickmap(key, nslice, fname, norm_energy) Reads a kickmap from a file or dict and stores it under a string key without activating it.
- use_kickmap(key) Activates a stored kickmap: swaps xtable, ytable, xkick, ykick, xkick1, ykick1, Nslice and Length to the data under that key. Raises KeyError if the key is not found.
- active_kickmap (property) Returns the key of the currently active kickmap, or None.
- list_kickmaps() Returns the list of stored keys, excluding internal metadata.

The store is held in a plain _kickmap_store dict attribute and is invisible to AT serialisation. PassMethod (IdTablePass / DriftPass) is intentionally untouched by use_kickmap.

This allows to do something like:
```
hu44.add_kickmap('LH', nslice=25, fname='HUS2_44_II10_g11500_Jaw0_kicks.txt', norm_energy=2.75)
hu44.add_kickmap('LV', nslice=25, fname='HUS2_44_II122000_g11500_Jaw0_kicks.txt', norm_energy=2.75)
hu44.add_kickmap('CR', nslice=25, fname='HUS2_44_II113250_g11500_Jaw0_kicks.txt', norm_energy= 2.75)
hu44.add_kickmap('LT', nslice=25, fname='HUS2_44_X112200_g11500_Jaw0_kicks.txt', norm_energy= 2.75)
hu44.list_kickmaps()
hu44.use_kickmap('LT')
hu44.active_kickmap
```

Also fix the docstring: the element supports first *and* second-order kick maps.

Tests
-----
Add pyat/test/test_idtable_kickmap.py with 19 tests covering:
- store lifecycle (empty store, add, list, active key)
- field swap correctness and exact round-trip restore
- KeyError on missing or empty store
- internal _active sentinel not exposed to users
- add_kickmap does not activate
- PassMethod preserved across use_kickmap calls (both IdTablePass and DriftPass)
- set_DriftPass / set_IdTablePass still work after swapping
- kick arrays unchanged after DriftPass → IdTablePass round-trip
- full cycle: swap → Drift → IdTable → swap

Fixes #1105. @oscarxblanco, please have a look and tell me if you like the idea. I want a workflow where I just store the information in the lattice and can swap between different "modes" of the same ID.

Some tests are not very smart (changing the energy of an ID is not very physical, unless you have a ramped "storage" ring)

I would also note that readthedocs documentation of pyAT does not display some of the InsertionDeviceKickMap docstrings correctly.


oscarxblanco:
@gubaidulinvadim , great ! At first glance it looks nice. It seems to provide the possibility to have different ID models. This is much more than what I understood you requested in https://github.com/atcollab/at/issues/1105#issuecomment-5045103538 about swapping tables.

oscarxblanco:
why is the list of kick maps empty when the element is created ?
image

oscarxblanco:
Also, why is the `active_kickmap` empty ?

gubaidulinvadim:
> Also, why is the `active_kickmap` empty ?

I've noticed this yesterday when testing. I need to fix this. The first kickmap (when object is created) is not registered at the moment. It's a bug.

oscarxblanco:
I see that kickmaps are listed by name, what happens if one adds twice the same ID ? or if two different IDs had the same name ? or if the name is empty ?
image

oscarxblanco:
I think that listing the IDs by name leads to several issues. Are you sure this is OK ?

oscarxblanco:
saving the ring in .mat format looks incompatible with `_kickmap_store`. Could you check this ?
image

gubaidulinvadim:
> I think that listing the IDs by name leads to several issues. Are you sure this is OK ?

What do you propose instead? The kickmap filename is also not always unique (and not always intelligible for beam dynamics people). And the "name"/key are up to the user to define however they like it. At SOLEIL, we would typically have names with different polarisations or different gap value. But I'm open to suggestions on how to improve this idea.

P.S I can fix any bugs you find tomorrow.

oscarxblanco:
> > I think that listing the IDs by name leads to several issues. Are you sure this is OK ?
> 
> What do you propose instead? The kickmap filename is also not always unique (and not always intelligible for beam dynamics people). And the "name"/key are up to the user to define however they like it. At SOLEIL, we would typically have names with different polarisations or different gap value. But I'm open to suggestions on how to improve this idea.
> 

Here is my suggestion
- create a unique index for the table when the element is created or when adding a table.
- listing should display the index and name, where name is optional, i.e. it could be empty.
- the user should be able to select any of the tables through `use_kickmap` that accepts index or name. If the name appears more than once, it should raise a warning.

Also, I noticed that FamName is never updated, is this behaviour on purpose ? I think it is convenient,  but, it comes with the risk of not knowing which is the active table in when working in matlab.

There might be other solutions, maybe someone else could advise (@lfarv , @swhite2401 ) ?

> P.S I can fix any bugs you find tomorrow.

Ok, no problem.

gubaidulinvadim:
I think the duplicate name is not possible currently because the name and kickmap parameters are stored in a Python dictionary. When you try to create a duplicate name, it overwrites the existing kickmap.

oscarxblanco:
> I would also note that readthedocs documentation of pyAT does not display some of the InsertionDeviceKickMap docstrings correctly.

Please, do the changes you consider adequate.

gubaidulinvadim:
> Also, why is the `active_kickmap` empty ?



> why is the list of kick maps empty when the element is created ? image

Now on object initiation, a 'default' kickmap is created and registered in kickmap list. The new behaviour sets it as active. So, you can see which kickmap is active and that the list of kickmaps only has a default one. 



> I see that kickmaps are listed by name, what happens if one adds twice the same ID ? or if two different IDs had the same name ? or if the name is empty ? image

If you add twice the same key, for example 'LH' and 'LH' the old one will be overwritten. The kickmaps are stored in a simple Python dictionary. The keys of a dictionary are unique. 

> > > I think that listing the IDs by name leads to several issues. Are you sure this is OK ?
> > 
> > 
> > What do you propose instead? The kickmap filename is also not always unique (and not always intelligible for beam dynamics people). And the "name"/key are up to the user to define however they like it. At SOLEIL, we would typically have names with different polarisations or different gap value. But I'm open to suggestions on how to improve this idea.
> 
> Here is my suggestion
> 
>     * create a unique index for the table when the element is created or when adding a table.
> 
>     * listing should display the index and name, where name is optional, i.e. it could be empty.
> 
>     * the user should be able to select any of the tables through `use_kickmap` that accepts index or name. If the name appears more than once, it should raise a warning.
> 
> 
> Also, I noticed that FamName is never updated, is this behaviour on purpose ? I think it is convenient, but, it comes with the risk of not knowing which is the active table in when working in matlab.
> 
> There might be other solutions, maybe someone else could advise (@lfarv , @swhite2401 ) ?
> 
> > P.S I can fix any bugs you find tomorrow.
> 
> Ok, no problem.

I do not like an index approach. If someone defines four kickmaps without a name and they are assigned an index, then there's no way for the user to know what each index means (gap value? phase value? polarisation?). But we can implement it if everyone agrees that this is necessary. 

For Matlab, I do not know anything because I do not intend to ever use it :) Do you think the FamName should always change to reflect that the kickmap is different? I'd imagine changing FamName dynamically is undesirable because it should serve as an ID for the element. I was imagining that an interface similar to mine here can be implemented in Matlab.

> saving the ring in .mat format looks incompatible with `_kickmap_store`. Could you check this ? image

I'm working on this and loking for a solution for Matlab formats. There's an issue with saving and loading lattice in .mat and .m formats. The kickmap dictionary is not saved or loaded properly and the information is lost for all kickmaps but the currently active one (it gets loaded as "default").  No issue with JSON format, which seems to be more powerful.

oscarxblanco:
Dear @gubaidulinvadim , typically names are optional in AT. Empty names are allowed for many other elements.

It may be that in your case all kickmaps have a non-empty and unique strings to be used as names, but, it would be unusual in AT to force it.

There might be another alternative, e.g. if name is not given assign the filename as the kickmap name. This is also a bit risky because you will have to deal with spaces and special characters.

gubaidulinvadim:
> Dear @gubaidulinvadim , typically names are optional in AT. Empty names are allowed for many other elements.
> 
> It may be that in your case all kickmaps have a non-empty and unique strings to be used as names, but, it would be unusual in AT to force it.
> 
> There might be another alternative, e.g. if name is not given assign the filename as the kickmap name. This is also a bit risky because you will have to deal with spaces and special characters.

Right now the name is always assigned as 'default'

Yes, in the end this is maybe a better solution. I've managed to fix the saving and loading in .m and .mat formats (which have limitations not present in JSON). There's a hack with 100+ lines of code that saves everything as arrays and reconstructs things... I'm not very happy with it. I will try to see if indexing it instead can help. 

Compare commits e7d4353745ee0e6e047c53b152b067b34b651fe9 (.m, .mat and .json is working) and b4ebb16ed2679e8424635e8d7f15a3f9d31b7cd3 (.json is working; .m and .mat are not working). @swhite2401 @lfarv maybe some of you would know a better solution? Or maybe there's a solution on the side of saving and loading the AT lattice. I'm not really familiar with MATLAB formats or their limitations.

swhite2401:
@gubaidulinvadim I am just picking up on this.
Could you quickly summarize where help is needed, is just the file saving to .mat .m files?
Can this proposal as it is for testing?

gubaidulinvadim:
> @gubaidulinvadim I am just picking up on this. Could you quickly summarize where help is needed, is just the file saving to .mat .m files? Can this proposal as it is for testing?

Yes, the branch can be tested as is. Either of the last two commits can be tested. 
- https://github.com/atcollab/at/commit/b4ebb16ed2679e8424635e8d7f15a3f9d31b7cd3 original fix that doesn't work for .m and .mat files
- https://github.com/atcollab/at/commit/e7d4353745ee0e6e047c53b152b067b34b651fe9 solution that I came up with to fix saving to .m and .mat

The main issue is saving in .m and .mat files. The new interface can store several kickmaps in an Python dictionary for a single object 
```
hu44.add_kickmap('LH', nslice=25, fname='HUS2_44_II10_g11500_Jaw0_kicks.txt', norm_energy=2.75)
hu44.add_kickmap('LV', nslice=25, fname='HUS2_44_II122000_g11500_Jaw0_kicks.txt', norm_energy=2.75)
hu44.add_kickmap('CR', nslice=25, fname='HUS2_44_II113250_g11500_Jaw0_kicks.txt', norm_energy= 2.75)
hu44.add_kickmap('LT', nslice=25, fname='HUS2_44_X112200_g11500_Jaw0_kicks.txt', norm_energy= 2.75)
hu44.list_kickmaps()
hu44.use_kickmap('LT')
hu44.active_kickmap
```
There's also a question if such an interface needs to be in Matlab or if Python-only changes are accepted.

swhite2401:
> There's also a question if such an interface needs to be in Matlab or if Python-only changes are accepted.

Python only changes are acceptable, we already have those for many things. This is true as long as you do not modify passmethod (C) which can be called by either interface, but looking at your PR this is not the case.

swhite2401:
For the other problem, I first need to understand your code....

gubaidulinvadim:
> For the other problem, I first need to understand your code....

The original fix in https://github.com/atcollab/at/commit/b4ebb16ed2679e8424635e8d7f15a3f9d31b7cd3 is much easier to understand :)

swhite2401:
Ok, could you send me a kick map file for testing?

gubaidulinvadim:
> Ok, could you send me a kick map file for testing?

I've attached the files. The kickmaps do not correspond to anything in SOLEIL II project :)

[HUS2_44_.zip](https://github.com/user-attachments/files/31373772/HUS2_44_.zip)

swhite2401:
@gubaidulinvadim, @oscarxblanco  , I am going to post questions / comments / suggestions along the way as I test this, here the first few (not necessarily from this PR but since this is the first time I use it here they are):

-If the idtable device becomes a list it should be possible to instantiate an empty element and then add tables or give a list of files (similar to a list or an array)
-Could you explicitly put all required input argument in the function signature otherwise these are not reflected in the API doc. Typically using *args is not helping the user
-The docstring is not very detailed, each positional and keyword argument should be described
_ from_user() shouldn't be private? Also the docstring is confusiong here: it looks like if should belong to the __init__() function of InsertionDeviceKickMap
- are set_DriftPass, set_IdTablePass and get_Passmethod really usefull? Can't you just read/set the Element attributes as for any other AT element?

But maybe I also misunderstood some of these things.... more to come

oscarxblanco:
@swhite2401 , you are right about these observations. Most of them are the consequence of me not having much experience with AT when I created this element about 5 years ago, so, I added what I thought could be useful.

> -If the idtable device becomes a list it should be possible to instantiate an empty element and then add tables or give a list of files (similar to a list or an array)

I don't think that having a list of tables inside one element is of any advantage wrt creating an element per table and changing the element.

However, this is @gubaidulinvadim 's idea and I think he develop it to have a simplified way of simulating several ID types and gaps, making of this element something closer to a library of IDs.

On the other hand, it could be a step to consider tracking when ID gaps are constantly changing.

>-Could you explicitly put all required input argument in the function signature otherwise these are not reflected in the API doc. Typically using *args is not helping the user

Right, it may be done in this PR ? otherwise I can fix it after this one is merged.

>-The docstring is not very detailed, each positional and keyword argument should be described
Right, I think it is just misplaced. I could reorganize it if not done in this PR.

> _ from_user() shouldn't be private? Also the docstring is confusiong here: it looks like if should belong to the init() function of InsertionDeviceKickMap

from_user was supposed to provide a way to inspect a table without creating the element. Sometimes the tables are a long to display and the info is difficult to plot. This was a way to read the file and get the data without creating the element. But, I agree that the documentation is not helpful. I could re organize it if not done in this PR.


>    are set_DriftPass, set_IdTablePass and get_Passmethod really usefull? Can't you just read/set the Element attributes as for any other AT element?

I thought they might be useful instead of manipulating the pass method names directly, but, this was just my lack of experience with AT. It was difficult to memorize and write many times the correct pass method per element with and without radiation, active or identity or drift, so, I decided to create something that sets the pass method name instead of having to type it every time.

gubaidulinvadim:
Most of your questions are for the original implementation of @oscarxblanco. I don't mind making the changes in this PR to avoid handling the conflict between changes. 

> @gubaidulinvadim, @oscarxblanco , I am going to post questions / comments / suggestions along the way as I test this, here the first few (not necessarily from this PR but since this is the first time I use it here they are):
> 
> - If the idtable device becomes a list it should be possible to instantiate an empty element and then add tables or give a list of files (similar to a list or an array)

Yes, we can do that and instantiate an ID without a kickmap that will have DriftPass PassMethod. 

> - Could you explicitly put all required input argument in the function signature otherwise these are not reflected in the API doc. Typically using *args is not helping the user

@oscarxblanco feel free to change the doc, I can also do it in this PR. 

> - The docstring is not very detailed, each positional and keyword argument should be described _ from_user() shouldn't be private? Also the docstring is confusiong here: it looks like if should belong to the **init**() function of InsertionDeviceKickMap

@oscarxblanco feel free to change the doc, I can also do it in this PR. from_user wasn't private originally. I was also thinking that it's only ever used in the __init__. We can make it public if necessary. My idea was that add_kickmap() kind of replaces from_user. 

 
>     * are set_DriftPass, set_IdTablePass and get_Passmethod really usefull? Can't you just read/set the Element attributes as for any other AT element?

This is from Oscar, but I find it useful. For IDs, one needs to set it to drift to study the impact of an ID and having a method for it is handy. get_Passmethod is maybe not so necessary. 

> 
> 
> But maybe I also misunderstood some of these things.... more to come

The idea of storing several tables is to have all necessary information in the lattice. And I find it a bit annoying having to create N different objects and swap them in a lattice by hand. But this is maybe my lack of experience with pyAT.

oscarxblanco:
@gubaidulinvadim , I will wait until this PR is closed to make any change that is left pending.

swhite2401:
Ok thanks for the clarifications!

I am also biased by my usage.... I realize that it is not so easy for first time user to disable an element.
So you think a generic `.enable() / .disable()` function for the `Element` object would be useful?

It would replace the passmethod by `DriftPass` or `IdentityPass` depending on the length. It is rather generic and would prevent have specific function like you proposed in this element.

I haven't started looking at your problem though...

gubaidulinvadim:
> Ok thanks for the clarifications!
> 
> I am also biased by my usage.... I realize that it is not so easy for first time user to disable an element. So you think a generic `.enable() / .disable()` function for the `Element` object would be useful?
> 
> It would replace the passmethod by `DriftPass` or `IdentityPass` depending on the length. It is rather generic and would prevent have specific function like you proposed in this element.
> 
> I haven't started looking at your problem though...

I think a standard enable/disable is good.

swhite2401:
Hi both,

So I had a quick and the problem seems to come from private attributes, starting with "_", `_kickmap_store` and `_active` are concerned.

You may simply remove the underscore to make it work, otherwise, you can use `_drop_attr` of the element class to exclude them. This option works only if these attributes are re-generated at initialization of the object, which seems to be the case

swhite2401:
> I think a standard enable/disable is good.

Ok I will work on that

gubaidulinvadim:
> Hi both,
> 
> So I had a quick and the problem seems to come from private attributes, starting with "_", `_kickmap_store` and `_active` are concerned.
> 
> You may simply remove the underscore to make it work, otherwise, you can use `_drop_attr` of the element class to exclude them. This option works only if these attributes are re-generated at initialization of the object, which seems to be the case

I'll try it out and see if this is better than the current solution.

swhite2401:
> I'll try it out and see if this is better than the current solution.

You can also use `_convert_attr`, see line 72 of `element_object.py`, this is done for alignment errors attributes.
If you use this you have to make sure that the constructor uses keywords without the underscore.

gubaidulinvadim:
@swhite2401, thanks for the tips. It is indeed better without underscore _kickmaps. Now this is called KickmapStore and some interface for MATLAB part of the code was also added by an LLM. There were still some issues with loading and saving .m and .mat files that required changes.

I've kept things backwards compatible with old interface, in case user provides only a single kickmap. Backward compatibility can be removed if this is not necessary. This will reduce the number of lines in the code :)

@oscarxblanco could you test if matlab part of the code is working properly?

pyat/at/load/matfile.py
```diff
@@ -317,6 +319,26 @@ def convert(value):
     matcls = line[:left].strip()[2:]
     build_attrs = _CLASS_MAP[matcls]._BUILD_ATTRIBUTES
     arguments = argsplit(line[left + 1 : right])
+    if (
```
oscarxblanco: @gubaidulinvadim , this function is supposed to use the class build attributes.
Could you explain why is it necessary to move Insertion Device attributes inside it and within an if ?

oscarxblanco:
Hi @gubaidulinvadim ,
the help message needs more info. It looks too generic
image

oscarxblanco:
@gubaidulinvadim ,
the code written in MATLAB by the LLM is using special names that already have other purposes, e.g. `legacy` or `options`.

Apart from that there seems to be no structure nor comments in other to understand what are those variables for. The only way I know to test LLM code is to put it through extensive LLM test and I don't have the information on what is legacy for and what are the new options for.

Could you provide this information with examples ?

gubaidulinvadim:
> Hi @gubaidulinvadim , the help message needs more info. It looks too generic image

Thanks, I've improved the 'matlab docstring' 
image
 

> @gubaidulinvadim ,
> the code written in MATLAB by the LLM is using special names that already have other purposes, e.g. legacy or options.
> 
> Apart from that there seems to be no structure nor comments in other to understand what are those variables for. The only way I know to test LLM code is to put it through extensive LLM test and I don't have the information on what is legacy for and what are the new options for.
> 
> Could you provide this information with examples ?

There are many issues for me to test matlab code: I do not use, I do not have matlab installed, so even an LLM cannot run it for me. I saw an easy solution and thought that this might work. This is only intended to change the interface to allow multiple ways to create a kickmap, which is now necessary if you load a lattice where InsertionDeviceKickmap stores several files ,namely, the same as in Python version.

```
ELEM=atinsertiondevicekickmap(FAMNAME)
    Create an empty insertion device with PASSMETHOD='DriftPass'.
 
  ELEM=atinsertiondevicekickmap(FAMNAME,PASSMETHOD,'FIELD1',VALUE1,...)
  ELEM=atinsertiondevicekickmap(FAMNAME,'FIELD1',VALUE1,...)
    Create an element from field/value pairs. PASSMETHOD defaults to
    'DriftPass'. This is the standard AT syntax and is used for elements
    written to an M-file by pyAT. Any element field may be supplied,
    including KickmapStore and ActiveKickmap.
 
  ELEM=atinsertiondevicekickmap(FAMNAME,PASSMETHOD,FILENAME,...
    NORMALIZATION_ENERGY,NSLICE,LENGTH,XKICK,YKICK,XKICK1,YKICK1,...
    XTABLE,YTABLE)
    Create an element with the original positional syntax. This form is
    retained for compatibility with existing MATLAB code.
```

Edit: I'll ask a colleague at SOLEIL to test MATLAB interface

pyat/at/load/utils.py
```diff
@@ -357,6 +357,40 @@ def split_ignoring_parentheses(
 
     Example: "l=0,hom(4,0.0,0)" -> ["l=0", "hom(4,0.0,0)"]
     """
+    if fence == ("\\(", "\\)"):
```
swhite2401: Is this string parsing really needed? This does not look very nice ....

swhite2401:
> > Hi @gubaidulinvadim , the help message needs more info. It looks too generic image
> 
> Thanks, I've improved the 'matlab docstring' image
> 
> > @gubaidulinvadim ,
> > the code written in MATLAB by the LLM is using special names that already have other purposes, e.g. legacy or options.
> > Apart from that there seems to be no structure nor comments in other to understand what are those variables for. The only way I know to test LLM code is to put it through extensive LLM test and I don't have the information on what is legacy for and what are the new options for.
> > Could you provide this information with examples ?
> 
> There are many issues for me to test matlab code: I do not use, I do not have matlab installed, so even an LLM cannot run it for me. I saw an easy solution and thought that this might work. This is only intended to change the interface to allow multiple ways to create a kickmap, which is now necessary if you load a lattice where InsertionDeviceKickmap stores several files ,namely, the same as in Python version.
> 
> ```
> ELEM=atinsertiondevicekickmap(FAMNAME)
>     Create an empty insertion device with PASSMETHOD='DriftPass'.
>  
>   ELEM=atinsertiondevicekickmap(FAMNAME,PASSMETHOD,'FIELD1',VALUE1,...)
>   ELEM=atinsertiondevicekickmap(FAMNAME,'FIELD1',VALUE1,...)
>     Create an element from field/value pairs. PASSMETHOD defaults to
>     'DriftPass'. This is the standard AT syntax and is used for elements
>     written to an M-file by pyAT. Any element field may be supplied,
>     including KickmapStore and ActiveKickmap.
>  
>   ELEM=atinsertiondevicekickmap(FAMNAME,PASSMETHOD,FILENAME,...
>     NORMALIZATION_ENERGY,NSLICE,LENGTH,XKICK,YKICK,XKICK1,YKICK1,...
>     XTABLE,YTABLE)
>     Create an element with the original positional syntax. This form is
>     retained for compatibility with existing MATLAB code.
> ```
> 
> Edit: I'll ask a colleague at SOLEIL to test MATLAB interface

Well if you can't test it then you should ask a matlab user to take care of this part.... or install matlab...
PRs involving LLM code ( or not LLM in fact) without testing it is really not good practice and should be avoided... it is already time consuming to review code but if now the reviewer is also responsible for basic debugging and testing this is really not nice

gubaidulinvadim:
> > > Hi @gubaidulinvadim , the help message needs more info. It looks too generic image
> > 
> > 
> > Thanks, I've improved the 'matlab docstring' image
> > > @gubaidulinvadim ,
> > > the code written in MATLAB by the LLM is using special names that already have other purposes, e.g. legacy or options.
> > > Apart from that there seems to be no structure nor comments in other to understand what are those variables for. The only way I know to test LLM code is to put it through extensive LLM test and I don't have the information on what is legacy for and what are the new options for.
> > > Could you provide this information with examples ?
> > 
> > 
> > There are many issues for me to test matlab code: I do not use, I do not have matlab installed, so even an LLM cannot run it for me. I saw an easy solution and thought that this might work. This is only intended to change the interface to allow multiple ways to create a kickmap, which is now necessary if you load a lattice where InsertionDeviceKickmap stores several files ,namely, the same as in Python version.
> > ```
> > ELEM=atinsertiondevicekickmap(FAMNAME)
> >     Create an empty insertion device with PASSMETHOD='DriftPass'.
> >  
> >   ELEM=atinsertiondevicekickmap(FAMNAME,PASSMETHOD,'FIELD1',VALUE1,...)
> >   ELEM=atinsertiondevicekickmap(FAMNAME,'FIELD1',VALUE1,...)
> >     Create an element from field/value pairs. PASSMETHOD defaults to
> >     'DriftPass'. This is the standard AT syntax and is used for elements
> >     written to an M-file by pyAT. Any element field may be supplied,
> >     including KickmapStore and ActiveKickmap.
> >  
> >   ELEM=atinsertiondevicekickmap(FAMNAME,PASSMETHOD,FILENAME,...
> >     NORMALIZATION_ENERGY,NSLICE,LENGTH,XKICK,YKICK,XKICK1,YKICK1,...
> >     XTABLE,YTABLE)
> >     Create an element with the original positional syntax. This form is
> >     retained for compatibility with existing MATLAB code.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Edit: I'll ask a colleague at SOLEIL to test MATLAB interface
> 
> Well if you can't test it then you should ask a matlab user to take care of this part.... or install matlab... PRs involving LLM code ( or not LLM in fact) without testing it is really not good practice and should be avoided... it is already time consuming to review code but if now the reviewer is also responsible for basic debugging and testing this is really not nice

I'll test it, no problem. I did not express myself correctly; I just wanted to get a quick check here if there's something unmatlabish.

pyat/at/load/utils.py
gubaidulinvadim: Yes, I do not like it either. But saving/loading the lattice in all formats does not seem to work without it. For example, I am saving a lattice in .m format, which produced the following snippet in .m (both .mat and .json are working without it):

```
atinsertiondevicekickmap('HU44', 'IdTablePass', 'Length', 2.156, 'ActiveKickmap', 'default', 'Filename_in', 'HUS2_44_X112200_g11500_Jaw0_kicks.txt', 'KickmapStore', struct('default', struct('Nslice', 25, 'Length', 2.156, 'xkick', [[8.4657999149389824e-08  6.8457706667608217e-08  3.4881100745141745e-08 -1.1677519567892139e-09 -3.9882960507044207e-08 -8.1334576649930687e-08 -1.2563285110816162e-07 -1.7309498162432923e-07 -2.2392616925076071e-07 -2.7812826277330827e-07 -3.3600793298380909e-07 -3.9787814856191082e-07 -4.6382005453662916e-07 -5.3417470090350688e-07 -6.0925733744305220e-07 -6.8928804001670829e-07 -7.7459279743541844e-07 -8.6549107080109636e-07 -9.6233791810125664e-07  ...
```

So, KickmapStore has nested parentheses, and loading fails if you try to load the lattice with it. I'm attaching the problematic lattice file (as a .txt because GitHub does not allow .m) and I'm also attaching .json. Maybe there's a better solution for it. If you have some other ideas, I can try them out. 

[test_saving_with_multiple_kickmaps.txt](https://github.com/user-attachments/files/32243573/test_saving_with_multiple_kickmaps.txt)
[test_saving_with_multiple_kickmaps.json](https://github.com/user-attachments/files/32245052/test_saving_with_multiple_kickmaps.json)

gubaidulinvadim:
Thank you for your comments @oscarxblanco @swhite2401. This is now ready for review. The main goal is to be able to have a single object in the lattice where kickmaps are stored for a particular insertion device, that can have different polarisation modes or different gap values. Then this data can be stored in the lattice file. Currently, lattice file only stores a single configuration of an ID. With this PR, several configuration can be stored and chosen. 

@swhite2401 , the issue with nested stuct() in .m file is solved in a different way (pyat/utils/load.py). I propose that we can move this to a separate PR if it needs more work, which I think is likely. I also do not want to break any existing functionality because I'm not sure the modified function is used only to parse .m files. 

I tested also that things are loadable in Matlab and added a utility function to swap active kickmap in Matlab. I've not modified the existing atidtable_dat.m that loads from file a single thing. 

You can test the code in Python by doing the following:

```
# Empty ID can be created and later filled-in
your_empty_id = at.idtable_element.InsertionDeviceKickMap('empty_id')
print(your_empty_id, your_empty_id.PassMethod)
>>>(InsertionDeviceKickMap('empty_id'), 'DriftPass')

# The old way of initialisation is preserved
your_id = at.idtable_element.InsertionDeviceKickMap('HU44', 25, 'data/kick_maps/2026-09-15/HUS2_44_II1_LV_Kicks.txt', 2.75)
print(your_id.active_kickmap, your_id.list_kickmaps())
>>> ('default', ['default'])

# This created an InsertionDeviceKickMap with several kickmaps stored inside, the first one listed is ActiveKickmap
your_id = at.idtable_element.InsertionDeviceKickMap('HU44', kickmaps={'LV': (25, 'kick_maps/2026-09-15/HUS2_44_II1_LV_Kicks.txt', 2.75),
																	  'LH': (25, 'kick_maps/2026-09-15/HUS2_44_II1_LH_Kicks.txt', 2.75)})
print(your_id)
>>>InsertionDeviceKickMap:
>>>       FamName: HU44
>>>        Length: 2.156
>>>    PassMethod: IdTablePass
>>> ActiveKickmap: LV
>>>   Filename_in: data/kick_maps/2026-09-15/HUS2_44_II1_LV_Kicks.txt
>>>  KickmapStore: {'LV': {'Nslice': 25, 'Length': 2.156, 'xk ...
>>>  ...
>>>  Normalization_energy: 2.75
>>>  Nslice: 25
>>>  xkick: ...
>>>  ...
print(your_id.ActiveKickmap)
>>>'LV'
print(your_id.list_kickmaps())
>>>['LV', 'LH']
your_id.use_kickmap('LH')
print(your_id.ActiveKickmap)
>>>'LH'
# It is possible to add kickmaps to an existing InsertionDeviceKickMap
your_id.add_kickmap('CR', nslice=25, fname='kick_maps/2026-09-15/HUS2_44_II1_C_kicks.txt', norm_energy= 2.75)
print(your_id.list_kickmaps())
>>>['LV', 'LH', 'CR']
# Adding to an existing key will override it
your_id.add_kickmap('CR', nslice=25, fname='kick_maps/2026-09-15/HUS2_44_II1_C_kicks.txt', norm_energy= 3.0)


```

You can test in MATLAB by doing the following
```
# Easiest is to first load a lattice with an ID
>> ring = atloadlattice('your_lattice_with_id.mat')
>> id_index = 1
>> your_id = ring{id_index}
>> ring{2159}

ans = 

  struct with fields:

                 FamName: 'HU44'
                  Length: 2.1560
              PassMethod: 'IdTablePass'
           ActiveKickmap: 'CR'
             Filename_in: 'HUS2_44_II10_g11500_Jaw0_kicks.txt'
            KickmapStore: [1×1 struct]
    Normalization_energy: 2.7500
                  Nslice: 25
                   xkick: [113×113 double]
                  xkick1: [113×113 double]
                  xtable: [1×113 double]
                   ykick: [113×113 double]
                  ykick1: [113×113 double]
                  ytable: [1×113 double]
                   Class: 'InsertionDeviceKickMap'

>> ring{2159}.KickmapStore

ans = 

  struct with fields:

    default: [1×1 struct]
         LH: [1×1 struct]
         LV: [1×1 struct]
         CR: [1×1 struct]
         LT: [1×1 struct]
>> ring{2159}.KickmapStore.LH

ans = 

  struct with fields:

    Nslice: 25
    Length: 2.1560
     xkick: [113×113 double]
     ykick: [113×113 double]
    xkick1: [113×113 double]
    ykick1: [113×113 double]
    xtable: [1×113 double]
    ytable: [1×113 double]
# Active kickmap can be swapped with a dedicated function (or manually if someone wants to do it that way)
>> swapKickmapData(ring{2159}, 'CR')

ans = 

  struct with fields:

                 FamName: 'HU44'
                  Length: 2.1560
              PassMethod: 'IdTablePass'
           ActiveKickmap: 'CR'
             Filename_in: 'HUS2_44_II10_g11500_Jaw0_kicks.txt'
            KickmapStore: [1×1 struct]
    Normalization_energy: 2.7500
                  Nslice: 25
                   xkick: [113×113 double]
                  xkick1: [113×113 double]
                  xtable: [1×113 double]
                   ykick: [113×113 double]
                  ykick1: [113×113 double]
                  ytable: [1×113 double]
                   Class: 'InsertionDeviceKickMap'

**Comment by oscarxblanco** [comment]
Hi @gubaidulinvadim ,
I still don't understand why all the insertion device build parameters need to be inside `matfile.py`
image

**Comment by gubaidulinvadim** [comment]
> Hi @gubaidulinvadim , I still don't understand why all the insertion device build parameters need to be inside `matfile.py` image

You are right! This is no longer necessary. I've had some issue with saving and loading .mat file with the new InsertionDeviceKickmap. This code was fixing this bug, but with the new changes it is not necessary and everything is loadable. I've removed it in the last commit.
**Edit: maybe not, the tests are failing without it. I'll check this.**

**Comment by gubaidulinvadim** [comment]
@oscarxblanco Yes, in fact there's an issue if you load an element using Lattice.load() because InsertionDeviceKickmap right now has different build attributes (ActiveKickmap, KickmapStore were added). So, if you try to load an element saved with the old format (without ActiveKickmap, KickmapStore). This will fail. 

I think the best would be to remove this code and have no backward compatibility for Lattice.load(). Would you agree?

**Comment by oscarxblanco** [comment]
Hi @gubaidulinvadim , I would prefer to keep backwards compatibility. 

My suggestion would be to change the element init, and build parameters so that it accepts the single kickmap as usual and all your modifications.

When I created the element I was more strict on the build parameters list, partially to understand pyat and also to make sure I didn't left anything behind, but it may change to your needs. They could be relaxed down to needing only Length, and pass all other parameters as keyword arguments.

**Comment by gubaidulinvadim** [comment]
> Hi @gubaidulinvadim , I would prefer to keep backwards compatibility.
> 
> My suggestion would be to change the element init, and build parameters so that it accepts the single kickmap as usual and all your modifications.
> 
> When I created the element I was more strict on the build parameters list, partially to understand pyat and also to make sure I didn't left anything behind, but it may change to your needs. They could be relaxed down to needing only Length, and pass all other parameters as keyword arguments.

I'm not sure I understand your suggestion. The init is changed already. That's the thing that causes backward compatibility issue. _build_attrs includes KickmapStore and ActiveKickmap now, so in matlab format it now has 4 more attributes. I'm not sure how a keyword argument would be defined in .m format.

**Comment by oscarxblanco** [comment]
The `InsertionDeviceKickMap` class inherits from the `Element` class the build attribute `FamName`.
https://github.com/atcollab/at/blob/84b3efc5cf7d97c85827560a0c400e375265783c/pyat/at/lattice/elements/element_object.py#L42-L45

Therefore, as long as you keep FamName you could modify  the build attributes specific for the insertion device, and deal inside the init all diferent cases, the ones I recall are: the case when the element is created as it is now, the case where the element exists in memory and needs to be represented, and then again both of the previous cases with all the new attributes you are adding.

Three years ago @lfarv told me that this could be done counting the number of args because it is a different number when you create the element and when you read the element.
image
So, after removing the build attributes, you could for example count the number of keyword arguments kwargs, or check if  they are defined.

The saving and loading of the file won't need to be changed as long as you keep the construction of the element from a dictionary. For example,
image

Keep in mind that the element class also uses `_conversions` to keep the type consistent inside of pyat disregarding the input format from mat, m, input from terminal or any other.
https://github.com/atcollab/at/blob/84b3efc5cf7d97c85827560a0c400e375265783c/pyat/at/lattice/elements/element_object.py#L46

**Comment by oscarxblanco** [comment]
> _build_attrs includes KickmapStore and ActiveKickmap now, so in matlab format it now has 4 more attributes. I'm not sure how a keyword argument would be defined in .m format.

Right, what I suggest is not to add more build attributes but to remove the build attributes and deal with the element creation using the kwargs.

**Comment by gubaidulinvadim** [comment]
@oscarxblanco thanks, i'll try. Maybe both Length and Energy need to stay as args to avoid accidentally wrong length or energy of the element.

**Comment by oscarxblanco** [comment]
@gubaidulinvadim ,
here is a second example where there is a nested dictionary
image

**Comment by gubaidulinvadim** [comment]
@oscarxblanco doesn't this also break backward compatibility of .m file? With a new fix, where energy and length are positional args I have the following in the .m file:
```
atinsertiondevicekickmap('HU44', 2.156, 2.75, 'IdTablePass', 'ActiveKickmap', 'default', 'Filename_in', 'data/kick_maps/2026-09-15/HUS2_44_II1_LV_Kicks.txt',...
```
And before it was
```
atinsertiondevicekickmap('HU44', 'IdTablePass', 'data/kick_maps/2026-09-15/HUS2_44_II1_LV_Kicks.txt', 2.75, 25, 2.156, 
...
```
And just to be clear, .mat file is always OK for save/load. It's only the .m that poses issues.

**Comment by oscarxblanco** [comment]
> @oscarxblanco doesn't this also break backward compatibility of .m file? With a new fix, where energy and length are positional args I have the following in the .m file:
> 
> ```
> atinsertiondevicekickmap('HU44', 2.156, 2.75, 'IdTablePass', 'ActiveKickmap', 'default', 'Filename_in', 'data/kick_maps/2026-09-15/HUS2_44_II1_LV_Kicks.txt',...
> ```
> 
> And before it was
> 
> ```
> atinsertiondevicekickmap('HU44', 'IdTablePass', 'data/kick_maps/2026-09-15/HUS2_44_II1_LV_Kicks.txt', 2.75, 25, 2.156, 
> ...
> ```
> 
> And just to be clear, .mat file is always OK for save/load. It's only the .m that poses issues.

The matlab function `atinsertiondevicekickmap` would need to be modified also to accept both formats.

FYI, in matlab the kick map table in .dat and .mat formats is read by `atidtable_dat.m` and it creates the element `atinsertiondevicekickmap`. There is another function called `atidtable.m` that used to be only for .mat kick map tables but now uses `atidtable_dat.m`.

**Comment by oscarxblanco** [comment]
There are two more functions involved in the .m file creation: `atguesclass.m` and `at2str`.

https://github.com/atcollab/at/blob/84b3efc5cf7d97c85827560a0c400e375265783c/atmat/lattice/at2str.m#L107-L121


https://github.com/atcollab/at/blob/84b3efc5cf7d97c85827560a0c400e375265783c/atmat/lattice/atguessclass.m#L67-L68

**Comment by gubaidulinvadim** [comment]
@oscarxblanco, I've checked that saving and loading the lattice with the new kickmap format works also in MATLAB. Guessclass was modifed slightly to just check when an empty ID is created (no kickmap, driftPass) because it should be recognised still as an ID (with Drift passMethod).

atidtable_dat.m was updated but only to register the kickmap as "default". I'm not sure I can update it to read many kickmaps. (this is mostly because this function does not work for me in matlab (error below is present for me on main branch of AT), so any modification that is not simple I cannot test.) This function is also optional, as the object (atinsertiondevicekickmap) can always be created; it is just not very convenient.

```
>> el = atidtable_dat('hu44', 25, 'HUS2_44_II10_g11500_Jaw0_kicks.txt', 2.75)
Index in position 1 exceeds array bounds (must not exceed 112).

Error in atidtable_dat (line 183)
        ykick = ykick(indy, indx);
```

oscarxblanco:
Could you please send me the file that is failing ?  ( 'HUS2_44_II10_g11500_Jaw0_kicks.txt')

gubaidulinvadim:
> Could you please send me the file that is failing ? ( 'HUS2_44_II10_g11500_Jaw0_kicks.txt')

I don't think the error is reproducible. SOLEIL colleagues use the MATLAB function without 
[HUS2_44_II113250_g11500_Jaw0_kicks.txt](https://github.com/user-attachments/files/32599476/HUS2_44_II113250_g11500_Jaw0_kicks.txt)
issue (but I cannot be sure because I use AT and they use their bootleg AT). And for me, I would never use matlab to create an ID. So, if it's only not working on my side, it's fine.

Issue #1131, opened 2026-08-19
VariableThinMultipole keeps PolynomA and PolynomB last value after tracking


oscarxblanco:
Dear all,
the VariableThinMultipole leaves the PolynomA and PolynomB in the state defined by the last turn tracked.
Note the difference between PolynomB before and after the track.

This behaviour could be incompatible with other functions and difficult to diagnose.

<img width="878" height="724" alt="Image" src="https://github.com/user-attachments/assets/f72f0b38-edf7-451d-b7d5-44497ac400ea" />

oscarxblanco:
Solved in #1132

PR #1132, opened 2026-08-19
Reset PolynomA and PolynomB of VariableThinMultipole after tracking

oscarxblanco:
This PR is a possible solution to issue #1131 .

PolynomA and PolynomB are set to zero at the end of the tracking function.

### DIFF
--- a/atintegrators/VariableThinMPolePass.c
+++ b/atintegrators/VariableThinMPolePass.c
@@ -20,6 +20,8 @@ struct elemab {
 struct elem {
     double* PolynomA;
     double* PolynomB;
+    double* PolynomA0;
+    double* PolynomB0;
     struct elemab* ElemA;
     struct elemab* ElemB;
     int Seed;
@@ -146,6 +148,12 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
             if (T2) ATaddvv(r6,T2);
         }
     }
+
+    /* restore polynoms to initial value.*/
+    for (i = 0; i < maxorder + 1; i++) {
+        pola[i] = Elem->PolynomA0[i];
+        polb[i] = Elem->PolynomB0[i];
+    }
 }
 
 #if defined(MATLAB_MEX_FILE) || defined(PYAT)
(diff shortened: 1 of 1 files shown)


oscarxblanco:
Dear all, who could review this pull request ?

atintegrators/VariableThinMPolePass.c
```diff
@@ -146,6 +146,13 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
             if (T2) ATaddvv(r6,T2);
         }
     }
+
+    /* reset the polynom values.
+    This is done to return the element to an unperturbed configuration.*/
+    for (i = 0; i < maxorder + 1; i++){
+        Elem->PolynomA[i] = 0;
```
swhite2401: Do you want to reset to zero or the initial value? 
I am not sure I can think of a case where the initial value would not be zero but who knows...

oscarxblanco: Understood, it is not a case I would imagined. I'll try to implement it.

swhite2401: If you think it is useless like this it is fine too, was more a question than a request

oscarxblanco:
@swhite2401 , I have implemented a possible solution to restore the polynoms. Here is the test and result it pyat and AT matlab.
```python
# test.py
import at
import numpy as np

v = at.VariableThinMultipole('v', AmplitudeB=[1e-3,2e-3,3e-3], FrequencyB=100, PhaseB=2*np.pi/8)
zin = np.zeros((6,1))
v.PolynomB[0] = -2
v.PolynomB[1] = -3
v.PolynomB[2] = -4
zout = v.track(zin)
print(zout)
print(v)
```
The pyat output
```bash
>>> exec(open('test.py').read())
>>> exec(open('test.py').read())
[[ 0.        ]
 [-0.00070711]
 [ 0.        ]
 [ 0.        ]
 [ 0.        ]
 [ 0.        ]]
VariableThinMultipole:
       FamName: v
        Length: 0.0
    AmplitudeB: [0.001 0.002 0.003]
    FrequencyB: 100.0
      MaxOrder: 2
          Mode: 0
      ModeName: SINE
      Periodic: True
        PhaseB: 0.7853981633974483
      PolynomA: [0. 0. 0.]
      PolynomB: [-2. -3. -4.]
```
And for AT matlab
```matlab
%% matlab
v = atvariablethinmultipole('v','AmplitudeB',1e-3*[1,2,3],'FrequencyB',100,'PhaseB',2*pi/8);
v.PolynomB = -2:-1:-4;
zin = zeros(6,1);
zout = elempass(v,zin);
v,zout'
```
the output is
```matlab
v = 
  struct with fields:
       FamName: 'v'
        Length: 0
          Mode: 0
      ModeName: 'SINE'
      PolynomA: [0 0 0]
      PolynomB: [-2 -3 -4]
    AmplitudeB: [1.0000e-03 0.0020 0.0030]
    FrequencyB: 100
        PhaseB: 0.7854
      MaxOrder: 2
ans =
   1.0e-03 *
         0   -0.7071         0         0         0         0
```

swhite2401 (approved):
Fine for me

PR #1135, opened 2026-08-25
Response fixes

swhite2401:
This PR propose 2 minor improvement for the response matrix interface:

1-OrbitResponseMatrix presently works only with KickAngle that is not defined in magnets elements by default. The user therefore has to first modify the lattice to use this function, this is not good. When the steerers list is provided by the user the default is now to act on K[ns]0L that are always defined.
A flag is provided to restore the previous behaviour when steerers are provided, when they are not the present behavior (use all element having KickAngle defined) is preserved

2-A warning is added in the calculation of the deviation when the target is not provided, this allows the user to correctly diagnose problems when the .correct() function is ineffective (targets not defined)

### DIFF
--- a/pyat/at/latticetools/observables.py
+++ b/pyat/at/latticetools/observables.py
@@ -66,8 +66,9 @@ from typing import ClassVar, Any
 
 import numpy as np
 import numpy.typing as npt
+import warnings
 
-from ..lattice import AtError, AxisDef, axis_, plane_
+from ..lattice import AtError, AxisDef, axis_, plane_, AtWarning
 from ..lattice import Lattice, Refpts, End
 
 RefIndex = int | tuple[int, ...] | slice
@@ -525,6 +526,9 @@ class Observable:
         if vnow is None:
             deviation = None
         elif self.target is None:
+            msg = (f"Target for observable {self.name} is None,"
+                   "this observable is therefore excluded from correction")
+            warnings.warn(AtWarning(msg))
             deviation = np.broadcast_to(0.0, vnow.shape)
         else:
             vsh = vnow.shape
--- a/pyat/at/latticetools/response_matrix.py
+++ b/pyat/at/latticetools/response_matrix.py
@@ -787,10 +787,12 @@ class OrbitResponseMatrix(ResponseMatrix):
     and optionally the sum of steerer angles named ``sum(h_kicks)`` or
     ``sum(v_kicks)``
 
-    The variable elements must have the *KickAngle* attribute used for correction.
-    It's available for all magnets, though not present by default
-    except in :py:class:`.Corrector` magnets. For other magnets, the attribute
-    should be explicitly created.
+    By default, and when the steerer are explicitly selected, integrated dipole strengh
+    *Kn0L* and *Ks0L* are used. In case *use_polynoms=False* and the list of steerers is
+    not provided the variable elements must have the *KickAngle* attribute used for correction.
+    It's available for all magnets, though not present by default except in
+    :py:class:`.Corrector` magnets. For other magnets, the attribute should be explicitly
+    created.
 
     By default, the observables are all the :py:class:`.Monitor` elements, and the
     variables are all the elements having a *KickAngle* attribute.
@@ -809,7 +811,7 @@ class OrbitResponseMatrix(ResponseMatrix):
         ring: Lattice,
         plane: AxisDef,
         bpmrefs: Refpts = Monitor,
-        steerrefs: Refpts = _orbit_correctors,
+        steerrefs: Refpts | None = None,
         *,
         cavrefs: Refpts = None,
         bpmweight: float | Sequence[float] = 1.0,
@@ -818,6 +820,7 @@ class OrbitResponseMatrix(ResponseMatrix):
         cavdelta: float | None = None,
         steersum: bool = False,
         stsumweight: float | None = None,
+        use_polynoms: bool = True,
     ):
         """
         Args:
@@ -827,8 +830,10 @@ class OrbitResponseMatrix(ResponseMatrix):
             bpmrefs:    Location of closed orbit observation points.
               See ":ref:`Selecting elements in a lattice <refpts>`".
               Default: all :py:class:`.Monitor` elements.
-            steerrefs:  Location of orbit steerers. Their *KickAngle* attribute
-              is used and must be present in the selected elements.
+            steerrefs:  Location of orbit steerers. If *use_polynoms=False,
+              their *KickAngle* attribute is used and must be present in the
+              selected elements, otherwise integrated dipole strength *Kn0L* and
+              *Ks0L* are used, they are present in all magnets.
               Default: All Elements having a *KickAngle* attribute.
             cavrefs:    Location of RF cavities. Their *Frequency* attribute
               is used. If :py:obj:`None`, no cavity is included in the response.
@@ -854,9 +859,9 @@ class OrbitResponseMatrix(ResponseMatrix):
 
         """
 
-        def steerer(ik, delta):
+        def steerer(ik, delta, attr_name, idx):
             name = f"{plcode}{ik:04}"
-            return RefptsVariable(ik, "KickAngle", index=pl, name=name, delta=delta)
+            return RefptsVariable(ik, attr_name, index=idx, name=name, delta=delta)
 
         def set_norm():
             bpm = LocalOpticsObservable(bpmrefs, "beta", plane=pl)
@@ -878,6 +883,19 @@ class OrbitResponseMatrix(ResponseMatrix):
             return cd, sw
 
         pl = plane_(plane, key="index")
+        if (pl!=0) and (pl!=1):
+            msg = "Orbit response can only be horizontal or vertical"
+            raise AtError(msg)
+        attr_name = "KickAngle"
+        idx = pl
+        if use_polynoms and steerrefs is not None:
+            if pl==0:
+                attr_name = "Kn0L"
+            if pl==1:
+                attr_name = "Ks0L"
+            idx = None
+        if steerrefs is None:
+            steerrefs = _orbit_correctors
         plcode = plane_(plane, key="code")
         ids = ring.get_uint32_index(steerrefs)
         nbsteers = len(ids)
@@ -892,10 +910,10 @@ class OrbitResponseMatrix(ResponseMatrix):
             # noinspection PyUnboundLocalVariable
             sumobs = LatticeObservable(
                 steerrefs,
-                "KickAngle",
+                attr_name,
                 name=f"{plcode}_kicks",
                 target=0.0,
-                index=pl,
+                index=idx,
                 weight=stsumweight if stsumweight else stsw / 2.0,
                 statfun=np.sum,
             )
@@ -903,7 +921,7 @@ class OrbitResponseMatrix(ResponseMatrix):
 
         # Variables
         variables = VariableList(
-            steerer(ik, delta) for ik, delta in zip(ids, deltas, strict=True)
+            steerer(ik, delta, attr_name, idx) for ik, delta in zip(ids, deltas, strict=True)
         )
         if cavrefs is not None:
             active = (el.longt_motion for el in ring.select(cavrefs))


pyat/at/latticetools/response_matrix.py
```diff
@@ -878,6 +883,19 @@ def set_norm():
...
+        if (pl!=0) and (pl!=1):
+            msg = "Orbit response can only be horizontal or vertical"
+            raise AtError(msg)
+        attr_name = "KickAngle"
+        idx = pl
+        if use_polynoms and steerrefs is not None:
+            if pl==0:
+                attr_name = "Kn0L"
+            if pl==1:
+                attr_name = "Ks0L"
```
lfarv, line 895: The problem is that `Kn0L` is the opposite of the horizontal deviation angle. So you will get the opposite of the response matrix.

lfarv:
I am thinking at another solution which is to redefine `KickAngle` as a property derived from `B[0]`, `A[0]` and `Length`, instead of a separate attribute. That way, it would always be available. That's a larger modification that can be part of the `magnet_integrators` branch. It would also make the integrators simpler.

pyat/at/latticetools/observables.py
```diff
@@ -525,6 +526,9 @@ def deviation(self) -> npt.NDArray[float]:
         if vnow is None:
             deviation = None
         elif self.target is None:
+            msg = (f"Target for observable {self.name} is None,"
+                   "this observable is therefore excluded from correction")
+            warnings.warn(AtWarning(msg))
```
lfarv, line 531: The response matrix computation works perfectly without targets so this warning does not look useful there.

But it generates some pollution: for instance the `observables.ipynb` notebook used in the documentation displays 14 such warnings, though it does not do any correction. In fact, simply printing observables (`print(obslist)`) generates the warning.

If necessary, this warning should be moved in places where a target is expected.

pyat/at/latticetools/response_matrix.py
swhite2401: Is this really a problem? RM algorithms work regardless of the sign or scale, in fact we may even allow for PolynomA/B[0].
I do not think we should hard code the attribute used to compute the RM, KickAngle could be the defaults why not but it should not be the only option.

swhite2401:
> I am thinking at another solution which is to redefine `KickAngle` as a property derived from `B[0]`, `A[0]` and `Length`, instead of a separate attribute. That way, it would always be available. That's a larger modification that can be part of the `magnet_integrators` branch. It would also make the integrators simpler.

I am fully in favor of redefining KickAngle, from what I know it is only used by historical and is really a duplication of functionalities as far as I am concerned. If it was just me I would even suppress it.

pyat/at/latticetools/observables.py
swhite2401: It works but it does not nothing and does not tell the user, presently if you follow the tutorial calling the function `correct()` does not correct because targets are not defined, this is why this warning is needed.

However, you are right this should only impact response matrices corrections, I will move it to the right place.

pyat/at/latticetools/response_matrix.py
lfarv: It is a problem: what you get is the response to a unit negative deviation angle.  So if you compute a correction, you will get the opposite of the desired correction angle. Try to inverse the sign of the response matrix in the ESRF autocor: you'll throw the beam away! If you want to correct with B0, this opposite angle is what you need. But the usual definition of an orbit response matrix is the response to a unit angle.

pyat/at/latticetools/observables.py
lfarv: Got it. So the check, if necessary, should be done on the `correct` method. It already checks for NaN deviations, it could also check that targets are defined.

Note that a missing target does not mean "ignore the value" but rather "do not modify the value".

This is not only for response matrix corrections. It applies to any kind of matching: a missing target mean "do not modify". Do you think that returning NaN instead of 0 would make more sense ?

swhite2401: Ah I thought missing target meant that the target is already achieved, so in effect if no target is defined nothing happens...

Yes maybe we could think of a different logic, in fact we could set the default to NaN and let the user define the "do not modify" behavior explicitly by setting it to None

pyat/at/latticetools/response_matrix.py
swhite2401: I disagree, AT is a simulation code and shall not be driven by SRCO considerations, in simulation you can correct the orbit with anything you want as long as you apply the the corrections to the same variable used to derive the RM, and in fact in this proposal using K0 everything works perfectly fine.

SO I insist that there should not be any strict requirement to use KickAngle.

lfarv: > in simulation you can correct the orbit with anything you want as long as you apply the the corrections to the same variable 

I fully agree with you. You can imagine building a response matrix from field integral (Kn0L) to closed orbit, or from quadrupole displacement to closed orbit, etc. Any can be used to correct the orbit, and all can be easily implemented in AT: you instantiate a `ResponseMatrix` with the desired variables and observables.

But none of these can be called "orbit response matrix". They have neither the right dimension (m/rad) not the right value. The "orbit response matrix" relates beam position to deviation angle, with the orientations non-ambiguously defined by the AT convention.

To turn it another way: for `OrbitResponseMatrix`, AT also provides a `build_analytical` method. It is not acceptable to have a `build` (tracking) and a `build_analytical` methods giving different results. And don't even think of changing the sign of the analytical response matrix: it's so widely described in the literature that changing that would introduce a huge confusion.

I think that for the use you mention (correct the orbit with anything you want), the ResponseMatrix class is perfectly suited. Depending on the chosen variables, the step size (for building the response), the range (for correction) are very different, and the Variable class allows to specify that easily.

lfarv: Afterthought: I see that the `OrbitResponseMatrix` provides special features for introducing the sum of steerer angles and RF frequency into the matrix. And a dedicated normalisation procedure.
These are not included in the general `ResponseMatrix` class. While the RF frequency can be easily introduced in the variables list, the rest is more tricky.

Maybe we can thing of a way th make this easier

swhite2401: Good, your definition of the ORM is just a convention and I do not see any reason to be so restrictive, the user should be allowed to correct orbit with whatever variable and profit from all the functionalities of the OrbitResponseMatrix class.
Being over-restrictive is unjustified and will just stop users from using this very nice functionality, which would be a pity...

I would just keep KickAngle as default and allow for any other attribute with proper documentation of the concerns combined with sanity checks when using the analytical method.

swhite2401: Question: is the sign of K0nL in AT following the same convention as other codes such as XSuite? If not we should consider aligning AT with the most common convention.

I think it is still time to do it for these attributes that are probably not very much used since they were introduced only recently....

lfarv: In all codes I know, a positive vertical field integral gives a negative horizontal kick. The AT field expansion is identical to the one of Xsuite, except for the factorial n! at denominator. Same for Elegant.

Anyway, if you get a + in one plane, you get a minus in the other one. Maxwell dixit

swhite2401: Ok fair enough, I'll get back at this PR next week and implement your comments, for the moment I am busy....

lfarv: > your definition of the ORM is just a convention

Yes, but it's the one used everywhere. I did a search on internet. The first ten papers (DESY, LOCO, Cornell, CERN…) define the ORM as:

$[\overrightarrow{\Delta x}] = M . [\overrightarrow{\theta_x}] \; \;$

Here is a typical definition (from BNL):

--------
capture

--------

And that's what the analytical formula computes. Again, it's unacceptable to have tracking and analytical giving different results.

What is the problem in using the general `ResponseMatrix` flexible class ? Would you like a more specific "OrbitCorrectionMatrix" similar but taking into account the RF frequency and sum of steerers ?

swhite2401: > Again, it's unacceptable to have tracking and analytical giving different results.

And this is why the default will provide the same result, in case user decides to  use different variable it is his responsibility and analytical calculation will not be available.

I do not understand the problem with that, why do you insist on being so strict? The OrbitResponseMatrix interface provides many features and simplications that are very useful, why do you want to force the user to re-implement these in a general class when they are already available.
I advocate to give flexibility to the user, as long as it is correctly documented and limits are clearly sated this perfectly acceptable, we should not be dogmatic with our own vision and conventions.

I am a user and I need this feature, should I make a new that inherits from  OrbitResponseMatrix class with a different name and analytical method disabled to satisfy your requirements?

lfarv: So let's look at an interface which would give the full flexibility you need, and will see at the end if it should be added to the existing OrbitResponseMatrix or if it's better implemented as a new class.

Concerning variables, for full flexibility I would go to a VariableList: you can mix different attributes and everything dealing with step size and range is already taken into account. Optionally adding the RF frequency is easy. Normalisation is more difficult: at the moment this is done using an analytical estimate of the magnitude of each column. For arbitrary attributes, this cannot be done, so then the normalisation would need a 1st computation of the matrix. Very time-consuming!
For the energy preservation, it's also a problem: what has to guaranteed is that the sum of horizontal **deviations** is zero. For arbitrary attributes, how to convert to deviations?

Therefore I think we should separate 3 use cases:

1. for looking at errors and their consequences, you could mix everything: magnet displacements, field errors…: then you need a full flexibility for variables. But no need for normalisation or energy preservation. There the general ResponseMatrix is well suited

2. for "usual" correction, all the attributes should be identical, and proportional to deviation. Then normalisation is easy. But assuming that an angular kick is available for all elements, why would you use something else ? We are back to small adaptations of the present OrbitResponseMatrix

3. for fancy corrections using exotic variables, it's probably to the user to take care of normalisation with proper deltas, and of energy conservation with a dedicated observable. I do not see how this can be automated. So back to the general ResponseMatrix.

lfarv:
@swhite2401 : here is a proposal which combines with minimal work what you propose in this PR and the evolution of magnet integrators:

I wrote above that the `KickAngle` attribute was creating problems in integrators: incompatibility with free-threaded python, problems with multiprocessing. Solving that implies a complexity slowing down the integrators. So in the `magnet_interators` branch I  remove the attribute and replace it with a property acting on PolynomA/B[0].

My proposal is to partially anticipate this change: I can right now create the new property replacing `KickAngle`, without changing anything else. The `KickAngle` attribute remains as it is for the moment, the integrators are unchanged. I propose 2 properties `HKick/VKick`, or `HAngle/VAngle` (to be chosen).

For the response matrix, these new properties replace `KickAngle`. This avoids all the discussion about creating the attributes when necessary.

Then, you can open the possibility for the user to replace `HKick/VKick` be the attribute of his choice, disabling the analytical computation in this case.

I can propose the new properties for all magnets in a dedicated PR some time to-day.

swhite2401:
> @swhite2401 : here is a proposal which combines with minimal work what you propose in this PR and the evolution of magnet integrators:
> 
> I wrote above that the `KickAngle` attribute was creating problems in integrators: incompatibility with free-threaded python, problems with multiprocessing. Solving that implies a complexity slowing down the integrators. So in the `magnet_interators` branch I remove the attribute and replace it with a property acting on PolynomA/B[0].
> 
> My proposal is to partially anticipate this change: I can right now create the new property replacing `KickAngle`, without changing anything else. The `KickAngle` attribute remains as it is for the moment, the integrators are unchanged. I propose 2 properties `HKick/VKick`, or `HAngle/VAngle` (to be chosen).
> 
> For the response matrix, these new properties replace `KickAngle`. This avoids all the discussion about creating the attributes when necessary.
> 
> Then, you can open the possibility for the user to replace `HKick/VKick` be the attribute of his choice, disabling the analytical computation in this case.
> 
> I can propose the new properties for all magnets in a dedicated PR some time to-day.

Ok sounds good thanks!

swhite2401:
I will take care of displacing the warning in this PR and change KickAngle to H/VKick as you propose.
I can create these additional properties in this PR too, it is straight forward.

Is that ok or you already started a new PR??

swhite2401:
Concerning weights it should be ok, for the steerer sum I will disable it together with the analytical calculation at instantiation.

lfarv:
> I can create these additional properties in this PR too, it is straight forward.

Sorry, @swhite2401 : I had It ready and created the PR before reading you comment…

lfarv:
Now 2 questions:

- for choosing an alternate attribute: you can keep the `use_polynom` flag, but it's restricted to a predefined attribute and for me it needs more explanation. Or you can propose an attribute name as optional input. 
- for the default element selection: though it can stay as it is, I think being based on `KickAngle` is no more a useful choice. Since now no one would explicitly create `KickAngle` attributes on magnets one could replace
  ```python
  _orbit_correctors = checkattr("KickAngle")
  ```
  by
  ```python
  _orbit_correctors = checktype("Corrector")
  ```
  Any other idea ?

swhite2401:
> Now 2 questions:
> 
>     * for choosing an alternate attribute: you can keep the `use_polynom` flag, but it's restricted to a predefined attribute and for me it needs more explanation. Or you can propose an attribute name as optional input.
> 
>     * for the default element selection: though it can stay as it is, I think being based on `KickAngle` is no more a useful choice. Since now no one would explicitly create `KickAngle` attributes on magnets one could replace
>       ```python
>       _orbit_correctors = checkattr("KickAngle")
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       by
>       ```python
>       _orbit_correctors = checktype("Corrector")
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       Any other idea ?

I would go for name attribute as optional input with of course all the explanantions needed to clarify why some features are disabled in this case.

Concerning the default selection, we could even remove it right? It is simple enough to provide widlcards no?
In any case if we want to keep it using Correctors is fine

swhite2401:
> > I can create these additional properties in this PR too, it is straight forward.
> 
> Sorry, @swhite2401 : I had It ready and created the PR before reading you comment…

No worries, I will rebase master once it is merged

lfarv:
No problem with removing the default steerer selection!

swhite2401:
Hello @lfarv , I am having difficulties understanding the logic you put in the response matrices. As suggested I set the default target to nan that results in a nan deviation. If I do nothing which returns an error as expected (for the moment I would like to have an option to automatically exclude nans).

However, I would expect that if I exclude the observable with nan target, this computation runs without errors. This is presently not the case, here is the code I used to test it:

```
target = np.zeros(320)
target[0] = np.nan    
resph = at.OrbitResponseMatrix(ring, 'x', "BPM*", "S[HFD]*", bpmtarget=target)
resph.exclude_obs(obsid=0, refpts="BPM_01")
resph.build(use_mp=True)

print(resph.response.shape)
#plt.contour(resph.response)
#plt.show()

q0.dx = 100.0e-6
_, orb0 = ring.find_orbit(refpts="BPM*")
resph.correct(ring, apply=True)
_, orb1 = ring.find_orbit(refpts="BPM*")
```

By the way the interface is really not user friendly in my opinion, and the code very difficult to follow as you have many interleaved functions... I do not see this as really usable for standard user for the moment but this is something else.

First point is when you have a combination of Observables, and ObservableList read/write should be done on the flattened array of all observables, when should not need `obsid` in `exclude_obs()` for instance.

swhite2401:
@lfarv in fact I think there is a bug with exclusion regardless of target values. Seeting all targets to 0.

With BPM_01 excluded:
image

With all BPM:
image


This is not what I expect at all....

lfarv:
@swhite2401 : From your plot, it looks like BPM_01 appears in several places in your lattice, and they are all excluded. If so, you should specify in refpts the location of the 1st BPM_01… Not easy, I agree

> First point is when you have a combination of Observables, and ObservableList read/write should be done on the flattened array of all observables, when should not need obsid in exclude_obs() for instance.

Specifying exclusions is indeed difficult and can be discussed. Internally, you are right, a flattened array of observables is used for computation. But in the general case where there are an arbitrary number of observables with any length each, you cannot ask the user to give the index in the flattened array. He would have to figure out the length of each observable, and then add the exclusion index in the desired observable to length of all previous ones. That's the reason for the `obsid` argument. For orbit response matrix, obsid = 0 is the default, you can forget it.

For the particular case of orbit response matrix, we could imagine another solution. And also, maybe the `refpts` argument is not the best choice. It could be the index inside the observable (bpm number in that case). At the moment, if you give an integer in refpts, it's the location in the ring, not the index in the BPM array

swhite2401:
> From your plot, it looks like BPM_01 appears in several places in your lattice, and they are all excluded. If so, you should specify in refpts the location of the 1st BPM_01… Not easy, I agree

Oh ok, I thought this was pointing to to observable names, and that unique names were automatically generated...

In fact using index in the observable list would be better than lattice refpts, similarly using observable name is better, once the response object is defined it is more natural to work with attributes belonging to it

lfarv:
> Oh ok, I thought this was pointing to to observable names, and that unique names were automatically generated...

In fact, there is a single `OrbitObservable`, obsid = 0, with multiple values (a numpy array). It is named "orbit[x]". But each value in the array has no specific name.

Note also that there is no correlation between exclusion and target values. This is because an exclusion (or re-inclusion) need to force a new inversion on the matrix. On the other hand, modifying the targets with the `set_target` method, for instance to set a "golden orbit", does not require a new inversion.

swhite2401:
> In fact, there is a single `OrbitObservable`, obsid = 0, with multiple values (a numpy array). It is named "orbit[x]". But each value in the array has no specific name.

So the observables in an ObservableList have no individual names?

swhite2401:
> Note also that there is no correlation between exclusion and target values. This is because an exclusion (or re-inclusion) need to force a new inversion on the matrix. On the other hand, modifying the targets with the `set_target` method, for instance to set a "golden orbit", does not require a new inversion.

Ok, understood, however, if a nan is detected when the correct() function is called we may as an option exclude that observable and recompute the inverse matrix.

Alternatively, we could automatically set the target to None as we would do in operation.

An easy thing to do as a start would be to put a more explicit error, saying which observables are nan and suggesting to exlclude them or set their target correctly.

lfarv:
> So the observables in an ObservableList have no individual names?

Oh yes, each `Observable` in an `ObservableList` keeps its name. But here the list contains 2 observables:
`resph.observables[0]` is an `OrbitObservable` with name "orbit [x]"
`repsh.observables[1]` is a `LatticeObservable` with name "sum(x_kicks)"

You can check this with `print(reps.observables)`

lfarv:
> Ok, understood, however, if a nan is detected when the correct() function is called we may as an option exclude that observable and recompute the inverse matrix.

> Alternatively, we could automatically set the target to None as we would do in operation.

> An easy thing to do as a start would be to put a more explicit error, saying which observables are nan and suggesting to exclude them or set their target correctly.

All this can be done. But the idea is not to act on the target. The BPM target is at 0 by default, it could be set to a golden orbit. But there is no reason to change the target of an undesired BPM: the idea is to disable it in the response, not to change its target.

But more generally a warning on non-numeric (NaN or None) target can of course be added when a correction is attempted.

swhite2401:
> > Ok, understood, however, if a nan is detected when the correct() function is called we may as an option exclude that observable and recompute the inverse matrix.
> 
> > Alternatively, we could automatically set the target to None as we would do in operation.
> 
> > An easy thing to do as a start would be to put a more explicit error, saying which observables are nan and suggesting to exclude them or set their target correctly.
> 
> All this can be done. But the idea is not to act on the target. The BPM target is at 0 by default, it could be set to a golden orbit. But there is no reason to change the target of an undesired BPM: the idea is to disable it in the response, not to change its target.
> 
> But more generally a warning on non-numeric (NaN or None) target can of course be added when a correction is attempted.

Well there are 2 ways to exclude an observable:
1-Remove it from the response matrix, this requires to recompute the inverse as you mention
2-Set its deviation to zero, this is not a complete removal but rather the equivalent of setting a weight to zero. It is often used in operation to avoid recomputing the matrix

What I was thinking was that we could have an option "Ignore_Nans" that would automatically set the deviation to zero for NaN observables

swhite2401:
> Oh yes, each `Observable` in an `ObservableList` keeps its name.

Yes but then there has to be a wait to point to individual observable names instead of observablelist names...

lfarv:
> 2-Set its deviation to zero, this is not a complete removal but rather the equivalent of setting a weight to zero. It is often used in operation to avoid recomputing the matrix

I repeat: setting the deviation to zero is not at all equivalent to setting the weight to zero : if you multiply the vector with a 0 deviation by the inverse matrix, you will get a zero displacement on the "excluded" BPM : setting the deviation to zero mathematically means: don't move !

On the other hand, setting the weight to zero roughly means: don't take into account. That's completely different. But changing the weights implies a new inversion, because what is inverted is the weighted matrix.

> What I was thinking was that we could have an option "Ignore_Nans" that would automatically set the deviation to zero for NaN observables

If you set the target to None, you get a zero deviation, whatever the value. 

> Yes but then there has to be a wait to point to individual observable names instead of observablelist names...

Did you try the `print` command I suggested above? There you see that the position, deviation etc. are printed facing the BPM name. So of course you can identify each value. You need the observable name + an index inside its array.
But as you experienced above, several elements in a lattice may have the same name. Using an index is more selective. The question is to decide if it better to use an index with respect to the lattice, which is how it works presently with the `refpts` argument of `exclude_obs`, or an index with respect to the observable (BPM number).

swhite2401:
No I haven't had time to look further into this. Probably next week.
I'll make a proposal and we can discuss further then.

PR #1136, opened 2026-08-25
Variablethinmultipole WHITENOISE buffer

oscarxblanco:
The PR adds the possibility to have a buffer for the WHITENOISE mode of the VariableThinMultipole.

Two new parameters are added: `BufferSize` and `BufferOffset`. The buffer size sets the lenght of the buffer and allows to store in an array the random value used in every turn, while the buffer offset sets the first turn to be stored in the array.

The parameter seed is removed because it had no use in this pass method. Instead, the seed for the random number generator should be set through the tracking functio, and it is mentioned in the help message of the element.

### DIFF
--- a/atintegrators/VariableThinMPolePass.c
+++ b/atintegrators/VariableThinMPolePass.c
@@ -16,6 +16,7 @@ struct elemab {
     double Sinmin, Sinmax;
     int NSamples;
     double* Func;
+    double* Buffer;
 };
 
 struct elem {
@@ -35,9 +36,7 @@ struct elem {
     double *T2;
     double *EApertures;
     double *RApertures;
-    double *Buffer;
-    double BufferSize;
-    double BufferOffset;
+    int BufferSize, BufferOffset;
 };
 
 double get_amp(double amp, double* ramps, double t)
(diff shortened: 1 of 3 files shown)


oscarxblanco:

swhite2401:
@oscarxblanco , isn't this already covered by FuncA, FuncB? -> the user can provide an arbitrary buffer

oscarxblanco:
Hi @swhite2401 ,
I understand that this history buffer could be used to calculate linear optics parameters from gaussian noise. You mentioned it in this conversation
https://github.com/atcollab/at/pull/839#issuecomment-2592947438
I will quote it here again,
> ... The other one is a bit more tricky, because you want to be able to access what is applied to the beam. However, I understand the complaints, we could instead think of adding a history buffer attribute that is filled during the tracking. This already exists for other elements like the beam monitor or beam loading for example, what do you think?
I you decide to go for this option you have to be very careful: the buffer needs to be properly initialized depending on the number of turns otherwise you will get segmentation faults

swhite2401:
These are 2 different things, this comments was relating to the possible of saving a history of Polynom[A/B], to record was was applied to the beam turn by turn.

What I mention here is case 2 in the passmethod, it allow to load any buffer that would then be applied periodically to the beam, which I believe covers what you implemented here

oscarxblanco:
@swhite2401 , ok, this is my mistake. Sorry.

oscarxblanco:
@swhite2401 , I will close this PR. Thank you again, and sorry for the distraction.

Issue #1139, opened 2026-08-26
Creating a VariableThinMultipole mode with double quotes in MATLAB is not loaded in pyat.


oscarxblanco:
When creating a variable thin multipole in matlab with mode in double quotes (see "WHITENOISE"), and saving it .mat, the file can not be loaded in pyat.
```matlab
v = atvariablethinmultipole('v',"WHITENOISE",'AmplitudeA',1e-3);
rp = {atringparam('ring',1e9)};
ring = [rp; v];
ring = atsetenergy(ring,1e9);
ring = atenable_6d(ring);
save("ring.mat",'ring');
```

In pyat

<img width="1211" height="390" alt="Image" src="https://github.com/user-attachments/assets/94316b41-9809-4614-91a0-1ff457ee2465" />

oscarxblanco:
Solved in #1142

PR #1140, opened 2026-08-26
allow double quotes in mode name for atvariablethinmultipole

oscarxblanco:
This PR solves issue #1139 , where a mode name double quoted created a .mat file incompatible with pyat.

The string is converted to characters to make it readable in pyat.

### DIFF
--- a/atmat/lattice/element_creation/atvariablethinmultipole.m
+++ b/atmat/lattice/element_creation/atvariablethinmultipole.m
@@ -63,7 +63,7 @@ function elem=atvariablethinmultipole(fname,varargin)
 
 [modename, rsrc] = getargs(varargin,'SINE', ...
                    'check',@(arg) any(strcmpi(arg,{'SINE','WHITENOISE','ARBITRARY'})));
-[modename, rsrc] = getoption(rsrc,'ModeName',modename);
+[modename, rsrc] = getoption(rsrc,'ModeName',convertStringsToChars(modename));
 if ~any(strcmpi(modename,{'SINE','WHITENOISE','ARBITRARY'}))
   error("ModeName should be 'SINE', 'WHITENOISE' or 'ARBITRARY'");
 end
--- a/atmat/lattice/element_creation/atvariablethinmultipole.m
+++ b/atmat/lattice/element_creation/atvariablethinmultipole.m
@@ -62,7 +62,8 @@ function elem=atvariablethinmultipole(fname,varargin)
 
 [modename, rsrc] = getargs(varargin,'SINE', ...
                    'check',@(arg) any(strcmpi(arg,{'SINE','WHITENOISE','ARBITRARY'})));
-[modename, rsrc] = getoption(rsrc,'ModeName',char(modename));
+[modename, rsrc] = getoption(rsrc,'ModeName',modename);
+modename = char(modename);
 [~, rsrc] = getoption(rsrc,'Mode',2); % remove Mode, the element is set by ModeName
 [method,rsrc]   = getargs(rsrc,'VariableThinMPolePass', ...
                   'check',@(arg) (ischar(arg) || isstring(arg)) && endsWith(arg,'Pass'));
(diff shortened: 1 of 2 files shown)


oscarxblanco:

oscarxblanco:
> Hi all. I’m away for another 2 weeks and I cannot look at the code… Matlab « strings » create problems, did you look at what happens for element names in double quotes ? Are they converted to char arrays ?Le 26 août 2026 à 13:52, oscarxblanco ***@***.***> a écrit :﻿oscarxblanco left a comment ([atcollab/at#1142](https://github.com/atcollab/at/pull/1142)) Dear all, who could review this PR ? —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>

Hi @lfarv !
I just checked and FamName has the same behaviour. If I double quote the FamName of an element I get an error in pyat.

For example, saving a ring with a drift as d = atdrift("d",0.1); where the name is double quoted makes pyat fail when loading the .mat file.

oscarxblanco:
@lfarv ,
the function `convertStringsToChars` could be used to convert the MATLAB strings to a char array which is compatible with the pyat read_mat method.

atmat/lattice/element_creation/atvariablethinmultipole.m
```diff
@@ -63,7 +63,7 @@
 
 [modename, rsrc] = getargs(varargin,'SINE', ...
                    'check',@(arg) any(strcmpi(arg,{'SINE','WHITENOISE','ARBITRARY'})));
-[modename, rsrc] = getoption(rsrc,'ModeName',modename);
+[modename, rsrc] = getoption(rsrc,'ModeName',convertStringsToChars(modename));
```
lfarv: I would rather use
```Matlab
[modename, rsrc] = char(getoption(rsrc,'ModeName',modename));
```
- using `char()` instead of `convertStringToChars()` avoids bumping the minimum Matlab version required to 2017b where this function was introduced. `convertStringToChars()` leaves "non-text" arguments unchanged while `char()` tries to convert any argument: that's good enough here.
- converting the result of `getoption` also protects the case where `ModeName` would be given as a keyword argument.

atmat/lattice/element_creation/atvariablethinmultipole.m
```diff
@@ -63,7 +63,7 @@
...
 if ~any(strcmpi(modename,{'SINE','WHITENOISE','ARBITRARY'}))
   error("ModeName should be 'SINE', 'WHITENOISE' or 'ARBITRARY'");
 end
```
lfarv: This should already have been tested in `getargs()` above.

lfarv:
```Matlab
    function rsrc = setparams(rsrc,modename,ab)
        amplarg=strcat('Amplitude',ab);
        if isfield(rsrc,amplarg)
            switch modename
                case "SINE"
                    rsrc = setsine(rsrc,ab);
                case "ARBITRARY"
                    rsrc = setarb(rsrc,ab);
                case "WHITENOISE"
                    rsrc = setwhitenoise(rsrc,ab);
            end
        end
    end
```

Since ModeName has been converted to `char`, one should use single quotes here.

lfarv:
To protect also `FamName` and `Method` in any element, you could change line 18 in `atbaselem():`

From
```Matlab
elem           = struct('FamName',famname,'PassMethod',method,'Length',lg,rsrc{:});
```
to
```Matlab
elem           = struct('FamName',char(famname),'PassMethod',char(method),'Length',lg,rsrc{:});

```

atmat/lattice/element_creation/atvariablethinmultipole.m
oscarxblanco: There should be two outputs but char returns one.
image

atmat/lattice/element_creation/atvariablethinmultipole.m
oscarxblanco: Right, I might have done some test with this before having a default. I will remove it.

atmat/lattice/element_creation/atvariablethinmultipole.m
lfarv: Oh yes sorry. So it must be
```Matlab
[modename, rsrc] = getoption(rsrc,'ModeName',modename);
modename=char(modename);
```

oscarxblanco:
@lfarv , all changes have been implemented.

lfarv (approved):
If you checked that everything is now working, that's ok for me.

PR #1143, opened 2026-08-27
function to enable/disable elements

swhite2401:
This PR allows to disable / enable elements in a lattice, converting its pass method to IdentityPass or DriftPass, with a function call.
When call the enable function the pass method is restored to the value before `disable()` was called.
This value is generally set at initialization, but could be changed by the user.

### DIFF
--- a/pyat/at/lattice/elements/element_object.py
+++ b/pyat/at/lattice/elements/element_object.py
@@ -256,7 +256,6 @@ class Element:
             if not np.array_equal(v, getattr(defelem, k, None))
             and k not in self._drop_attr
         }
-        print(keywords)
         return self.__class__.__name__, arguments, keywords
 
     def items(self) -> Generator[tuple[str, Any], None, None]:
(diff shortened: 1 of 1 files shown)


gubaidulinvadim (approved):

swhite2401:
Sorry I had left a print statement .... needs re-approval

pyat/at/lattice/elements/element_object.py
```diff
@@ -91,6 +94,11 @@ def __init__(self, family_name: str, **kwargs):
         self.FamName = family_name
         self.Length = kwargs.pop("Length", 0.0)
         self.PassMethod = kwargs.pop("PassMethod", "IdentityPass")
+        self._enable_passmethod = self.PassMethod
+        if self.Length==0.0:
```
oscarxblanco: @swhite2401 , shouldn't this depend on  the class LongElement ?

gubaidulinvadim (approved):

PR #1144, opened 2026-08-27
Atvariablethinmultipole func derivatives

oscarxblanco:
This PR implements the posibility to create a function FUNC in the ARBITRARY mode of the VariableThinMultipole with any order of derivatives with respect to time $c\tau$.

### DIFF
--- a/atintegrators/VariableThinMPolePass.c
+++ b/atintegrators/VariableThinMPolePass.c
@@ -179,7 +179,7 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
     /* mode 1 : random value applied to all particles */
     /* mode 2 : custom function */
 
-    if (mode != 0) {
+    if (mode == 1) {
         for (i = 0; i < maxorder + 1; i++) {
             /* calculate the polynom to apply on all particles */
             /* the particle delay time, tpart, is not used */
--- a/atintegrators/VariableThinMPolePass.c
+++ b/atintegrators/VariableThinMPolePass.c
@@ -312,7 +312,7 @@ ExportMode struct elem* trackFunction(const atElem* ElemData, struct elem* Elem,
     }
     double t0 = Param->T0; // revolution time of the nominal ring
     int turn = Param->nturn;  // current turn
-    VariableThinMPolePass(r_in, Elem, t0, turn, num_particles, Param->thread_rng);
+    VariableThinMPolePass(r_in, Elem, t0, turn, num_particles, Param->common_rng);
     return Elem;
 }
 
--- a/atintegrators/VariableThinMPolePass.c
+++ b/atintegrators/VariableThinMPolePass.c
@@ -98,7 +98,7 @@ double get_pol(struct elemab* elem, double* ramps, int mode,
     double* amp = elem->Amplitude;
     double funcdelay, functot;
     double tpow;
-    int dorder;
+    int allorders;
     if (!amp) {
         return 0.0;
     }
@@ -122,19 +122,19 @@ double get_pol(struct elemab* elem, double* ramps, int mode,
             funcdelay = elem->FuncDelay;
             turnidx_modn = turn % elem->NSamples;
             ampt *= func[turnidx_modn];
-            dorder = elem->Dorder;
+            allorders = elem->Dorder + 1;
             /* Calculate the amplitude of function func in sample turnidx_modn */
-            functot = func[dorder*turnidx_modn]; //order zero
-            /* Add dorder derivatives for the sample turnidx mod NSamples*/
+            functot = func[(allorders)*turnidx_modn]; //order zero
+            /* Add derivatives for the sample turnidx mod NSamples */
             tpow = 1;
             t = t*C0 - funcdelay; // change to ctau coordinates and offset
-            for (i=1;i<dorder;i++){
+            for (i=1;i<allorders;i++){
               /* indexing is fortran-like. We start with columns.
                  cols are derivative components.
                  rows are turn samples.
               */
               tpow = tpow * t;
-              functot = functot + tpow * func[i + dorder*turnidx_modn];
+              functot = functot + tpow * func[i + allorders*turnidx_modn];
             };
             return ampt * functot;
         } else {
@@ -268,8 +268,8 @@ ExportMode struct elem* trackFunction(const atElem* ElemData, struct elem* Elem,
         Seed=atGetOptionalLong(ElemData, "Seed", 0); check_error();
         NSamplesA=atGetOptionalLong(ElemData, "NSamplesA", 1); check_error();
         NSamplesB=atGetOptionalLong(ElemData, "NSamplesB", 1); check_error();
-        DorderA=atGetOptionalLong(ElemData, "DorderA", 1); check_error();
-        DorderB=atGetOptionalLong(ElemData, "DorderB", 1); check_error();
+        DorderA=atGetOptionalLong(ElemData, "DorderA", 0); check_error();
+        DorderB=atGetOptionalLong(ElemData, "DorderB", 0); check_error();
         FuncDelay=atGetOptionalDouble(ElemData,"FuncDelay", 0); check_error();
         FuncA=atGetOptionalDoubleArray(ElemData,"FuncA"); check_error();
         FuncB=atGetOptionalDoubleArray(ElemData,"FuncB"); check_error();
@@ -365,8 +365,8 @@ void mexFunction(int nlhs, mxArray* plhs[], int nrhs, const mxArray* prhs[])
         Seed=atGetOptionalLong(ElemData, "Seed", 0); check_error();
         NSamplesA=atGetOptionalLong(ElemData, "NSamplesA", 0); check_error();
         NSamplesB=atGetOptionalLong(ElemData, "NSamplesB", 0); check_error();
-        DorderA=atGetOptionalLong(ElemData, "DorderA", 1); check_error();
-        DorderB=atGetOptionalLong(ElemData, "DorderB", 1); check_error();
+        DorderA=atGetOptionalLong(ElemData, "DorderA", 0); check_error();
+        DorderB=atGetOptionalLong(ElemData, "DorderB", 0); check_error();
         FuncDelay=atGetOptionalDouble(ElemData,"FuncDelay", 0); check_error();
         FuncA=atGetOptionalDoubleArray(ElemData,"FuncA"); check_error();
         FuncB=atGetOptionalDoubleArray(ElemData,"FuncB"); check_error();
--- a/atintegrators/VariableThinMPolePass.c
+++ b/atintegrators/VariableThinMPolePass.c
@@ -124,7 +124,7 @@ double get_pol(struct elemab* elem, double* ramps, int mode,
             functot = func[allorders*turnidx_modn]; //order zero
             /* Add derivatives for the sample turnidx mod NSamples */
             tpow = 1;
-            t = t*C0 - funcdelay; // change to ctau coordinates and offset
+            t = t*C0 - funcdelay; // change to ct coordinates and offset
             for (i=1;i<allorders;i++){
               /* indexing is fortran-like. We start with columns.
                  cols are derivative components.
--- a/atintegrators/VariableThinMPolePass.c
+++ b/atintegrators/VariableThinMPolePass.c
@@ -49,7 +49,6 @@ struct elem {
     double* PolynomBstart;
     struct elemab* ElemA;
     struct elemab* ElemB;
-    int Seed;
     int Mode;
     int MaxOrder;
     double* Ramps;
@@ -90,7 +89,7 @@ double get_amp(double amp, double* ramps, double t)
    If the mode is 0 or 2, the polynom depends on the particle
    time delay*/
 double get_pol(struct elemab* elem, double* ramps, int mode,
-    double t, int turn, int seed, int order, int periodic)
+    double t, int turn, int order, int periodic, pcg32_random_t* rng)
 {
     int i, turnidx_modn;
     double ampt, freq, ph, val;
@@ -113,8 +112,7 @@ double get_pol(struct elemab* elem, double* ramps, int mode,
         ampt *= val;
         return ampt;
     case 1:
-        val = atrandn(0.0, 1.0);
-        ampt *= val;
+        ampt *= atrandn_r(rng, 0, 1);
         return ampt;
     case 2:
         if (periodic || turn < elem->NSamples) {
@@ -147,7 +145,8 @@ double get_pol(struct elemab* elem, double* ramps, int mode,
 
 /* This function tracks a particle through a thin element with
    variable PolynomB and PolynomA per turn.*/
-void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, int num_particles)
+void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, int num_particles,
+    pcg32_random_t* rng)
 {
 
     int i; // order of polynom, counter
@@ -161,7 +160,6 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
     int periodic = Elem->Periodic;
     double* pola = Elem->PolynomA;
     double* polb = Elem->PolynomB;
-    int seed = Elem->Seed;
     int mode = Elem->Mode;
     struct elemab* ElemA = Elem->ElemA;
     struct elemab* ElemB = Elem->ElemB;
@@ -185,8 +183,8 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
         for (i = 0; i < maxorder + 1; i++) {
             /* calculate the polynom to apply on all particles */
             /* the particle delay time, tpart, is not used */
-            pola[i] = get_pol(ElemA, ramps, mode, 0, turn, seed, i, periodic);
-            polb[i] = get_pol(ElemB, ramps, mode, 0, turn, seed, i, periodic);
+            pola[i] = get_pol(ElemA, ramps, mode, 0, turn, i, periodic, rng);
+            polb[i] = get_pol(ElemB, ramps, mode, 0, turn, i, periodic, rng);
         };
     };
 
@@ -207,8 +205,8 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
                 tpart = time_in_this_mode + r6[5] / C0;
                  /* calculate the polynom A and B components seen by the particle */
                 for (i = 0; i < maxorder + 1; i++) {
-                    pola[i] = get_pol(ElemA, ramps, mode, tpart, turn, seed, i, periodic);
-                    polb[i] = get_pol(ElemB, ramps, mode, tpart, turn, seed, i, periodic);
+                    pola[i] = get_pol(ElemA, ramps, mode, tpart, turn, i, periodic, rng);
+                    polb[i] = get_pol(ElemB, ramps, mode, tpart, turn, i, periodic, rng);
                 };
             };
             /*  misalignment at entrance  */
@@ -236,7 +234,7 @@ ExportMode struct elem* trackFunction(const atElem* ElemData, struct elem* Elem,
     double* r_in, int num_particles, struct parameters* Param)
 {
     if (!Elem) {
-        int MaxOrder, Mode, Seed, NSamplesA, NSamplesB, DorderA, DorderB, Periodic;
+        int MaxOrder, Mode, NSamplesA, NSamplesB, DorderA, DorderB, Periodic;
         double *R1, *R2, *T1, *T2, *EApertures, *RApertures;
         double *PolynomA, *PolynomB, *AmplitudeA, *AmplitudeB;
         double *Ramps, *FuncA, *FuncB;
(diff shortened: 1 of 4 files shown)


oscarxblanco:

oscarxblanco:
This PR is ready to be review @swhite2401 . `Seed` is removed from the element parameters, and the help message refers to the seed parameter in the tracking options.

swhite2401:
This looks fine for me, just a minor comment the 5th coordinate in AT is `ct` and not `ctau`, it is pprobably better to stick to this convention to avoid confusion

oscarxblanco:
> This looks fine for me, just a minor comment the 5th coordinate in AT is `ct` and not `ctau`, it is pprobably better to stick to this convention to avoid confusion

ok, but it is the 6th component with index 5 in python and C, right ?  There is not 0th component.

swhite2401:
> > This looks fine for me, just a minor comment the 5th coordinate in AT is `ct` and not `ctau`, it is pprobably better to stick to this convention to avoid confusion
> 
> ok, but it is the 6th component with index 5 in python and C, right ? There is not 0th component.

Yes index 5 is what I ment

oscarxblanco:
@swhite2401 , $c\tau$ is now $ct$.

atintegrators/VariableThinMPolePass.c
```diff
@@ -123,21 +175,38 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
...
+
+    /* cycle over all particles */
     for (c = 0; c < num_particles; c++) {
         r6 = r + c * 6;
+        /* check if the particle is alive */
         if (!atIsNaN(r6[0])) {
-            if (mode == 0) {
-                double tpart = t + r6[5] / C0;
+            /* mode 0 and mode 2 take into account the particle delay time */
+            if (mode == 0 || mode == 2) {
```
swhite2401: For mode ==2 `get_pol` is called twice lines 186 and 208

atintegrators/VariableThinMPolePass.c
```diff
@@ -123,21 +175,38 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
...
+            pola[i] = get_pol(ElemA, ramps, mode, 0, turn, i, periodic, rng);
+            polb[i] = get_pol(ElemB, ramps, mode, 0, turn, i, periodic, rng);
         };
     };
 
+    // offset the time when applying the sin function
+    // branchless if
+    //   mode == 0      -> time_in_this_mode = t0 *turn
+    //   any other mode -> 0
+    time_in_this_mode = t0 * turn * (mode == 0);
```
swhite2401: Is this very useful? Can't you just use the PhaseA/B argument to apply this offset?

swhite2401:
@oscarxblanco looking again at you code a question came to my mind for the arbitrary function. What you propose here is an implementation of time dependent table using function derivative.

Would it be simpler to provide a 2D array with amplitude and times and interpolate to the particle ct coordinate? My feeling is that the user interface would be simpler that way and also the table generation.
You may look at the WakeFieldPassMethod as an example, we also have to interpolate the wake to the ct coordinate for this element.

Sorry this comment comes a bit late.... but I thought of it just now.

swhite2401:
> @oscarxblanco looking again at you code a question came to my mind for the arbitrary function. What you propose here is an implementation of time dependent table using function derivative.
> 
> Would it be simpler to provide a 2D array with amplitude and times and interpolate to the particle ct coordinate? My feeling is that the user interface would be simpler that way and also the table generation. You may look at the WakeFieldPassMethod as an example, we also have to interpolate the wake to the ct coordinate for this element.
> 
> Sorry this comment comes a bit late.... but I thought of it just now.

One drawback for this option my be speed...

atintegrators/VariableThinMPolePass.c
oscarxblanco: Fixed, thank you !

atintegrators/VariableThinMPolePass.c
oscarxblanco: I don't understand

oscarxblanco: I have removed this part of the code at the expense of a separate if for mode 0 and mode 2.

swhite2401: > I don't understand

Well a time offset of a sine wave is the same as a change of phase, you can change it through the PhaseA and PhaseB attributes

oscarxblanco:
> > @oscarxblanco looking again at you code a question came to my mind for the arbitrary function. What you propose here is an implementation of time dependent table using function derivative.
> > Would it be simpler to provide a 2D array with amplitude and times and interpolate to the particle ct coordinate? My feeling is that the user interface would be simpler that way and also the table generation. You may look at the WakeFieldPassMethod as an example, we also have to interpolate the wake to the ct coordinate for this element.
> > Sorry this comment comes a bit late.... but I thought of it just now.
> 
> One drawback for this option my be speed...

The issue I see with interpolation is that it only works well for a restricted interval, otherwise, you will need to also implement extrapolation.

This may not be an issue for the WakefieldPass method but there are cases where the particles are debunched and one would like to see the effect of a kick on a beam stretched along more than one bucket, for example in cases when the RF is off one could track the effect of a kicker to bring the beam to a different location, or for a beam dump. In those cases, as long as the derivative represents well the behavior of the element there should be no issues on the valid interval for particle 6th coordinate.

In any case, I think you are right about making it more simple for the user. Another option is to accept the user amplitude vs time, calculate the derivatives during the element creation in python and matlab, and calculate the table with derivatives for the pass method in C.

swhite2401:
> > > @oscarxblanco looking again at you code a question came to my mind for the arbitrary function. What you propose here is an implementation of time dependent table using function derivative.
> > > Would it be simpler to provide a 2D array with amplitude and times and interpolate to the particle ct coordinate? My feeling is that the user interface would be simpler that way and also the table generation. You may look at the WakeFieldPassMethod as an example, we also have to interpolate the wake to the ct coordinate for this element.
> > > Sorry this comment comes a bit late.... but I thought of it just now.
> > 
> > 
> > One drawback for this option my be speed...
> 
> The issue I see with interpolation is that it only works well for a restricted interval, otherwise, you will need to also implement extrapolation.
> 
> This may not be an issue for the WakefieldPass method but there are cases where the particles are debunched and one would like to see the effect of a kick on a beam stretched along more than one bucket, for example in cases when the RF is off one could track the effect of a kicker to bring the beam to a different location, or for a beam dump. In those cases, as long as the derivative represents well the behavior of the element there should be no issues on the valid interval for particle 6th coordinate.
> 
> In any case, I think you are right about making it more simple for the user. Another option is to accept the user amplitude vs time, calculate the derivatives during the element creation in python and matlab, and calculate the table with derivatives for the pass method in C.

Ok that is in fact an alternative, but you may define the time steps in your table as you want, for the wake we have multi-turn/ multibunch and intra-bunch so it is a bit the same.  We nevertheless run into issues with the size of this table when we put to much granularity over long time intervals...your proposal probably solves this.

How will you compute the derivatives? Numerical derivatives can become inaccurate for large time steps...

atintegrators/VariableThinMPolePass.c
```diff
@@ -123,21 +175,41 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
...
+    if (mode == 1) {
         for (i = 0; i < maxorder + 1; i++) {
-            pola[i] = get_pol(ElemA, ramps, mode, t, turn, seed, i, periodic);
-            polb[i] = get_pol(ElemB, ramps, mode, t, turn, seed, i, periodic);
+            /* calculate the polynom to apply on all particles */
+            /* the particle delay time, tpart, is not used */
+            pola[i] = get_pol(ElemA, ramps, mode, 0, turn, i, periodic, rng);
+            polb[i] = get_pol(ElemB, ramps, mode, 0, turn, i, periodic, rng);
         };
     };
```
swhite2401, line 189: Sorry I had misunderstood your code in my previous comment... I gave you misleading review.

In fact you may write here:

```
double t = 0.0;
if mode == 0 {
    t = t0*turn;
}
```

and then in the particle loop have a single if block:

```
if (mode != 1) {
    tpart = t + r6[5] /C0;
    ...
}
```

but this is cosmetic to avoid code duplication...

swhite2401: Which is basically what you had just written differently

oscarxblanco:
> > > > @oscarxblanco looking again at you code a question came to my mind for the arbitrary function. What you propose here is an implementation of time dependent table using function derivative.
> > > > Would it be simpler to provide a 2D array with amplitude and times and interpolate to the particle ct coordinate? My feeling is that the user interface would be simpler that way and also the table generation. You may look at the WakeFieldPassMethod as an example, we also have to interpolate the wake to the ct coordinate for this element.
> > > > Sorry this comment comes a bit late.... but I thought of it just now.
> > > 
> > > 
> > > One drawback for this option my be speed...
> > 
> > 
> > The issue I see with interpolation is that it only works well for a restricted interval, otherwise, you will need to also implement extrapolation.
> > This may not be an issue for the WakefieldPass method but there are cases where the particles are debunched and one would like to see the effect of a kick on a beam stretched along more than one bucket, for example in cases when the RF is off one could track the effect of a kicker to bring the beam to a different location, or for a beam dump. In those cases, as long as the derivative represents well the behavior of the element there should be no issues on the valid interval for particle 6th coordinate.
> > In any case, I think you are right about making it more simple for the user. Another option is to accept the user amplitude vs time, calculate the derivatives during the element creation in python and matlab, and calculate the table with derivatives for the pass method in C.
> 
> Ok that is in fact an alternative, but you may define the time steps in your table as you want, for the wake we have multi-turn/ multibunch and intra-bunch so it is a bit the same. We nevertheless run into issues with the size of this table when we put to much granularity over long time intervals...your proposal probably solves this.
> 
> How will you compute the derivatives? Numerical derivatives can become inaccurate for large time steps...

We could use the gradient.
https://numpy.org/doc/stable/reference/generated/numpy.gradient.html
https://es.mathworks.com/help/matlab/ref/gradient.html

swhite2401:
> We could use the gradient.
> https://numpy.org/doc/stable/reference/generated/numpy.gradient.html
> https://es.mathworks.com/help/matlab/ref/gradient.html

Ok let's try and see how its works!

oscarxblanco:
@swhite2401 , this is an example to create the func from a 2D table and some additional user input.
The blue x are 2D user input
The orange crosses are the approximation with one gradient.
The green dots are the approximation with func, using the gradient derivative and 10 turns.

image

```python
import numpy as np

# gaussian
def gauss(x, mu, sigma):
    return 1/np.sqrt(2*np.pi) * np.exp(0.5 * -(x - mu)**2 / sigma**2)

# input from user
T0 = 268.8/3e8 # revolution time
nturns = 10 # turns for the variable thin multipole
dorder = 1 # number of derivatives

#create dummy curve, it will be input from user
# the input from user is a (2,n) array with time and amplitude
steps = 10
tmax = 10e-6
tmin = 0
tstep = (tmax - tmin) / steps
# finally, the t and amplitude
t = np.linspace(tmin,tmax,steps) # seconds
amplitude = gauss(t, 3e-6, 2e-6)

# calculate table of derivatives
tderiv = np.copy(amplitude)
tderiv = tderiv[np.newaxis,:]
for i in range(dorder):
    tderiv = np.vstack((tderiv, np.gradient(tderiv[i-1,:])))

powderiv = (tderiv.T * tstep ** np.arange(dorder+1)).T
curve_approx = np.sum(powderiv, axis=0)

# interpolate table of derivatives to turns
aturns = np.arange(nturns)*T0
func = np.empty((dorder+1,nturns))
for i in range(dorder+1):
    func[i,:] = np.interp(aturns,t,tderiv[i,:])


powfunc = (func.T * tstep ** np.arange(dorder+1)).T
curve_func = np.sum(powfunc, axis=0)

plt.plot(t,amplitude,'x',label='input')
plt.plot(t,curve_approx, '+',label='approx with gradient')
plt.plot(aturns,curve_func, '.', label="turns from func")
plt.legend()
plt.grid()
plt.xlabel('t [s]')
plt.ylabel('Amplitude [a.u.]')
plt.show(block=False)

print(f'The func shape is {(func.shape)}')
print(f'The output func {func}')
```
Here is the terminal output
```python
The func shape is (2, 10)
The output func [[ 0.1295176   0.23102879  0.32555788  0.37976071  0.37401976  0.303166
   0.19751918  0.11055461  0.05115881  0.01842058]
 [ 0.12588193  0.12128797  0.08883117  0.02698071 -0.04869933 -0.1088012
  -0.11523439 -0.09082908 -0.05766439 -0.02849363]]
```

oscarxblanco:
A sliightly more complex example
image

```python

import numpy as np
import at


# gaussian
def gauss(x, mu, sigma):
    return 1/np.sqrt(2*np.pi) * np.exp(0.5 * -(x - mu)**2 / sigma**2)

# input from user
T0 = 268.8/3e8 # revolution time
nturns = 15 # turns for the variable thin multipole
dorder = 1 # number of derivatives

#create dummy curve, it will be input from user
# the input from user is a (2,n) array with time and amplitude
steps = 100
tmax = 20e-6
tmin = 0
tstep = (tmax - tmin) / steps
# finally, the t and amplitude
t = np.linspace(tmin,tmax,steps) # seconds
amplitude1 = gauss(t, 3e-6, 2e-6)
amplitude2 = gauss(t, 13e-6, 3e-6)
amplitude = amplitude1 + amplitude2

# calculate table of derivatives
tderiv = np.empty((dorder+1,len(amplitude)))
tderiv[0,:] = np.copy(amplitude)
for i in range(dorder):
    tderiv[i+1,:] = np.gradient(tderiv[i,:])

powderiv = (tderiv.T * tstep ** np.arange(dorder+1)).T
curve_approx = np.sum(powderiv, axis=0)

# interpolate table of derivatives to turns
aturns = np.arange(nturns)*T0
func = np.empty((dorder+1,nturns))
for i in range(dorder+1):
    func[i,:] = np.interp(aturns,t,tderiv[i,:])


powfunc = (func.T * tstep ** np.arange(dorder+1)).T
curve_func = np.sum(powfunc, axis=0)

plt.plot(t,amplitude,'x',label='input')
plt.plot(t,curve_approx, '+',label='approx with gradient')
plt.plot(aturns,curve_func, '.', label="turns from func",markersize=24)
plt.legend()
plt.grid()
plt.xlabel('t [s]')
plt.ylabel('Amplitude [a.u.]')
plt.show(block=False)

print(f'The func shape is {(func.shape)}')
print(f'The output func {func}')

```

swhite2401:
@oscarxblanco , this seems to work well! I still think interpolation would be simpler as you do not need to specify the number of derivatives... but if you prefer this fine for me...

One question though, why are you interpolating the table of derivative? If you have a function of `t` you should be able to compute it for any value of `t` exactly right?
If an interpolation is anyway required I do not see much interest in this method w.r.t. simple interpolation...

oscarxblanco:
@swhite2401 , the difference is when would you prefer to interpolate, either at the moment of tracking or at the moment of the element creation.

If the pass method receives the input function with any time and amplitude then it will need to interpolate the function to any time t composed by the revolution period plus the particle delay, everything is done in C. The cost of this choise is the calculation time to interpolate every time the element is tracked and the space to save the non-sampled function inside the element to be able later to sample it on every turn. Another cost is the time to implement a reliable interpolation in C, but his point might be already solved as you mentioned before, so, it is reduced to reuse the required code and test it.

If the input function is sampled and interpolated in python/matlab at the moment of creation of the element, and the n degree polynom is provided to the pass method in C then you only interpolate once. The cost of interpolation time is paid once the element is created, and the cost of saving the element is reduced to the number of samples times the order of the polynom.

The complexity for the user is also different. In the first case the user only needs to provide a 2D array with time and amplitude, and number of turns. In the second case, it needs to provide the previous, plus the order of the derivative and the revolution time because it is is unknown to the element.

I implemented this method because the polynomial approximation on every turn is relatively simple to calculate and allows to handle well many functions including the behavior of particles with small or big delays with respect to the revolution time, and with relatively small cost in tracking time and additional complexity for the user.

swhite2401:
> the revolution time

Ok fair enough. Side comment: the revolution time is accessible in the `struct params`, the user does not need to provide it.

oscarxblanco:
@swhite2401, this is a possible implementation in pyat.
Instead of the function `func`,  the user could provide a curve over a time range in a dictionary with the following entries:
```python
datainput['curve'] = curve # (2,n) array, [0,:] is time, [1,:] is amplitude
datainput['turns'] = nturns
datainput['revolution_time'] = T0
datainput['dorder']= derivative_order
# nturns*T0 < max(datainput['curve'][0,:])


v = at.VariableThinMultipole('v',at.ACMode.ARBITRARY, AmplitudeB=1e-3, inputcurveB=datainput)
```

swhite2401:
> @swhite2401, this is a possible implementation in pyat. Instead of the function `func`, the user could provide a curve over a time range in a dictionary with the following entries:
> 
> ```python
> datainput['curve'] = curve # (2,n) array, [0,:] is time, [1,:] is amplitude
> datainput['turns'] = nturns
> datainput['revolution_time'] = T0
> datainput['dorder']= derivative_order
> # nturns*T0 < max(datainput['curve'][0,:])
> 
> 
> v = at.VariableThinMultipole('v',at.ACMode.ARBITRARY, AmplitudeB=1e-3, inputcurveB=datainput)
> ```

@oscarxblanco , I would stick to keyword arguments for the input arguments, this is much easier to document. You can still use a dictionary to load them with **datainput as long as you keep correct key names.

Then I do not understand why you constraint the length of the input, the idea is to have a buffer with arbitrary duration that is repeated periodically or not depending on the Periodic kwarg. This feature is required to model a noise excitation buffer for example and we need this for ESRF.

With the time buffer you have all the information you need, why introduce nturns and T0?

swhite2401:
I would like to propose the following implementation.

Only three inputs needs:
-1D vector of amplitude (required) same as now
-1D vector of time optional - same length as amplitude vector - if absent same behavior as present: 1 kick per turn for all bunches
-Fit order: derivative, or polynomial fit order to determine f(t)

-> if time vector is present, compute derivative versus time to get amplitude = f(t), could be a polynomial fit by the way

To apply the kick use a modulo(total time) if Periodic =True such that t is always within the boundaries of the initial vector.
If Periodic = False set polynoms to zero when t> tmax (time vector)

oscarxblanco:
> @oscarxblanco , I would stick to keyword arguments for the input arguments, this is much easier to document. You can still use a dictionary to load them with **datainput as long as you keep correct key names.

OK


> Then I do not understand why you constraint the length of the input, the idea is to have a buffer with arbitrary duration that is repeated periodically or not depending on the Periodic kwarg. This feature is required to model a noise excitation buffer for example and we need this for ESRF.

There is no constraint, just  a check to be sure the number of turns to interpolate is smaller than the curve vs time provided. Once the number of samples/turns is defined, they repeat as usual.

oscarxblanco:
> I would like to propose the following implementation.
> 
> Only three inputs needs: -1D vector of amplitude (required) same as now -1D vector of time optional - same length as amplitude vector - if absent same behavior as present: 1 kick per turn for all bunches -Fit order: derivative, or polynomial fit order to determine f(t)
> 
> -> if time vector is present, compute derivative versus time to get amplitude = f(t), could be a polynomial fit by the way
> 
> To apply the kick use a modulo(total time) if Periodic =True such that t is always within the boundaries of the initial vector. If Periodic = False set polynoms to zero when t> tmax (time vector)

@swhite2401 , I think I don't understand these requirements because they seem to me incompatible.

 -1D vector of amplitude (required) same as now : This is OK.
 -1D vector of time optional - same length as amplitude vector: This is the confusing one for me. This means that the fit needs to be done in C in the pass method at the time of tracking because it is the only moment when the period T0 is known.

Therefore, the fit/derivatives in pyat should be removed, correct ?

oscarxblanco:
Oh, I see a big difference. You are asking to ignore the time of the particle !
>  1 kick per turn for all bunches

oscarxblanco:
I had in mind a very different use for this mode. For me the modelling of the pulse needs to include the time of flight of every particle because having the same kick for all particles does not model well the injection of a long bunch.

oscarxblanco:
In other words, I wanted the mode ARBITRARY to model a different function per turn to arbitrary order in ct, and the simplest way to implement it is to include an arbitrary number of derivatives. In this way the effect of the time of flight of the particle can be studied with the only limitation being the representation of a function as a polynomial in every turn.

What you are asking is different. You are saying that the time of flight should be ignored. The function should be a kick per turn, the same for all particles. In that case there is no need of derivatives. And it is enough to interpolate or fit. The only novelty here is to provide a general curve to fit, instead of the rigid sampling per turn. The dynamics of particles with different time of flight is not visible.

oscarxblanco:
@swhite2401 , I will go back one step in this review.

The function `func` in the master is of shape(1,m) with m the number of samples/turns.
In this PR the function `func` is modified to be of shape (n,m) with n the order of the function in the coordinate ct, and m the number of turns.

Is this limiting for your request on the alternative function input with curves and time ?
If it is I will close this PR, and if it is not I will propose to merge up to the point just before the modifications to accept a 2D array with time and amplitude, and any other parameter if necessary.

swhite2401:
> There is no constraint, just a check to be sure the number of turns to interpolate is smaller than the curve vs time provided. Once the number of samples/turns is defined, they repeat as usual.

But why the signal needs to be a integer number of turns, I can be of arbitrary duration and repeat over time without necessarily being synchronized with the revolution time.

Also once you have f(t), defined by you derivatives of order n for why do you need to interpolate? You just need to compute f(t)....

swhite2401:
> @swhite2401 , I will go back one step in this review.
> 
> The function `func` in the master is of shape(1,m) with m the number of samples/turns. In this PR the function `func` is modified to be of shape (n,m) with n the order of the function in the coordinate ct, and m the number of turns.
> 
> Is this limiting for your request on the alternative function input with curves and time ? If it is I will close this PR, and if it is not I will propose to merge up to the point just before the modifications to accept a 2D array with time and amplitude, and any other parameter if necessary.

Haha, well it looks like we don't understand each other!! How about we try to discuss this quickly over zoom?

oscarxblanco:
@swhite2401 , I'll be out of the office until Thursday afternoon. After that I am available at any time.

PR #1155, opened 2026-09-09
Move binary search and table interpolation

oscarxblanco:
This PR creates a file called `atsearch.c` which contains the implementation of the binarySearch function. The duplicated code in `ImpedanceTablePass.c` and `atimplib.c` is removed.


oscarxblanco:

atintegrators/atimplib.c
```diff
@@ -9,25 +10,6 @@
...
-        lower = pivot;
-        nStep+=1;
-        return binarySearch(array,value,upper,lower,nStep);
-    }else{
-        return pivot;
-    };
-};
-
-
 static double getTableWake(double *waketable,double *waketableT,double distance,int index){
```
swhite2401: Could you move this one as well and call it interpolTable for example>

oscarxblanco: Sure, there is a file called `interpolate.c`, I will put it there.

atintegrators/atimplib.c
```diff
@@ -1,5 +1,7 @@
 #include "atconstants.h"
 #include "atelem.c"
+#include "atsearch.c"
+#include "interpolate.c"
```
swhite2401, line 3: I would combine these 2 in a single C library, we may add other things in the future to it.
We do not need to follow the matlab style of 1 function -> 1 file

oscarxblanco: binarySearch has been moved to interpolate.c

atintegrators/ImpedanceTablePass.c
swhite2401: replace by interpolTable

swhite2401: This has to be removed as well it is interpolTable

swhite2401 (approved):

PR #1156, opened 2026-09-15
set binarySearch max number of steps to 200

oscarxblanco:
This Pull Request limits to 200 the maximum number of times the recurrent calls to binarySearch.


swhite2401:
@oscarxblanco , is there any reason to do this? I am afraid that for very large tables (for impedance we can have thousands of points) this will stop before finding the correct interval without warning and return wrong result.

This has been used and worked without any issues for years so I would leave it as it is to avoid unwanted behavior. If really needed this protection has to adapt to the table size and not be hard-coded to an arbitrary value.

oscarxblanco:
- The algorithm divides in two the array in every step, so, it will work with arrays as long as 2^200=1.6069e+60 elements.
- At the same time it prevents to iterate more than 200 times over an array where this algorithm does not converge, for example, if the array is not sorted it is not certain if binarySearch will end.
- Also, 200 hundred is small enough to be represented in 8 bits. 

The number 200 is just arbitrary, it could be any number.

swhite2401:
Ok that is fine then, maybe you could print a warning or issue an error if the limit is reached?
Even thought there is very little chance it happens....

swhite2401:
Or return NaN?

oscarxblanco:
@swhite2401 , I have included an atWarning.

About returning NAN it seems complicated because the function is supposed to return a positive integer to be used as index, and the behaviour of NAN cast to an int is not well defined. In the number of steps is exceeded it nows returns -1 which is an invalid index value in C.

oscarxblanco:
If I tried to include the atWarning inside the binarySearch function but the MACOS compilation complains because the return type for check_error is different, binarySearch returns an int, and check_error a NULL.

```terminal
       atintegrators/atelem.c:195:52: note: expanded from macro 'check_error'
        195 | #define check_error() if (PyErr_Occurred()) return NUL
```

Although this is only for MACOS and the others OSs have no issue in compiling it, I prefer to remove the atWarning because the only way I know to solve it is to return NULL in binarySearch and pass the index by reference to be modified. I could try it but changes the logic on how to use it which I prefer to avoid.

So, I have a question, @swhite2401 . Would it be OK for you to return -1, which is an invalid index, if the number of steps is exceeded ? It means that there will be no warning, just an invalid index. Otherwise I will close this PR and leave it without a maximum number of steps as you said from the beginning.

swhite2401:
Yes this is fine for me, additionally you may just print a message like "BinarySearch could not converge, -1 returned" which would help the user diagnose the problem, otherwise they will just get an obscure C error

atintegrators/interpolate.c
```diff
@@ -17,7 +17,7 @@
 int binarySearch(double *array,double value,int upper,int lower,int nStep){
     int pivot = (int)(lower+upper)/2;
     if (nStep > 200){
-       printf("Warning: binarySearch could not converge, -1 returned.");
+       printf("Warning: binarySearch could not converge, -1 returned.\n");
```
lfarv: `printf` will not appear in Matlab. You should use `atPrintf`

lfarv: By the way, shouldn't the returned index be tested, to generate an error rather than a crash, or even worse. impredictible behaviour?

oscarxblanco: @lfarv , do you mean to test it outside `binarySearch`, after every call to `binarySearch`, right ?

lfarv:
> binarySearch returns an int, and check_error a NULL.

`check_error` must be used only in the top-level function (`struct elem *trackFunction`). Never in sub-functions. Errors in sub-functions set the error code (`atError`) and return up to the top-level where check_error returns NULL to python.

atintegrators/interpolate.c
lfarv: Ideally yes, otherwise reading for an address outside the table will return garbage with unpredictable consequences…

oscarxblanco: Ok, I will do it.

oscarxblanco:
> > binarySearch returns an int, and check_error a NULL.
> 
> `check_error` must be used only in the top-level function (`struct elem *trackFunction`). Never in sub-functions. Errors in sub-functions set the error code (`atError`) and return up to the top-level where check_error returns NULL to python.

@lfarv, now that you have cleared it up , do you prefer that I change the atPrintf for an atWarning ?

lfarv:
`atWarning` prints a warning but goes on. If there is a safe way to continue, that's OK.
Otherwise, if you want to stop and raise an exception, you need to call `atError`.

In Matlab, `atError` never returns. It goes directly back to Matlab while taking care of freeing memory etc. That's easy
In python, atError sets a python error code but returns to the caller. It's up to you to return safely to the top-level function and call `check_error`. Error handling in python is more tricky…

oscarxblanco:
> `atWarning` prints a warning but goes on. If there is a safe way to continue, that's OK. Otherwise, if you want to stop and raise an exception, you need to call `atError`.
> 
> In Matlab, `atError` never returns. It goes directly back to Matlab while taking care of freeing memory etc. That's easy In python, atError sets a python error code but returns to the caller. It's up to you to return safely to the top-level function and call `check_error`. Error handling in python is more tricky…

OK, I think `atWarning` is enough. One could have a case where not finding the right index in binarySearch is not a big issue. And also the index checking is already implemented in the previous commit for the cases when it is necessary.

lfarv:
`InterpolTable(..., -1)` may return unpredictable values, let's hope it never happens… But it's true that a correct error checking may be a lot of work…

oscarxblanco:
@lfarv , in the last commit I am checking that index is larger than zero before the `InterpolTable` call
image

lfarv:
This is not enough. As I said above, in python `atError` returns, so the computation will go on normally with the negative index.

In atimplib, instead of calling `atError`, which has already been called in `binarySearch`, you should immediately return a flag, test in in the calling function and immediately return it, and so on until you reach the top level, where `check_error` returns the control to python.

If the error is deeply nested in a long chain of calls, this is awfully complex. For this PR, I suggest that you give up…

A possible solution, not tested yet, is to use setjmp/longjmp in <stdjmp.h>. The code would look like that:

In interpolate.c:
```C
#include <stdjmp.h>

int binarySearch(..) {
    ...
    if (nStep > 200){
        /* set the error code for python */
       atError("Warning: binarySearch could not converge, -1 returned.\n");
        /* return directly to setjmp */
       longjmp(env, -1);
    };
    ...
}
```
In VariableThinMPolePass.c:
```C
#include <stdjmp.h>
static jmp_buf env;
#include "interpolate.c"

ExportMode struct elem* trackFunction(...) {


    ...
    int turn = Param->nturn;
    /* setup jmp_buf and return 0 */
    if (setjmp(buf)) {
        /* error return from longjmp*/
        return NULL;  /* indicates an error to python */
    }
    else {
        /* normal behaviour */
        VariableThinMPolePass(r_in, Elem, t0, turn, num_particles, Param->common_rng);
        return Elem;
    }
}
```
This is also delicate: calling `longjmp` if `setjmp` has not been called previously will crash since `env` has not been set. Any memory allocation will not be released...

I think this is worth dedicated tests.

oscarxblanco:
> If the error is deeply nested in a long chain of calls, this is awfully complex. For this PR, I suggest that you give up…

Alright, I give up. In any case,  thank you for all the explanations @lfarv .

lfarv:
Sorry @oscarxblanco : I meant give up with error handling, not with the PR ! The worse which may happen is a crash, but the warning should give an explanation…

oscarxblanco:
@lfarv, I had a look on how to use stdjmp and seemed straight forward if everything stays inside C.

I don't know how this will behave when being call from python and matlab, but I could give it a try ! Do you want to keep it in this PR ?

I have a comment apart from the error handling. The name `binarySearch` doesn't match exactly what the function is doing. At first sight I was tempted to change it by `bsearch` from the standard C library, but in fact, this is returning a value that I would call `find_closest_index` which implements a bipartition as in binary search. @swhite2401 and @lfarv , Would it be OK to change the name of the function `binarySearch` to `find_closest_index`  or similar ?

lfarv:
> I don't know how this will behave when being call from python and matlab, but I could give it a try ! Do you want to keep it in this PR ?

For Matlab, this does not apply: `atError` never returns, it gives immediately the control back to Matlab. That's much easier, and Matlab takes care of releasing any allocated memory, provided it has been allocated with `atMalloc` (not `malloc`). In that case, the `longjmp` which should immediately follow `atError` is not called.

For python, everything remains inside the C code, as you said, and the control is given back to python with the normal return statement. So I don't expect any problem.

Concerning this PR, I see 2 difficulties (no luck…)

- the `longjmp` (equivalent of throwing an exception) happens where the error occurs, that is in `binarySearch` (ideally). In the case of `VariableThinMPolePass`, this is called in the body of the `VariableThinMPolePass` function, after `pola/b` have been allocated. So they will never be released. This concern a small memory (a few bytes), and for an extremely rare event. So we can probably forget about it, though it could be mentioned in a comment.
- `binarySearch` is in the `interpolate.c` file, included in many integrators `IdTablePass`, `ImpedanceTablePass`, `VariableThinMPolePass`, `BeamLoadingCavityPass`, `BeamMomentsPass`, `SliceMomentsPass`. `WakeFieldPass`. All must be modified to add the `setjmp` (equivalent of try…catch…). Or one could take a single one  as a proof of concept (VariableThinMPolePass for instance) and throw the error on the return of `binarySearch` instead of within it. Other integrators are not affected.
- in my sample code above, I use a static variable to make `env` accessible from any place. This is not thread-safe: it's a single variable for all threads. To avoid that, `env` should be transmitted with function arguments to all places where a `longjmp` happens: this is a big reorganisation. Thread-safety is not an urgent problem: multi-threaded python is now possible since python 3.14 and would be more efficient than multi-processing presently used. But the main C-file `at.c` is far from being thread-safe, that's large amount of work, if ever possible… My recommendation is to have at least the integrators thread-safe from now on, to prepare the future, but that's not critical. For the moment, we could accept a static variable.

With that, you can consider this PR as a good or bad opportunity to test error handling. Otherwise, building a dummy integrator where you could control the appearance of errors is another possibility.

lfarv:
> Would it be OK to change the name of the function binarySearch to find_closest_index or similar ?

Nothing against the change, but it's some work… Up to you

PR #1157, opened 2026-09-15
Variablethinmultipole INTERPOLATION_TABLE

oscarxblanco:
This PR implements the INTERPOLATION_TABLE mode for the Variable Thin Multipole in AT MATLAB and pyat.

A table with two rows must be defined, the first row is time in seconds, and the second row is the function values. The element interpolates linearly the input table to calculate the function value using the particle time delay.

Here is an example in python of a triangular function that with a period of 2.2 seconds.
```python
>>> v = at.VariableThinMultipole('v',at.ACMode.INTERPOLATION_TABLE, AmplitudeA=1e-3,FuncA=np.array([[0,1.1,2.2],[1,0,1]]))
VariableThinMultipole:
       FamName: v
        Length: 0.0
    AmplitudeA: [0.001]
 FinterpolateA: [1. 0. 1.]
         FuncA: [[0.  1.1 2.2]
 [1.  0.  1. ]]
      MaxOrder: 0
          Mode: 3
      ModeName: INTERPOLATE
     NSamplesA: 3
      Periodic: True
      PolynomA: [0.]
      PolynomB: [0.]
 TinterpolateA: [0.  1.1 2.2]

```
Tracking a particle with a time offset of 0.5 seconds gives the following
```python
v.track(np.array([0,0,0,0,0,0.5*scipy.constants.speed_of_light],dtype=np.float64))
array([0.00000000e+00, 0.00000000e+00, 0.00000000e+00, 5.45454545e-04,
       0.00000000e+00, 1.49896229e+08])
```

And here is the equivalent in matlab
```matlab
v = atvariablethinmultipole('v','INTERPOLATION_TABLE','AmplitudeA',1e-3,...
    'FuncA',[0 1.1 2.2; 1 0 1]);
v = 
  struct with fields:
          FamName: 'v'
           Length: 0
             Mode: 3
         ModeName: 'INTERPOLATION_TABLE'
         PolynomA: 0
         PolynomB: 0
       AmplitudeA: 1.0000e-03
            FuncA: [2×3 double]
         MaxOrder: 0
    TinterpolateA: [0 1.1000 2.2000]
    FinterpolateA: [1 0 1]
        NSamplesA: 3
```
And the result of tracking 
```matlab
>> elempass(v,[zeros(5,1); 0.5*physconst('LightSpeed')])'
ans =
   1.0e+08 *
  Columns 1 through 5
                   0                   0                   0   0.000000000005455                   0
  Column 6
   1.498962290000000
```


oscarxblanco:
@swhite2401 or @lfarv , could you please review this PR ?

pyat/at/lattice/elements/variable_elements.py
```diff
@@ -36,8 +38,12 @@ class VariableThinMultipole(Element):
         Sinmax=float,
         NSamplesA=int,
         NSamplesB=int,
-        FuncA=_array,
-        FuncB=_array,
+        FuncA=_anyarray,
+        FuncB=_anyarray,
```
lfarv, line 42: What is the reason not to use the standard `_array` ?

oscarxblanco: Func will be of 2 dimensions and shape (2,n). `_array` reshapes it to be of 1 dimension.

lfarv: Then it should be `FuncA=lambda v: _array(v, (2, -1)),` to ensure the two lines.

lfarv:
I see a few general problems in this integrator:

- memory allocation: the memory allocated to `PolynomAstart` is never released. And it cannot be… Practically, it means that each new tracking (unless `keep_lattice` is True) will create a new element, with a new allocation. The `struct elem` itself is released, but any pointer to allocated memory that it contains cannot be. Solving that needs a modification of the interface of integrators. Or allocating memory by extending the 'struct elem'.
- multithreading: this integrator is not compatible with the upcoming free-thread python (from 3.14 on), nor with present multi-processing (`use_mp=True`) in case the startup mode is "fork", which was the default on Linux up to 3.13. In both those cases, the lattice structure is shared between all processes or threads. So several threads may modify the common Elem->PolynomA/B simultaneously without any synchronisation.

oscarxblanco:
@lfarv  Thank you, I was not aware of these two problems.

About the PolynomAstart and PolynomBstart, they were recently created to store the initial value of the polynoms, and both PolinomAstart and PolynomBstart could be eliminated as long as we all accept for example to set PolynomA and PolynomB to zeros at the end of the tracking function. Would that be acceptable ?

About the second point on multithreading, I don't have any immediate solution or idea.

lfarv:
> About the second point on multithreading, I don't have any immediate solution or idea.

A simple rule is to never modify the contents of the `struct elem` structure, but it's not always possible.

In this particular case, is not it possible to have `pola` and `polb` local to the `void VariableThinMPolePass()` function?
`get_pol()` would allocate the space for the polynomial, fill it and return a pointer to it. `VariableThinMPolePass()` would `free(pola)` and `free(polb)` before returning.

Something like:
```C
double *get_pol(...) {
    double *polynom = atMalloc(..);
    // fill polynom
    return polynom;
}

void VariableThinMPolePass() {
    double *pola;
    ...
    pola = get_pol(...)
    ...
    atFree(pola);
}
```

I am not familiar with the code, I may miss something…

oscarxblanco:
@lfarv , it seems possible. I don't know how the element would behave if it does not have PolynomA and PolynomB and one for example changes the pass method to IdentityPass, but I could try.

lfarv:
Another point: `ElemA` and `ElemB` are simple structures, they do not need dynamic allocation. They can be statically defined in `struct Elem`.

Switching the PassMethod to IdentityPass is not a problem: `VariableThinMPolePass()` is not called, no allocation, no problem…

oscarxblanco:
@lfarv , I have done the pola and polb allocation as you said would be OK for multithreading. And also, as a result, PolynomA and PolynomB are never modified so they don't need to be restored and the PolynomAstart and PolynomB start are removed.

The modification to allocation statically ElemA and ElemB is also implemented in a separate commit, the last one.

Is there anything else you would prefer to change ?

atintegrators/VariableThinMPolePass.c
```diff
@@ -167,10 +202,14 @@ ExportMode struct elem* trackFunction(const atElem* ElemData, struct elem* Elem,
         double *R1, *R2, *T1, *T2, *EApertures, *RApertures;
         double *PolynomA, *PolynomB, *AmplitudeA, *AmplitudeB;
         double *Ramps, *FuncA, *FuncB;
+        double *FinterpolateA, *FinterpolateB;
+        double *TinterpolateA, *TinterpolateB;
         double FrequencyA, FrequencyB;
         double PhaseA, PhaseB;
         double Sinmin, Sinmax;
-        struct elemab *ElemA, *ElemB;
+        Elem = (struct elem*)atMalloc(sizeof(struct elem));
```
lfarv: Elem should be allocated after all the arguments are processed (where it was before). Otherwise if check_error returns, the Elem structure will stay allocated.

lfarv:
I have two questions, but I may misunderstand something.

First: I do not see `PolynomA` and `PolynomB` referred to anywhere. Are they used at all ? 

Second: it seems that for all modes (except RANDOM, I come to that later), the whole amplitude vector (`AmplitudeA/B`) is multiplied be a single scalar factor. In other words `get_pol` computes the same scaling factor (the `val` variable) independently of `order`. For instance, the binary search will be called for each order, while it does not depend on it and returns the same value. Would it be possible to take `get_pol` out of the `for (i = 0; i < maxorder + 1; i++)` loop and multiply the whole amplitudeA/B by its result ? For RANDOM, its another question: at the moment, each order is scaled by a different value so `get_pol` in that case would return an array of scaling factors. But the other option (single random scaling for the whole amplitude vector) also make sense: for a magnet ripple, all multipoles vary together.

I understand that in most cases, the Amplitude vector may have size 1 (dipole only), and the gain would be small. But for a higher order multipole, even orders with 0 amplitude will make the full scaling computation, which is really useless.

Maybe I'm completely wrong, if so, forget all that…

oscarxblanco:
> First: I do not see PolynomA and PolynomB referred to anywhere. Are they used at all ?

Not anymore. They used to be modified in every turn but now the are not used, instead pola and polb are allocated during the tracking. The initial value of the PolynomA and PolynomB is not changed, nor used.

oscarxblanco:
@lfarv , should I remove PolynomA and PolynomB ?

oscarxblanco:
> Second: it seems that for all modes (except RANDOM, I come to that later), the whole amplitude vector (AmplitudeA/B) is multiplied be a single scalar factor. In other words get_pol computes the same scaling factor (the val variable) independently of order. For instance, the binary search will be called for each order, while it does not depend on it and returns the same value. Would it be possible to take get_pol out of the for (i = 0; i < maxorder + 1; i++) loop and multiply the whole amplitudeA/B by its result ? For RANDOM, its another question: at the moment, each order is scaled by a different value so get_pol in that case would return an array of scaling factors. But the other option (single random scaling for the whole amplitude vector) also make sense: for a magnet ripple, all multipoles vary together.

>I understand that in most cases, the Amplitude vector may have size 1 (dipole only), and the gain would be small. But for a higher order multipole, even orders with 0 amplitude will make the full scaling computation, which is really useless.

you are right, I had a look and it seems possible to move `get_pol` outside the for loop. I have to implement a way to check that Amplitude exist, `get_pol` was checking it at the start and returning zero if it was not definedd.  Now I would have to check it somewhere else...

atintegrators/VariableThinMPolePass.c
```diff
@@ -121,23 +139,47 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
...
-            polb[i] = get_pol(ElemB, ramps, mode, t, turn, i, periodic, rng);
+    // create thread safe polynoms. Each thread will allocate and free pola and polb
+    double *pola = (double *) atMalloc((maxorder+1)*sizeof(double));
+    double *polb = (double *) atMalloc((maxorder+1)*sizeof(double));
+
+    // initialize pola and polb
+    for (i = 0; i < maxorder + 1; i++){
+      pola[i] = 0;
+      polb[i] = 0;
+    };
```
lfarv: Alternately, you could use
```C
double *pola = (double *) atCalloc(maxorder+1, sizeof(double));
```
which initialises the allocated space with zero.

lfarv:
That looks nice. Finally, you could get rid of `PolynomA` and `PolynomB`. and restore `conversion.py` to its initial state.

Final question: do you have a way to check that results in the different modes are still correct after all these modifications ?

atintegrators/VariableThinMPolePass.c
```diff
@@ -121,23 +139,47 @@ void VariableThinMPolePass(double* r, struct elem* Elem, double t0, int turn, in
...
     for (c = 0; c < num_particles; c++) {
         r6 = r + c * 6;
         if (!atIsNaN(r6[0])) {
-            if (mode == 0) {
-                double tpart = t + r6[5] / C0;
-                for (i = 0; i < maxorder + 1; i++) {
-                    pola[i] = get_pol(ElemA, ramps, mode, tpart, turn, i, periodic, rng);
-                    polb[i] = get_pol(ElemB, ramps, mode, tpart, turn, i, periodic, rng);
+            if (mode != 1){
+                tpart = t*(mode == 0) + t*(mode == 3) + r6[5] / C0;
```
lfarv: The tests are useless : `tpart = t + r6[5] / C0;`

oscarxblanco:
> Final question: do you have a way to check that results in the different modes are still correct after all these modifications ?

Yes, but I haven't done it for the last modifications. 

As I was modifying parts of the element memory handling in C I decided to run in a terminal the default element with few parameters and tracking something to get a known result. I did it to be sure in short time that it compiled without issues, that tracking did not crash and that the modifications where not obviously wrong.

However, I wrote several tests to create, save, load, and track the different modes that I could reuse now that the C part is modified. It could be worth to put some of them in the list done by pytest, so, I'll do it now.

atintegrators/VariableThinMPolePass.c
oscarxblanco: But, for mode 2 it should be `tpart = r6[5] / C0`, while for mode 3 and mode 0 `tpart = t + r6[5] / C0`.
And for mode 1 time is ignored.

atintegrators/VariableThinMPolePass.c
```diff
@@ -56,38 +56,54 @@ double get_amp(double amp, double* ramps, double t)
...
             return 0.0;
         }
+    case 3:
+       titp = elem->Tinterpolate;
+       fitp = elem->Finterpolate;
+       if (periodic){
+         while (t < titp[0]){t = t+titp[nsamples-1];};
+         t = fmod(t, titp[nsamples-1]);
+       };
+       idx = binarySearch(titp , t, nsamples, 0, 0);
```
swhite2401: Shouldn't this be in the else statement below where idx is needed?
If the value is outside the table don't you want to return 0?

atmat/lattice/element_creation/atvariablethinmultipole.m
```diff
@@ -9,7 +9,7 @@
 %    FNAME          Family name
 %
 %  OPTIONS (order does not matter)
-%    MODENAME       'SINE', 'WHITENOISE' or 'ARBITRARY'.
+%    MODENAME       'SINE', 'WHITENOISE', 'ARBITRARY' or 'INTERPOLATION_TABLE'
```
swhite2401, line 12: Is arbitrary still the correct name? Would POLYNOMIAL_TABLE or DERIVATIVE_TABLE be more descriptive?

atintegrators/VariableThinMPolePass.c
lfarv: Apparently, for mode 2 `tpart` is ignored

pyat/at/lattice/elements/variable_elements.py
```diff
@@ -174,7 +188,23 @@ def _set_sine(self, ab, **kwargs):
 
     def _set_arb(self, ab, **kwargs):
         func = kwargs.pop("Func" + ab, None)
-        nsamp = len(func)
         assert func is not None, "Please provide a value for Func" + ab
+        nsamp = len(func)
         setattr(self, "Func" + ab, func)
         setattr(self, "NSamples" + ab, nsamp)
+
+    def _set_interpolate(self, ab, **kwargs):
```
swhite2401, line 206: this is nice but you have to sort all array not just time. It does not look like you are doing it now.
You can easily do it with the 2D array interpolate.

You should probably check that t values are unique as well

atmat/lattice/element_creation/atvariablethinmultipole.m
lfarv: ARBITRARY is a turn-by-turn definition, right ? I do not understand your proposals. Can you explain ?

atintegrators/VariableThinMPolePass.c
oscarxblanco: Yes, it also works and it is better to keep it close `interpolTable`, thank you.

The reason why it also works here is because `bynarySearch` always returns an index, and it returns the first or the last index in the case when `t` is outside the range. Therefore, it is safe to use anywhere, one should be careful just when to use it for interpolation.

atmat/lattice/element_creation/atvariablethinmultipole.m
swhite2401: Yes it is turn by turn but @oscarxblanco introduces turn by turn derivatives or polynomials what ever you would like to call it to apply the kick for an t value, not just turn multiples, I proposed to implement the interpolation as a gneralization to this for multi-turn, multi-bunch , intra-bunch.

I was just thinking that arbitrary could apply to both INTERPOLATION and TBT so I was just trying to propose something more descriptive

atintegrators/VariableThinMPolePass.c
oscarxblanco: Sorry, I still don't see it. From these two lines:
```C
if (mode != 1){
                tpart = t*(mode == 0) + t*(mode == 3) + r6[5] / C0;
```
mode 0 will enter the if, and `tpart = t*(1) + t*(0) + r6[5] / C0`
mode 1 won't enter the if.
mode 2 will enter the if, and `tpart = t*(0) + t*(0) + r6[5] / C0`
mode 3 will enter the if, and `tpart = t*(0) + t*(1) + r6[5] / C0`

Do you agree ?

atmat/lattice/element_creation/atvariablethinmultipole.m
oscarxblanco: Sorry @lfarv , this comment comes from an off-line communication.
There are two different implementations for the same result. The first is interpolating which is the one is implemented in this PR.

But we also discuss extending the table used in ARBITRARY mode to contain the derivatives of Func, therefore, any function that can be expand to a Taylor series at every turn would be covered. @swhite2401 refers to that second one which he proposes to call DERIVATIVE_TABLE and it is not in this PR.

swhite2401: Ah sorry for the confusion, I was not paying attention you removed it from this PR, forget about my comment then!

pyat/at/lattice/elements/variable_elements.py
oscarxblanco: @swhite2401 , if I understand well your first phrase you are saying that `Finterpolate` should be also sorted when sorting the `Tinterpolate`, right ? if so, yes, I'll do it.

About your second phrase, yes, you are right, I'll do it.

swhite2401: Well yes, I assume that if the user is providing 1D vectors they are associated by index. That would be the logical thing to do at least

atmat/lattice/element_creation/atvariablethinmultipole.m
oscarxblanco: @lfarv  and @swhite2401 , DERIVATIVE_TABLE is not in this PR. Could we discuss it after finishing this one ?

atintegrators/VariableThinMPolePass.c
lfarv: If you write `tpart = t + r6[5] / C0;`, you save 2 tests and 2 multiplications. And
- for mode 0 and more 3 it's correct,
- for mode 1 it does not enter the test (and it's ignored anyway),
- for mode 2 it's ignored, so it's ok

atmat/lattice/element_creation/atvariablethinmultipole.m
lfarv: OK, up to you…

oscarxblanco: It will the next PR.

atintegrators/VariableThinMPolePass.c
oscarxblanco: Understood. I'll do it

atmat/lattice/element_creation/atvariablethinmultipole.m
swhite2401: Yes sorry, my fault

oscarxblanco:
@lfarv  and @swhite, `FuncA` and `FuncB` is 1D in ARBITRARY mode, while in INTERPOLATION_TABLE it should be 2D.

Should I go back to accept an ndarray for `FuncA` and `FuncB`? Or would you prefer that I change the name of the input table for the mode INTERPOLATION_TABLE, for example I could call it `TableA` and `TableB`, and change the element in C.

lfarv:
> Should I go back to accept an ndarray for FuncA and Func...

For me it's ok as it is, provided the documentation clearly describes the input format for each mode.

lfarv:
@oscarxblanco: I think that there is some misunderstanding between the `Func` argument and the `Element.Func` attribute. The `Func` argument may be either a 1-D array for ARBITRARY or a 2-D array for INTERPOLATION_TABLE.

But the `Element.Func` attribute is used only for ARBITRARY, and it must be a 1-D ndarray. There are 2 other attributes are used for INTERPOLATION_TABLE: `TInterpolate` and `Finterpolate` (also 1-D).

So I think that in `_conversions` you can set `FuncA=_array `and restore `conversions.py` to its initial state.

pyat/at/lattice/elements/variable_elements.py
lfarv, line 157: Is there any use for PolynomA/B ?

oscarxblanco:
> @oscarxblanco: I think that there is some misunderstanding between the `Func` argument and the `Element.Func` attribute. The `Func` argument may be either a 1-D array for ARBITRARY or a 2-D array for INTERPOLATION_TABLE.
> 
> But the `Element.Func` attribute is used only for ARBITRARY, and it must be a 1-D ndarray. There are 2 other attributes are used for INTERPOLATION_TABLE: `TInterpolate` and `Finterpolate` (also 1-D).
> 
> So I think that in `_conversions` you can set `FuncA=_array `and restore `conversions.py` to its initial state.

Sorry, I don't understand. 
In INTERPOLATION_TABLE if I set `FuncA=_array` the shape will change from (2,n) to (2n,) and I would not be able  to tell apart time and function arrays.

It is true that INTERPOLATION_TABLE is not using `Element.Func`, and this was done to avoid creating an extra keyword for this element, would you prefer to have another name ? Maybe it is simpler if I ask for `Ttable` and `Ftable` ? In tha case the user would input separatedly time and values, each of shape (n,) and there will be no modification on `Func` conversion. I could also change `Tinterpolation` by `Ttable` and `Finterpolation` by `Ftable` all along the code.

pyat/at/lattice/elements/variable_elements.py
```diff
@@ -174,7 +194,27 @@ def _set_sine(self, ab, **kwargs):
...
+        interpolate = kwargs.get("Func" + ab)
+        assert np.ndim(interpolate) == 2, "Func" + ab + " should be of 2 dimensions."
+        _, nsamp = np.shape(interpolate)
+        assert nsamp >= 2, "Func" + ab + " requires at least two points to interpolate."
+        assert ~np.any(np.isnan(interpolate)), "Function has nan values."
+        assert ~np.any(np.isinf(interpolate)), "Function has inf values."
+        tsort = interpolate[0, :]
+        assert len(tsort) == len(np.unique(tsort)), "Time array has repeated elements."
```
lfarv: As a general rule, it is not recommended to use `assert` for validation, it should be reserved for debugging (see [here](https://docs.python.org/3.14/reference/simple_stmts.html#the-assert-statement)).

If python is compiled with optimisation (which I have never seen), the `assert` statement is ignored.

oscarxblanco: Sometime ago it was signaled by flake8, but, it didn't complain this time. I may have changed something on my installatio.

I will change it.

pyat/at/lattice/elements/variable_elements.py
oscarxblanco: No, except if one changes the Pass Method, for example to `IdentityPass`.

lfarv:
>in INTERPOLATION_TABLE if I set FuncA=_array the shape will change from (2,n) to (2n,) and I would not be able to tell apart time and function arrays.

In `_set_interpolate` you never call `setattr("Func", ...)`. The `Func` attribute is only set in `set_arb`.

The conversion function is called inside `setattr`

oscarxblanco:
> Final question: do you have a way to check that results in the different modes are still correct after all these modifications ?

@lfarv ,I have included some numerical test in python for each one of the modes of the variable thin multipole and also a save and load test with the default mode.

Also, I have prepared 4 simple scripts to test the saving and loading from python/MATLAB to MATLAB/python. It all works now. I can upload to the comments in this PR a zip file with the 4 files, a README and the output ring in .m and .mat formats. I however did not try to implement an automatic test in MATLAB, I hope it is OK.

oscarxblanco:
> In _set_interpolate you never call setattr("Func", ...). The Func attribute is only set in set_arb.
> The conversion function is called inside setattr

I might be doing something wrong, but I still  get an error because the shape changes when loading a .m or .mat file.
Here is my modification to the last commit
image
Here is the code to test save and load in python
```python
"""This script saves and load an element in pyat."""
# test_save_load.py
# oblanco 2026sep18

import numpy as np, at

print("Create element")
#v = at.VariableThinMultipole('v', at.ACMode.SINE, AmplitudeA=1e-3, AmplitudeB=2e-3, FrequencyA=10, FrequencyB=100, PhaseA=0, PhaseB=2*np.pi/4)¶
#v = at.VariableThinMultipole('v', at.ACMode.WHITENOISE, AmplitudeB=[0,1e-3])¶
#v = at.VariableThinMultipole('v', at.ACMode.ARBITRARY, AmplitudeB=[0,0,1e-3],FuncB=[1,0,1,0,1])¶
t = np.arange(1000)
fa = np.zeros((2,1000))
fa[0,:] = 0.1* t
fa[1,:] = np.sin(fa[0,:])
v = at.VariableThinMultipole('v', at.ACMode.INTERPOLATION_TABLE, AmplitudeA=1e-3, FuncA=fa)
print(v)

print("Save and load mat")
r1 = at.Lattice([v],energy=1e9)
at.save_mat(r1, 'ring1.mat')
r1l = at.load_mat('ring1.mat')
print(r1l[0])

print("Save and load m")
at.save_m(r1, 'ring1.m')
r1m = at.load_m('ring1.m')
print(r1m[0])
```

But when loading from file there is the following error.
image


This does not happent if I set the `FuncA` and `FuncB` to convert to `_anyarray`

lfarv:
@oscarxblanco : the problem is in the file writing or reading. I'll take your script and have a look.

lfarv:
@oscarxblanco : I found the problem. In `_set_interpolate`, you use
```
    interpolate = kwargs.get("Func" + ab)
```
Instead of
```
    interpolate = kwargs.pop("Func" + ab)
```
So the element in the end has both the `Tinterp`, `Finterp` attributes **AND** the `Func` attribute. I did not expect that. When reading, it's the `Func` attribute which is interpreted…

This is redundant, but it looks complicated to sort this out. Now I understand what is happening and since it does not look harmful, I suggest that you leave it as it is.

Same for the assert statements: though it's not recommended, I have never seen a python version without __debug__ positioned, so leave it as it is.

oscarxblanco:
I have removed all assert and here are the scripts I used to test the save and load.
[test_variable_thin_multipole.zip](https://github.com/user-attachments/files/32386038/test_variable_thin_multipole.zip)

One note: I reused the `FuncA` and `FuncB` in INTERPOLATION_TABLE to avoid creating other keyword for the user. This made possible to keep separated a copy of the user function, and the arrays containing time and the function values. This could be redundant but at the same time allows the user to see if there were modifications because the function was not sorted. However, I could do more changes if you consider that names or any other changes are required.

@lfarv and @swhite2401 , is there any other modification pending ?

lfarv:
OK for me !

PR #1158, opened 2026-09-17
New magnet properties

lfarv:
All magnets have an optional `KickAngle` attribute meant to be used for orbit correction. At the moment, the attribute is combined in the integrators with `PolynomA/B` in a way that is not thread-safe. In the future, it is expected to replace this attribute by  a property based on `PolynomA/B[0]` and `Length`, so that it does not appear any more in the integrators.

This PR anticipates this evolution by introducing 2 properties HKick and VKick but keeps everything else unchanged. The `KickAngle` attribute and the integrators remain unchanged at the moment. But the new properties should be used from now on to provide deviation angles for orbit control. They are available in all magnets, including `Corrector` elements.

Note: `KickAngle` is expressed in angles while `HKick` and `VKick` are expressed in momentum kicks (slope).
So `HKick = tan(KickAngle[0])`. Instead, we could define `HAngle` and `VAngle` to be closer to the legacy behaviour. It's a choice, please give advice.


swhite2401:
Looks good to me, I like the momentum kick personally

swhite2401 (approved):

lfarv:
> I like the momentum kick personally

I agree with you , let's go with `H/VKick`

